PR #301, opened 2018-05-15
Use rollup, ship es6, and update core / layers versions.

nsthorat:
This change does the following:
- Convert to rollup for bundling. Remove browserify.
- Use "extend: true" to extend a global "tf" object. There is now no "tfc".
- Ship es6 in a dist-es6 folder as well as es5.
- Remove models. They don't build anymore nor are they useful. We'll keep the KNN image classifier around and not unit test it just so we have the code. I'll remove this at some point in the future.

This PR goes along with:
https://github.com/tensorflow/tfjs-layers/pull/191
https://github.com/tensorflow/tfjs-core/pull/1037

I tested with yalc across all packages, and eventually all the way to the mobilenet example in tfjs-example and everything seems to work. Using the bundles I've also used the code on the homepage to train a simple layers model (tested both union and including layers and core independently).

Hosted bundles here:
https://storage.googleapis.com/dljs-test/rollup/tf.js
https://storage.googleapis.com/dljs-test/rollup/tf.min.js
https://storage.googleapis.com/dljs-test/rollup/tf-core.js
https://storage.googleapis.com/dljs-test/rollup/tf-core.min.js
https://storage.googleapis.com/dljs-test/rollup/tf-layers.js
https://storage.googleapis.com/dljs-test/rollup/tf-layers.min.js

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/301)


caisq:
:lgtm_strong:

LGTM modulo changing tfjs-layers from 0.6.0 to 0.6.1

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/301#-:-LCnmGsj0MGtjFNlWaL6:b-410nr6)*

PR #355, opened 2018-05-29
Ignore require("crypto").

mbostock:
I haven’t tested (yet), but I believe this should fix the broken require("crypto") in the AMD build.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/355)

PR #368, opened 2018-05-31
Generate an ES Module bundle, too.

### LINKED ISSUE #367: Unable to load tfjs as an ES Module.
To get help from the community, check out our [Google group](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs).

#### TensorFlow.js version

\@tensorflow/tfjs@0.11.2
\@tensorflow/tfjs-converter@0.4.0

Source here: https://unpkg.com/@tensorflow/tfjs@0.11.2/dist-es6/index.js?module

#### Browser version

Google Chrome Version 67.0.3396.62 (Official Build) (64-bit)

I was curious if I could load tfjs as an ES Module rather than as a transpiled UMD; this is supported by [unpkg](https://unpkg.com) and the basis of Observable’s dynamic `import` feature. However, it appears that at least one of the transitive imports is a CommonJS module, namely [url](https://www.npmjs.com/package/url).

https://github.com/tensorflow/tfjs-converter/blob/93ade97d6f3a79246f0e18c2421fcbdfdd3254cc/src/executor/frozen_model.ts#L19

Tensorflow uses rollup-plugin-commonjs to transpile this to an ES module when bundling, but unfortunately this transpilation step isn’t available if you try to load tfjs as an ES module directly.

I think the answer here is probably *not* to eliminate any dependencies on CommonJS packages (though in the long run that’s probably a good idea), but for each Tensorflow package to provide an ES module bundle generated by Rollup *in addition to* the existing UMD bundle. The current “dist-es6” is not a bundle—it’s TypeScript transpiled to ECMAScript, unbundled, and therefore has transitive dependencies on non-Tensorflow modules.

If you distribute a bundled ES module with only Tensorflow dependencies, users of Tensorflow can consume it as an ES module, even if Tensorflow internally uses some non-ES module code.

```js
tf = import("@tensorflow/tfjs")
```

Or more directly:

```js
url = import("url")
```

https://beta.observablehq.com/d/29e5220c6340819e

mbostock:
Fixes #367.

This produces a single ES module bundle with no dependencies. If you prefer, it might be slightly better to retain the dependencies on the other Tensorflow packages (*e.g.*, tfjs-converter). That way if you import those other packages directly as well as tfjs, you don’t get duplicate code. (This would require that the other Tensorflow packages likewise publish an ES module bundle as in this PR.)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/368)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "description": "An open-source machine learning framework.",
   "private": false,
   "main": "dist/index",
-  "jsnext:main": "dist-es6/index.js",
-  "module": "dist-es6/index.js",
+  "jsnext:main": "dist/tf.esm.js",
+  "module": "dist/tf.esm.js",
   "jsdelivr": "dist/tf.min.js",
   "unpkg": "dist/tf.min.js",
   "types": "dist/index.d.ts",
@@ -40,11 +40,11 @@
     "rollup-plugin-json": "~3.0.0",
     "rollup-plugin-node-resolve": "3.3.0",
     "rollup-plugin-typescript2": "0.13.0",
+    "rollup-plugin-uglify": "^3.0.0",
     "shelljs": "~0.8.1",
     "ts-node": "~4.1.0",
     "tslint": "~5.9.1",
     "typescript": "2.7.2",
-    "uglify-js": "~3.0.28",
     "yalc": "~1.0.0-pre.21"
   },
   "scripts": {
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -20,44 +20,83 @@ import typescript from 'rollup-plugin-typescript2';
 import commonjs from 'rollup-plugin-commonjs';
 import resolve from 'rollup-plugin-node-resolve';
 import json from 'rollup-plugin-json';
+import uglify from 'rollup-plugin-uglify';
 
-export default {
-  input: 'src/index.ts',
-  plugins: [
-    typescript(),
-    node(),
-    // Polyfill require() from dependencies.
-    commonjs({
-      ignore: ["crypto"],
-      include: 'node_modules/**',
-      namedExports: {
-        './node_modules/seedrandom/index.js': ['alea'],
-        './src/data/compiled_api.js': ['tensorflow'],
-        './node_modules/protobufjs/minimal.js': ['roots', 'Reader', 'util']
-      },
-    }),
-    json(),
-    // We need babel to compile the compiled_api.js generated proto file from es6 to es5.
-    babel()
-  ],
-  output: {
-    extend: true,
-    banner: `// @tensorflow/tfjs Copyright ${(new Date).getFullYear()} Google`,
-    file: 'dist/tf.js',
-    format: 'umd',
-    name: 'tf',
-    globals: {
-      'crypto': 'crypto'
+const copyright =  `// @tensorflow/tfjs Copyright ${(new Date).getFullYear()} Google`;
+
+function minify() {
+  return uglify({
+    output: {
+      preamble: copyright
     }
-  },
-  external: ['crypto'],
-  onwarn: warning => {
-    let {code} = warning;
-    if (code === 'CIRCULAR_DEPENDENCY' ||
-        code === 'CIRCULAR' ||
-        code === 'THIS_IS_UNDEFINED') {
-      return;
+  });
+}
+
+function config({plugins = [], output = {}}) {
+  return {
+    input: 'src/index.ts',
+    plugins: [
+      typescript(),
+      node(),
+      // Polyfill require() from dependencies.
+      commonjs({
+        ignore: ["crypto"],
+        include: 'node_modules/**',
+        namedExports: {
+          './node_modules/seedrandom/index.js': ['alea'],
+          './src/data/compiled_api.js': ['tensorflow'],
+          './node_modules/protobufjs/minimal.js': ['roots', 'Reader', 'util']
+        },
+      }),
+      json(),
+      // We need babel to compile the compiled_api.js generated proto file from es6 to es5.
+      babel(),
+      ...plugins
+    ],
+    output: {
+      banner: copyright,
+      ...output
+    },
+    external: ['crypto'],
+    onwarn: warning => {
+      let {code} = warning;
+      if (code === 'CIRCULAR_DEPENDENCY' ||
+          code === 'CIRCULAR' ||
+          code === 'THIS_IS_UNDEFINED') {
+        return;
+      }
+      console.warn('WARNING: ', warning.toString());
     }
-    console.warn('WARNING: ', warning.toString());
-  }
-};
+  };
+}
+
+export default [
+  config({
+    output: {
+      format: 'umd',
+      name: 'tf',
+      extend: true,
+      file: 'dist/tf.js'
+    }
+  }),
+  config({
+    plugins: [
+      minify()
+    ],
+    output: {
+      format: 'umd',
+      name: 'tf',
+      extend: true,
+      file: 'dist/tf.min.js'
+    }
+  }),
+  config({
+    plugins: [
+      minify()
+    ],
+    output: {
+      format: 'es',
+      file: 'dist/tf.esm.js'
+    }
+  })
+];
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -32,7 +32,7 @@ function minify() {
   });
 }
 
-function config({plugins = [], output = {}}) {
+function config({plugins = [], output = {}, external = []}) {
   return {
     input: 'src/index.ts',
     plugins: [
@@ -57,7 +57,10 @@ function config({plugins = [], output = {}}) {
       banner: copyright,
       ...output
     },
-    external: ['crypto'],
+    external: [
+      'crypto',
+      ...external
+    ],
     onwarn: warning => {
       let {code} = warning;
       if (code === 'CIRCULAR_DEPENDENCY' ||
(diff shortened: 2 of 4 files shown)


dsmilkov:
Thanks Mike! Assuming we drop the `Url` dependency in `tfjs-converter` (pretty easy to do), what's the benefit of having an ES module bundle vs having `dist-es6/index.js` unbundled. I'm asking since I see `d3-array`  doesn't bundle it's source files: https://unpkg.com/d3-array@1.2.1/index.js?module

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/368#-:-LDsIuDy7rtRXCbyQ4FV:b-8b0vyb)*

dsmilkov:
its* source files

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/368#-:-LDsJwmj3u0xROC1dzba:blle0ff)*

dsmilkov:
Just realized the answer might be related to your point about de-duplication: if everything is unbundled, you can load es6 module all the way to the file level without risking code duplication

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/368#-:-LDsKBd13cwxHwuuJF5m:bof0k9n)*

mbostock:
It’s my intention to distribute ES module bundles with the D3 modules, too—I just haven’t gotten around to it yet. 😳 Support for ES modules in browsers is relatively recent; previously, ES modules were only intended to be consumed by bundlers (such as Rollup), and so it didn’t matter that there were hundreds of small files. Now that we can consume them in browsers, and even with HTTP/2, it makes sense to bundle and minify them.

hk144:
:lgtm_strong:

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/368#-:-LDvvnkg17S4DvDBkbzw:b-jbsf8z)*

dsmilkov:
Sorry for the delay in reviewing this Mike. Just a couple more questions as I'm still learning about the js build world. Thank you!

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[package.json, line 7 at r1](https://reviewable.io/reviews/tensorflow/tfjs/368#-LE1IaSUEnqKNO94Kazg:-LE1IaSUEnqKNO94Kazh:b-8fciyb) ([raw file](https://github.com/tensorflow/tfjs/blob/f4faed6b3dca83dea37906f8b922ee951d14505c/package.json#L7)):*
> ```JSON
>   "private": false,
>   "main": "dist/index",
>   "jsnext:main": "dist/tf.esm.js",
> ```

Now that the module is bundled, would that impact how much bundlers like rollup can tree-shake? And also what happens if users that are using bundlers like rollup do the following:
```
import * as tfc from '@tensorflow/tfjs-core';
import * as tfl from '@tensorflow/tfjs-layers';
```

would rollup figure out to dedup since tfjs-layers imports bunch of code from tfjs-core?

Related q: is there a way in package.json to separate the module that rollup can use (unbundled) vs the module that the browser can use (bundled)?

---

*[rollup.config.js, line 23 at r1](https://reviewable.io/reviews/tensorflow/tfjs/368#-LE1JbKh-Tu2znUSHugc:-LE1JbKh-Tu2znUSHugd:b-khvb49) ([raw file](https://github.com/tensorflow/tfjs/blob/f4faed6b3dca83dea37906f8b922ee951d14505c/rollup.config.js#L23)):*
> ```JavaScript
> import resolve from 'rollup-plugin-node-resolve';
> import json from 'rollup-plugin-json';
> import uglify from 'rollup-plugin-uglify';
> ```

I tried this on my laptop and for some reason it gives [!] TypeError: uglify is not a function, but `import {uglify} from ....` works

---

*[rollup.config.js, line 49 at r1](https://reviewable.io/reviews/tensorflow/tfjs/368#-LE1Jv212W4_kqGyMNqe:-LE1Jv212W4_kqGyMNqf:bdoy1vb) ([raw file](https://github.com/tensorflow/tfjs/blob/025c24cb94bcd520337eead2ed11994fafaa94d4/rollup.config.js#L49)):*
> ```JavaScript
>     format: 'umd',
>     name: 'tf',
>     globals: {
> ```

curious the globals went away. Is that because tfjs-core took care of that and any libraries upstream shouldn't worry about it?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/368#-:-LE1K5_aDEl9rsQtTXg_:b-qb1ier)*

dsmilkov:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[rollup.config.js, line 49 at r1](https://reviewable.io/reviews/tensorflow/tfjs/368#-LE1Jv212W4_kqGyMNqe:-LE1L8rvD5Ern6TyDTCH:bj5wck2) ([raw file](https://github.com/tensorflow/tfjs/blob/025c24cb94bcd520337eead2ed11994fafaa94d4/rollup.config.js#L49)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

curious the globals went away. Is that because tfjs-core took care of that and any libraries upstream shouldn't worry about it?
</blockquote></details>

Ah I see, this is not needed since we ignore crypto (your previous PR)

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/368)*

mbostock:
Switching to

```js
import {uglify} from "rollup-plugin-uglify";
```

generates the following error for me

```
$ node_modules/.bin/rollup -c
[!] TypeError: rollupPluginUglify.uglify is not a function
TypeError: rollupPluginUglify.uglify is not a function
    at minify (/Users/mbostock/Development/tensorflow/tfjs/rollup.config.js:32:29)
    at Object.<anonymous> (/Users/mbostock/Development/tensorflow/tfjs/rollup.config.js:88:7)
    at Module._compile (module.js:643:30)
    at Object.require.extensions..js (/Users/mbostock/Development/tensorflow/tfjs/node_modules/rollup/bin/rollup:2914:24)
    at Module.load (module.js:556:32)
    at tryModuleLoad (module.js:499:12)
    at Function.Module._load (module.js:491:3)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at /Users/mbostock/Development/tensorflow/tfjs/node_modules/rollup/bin/rollup:2921:32
```

I’m not sure what would explain this difference, unless perhaps you didn’t run `yarn`?

mbostock:
I don’t think the global crypto (or the external crypto) is needed now that we ignore the require crypto per #355.

mbostock:
> Related q: is there a way in package.json to separate the module that rollup can use (unbundled) vs the module that the browser can use (bundled)?

You can direct the browser (or a bundler) to import the ES module directly rather than using the `module` entry point, but the default behavior of both unpkg and bundlers is to use the `module` entry point, and there isn’t a different entry point to indicate whether the `module` entry point is bundled or unbundled.

Related: I’ve pushed a commit that changes the ES module so that it doesn’t inline the other tfjs submodules; instead, it imports and re-exports them. This means that if you import tfjs in an ES module environment (either a browser or a bundler), you won’t get duplicate code if you also import a tfjs submodule (such as tfjs-core) directly.

I think this is what we want going forward, but note that it will require equivalents to this PR to all the tensorflow submodules. (As well as the crypto fix, tensorflow/tfjs-core#1092.) Happy to port this to the other repos if you’re supportive, or assist reviewing your PRs if you want to take over.

dsmilkov:
:lgtm_strong: Thanks a lot Mike! Now that I have the context, and this PR as an example, I can do the equivalent in the other repos - thanks so much for spending time on this.

---

Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/368#-:-LEeP5CM8Jr5BPIgAN0p:b-on3z3r)*

dsmilkov:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[rollup.config.js, line 23 at r1](https://reviewable.io/reviews/tensorflow/tfjs/368#-LE1JbKh-Tu2znUSHugc:-LEePyW53Er2yc18NRFa:b-vszo94) ([raw file](https://github.com/tensorflow/tfjs/blob/f4faed6b3dca83dea37906f8b922ee951d14505c/rollup.config.js#L23)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

I tried this on my laptop and for some reason it gives [!] TypeError: uglify is not a function, but `import {uglify} from ....` works
</blockquote></details>

Resolving this comment since I can't reproduce the problem again.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/368)*

lastmjs:
Has this been released yet? I'm running into issues because of the url dependency, I was hoping this wouldn't have been a problem with a pure es module bundle

lastmjs:
Having to import this file: `/node_modules/@tensorflow/tfjs-converter/dist-es6/executor/frozen_model.js` leads to importing `url`, which does not have a module bundle on my machine.

dsmilkov:
I'm about to submit https://github.com/tensorflow/tfjs-converter/pull/168 which will fix this. After that's in, we will release a new patch version 0.12.1 for `@tensorflow/tfjs`

lastmjs:
It works!!! Thank you very much

PR #526, opened 2018-07-18
Updated the gallery 

zaidalyafeai:
I added two notebooks to the gallery.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/526)


nsthorat:
Hi @zaidalyafeai, regarding the intro: I think we're going to air on the side of non-duplication, and there is pretty good coverage of the intro concepts of TensorFlow.js already. Regarding the activation reconstruction notebook, I think there may not be enough explanation of why you are doing what you're doing -- this may scare newer folks.

zaidalyafeai:
@nsthorat , fair enough.

Issue #787, opened 2018-10-12
Improve source map support for all packages, and built UMD bundle


tafsiri:
unminified / with source maps. Currently if there are exceptions in the lib when using

```
<script src="https://cdn.jsdelivr.net/npm/@tensorflow/tfjs@0.13.1/dist/tf.js"> </script>
```

These are harder to debug than when using 

```
<script src="https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-core@0.13.5/dist/tf-core.js"></script>
<script src="https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-layers@0.8.1/dist/tf-layers.js"> </script>    
```

(Note that in the above i explicitly load the non minified versions). The latter has sourcemap support and give a nice debugging experience.

tafsiri:
Opening the scope of this to just more broadly improve sourcemap support given some recent issues we've had with debugging.

Leaving a breadcrumb here to pick up the investigation next week. But basically I think we may need to put the ts sources in our npm bundle, or we could re-write the sourcemap urls to the compiled js. However i think the latter is not as ideal as its not as readable as the ts.

I've made a branch in a side repo so that we can discuss it more easily https://github.com/tafsiri/tfjs/tree/sourcemap-experiments/node_modules/%40tensorflow, if you trace the sourcemap urls for the union package they eventually reference the original ts sources which are not actually in the npm package.

cc @dsmilkov just fyi for now.

tafsiri:
Update: https://github.com/tensorflow/tfjs/pull/816 handles the umd bundle, while upgrading parcel fixes the issues we were having when bundling locally.

tafsiri:
https://github.com/tensorflow/tfjs-website/pull/199 updates the default link on the homepage to one that will correctly load the sourcemaps for the umd bundle.

tafsiri:
Fixed for tfjs-examples in https://github.com/tensorflow/tfjs-examples/pull/161

tafsiri:
I've tested sourcemap support for core, layers and converter umd bundles and they are working. The only file i noticed without a sourcemap is https://github.com/tensorflow/tfjs-converter/blob/master/src/data/compiled_api.js (check your node_modules for an installed tfjs to see). Not sure how critical it is and if i remember correctly this is not a file we generate ourselves.

@dsmilkov Going to close this unless you can think of other places to check/things to look at.

dsmilkov:
[...]

PR #816, opened 2018-10-22
incorporate sourceMappingUrls from dependencies

tafsiri:
This improves the sourcemaps for the union package by allowing it to bring in the sourcemaps generated by the subpackages, as opposed to generating sourcemaps for the compiled versions of the dependencies.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/816)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "rollup-plugin-commonjs": "9.1.3",
     "rollup-plugin-json": "~3.0.0",
     "rollup-plugin-node-resolve": "3.3.0",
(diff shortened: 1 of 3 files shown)


tafsiri:
[sourcemap-comparisons.zip](https://github.com/tensorflow/tfjs/files/2502070/sourcemap-comparisons.zip)

The above zip file has the old sourcemap and the new one for comparison. A quick check to compare between the two is to search for type annotations like `: number`

tafsiri:
closes #787

PR #1183, opened 2019-02-01
Update TS version to 2.9.2 in the union package.

nsthorat:
All the other packages are 2.9.2. Without this, the build fails (`Extract` type added in layers which is defined in 2.9.2).


nsthorat:
There should be a lock file change. For some reason my mac isn't updating it, working on it. Won't submit before the lock file change is in.

nsthorat:
There we go -- had to upgrade yarn.

PR #1361, opened 2019-03-12
Add more benchmarking targets.

annxingyuan:
Here is a link to a completed build: https://travis-ci.org/tensorflow/tfjs/builds/508099474

#### Changes

- Add Universal Sentence Encoder and CoCoSSD to benchmarks
- Add Safari target and iOS targets

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1361)


annxingyuan:
@dsmilkov 

If I run the browsers in parallel (`--browsers=bs_safari_mac,bs_chrome_mac`) instead of with separate calls to `run-browserstack`, how do I figure out which browser I'm currently targeting from within a test?

annxingyuan:
@caisq 

I've removed 'GPU' from the function names. 

The warmup now occurs here: https://github.com/tensorflow/tfjs/pull/1361/files#diff-1bd79e776a0f7e677d8a0b33dc9d0a7aR37

PR #1444, opened 2019-03-25
[benchmarks] Add new type hierarchies to benchmarks/types.ts

caisq:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1444)

### DIFF
--- a/integration_tests/benchmarks/benchmark_models_test.ts
+++ b/integration_tests/benchmarks/benchmark_models_test.ts
@@ -30,6 +30,7 @@ describe('benchmark models', () => {
     const runs = 20;
 
     const benchmark = new MobileNetV1GPUBenchmark();
+    await benchmark.loadModel();
 
     await test_util.benchmarkAndLog(
         'mobilenet_v1', size => benchmark.run(size), sizes,
@@ -41,6 +42,7 @@ describe('benchmark models', () => {
     const runs = 20;
 
     const benchmark = new UniversalSentenceEncoderBenchmark();
+    await benchmark.loadModel();
 
     await test_util.benchmarkAndLog(
         'use', size => benchmark.run(size), sizes, size => '41', runs);
@@ -51,6 +53,7 @@ describe('benchmark models', () => {
     const runs = 20;
 
     const benchmark = new CoCoSSDBenchmark();
+    await benchmark.loadModel();
 
     await test_util.benchmarkAndLog(
         'cocossd', size => benchmark.run(size), sizes, size => '224', runs);
--- a/integration_tests/benchmarks/benchmark_travis.sh
+++ b/integration_tests/benchmarks/benchmark_travis.sh
@@ -29,6 +29,7 @@ then
   echo 'Use latest version of tfjs-core'
   git clone https://github.com/tensorflow/tfjs-core.git --depth 5
   cd tfjs-core
+  HASH_CORE=`git rev-parse HEAD`
   rm -rf dist/ && yarn && yarn build && rollup -c && yalc push
 
   cd ..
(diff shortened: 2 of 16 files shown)


caisq:
Gentle ping @ other reviews.

caisq:
Gentle ping @davidsoergel for suggestions on schema design for fast database access.

nsthorat:
FYI you can index on fields with indexOn: https://firebase.google.com/docs/database/security/indexing-data

and it looks like you can do it on non-flat structures:
https://groups.google.com/forum/#!topic/firebase-talk/m3k51DOyfrs

You can use this tool to find out what the right rules are:
https://firebase.google.com/docs/database/usage/profile

annxingyuan:
> Plot metric vs time, given constraints (e.g. which model, which environment)

I feel like this is our primary use case. Would it be possible to achieve this using `indexOn` ([docs](https://firebase.google.com/docs/database/security/indexing-data)) and keeping the data hierarchical by date?

davidsoergel:
From what I understand, no.  `indexOn` can improve query performance, but it does not make new kinds of queries possible.  Realtime Database provides no query syntax that allows filtering within subtrees (https://firebase.google.com/docs/database/web/lists-of-data).

The top-level object in this PR is, I think,

    /** Benchmark logs from multiple days. */
    export type BenchmarkHistory = {
      [datetime: string]: SuiteLog
    };

Given this hierarchical schema, you **can**
* Select all SuiteLogs within a given date range, and load them fully.
* Select all SuiteLogs that meet a condition nested somewhere within them (e.g., a certain metric is present), and load them fully.  (In our case, this will generally cause loading the entire database).
* *Within a given SuiteLog*, select all TaskGroupLogs that meet a certain condition
* *Within a given TaskGroupLog*, select all MultiEnvironmentTaskLogs that meet a certain condition
* (etc.)

You **cannot**
* Select all TaskLogs that meet a certain condition--i.e., filtering on taskName or time range, and including TaskLogs nested within different SuiteLogs / TaskGroupLogs / MultiEnvironmentTaskLogs.

That last thing is exactly what you need for the primary use case.  If you think it's possible and I'm misunderstanding something, can you show the query that would achieve what you need using the RTDB ordering and filtering functions?

davidsoergel:
Oh, maybe I confounded your question about a hierarchical schema with the question of which database to use.  It may still be possible to do what you want using a hierarchical schema *in Firestore*.  (I don't see any reason to do that, though.)

caisq:
@davidsoergel I agree with @annxingyuan that the most common benchmark type is "selecting a certain date range of a given benchmark". For instance: "select all benchmarks results between 2019-04-01 and 2019-08-01 for the benchmark called MobileNetV2.predict". 

Can this be achieved through reorganizing the schema we have here from datetime-first to benchmark-first?

caisq:
@davidsoergel Ah - just saw your gist. Catching up. Sorry.

davidsoergel:
NP.  To be clear, yes you could rearrange the hierarchy to be metric-first to make the common query more efficient-- at the cost of the other query types, which would still require loading the whole db.

annxingyuan:
Benchmark-first makes sense to me!

davidsoergel:
If you're committed to maintaining a hierarchical structure, then yes, metrics-first is best for the primary use case.  But, I'm missing why you want to force the data into a hierarchy at all, when it is not naturally hierarchical.

For RTDB, the first best practice listed at https://firebase.google.com/docs/database/web/structure-data is 'avoid nesting data'.  For Firestore, the pros/cons at https://firebase.google.com/docs/firestore/manage-data/structure-data point in the same direction as well.

davidsoergel:
Oops the comment that taskType is redundant is obsolete— we agreed that’s
on purpose.

PR #1557, opened 2019-05-01
PERF: change map,foreach to for

aymen94:
change `map`, `foreach ` to `for...of`,  performance increase.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1557)

### DIFF
--- a/integration_tests/benchmarks/firebase.ts
+++ b/integration_tests/benchmarks/firebase.ts
@@ -16,7 +16,7 @@
  */
 
 import firebase from 'firebase/app';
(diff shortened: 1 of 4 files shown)


dsmilkov:
Hi @aymen94. Thanks for the contribution! It's unclear what performance improvements we will get from this optimization. Also when we deploy to npm, we actually transpile to es5 so `for...of` goes away. Also the files that you are changing are unrelated to the actual library. I'm going to close this PR but feel free to reply if you have perf numbers, and I can reopen if needed.

PR #1580, opened 2019-05-13
Add benchmark for tf.GraphModel

caisq:
This PR adds benchmarks for  the GraphModel version of MobileNetV2.

Such models are generated from Python TensorFlow in the following steps:
1. Load the MobileNetV2 application from tf.keras
2. Save them as SavedModels using `tf.keras.experimental.export_saved_model()`
3. Convert the SavedModels into tfjs_graph_model format using tensorflowjs_converter

The during the browser (karma) benchmarks, the tfjs_graph_model artifacts are loaded in the
testing page and their `predict()` methods are called with random data for performance benchmarking.

Like other model formats (LayersModel), the results are logged to Cloud Firestore, which can
be viewed in the frontend. 

For MobileNetV2, the four width values are benchmarked: 0.25, 0.5, 0.75, 1.0.
For each of the width values, the benchmark numbers consistently show a performance
advantage of GraphModel over LayersModel by about 15%-20%.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1580)

### DIFF
--- a/integration_tests/benchmarks/layers/layers_benchmarks.ts
+++ b/integration_tests/benchmarks/layers/layers_benchmarks.ts
@@ -355,6 +355,7 @@ describe('TF.js Layers Benchmarks', () => {
         model = await loadLayersModel(modelName);
       } else if (modelFormat === 'GraphModel') {
         if (isNodeJS) {
+          // TODO(cais): Benchmark GraphModel in Node.js.
           console.warn(
               `WARNING: Skipping GraphModel in Node.js benchmark: ` +
               `${modelName}`);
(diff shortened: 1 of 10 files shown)


caisq:
cc @annxingyuan

integration_tests/benchmarks/layers/layers_benchmarks.ts
```diff
@@ -27,6 +27,7 @@ import * as tfl from '@tensorflow/tfjs-layers';
 import {BrowserEnvironmentType, BrowserEnvironmentInfo, ModelBenchmarkRun, ModelFunctionName, ModelTrainingBenchmarkRun, VersionSet, PythonEnvironmentInfo, NodeEnvironmentType, NodeEnvironmentInfo, EnvironmentInfo} from '../types';
 // tslint:disable-next-line:max-line-length
 import {addEnvironmentInfoToFirestore, addOrGetTaskId, addBenchmarkRunsToFirestore, addVersionSetToFirestore} from '../firestore';
+import {NamedTensorMap} from '@tensorflow/tfjs-converter/dist/src/data/types';
```
pyu10055: maybe rename the file to model_benchmarks.ts and move to a different directory?

integration_tests/benchmarks/layers/layers_benchmarks.ts
```diff
@@ -127,24 +129,30 @@ function getRandomInputsAndOutputs(model: tfl.LayersModel, batchSize: number):
...
+async function syncData(tensors: tfc.Tensor|tfc.Tensor[]|NamedTensorMap) {
+  if (tensors instanceof tfc.Tensor) {
+    await (tensors as tfc.Tensor).data();
+  } else  if (Array.isArray(tensors)) {
     const promises = tensors.map(tensor => tensor.data());
     await Promise.all(promises);
   } else {
-    await tensors.data();
+    // tensors is a NamedTensorMap.
+    const promises = Object.entries(tensors).map(item => item[1].data());
```
pyu10055: should wait for data of all items for each entry.

caisq: This is already doing that. `item` is a [key, value] pair. The key is ignored, while the value is `item[1]` here.

integration_tests/benchmarks/layers/layers_benchmarks.ts
caisq: Good point. Done. Renamed to `model_benchmarks.ts` and moved one level up.

caisq:
Thanks for the review, @pyu10055

PR #1609, opened 2019-05-24
Mention tfjs-data in README

pvaneck:
Since tfjs-data is a part of the union package, it would be worthwhile to mention it in the README.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1609)


dsmilkov:
Thank you!

PR #1628, opened 2019-05-30
Fix benchmarks: let cp follow symlinks

caisq:
Related to https://github.com/tensorflow/tfjs-node/pull/252/files

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1628)


caisq:
cc @nkreeger

PR #1787, opened 2019-08-01
Add command to set-url

kangyizhang:
When creating /tmp/tfjs-release/tfjs the default remote url might be `https://github.com/xxxx.git`, which does not use ssh. Set it to `git@github.com/xxxx.git` before push.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1787)

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -207,6 +207,7 @@ async function main() {
       phase.scripts.forEach(script => $(script));
     }
 
+    $(`git remote set-url origin git@github.com:tensorflow/${repo}.git`);
     $(`git checkout -b b${newVersion}`);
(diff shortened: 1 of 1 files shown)


nsthorat:
Just curious you couldn't get this to work with the regular https? It should work in either mode.

kangyizhang:
No. When it's https, it asks for username and password in terminal, which breaks the release script.

nsthorat:
You can fix that with a global git login, changing the time to live of the login. I would prefer you do that to doing this change.

nsthorat:
Can we close this out?

kangyizhang:
Yes.

PR #1826, opened 2019-08-14
Update tfjs-core to 1.2.8.

nsthorat:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1826)


nsthorat:
tfjs-node and tfjs-layers failures are expected. I manually linked tfjs-data and tfjs-converter and they succeed (here they are timing out because of browserstack).

PR #1852, opened 2019-08-19
Fix document state check error suggests a Inappropriate event

y0ke:
https://github.com/tensorflow/tfjs/blob/aa92053a8d7c640fe9e685e254655f417480cf0d/tfjs-core/src/backends/webgl/backend_webgl.ts#L313-L321
The error message above suggests to use `DOMContentLoaded` though, Its if condition checks `document.readyState !== 'complete'`.
Because `DOMContentLoaded` is called before `document.readyState` becomes `'complete'`.  It's  an inappropriate suggestion. 
It should suggest `load` event instead.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1852)


nsthorat:
Thanks for the fix, this is correct!

PR #1854, opened 2019-08-19
Update tfjs-layers, tfjs-converter, tfjs-data to 1.2.8, 1.2.8, 1.2.8.

nsthorat:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1854)


nsthorat:
The test failure is coming from tfjs-data snippets depending on an older version of layers. This will be fixed once we do a release.

PR #1862, opened 2019-08-19
Consolidate make-version, and tag-version scripts.

nsthorat:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1862)


nsthorat:
Tests are still failing because of the tfjs-data snippet, this will get fixed once layers publishes.

PR #1865, opened 2019-08-19
[tfjs-layers] Fix memory leak with loaded models containing nested Sequential layers

pvaneck:
If a model loaded with loadLayersModel contains a nested Sequential layer, disposing of the main model leaves the tensors of the nested model behind. This is because the refCount of the nested model doesn't go to zero during the call to `dispose` since the refCount was incremented twice (once during submodel creation and once when the submodel is added as a layer) but only decremented once during.

This commit makes it so references to each internally constructed sub-container are tracked and disposed when the parent model is disposed to ensure that the refCount hits zero and tensors are disposed.

*Note:* Nested sequential layers constructed through the layers API still works as intended as tested by: https://github.com/tensorflow/tfjs/blob/master/tfjs-layers/src/engine/container_test.ts#L603

Original PR before re-org: https://github.com/tensorflow/tfjs-layers/pull/578

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1865)

### DIFF
--- a/tfjs-layers/src/engine/container.ts
+++ b/tfjs-layers/src/engine/container.ts
@@ -59,6 +59,8 @@ export abstract class Container extends Layer {
   layersByDepth: {[depth: string]: Layer[]};
   nodesByDepth: {[depth: string]: Node[]};
 
+  internalContainerRefs: Container[];
+
   containerNodes = new Set<string>();
 
   // TODO(michaelterry): Add cache support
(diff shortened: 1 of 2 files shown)


irealva:
This is _maybe_ happening in Teachable Machine? We have exactly five tensors that remain even after disposing all the models and haven't yet figured out where they're coming from.

dsmilkov:
And for the carefully written unit test!

PR #1878, opened 2019-08-21
[tfjs-react-native] Run core unit tests in react native app

tafsiri:

This PR runs tfjs-core unit tests in the react app. Another step on the road to automating the tests. Tests can run in both CPU and WebGL backends, but this PR only has fixes to get all tests passing on CPU. Precision/NaN stuff for GPU will follow in another PR.

It looks something like this.

Screen Shot 2019-08-16 at 5 18 35 PM

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1878)


tafsiri:
Note that is what it looks like with tests failing. But all tests currently pass on CPU. About 10 or so are failing on GPU primarily relating to NaN propagation.

PR #1879, opened 2019-08-21
[tfjs] Fix integration tests and benchmarks to work with the monorepo.

nsthorat:
We do this by copying local directories instead of cloning them. This keeps the same functionality as before.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1879)


nsthorat:
@caisq just for the interest of unblocking some other stuff I'm going to merge this -- please leave comments and I will send a follow up PR to fix anything you may ask for!

Issue #1894, opened 2019-08-22
Flaky dropout test in tfjs-layers


dsmilkov:
Chrome 76.0.3809 (Mac OS X 10.13.6) Dropout Layer cpu undefined tensor training=true, dropoutRate=0.5, noiseShape=[2,3,1] FAILED
Expected 24 to be less than 24.

See logs: https://pantheon.corp.google.com/cloud-build/builds/1cd1f7aa-190b-44a7-8464-162106b4a737?project=learnjs-174218

Related to this PR: https://github.com/tensorflow/tfjs-layers/pull/556

PR #1895, opened 2019-08-22
Fix flaky dropout test

dsmilkov:
It is not that unlikely to get ~20 heads in a row in 0.5 probability of flip.

Fixes https://github.com/tensorflow/tfjs/issues/1894

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1895)


caisq:
Alternative to consider: set a random seed here

PR #1899, opened 2019-08-23
[tfjs-node] Adds encodeJpeg and encodePng

justadudewhohacks:
This PR exposes the bindings to tf.io.encode_jpeg and tf.image.encode_png.

The image decoding operations have been added to tfjs-node a while ago, but currently we still need the node canvas package to save images. That's why I suggest to include these bindings.

I made the necessary changes after @kangyizhang's review in the original PR: https://github.com/tensorflow/tfjs-node/pull/296. 

I didn't include validation of encoded value since the encoded data array is pretty large. That's why I only check the data prefix to match the according encoding. If there are still adjustments to be done, let me know.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1899)


justadudewhohacks:
>  Is it possible to use tf.node.decodeImage() to get the decoded data and validate if it matches the original Uint8Array? If yes, it would be great to add this validation to one encodeJpeg and one encodePng test. No need to do it in all the tests.

That works for encodePng, because compression so I have added validation to the unit test. 

Doesn't work for encodeJpeg though, because of the lossy compression. Not even the length of both data arrays are the same. One can only validate this visually by decoding a jpeg image from disk, encoding it and saving it back to disk.

PR #1956, opened 2019-09-02
remove dist-import in node & layers

WenheLI:
This is about https://github.com/tensorflow/tfjs-node/pull/289.
This pr removes dist-importing in tfjs-node and tfjs-layers. 
The rest of the removing will be at a separate PR.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1956)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
     "clang-format": "~1.2.4",
     "chalk": "~2.4.2",
     "shelljs": "~0.8.3",
+    "tslint": "^5.11.0",
     "ts-node": "~4.1.0",
     "typescript": "3.5.3"
   },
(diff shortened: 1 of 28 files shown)


WenheLI:
@dsmilkov - Hey Daniel, it seems that the CI will run changes separately based on different folders they are in. 
Am I right about this?
 As I need to import `tslint` in the root folder, but the test in `react-native` does not have the change about `tslint` in the root folder.

dsmilkov:
Hi, yes, CI runs separately for each folder, however because you are touching the global `tslint.json`, which is shared across all projects, the CI will run the lint job across all folders. So you will need to fix the dist imports across all projects.

Issue #1866, opened 2019-08-19
Support op TF.unique


evancasey:
It would be really great to have support for [tf.unique](https://www.tensorflow.org/api_docs/python/tf/unique).

This has been brought up by a number of [previous](https://github.com/tensorflow/tfjs/issues/958) [issues](https://github.com/tensorflow/tfjs/issues/1693) and in general it is a very common operation that many applications rely on.

rthadur:
@evancasey thank you , may i know your use case ?

evancasey:
Hey @rthadur, I'm using tf.unique right now to do a type of one-hot encoding where I have an input vector with a discrete number of unique values but the values are scattered between [0,N] and N is potentially very large. I can't call tf.one_hot directly on this because I want the depth to be equal to the number of unique values.

rthadur:
@pyu10055 any plans to support TF.unique op ?

ClaytonSmith:
Poking this issue. There aren't any good alternatives to `unique`.

PR #1959, opened 2019-09-03
add unique op

WenheLI:
This pr is about #1866 
That adds unique op and tests.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1959)


tfjs-core/src/ops/array_ops.ts
```diff
@@ -1114,6 +1114,24 @@ function print<T extends Tensor>(x: T, verbose = false): void {
   console.log(x.toString(verbose));
 }
 
+/**
+ * Finds unique elements in a 1-D tensor.
+ *
+ * ```js
+ * const t = tf.tensor([1,1,2,2]);
+ * const [values, idx] = await tf.unique(t);
```
Lewuathe: The sample code runs at the test time. As `await` can block the main thread, it may be better to just print the result?
```suggestion

  * tf.unique(t).print();
```

WenheLI: I see your point! But unique returns an array of tensors. Thus, we can not print it directly.

evancasey:
Does this work on tensors with 1M+ elements?

WenheLI:
@evancasey - I have tested manually that it works on 1m+ elements. I will add another unit test to ensure that is fine.

evancasey:
Hey just wanted to bump this -- I'd really love to use this! It would be wonderful to get this mainlined @nsthorat

nsthorat:
@annxingyuan / @lina128 could one of you review this one :)

annxingyuan:
Thanks for this contribution! I left a few comments. Also wondering - do you happen to know how long the 1m+ test takes to run?

WenheLI:
Thanks for the reviews! Since this PR was drafted a while ago, I will propose a new PR later with the suggested changes!

lina128:
Unique has been added to the code base. Thank you for the contribution all the same, @WenheLI !

PR #1963, opened 2019-09-03
Remove the systemFetch export from platform_node

nsthorat:
Closure doesn't like exported objects.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1963)


rkirov:
Thanks! Note that you already were working against ES2015 module semantics (hence the ts-ignores that you can now remove), and things just happened to work because the actual code ran as a CommonJS module. So in terms of strictness - Closure modules > ES2015 modules > CommonJS modules.

nsthorat:
Ah totally makes sense @rkirov, thank you!

PR #1968, opened 2019-09-04
[WebGPU] Don't release the uniform buffer under delayed mode

qjia7:
BUG

In delayed mode, the command list hasn't been executed actually.
So, the uniform buffer can't be recycled.

This patch also fixes the issue that PostNet works incorrectly under
delayed mode.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1968)


qjia7:

PR #1981, opened 2019-09-05
Fix bug in converter to load saved model

dsmilkov:
BUG

Fix a bug that prohibits the conversion of a Saved Model produced by AutoML object detection.

The error was produced by an internal `convert_variables_to_constants() method`:
`ValueError: Cannot find the variable that is an input to the ReadVariableOp.`

Comparing our code to the `freeze_graph` util showed that we were not loading the graph properly into a Session object, before calling `convert_variables_to_constants()`.

The unit test emulates the Saved Model by adding a hash table op that is not used by the inference signature. Verified that this test fails at master.

Cloud will make an integration test internally that asserts that this conversion continuous to work.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/1981)


caisq:
@dsmilkov Per offline discussion, I'll await your new unit test(s).

dsmilkov:
@caisq 	PTAL. Added a unit test that emulates the Saved Model by adding a hash table op that is not used by the inference signature. Verified that this test fails at master. Simply creating a `variable(..., use_resource=False)` was not enough to repro the error.

PR #2002, opened 2019-09-09
webgpu: Fix the compile error in conv2d_naive shader

qjia7:
BUG

When switch to the naive method of conv2d, below error will be met:
Error: Shader compilation failed: file:124: error: 'getX' : no matching overloaded function found.
Error: Shader compilation failed: file:131: error: 'getW' : no matching overloaded function found.

The fixing uses the right get* methods. And uses int instead of uint as the parameter type.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2002)


qjia7:

FYI: I can't reproduce the performance regression we met before https://github.com/qjia7/tfjs-webgl2compute/pull/13 which I ever mentioned in our last meeting.  I am not sure if you ever did some changes in your current conv2d_mm method. Anyway it's a good news.

Issue #1996, opened 2019-09-07
Add relu6 op to core.


caisq:
To get help from the community, we encourage using Stack Overflow and the [`tensorflow.js`](https://stackoverflow.com/questions/tagged/tensorflow.js) tag.

#### TensorFlow.js version
1.2.9

23/26: mobilenet_v2_025_GraphModel
F

Failures:
1) TF.js Layers Benchmarks Benchmark models
  Message:
    Error: Activation: relu6 has not been implemented for the Node.js backend

kangyizhang:
hi @annxingyuan , can you add the relu6 op in tfjs-core? I'll add it in node once it's available in tfjs-core.

annxingyuan:
Hi @kangyizhang - yes I will add this to core - sorry I missed this issue until now!

cc @caisq

lewis617:
@annxingyuan 请问relu6加上了吗？我在"@tensorflow/tfjs": "^1.2.9"这个版本下，依然报错：
```
Error: Activation relu6 has not been implemented for the WebGL backend.
```

PR #2006, opened 2019-09-09
Add relu6 activation

kangyizhang:
fix: https://github.com/tensorflow/tfjs/issues/1996#event-2616415865

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2006)


kangyizhang:
No. Actually the relu6 op is not even supported in tfjs-core: https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/fused_util.ts#L21

caisq:
@kangyizhang In that case, I think we should add relu6 to tfjs-core first.

kangyizhang:
Closing this PR since the relu6 op need to be added in tfjs-core first.

PR #2013, opened 2019-09-11
webgpu: Use (16, 1, 1) as the work group size for binary and unary ops

qjia7:
With this change, the add benchmark has close to 100% speedup.
And PoseNet+ResNet demo improves from 3 fps to 6 fps.

We should try to avoid using (1, 1, 1) as the default work group size.

PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2013)


qjia7:

PR #2027, opened 2019-09-12
fix incorrect Erf op output

tedhtchang:
The Erf output should be between -1 and 1, but I got [-1.0235416, -7.3770967, 80427.2890625, 1.0000005] when I ran below code:
```
tf = require('@tensorflow/tfjs')
const x = tf.tensor1d([-1.4,-2.5,-3.1,-4.4]);
x.erf().print();
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2027)


tedhtchang:
Thank you for the quick merge!

Issue #1700, opened 2019-06-27
Implement greater and greaterEqual in WebGPU backend


nkreeger:
These are needed to get clip* ops to work properly. Should be mildly trivial.

axinging:
Hi, xing is willing to work on this.

PR #2028, opened 2019-09-12
Implement greater and greaterEqual

axinging:
Fix https://github.com/tensorflow/tfjs/issues/1700

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2028)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -21,6 +21,7 @@ import './flags_webgpu';
 
 import {backend_util, DataMover, DataType, ENV, KernelBackend, Rank, RecursiveArray, ShapeMap, Tensor, Tensor2D, Tensor3D, Tensor4D, TimingInfo, util} from '@tensorflow/tfjs-core';
 import * as shaderc from '@webgpu/shaderc';
+import * as webgpu_util from './webgpu_util';
 
 import {BufferManager} from './buffer_manager';
 import {ArgMinMaxProgram} from './kernels/argminmax_webgpu';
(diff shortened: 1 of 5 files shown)


math3d:
@annxingyuan PTAL.

axinging:
@annxingyuan  PTAL.

axinging:
Hi @annxingyuan,  I submit a fix for your comments.
But the checks reminds me that "Some checks were not successful". And I have no access right for the details failures.

annxingyuan:
Hi @axinging - it looks like there are 2 lint errors remaining. Try running `yarn lint` in the webgpu directory. The errors are:

ERROR: (prefer-const) /workspace/tfjs-backend-webgpu/src/backend_webgpu.ts[253, 11]: Identifier 'dataAsTypedArray' is never reassigned; use 'const' instead of 'let'.

ERROR: (triple-equals) /workspace/tfjs-backend-webgpu/src/backend_webgpu.ts[269, 20]: != should be !==

One more small change request - rather than adding a class method `convertInt32ToUint8`, could you add a method to `webgpu_util` called `float32ToTypedArray` that takes the result of `this.getBufferData` and converts it appropriately? This would make the WebGPU `read` method look more like the WebGL backend. 

Thank you!

axinging:
@annxingyuan , PTAL.
The input is ArrayBuffer, so I named it ArrayBufferToTypedArray instead of float32ToTypedArray.

annxingyuan:
Thank you @axinging !

By the way, would you mind downloading the [clang](https://clang.llvm.org/docs/ClangFormat.html) extension to your code editor? You can find more details about the linting extensions we use here: https://github.com/tensorflow/tfjs/blob/master/DEVELOPMENT.md (see the "Code editor" section)

These tools will auto-format your code upon save according to some style guidelines we've set. This way we won't have whitespace-related changes in the future if I edit the files you created.

axinging:
That's great! I will try.

PR #2032, opened 2019-09-12
Lazily instantiate the TextEncoder in platform_browser

esouthren:
Microsoft IE does not support TextEncoder in browser, so a workaround is needed to ensure it doesn't break tests which implement multiple browsers.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2032)

### DIFF
--- a/tfjs-core/src/platforms/platform_browser.ts
+++ b/tfjs-core/src/platforms/platform_browser.ts
@@ -19,16 +19,6 @@ import {ENV} from '../environment';
 import {Platform} from './platform';
 
 export class PlatformBrowser implements Platform {
-  private textEncoder: TextEncoder;
-
-  constructor() {
-
-    // Workaround for IE11 compatibility
-    // More info: https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder
-    if (!isIE()) {
-      this.textEncoder = new TextEncoder();
-    }
-  }
 
   fetch(path: string, init?: RequestInit): Promise<Response> {
     return fetch(path, init);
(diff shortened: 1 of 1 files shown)


esouthren:
@nsthorat 
> 
> > ```
> >     // Workaround for IE11 compatibility
> >     // More info: https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder
> >     if (!isIE()) {
> > ```
> 
> in general we try to do this kind of thing via feature testing, is it possible to check the existence of TextEncoder rather than checking the IE browser?

Previously I was doing `if (window.textEncoder) {` instead of  `if(!isIE())` (which I agree is cleaner), but this threw a `Variable window not found` error in the Safari tests. Any suggestions?

esouthren:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2032)_ status: 0 of 1 approvals obtained (waiting on @dsmilkov and @esouthren)
> 
> _[tfjs-core/src/platforms/platform_browser.ts, line 65 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2032#-LofPk7t5JcdYWe-NHeA:-LofPk7t5JcdYWe-NHeB:b-kvuehu) ([raw file](https://github.com/tensorflow/tfjs/blob/5afff7719c2fcb5ecfdd8529df561b87ca8bd3ba/tfjs-core/src/platforms/platform_browser.ts#L65)):_
> 
> > ```
> > if (ENV.get('IS_BROWSER')) {
> >   ENV.setPlatform('browser', new PlatformBrowser());
> > ```
> 
> The idea behind the design we did is so that can you register your own `PlatformIE` and reimplement the decode and encode with polyfills. Now, I see that because our library instantiates a PlatformBrowser() always, it causes a runtime error because we create a new TextEncoder in the c-tor of PlatformBrowser. Two things to address this problem:
> 
> * Move the creation of TextEncoder into the encode() method
> * As a user of IE, make a PlatformIE (in your use code)

Thanks for the help Daniel!

tfjs-core/src/platforms/platform_browser.ts
```diff
@@ -20,6 +20,14 @@ import {Platform} from './platform';
 
 export class PlatformBrowser implements Platform {
 
+  private textEncoder: TextEncoder;
+
+  constructor() {
+  // According to the spec, the built-in encoder can do only UTF-8 encoding.
+  // https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/TextEncoder
```
nsthorat: you can move this comment to the class field

tfjs-core/src/platforms/platform_browser.ts
```diff
@@ -33,7 +41,10 @@ export class PlatformBrowser implements Platform {
       throw new Error(
           `Browser's encoder only supports utf-8, but got ${encoding}`);
     }
-    return new TextEncoder().encode(text);
+    if (this.textEncoder == null) {
+      this.textEncoder = new TextEncoder();
+    }
+    return this.textEncoder().encode(text);
```
nsthorat: s/this.textEncoder()/this.textEncoder

PR #2034, opened 2019-09-12
Create a new tf.env() getter that returns the global ENV

nsthorat:
This is for the no exported objects rule which is against ES2015 module semantics.

This will be a replacement for anyone using tf.ENV as a global getter (we will not remove tf.ENV until the next major version bump).

We won't use exports = {} because it breaks the OSS build.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2034)


dsmilkov:
File a bug and assign it to the 2.0 project: https://github.com/tensorflow/tfjs/projects/10

nsthorat:
Issue already filed: https://github.com/tensorflow/tfjs/issues/2124

PR #2039, opened 2019-09-13
Bring back the dist importing lint rule

nsthorat:
This was lost during the move to the monorepo.

I removed dist importing where possible in packages. Now the custom lint rule runs for all packages (except data which doesn't share lint rules yet because it has a circular dependency).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2039)


nsthorat:
This failure seems to also be happening at head so I'm going to merge this.

PR #2045, opened 2019-09-17
[docs] Update setup instructions to reflect mandatory async-storage for React Native

kevinvangelder:
`async-storage` is mandatory as attempting to start a project without it installed gives a redbox error. Additional changes will need to be made to this library if it is going to become an optional dependency.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2045)


kevinvangelder:
@googlebot I signed it!

epinnock:
@kevinvangelder I'm trying to follow the setup instructions and had a question about installing asyncstorage since it requires linking would i not be able to use tfrn without ejecting my expo based projects?

kevinvangelder:
I don't use Expo, so I couldn't tell you, but I would expect Expo has Async Storage support through one of it's modules.

PR #2048, opened 2019-09-17
[WIP] webgpu: Add vec4 support

qjia7:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2048)


qjia7:
@annxingyuan @kainino0x  MatMulPackedV2Program is ported from https://www.ibiblio.org/e-notes/webgl/gpu/mul/intel.htm with Intel SLM_8x8_4x16 shader which is supposed the fastest one on windows. However, I can't see much improvement on mac device. So I did some adjustments for the tile size and work per thread. Currently, this version is better than MatMulPackedV2Program on my device with matmul (500, 500) (`6~7ms vs 8~9ms`). However, it becomes worse when the matrix increase to (1000, 1000).
Any comments on this? Do you want me to continue this way?

annxingyuan:
@qjia7 Thanks for doing this investigation! In general we're happy to compile different versions of shaders for different inputs - so if this is the fastest way to compute 500,500 matMul then we should use it. 

It would be great to better understand the exact conditions for improvement, and why the improvement disappears for larger dimensions. Do you have a sense for that? Is it a matter of tweaking the parameters for 1000,1000?

Issue #1987, opened 2019-09-06
Fallback to cpu when the inputs size is small for WebGPU backend


qjia7:

qjia7:
@annxingyuan FYI

annxingyuan:
Thanks @qjia7 ! By the way, if you wouldn't mind adding the 'webgpu' label and 'WebGPU' project board to these issues that would be awesome :)

qjia7:
>By the way, if you wouldn't mind adding the 'webgpu' label and 'WebGPU' project board to these issues that would be awesome

@annxingyuan I don't have the permission to change it. Is it ok to give me those permission?

annxingyuan:
@qjia7 Ah I see... in that case I will apply the labels! Thanks!

PR #2061, opened 2019-09-18
Fallback to cpu when the inputs size is small

axinging:
Fix https://github.com/tensorflow/tfjs/issues/1987.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2061)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -19,7 +19,7 @@
 
 import './flags_webgpu';
 
-import {backend_util, DataStorage, DataType, ENV, KernelBackend, Rank, RecursiveArray, ShapeMap, Tensor, Tensor2D, Tensor3D, Tensor4D, TimingInfo, util} from '@tensorflow/tfjs-core';
+import {backend_util, DataStorage, DataType, ENV, findBackend, KernelBackend, Rank, RecursiveArray, ShapeMap, Tensor, Tensor2D, Tensor3D, Tensor4D, TimingInfo, util} from '@tensorflow/tfjs-core';
 // TODO(annxingyuan): get ENGINE from tf.engine() once core 1.3.0 is released.
 // tslint:disable-next-line: no-imports-from-dist
 import {ENGINE} from '@tensorflow/tfjs-core/dist/engine';
@@ -81,6 +81,10 @@ export interface WebGPUTimingInfo extends TimingInfo {
   downloadWaitMs: number;
 }
 
+// Empirically determined constant used to determine size threshold for handing
+// off execution to the CPU.
+const CPU_HANDOFF_SIZE_THRESHOLD = 128;
+
 const DEFAULT_GPUBUFFER_USAGE =
     GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST;
 
@@ -107,6 +111,7 @@ export class WebGPUBackend extends KernelBackend {
   private activeTimers: TimerNode[];
   private uploadWaitMs = 0;
   private downloadWaitMs = 0;
+  private cpuBackend: KernelBackend;
 
   constructor(device: GPUDevice, shaderc: shaderc.Shaderc) {
     super();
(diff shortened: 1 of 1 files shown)


axinging:
@annxingyuan , PTAL.

BTW, I applied clang-format auo format on this changed file.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -78,6 +78,10 @@ export interface WebGPUTimingInfo extends TimingInfo {
   downloadWaitMs: number;
 }
 
+// Empirically determined constant used to determine size threshold for handing
+// off execution to the CPU.
+const CPU_HANDOFF_SIZE_THRESHOLD = 128;
```
annxingyuan: Just curious - did you test this constant? Is 128 optimal? Is it approximately the same across different ops?

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -476,6 +481,34 @@ export class WebGPUBackend extends KernelBackend {
...
+
+    if (this.cpuBackend == null) {
+      this.cpuBackend = findBackend('cpu');
+    }
+
+    return this.cpuBackend;
+  }
+
+  /*
+  Tests whether all the inputs to an op are small and on the CPU. This heuristic
```
annxingyuan: Let's remove this comment.

tfjs-backend-webgpu/src/backend_webgpu.ts
axinging: No, I didn't have any data about if 128 is good enough for webgpu. 128 is from WebGL backend.

And, I think this fallback may help unittest or benchmark test. But not real model. Below is my explanation for this. 
If run by yarn test,  "this.tensorMap.get(input.dataId).bufferInfo.buffer == null" will  always be true. This means when we run single unit test, we have a chance to fallback to CPU and get a better performance (So does benchmark test).
When run posnet(though it failed after a short time due to clip not implemented), "this.tensorMap.get(input.dataId).bufferInfo.buffer == null" always return false.  This means, for posnet model, it has no chance to fallback to CPU.

I am not sure if this is the expected behavioural. And did you notice this on WebGL (this.texData.get(input.dataId).texture == null)?

annxingyuan: @axinging Hmm, it is not expected for the buffer to always be null in the context of a model. Let me investigate this. You're right that if that's the case, the model does not have a chance to benefit from cpu forwarding.

axinging: Thanks @annxingyuan. And I have another thought about this issue.
For a model with three ops: op1 => op2 => op3. op1's output is op2's input. So does op2 and op3. op1 and op3 run in GPU mode, op2 fallback to CPU mode.

In order to fallback to CPU,  op1's GPU data should be copied into CPU for op2. When op2 is done, op2's CPU data should be copied into GPU for op3. 

For this simple scenario,  there are two copies: one from CPU to GPU, another is from GPU to CPU (And in TFJS's code, I didn't find the copies when fallback. Am I missing something?). We know that copies between CPU and GPU are really time consuming. We should evaluate the fallback mechanism with all these copies.

annxingyuan:
Hi @axinging - we've just merged https://github.com/tensorflow/tfjs/pull/2085 which enables lazy tensor uploading for WebGPU. This fixes the issue you pointed out that WebGPU buffers are never null.

Now that this fix is in, would you mind rebasing master and then investigating where the threshold should be for CPU forwarding, and whether the threshold is different across ops? Afterward that I'll be happy to merge.

axinging:
@annxingyuan,  rebase is done.
As to different threshold across ops, I think these thresholds are also different across different CPU/GPU.
A faster CPU, a slower GPU, the threshold may be bigger, and vice verse.

annxingyuan:
@axinging fair enough!

qjia7:
@axinging With this change, I have 4 tests regression in backend_webgpu_test.ts. Please have a look.

annxingyuan:
@qjia7 ah thank you for pointing this out - I think the behavior is correct we just need to adjust the tests to account for CPU forwarding - I will submit a patch soon.

annxingyuan:
@qjia7 FYI the regressions have been resolved with https://github.com/tensorflow/tfjs/pull/2176

PR #2064, opened 2019-09-18
ArrayBuffer should not be deep cloned

kangyizhang:
deepClone() is added in https://github.com/tensorflow/tfjs-data/commit/9d123814d30aca380bdcc35682df947997488554, so that when creating a Dataset from nested object with existing tensors, the dataset keeps working after user disposing the original tensor.

```
const ds = tf.data.array({a:tensorA, b:tensorB});
tensorA.dispose();
tensorB.dispose();
const result = ds.toArray();
```
But ArrayBuffer is treated as iterable object and got deep cloned. So a Float32Array is converted to {0:xxx, 1:xxx, 2:xxx ...}. This PR prevent from deep cloning ArrayBuffer.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2064)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/depthwise_conv2d_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/depthwise_conv2d_webgpu.ts
@@ -94,7 +94,7 @@ export class DepthwiseConv2DProgram implements WebGPUProgram {
 
             float xVal = getX(batch, xR, xC, d1);
             float wVal = getW(wR, wC, d1, q);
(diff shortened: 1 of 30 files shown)


kangyizhang:
cc @tafsiri  since this might be related.

tafsiri:
@kangyizhang ah interesting, thanks for the pointer. Different cause from what i was seeing.

dsmilkov:
Thank you for the quick fix Kangyi and unblocking the Teachable Machine team

PR #2087, opened 2019-09-24
webgpu: Fix argMax benchmark doesn't work

qjia7:
BUG

Fix below error message when you run 'yarn benchmark':
  TypeError: Cannot read property 'data' of undefined

The error is because we want to test 50 sets of data. However, we used the
default 'reps' which is 1, that will result only the first element of
maxes is valid. So we will get above error message for the following
elements in maxes.
Meanwhile, decrease n to 2 to avoid out of time.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2087)

### DIFF
--- a/tfjs-backend-webgpu/src/benchmark_ops_test.ts
+++ b/tfjs-backend-webgpu/src/benchmark_ops_test.ts
@@ -79,7 +79,7 @@ describeWebGPU('Ops benchmarks', () => {
   }
 
   it('argMax', async () => {
-    const n = 50;
+    const n = 2;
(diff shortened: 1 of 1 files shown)


qjia7:
I am not sure if I should set n = 1 since other cases all used default value reps = 1.
BTW, it seems that the bots don't run those tests and catch the failure. I notice there is another failure in fromPixels when running 'yarn test'.

annxingyuan:
@qjia7 our testing system doesn't include WebGPU yet - we will include it once WebGPU is available by default on regular Chrome.

PR #2090, opened 2019-09-25
Public API to get the list of backend names

Lewuathe:
Although the engine has a method to return the list of available backends, there is no public API to access it. The list of available backends is important information to decide the preferable backend manually at runtime.


tvanier:
@Lewuathe thoughts about exposing `getSortedBackends`? https://github.com/tensorflow/tfjs/pull/4491

PR #2094, opened 2019-09-26
Only call register_prelu_op with open-source TensorFlow 1.X

superbobry:
`register_prelu_op` is a workaround for the MetaGraph validity check which requires all GraphDef ops to be registered. This check has been lifted in tensorflow/tensorflow@8b287e23.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2094)

### DIFF
--- a/tfjs-converter/python/tensorflowjs/converters/fuse_prelu.py
+++ b/tfjs-converter/python/tensorflowjs/converters/fuse_prelu.py
@@ -33,9 +33,12 @@ from tensorflow.python.framework import op_def_registry
 
 from tensorflowjs.converters import common
 
+# The function is only needed for TensorFlow 1.X and 2.0. Remove once tfjs
+# no longer depends on these versions.
 def register_prelu_op():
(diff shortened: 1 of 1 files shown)


tfjs-converter/python/tensorflowjs/converters/fuse_prelu.py
```diff
@@ -197,4 +200,6 @@ def prelu_fn(*args):
   with graph.as_default():
     prelu_fn(tf.constant(1.0), tf.constant(1.0))
 
-register_prelu_op()
+
+if tf.version.GIT_VERSION != 'unknown' and tf.__version__ <= '2.1':
```
dsmilkov: This includes both 2.0 and 2.1. Is that intentional?

superbobry: Nope, changed to <. Thanks for spotting that!

dsmilkov:
I'm glad. Thanks for sending the fix on Github.

PR #2098, opened 2019-09-26
Annotated code snippets for Cloud docs

dsmilkov:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2098)


fhinkel:
In my opinion, this repo is the right place to publish this sample. The [nodejs-docs-samples repo](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/) is for samples that don't belong to a specific client library. But this sample only uses `tfjs`, I think this is where it fits.

PR #2102, opened 2019-09-27
Update tfjs-core to 1.2.10.

annxingyuan:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2102)

### DIFF
--- a/tfjs-core/package.json
+++ b/tfjs-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow/tfjs-core",
-  "version": "1.2.9",
+  "version": "1.2.10",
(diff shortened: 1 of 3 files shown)


annxingyuan:
@dsmilkov I've prepared a PR to address the node integration test failure: https://github.com/tensorflow/tfjs/pull/2119

#2119 depends on tfjs-core 1.2.10, and this PR will continue to fail CI without #2119. What do we do in these cases?

dsmilkov:
Thanks Ann. We force submit this PR in that case. (Previous when we were using Travis, we marked the test-integration part of the CI as optional so it doesn't block the PR), however Cloud Build lacks that functionality)

PR #2104, opened 2019-09-27
Update integration script to cd into subprojects rather than cloning them.

annxingyuan:
Now that we have a monorepo there's no need to clone projects we can just cd into them.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2104)

### DIFF
--- a/tfjs-core/scripts/test-integration.sh
+++ b/tfjs-core/scripts/test-integration.sh
@@ -26,32 +26,28 @@ echo 'Testing layers/converter/node/data against tfjs-core@master.'
 echo '######################'
 yarn build && yarn build-test-snippets && yarn yalc publish
 
-echo 'Cloning layers'
-git clone https://github.com/tensorflow/tfjs-layers.git --depth 1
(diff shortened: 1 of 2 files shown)


dsmilkov:
Can you do this change in the PR where you touch version.ts in core - otherwise you are not testing that this change works.

tfjs-core/scripts/test-integration.sh
```diff
@@ -26,32 +26,28 @@ echo 'Testing layers/converter/node/data against tfjs-core@master.'
 echo '######################'
 yarn build && yarn build-test-snippets && yarn yalc publish
 
-echo 'Cloning layers'
-git clone https://github.com/tensorflow/tfjs-layers.git --depth 1
+echo 'Testing layers'
 cd tfjs-layers
```
dsmilkov: I expected cd ../tfjs-layers (going to the parent first) but we’ll know for sure once you edit version.ts and let the CI run

Issue #1908, opened 2019-08-24
Cannot save then load LSTM layers model


jungwook:
To get help from the community, we encourage using Stack Overflow and the [`tensorflow.js`](https://stackoverflow.com/questions/tagged/tensorflow.js) tag.

#### TensorFlow.js version 
1.2.7

#### Browser version 
N/A. using node

#### Describe the problem or feature request
Cannot load a saved model with LSTM layers.

#### Code to reproduce the bug / link to feature request

    const model = tf.sequential();
    model.add(tf.layers.dense({units: 1, inputShape: [10, 20]}));
    const cells = [
        tf.layers.lstmCell({units: 4}),
        tf.layers.lstmCell({units: 8}),
    ];
    const rnn = tf.layers.rnn({cell: cells, returnSequences: true});
    model.add(rnn);
    
    model.save('file://test1').then( _ =>{
        tf.loadLayersModel('file://test1/model.json').then(m2=>{
            console.log(m2);
        })
    })  

Then you get something like the following.

(node:20145) UnhandledPromiseRejectionWarning: Error: Duplicate weight name: rnn_RNN1/kernel at new ValueError (/Users/jungwook/Documents/Development/Tensor/node_modules/@tensorflow/tfjs-layers/dist/errors.js:68:28)at Sequential.Container.loadWeights (/Users/jungwook/Documents/Development/Tensor/node_modules/@tensorflow/tfjs-layers/dist/engine/container. at /Users/jungwook/Documents/Development/Tensor/node_modules/@tensorflow/tfjs-layers/dist/models.js:302:27 at step (/Users/jungwook/Documents/Development/Tensor/node_modules/@tensorflow/tfjs-layers/dist/models.js:54:23) at Object.next (/Users/jungwook/Documents/Development/Tensor/node_modules/@tensorflow/tfjs-layers/dist/models.js:35:53) at fulfilled (/Users/jungwook/Documents/Development/Tensor/node_modules/@tensorflow/tfjs-layers/dist/models.js:26:58)

jungwook:
Still not working. I’m using the latest version of tfjs but same error persists

heitorrubira:
Hey guys, hope to contribute a little as well!

Version: `"@tensorflow/tfjs-node": "^1.2.8"`
Error:
```
Error: Duplicate weight name: rnn_RNN1/kernel
[...]

PR #2105, opened 2019-09-27
[layers] Fix RNNCell weight serialization

caisq:
The root cause of the problem is that the `originalName`s of the weights of different RNN cells clashed.
This PR fixes that by using a unique namescope for each cell in a stack of RNN cells.

Fixes https://github.com/tensorflow/tfjs/issues/1908

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2105)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,9 @@
 {
   "devDependencies": {
-    "clang-format": "~1.2.4",
+    "@types/node": "^12.7.5",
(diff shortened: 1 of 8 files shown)


caisq:
@davidsoergel Merged the PR. If you have more comments, please let me know and I'll be happy to address them in follow-ups.

PR #2106, opened 2019-09-27
Fix an internal tfjs conversion failure for prelu fusing

pyu10055:
Since the DeFun does not allow specifying node attr, when you don't register the op through register_op_list mothod, it will cause importer to failed on the mismatch of node def and the DeFun def.

Here removed the 'T' attr for prelu, since we are not using it anyway.
Changed the version check for registering op list with method existence check, since the version check is not robust enough to differentiate the nightly TF that integration test is depending on.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2106)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,9 @@
 {
   "devDependencies": {
-    "clang-format": "~1.2.4",
+    "@types/node": "^12.7.5",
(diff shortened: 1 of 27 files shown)


caisq:
cc myself

Issue #510, opened 2018-07-12
Intel GPU support for Celeron/Baytrail


justadudewhohacks:
Not sure if tfjs is going to support Intel GPUs, but it turns out that some ops seem to return different results on an Intel GPU. (I am referring to [this](https://github.com/justadudewhohacks/face-api.js/issues/43) issue).

Unfortunately I don't have access to an Intel GPU anymore, otherwise I could try to find out where this issue occurs, but I thought it would still be worth reporting it.

caisq:
cc @nsthorat @dsmilkov for comments.

dsmilkov:
Looks like some math operations are departing from the expected result. Unfortunately, we don't have an intel gpu at the moment, but if anyone with input gpu is interested in helping us debug this, we can share some pointers for how to debug.

seranus:
@dsmilkov I'll give it a go it became an issue for me since I can't warn my users if the site won't work

seranus:
@dsmilkov karma failed tests output on latest tfjs-core build

> Chrome 67.0.3396 (Linux 0.0.0) gramSchmidt-non-tiny test-webgl2 {"WEBGL_VERSION":2} 32x512 FAILED
	Error: Arrays differ: actual[0] = 0.9943403601646423, expected[0] = 1.
	    at <Jasmine>
	    at Object.expectArraysClose (src/test_util.ts:95:12 <- src/test_util.js:68:19)
	    at UserContext.<anonymous> (src/ops/linalg_ops_test.ts:93:4 <- src/ops/linalg_ops_test.js:56:21)
	    at <Jasmine>


> Chrome 67.0.3396 (Linux 0.0.0) transpose test-webgl1 {"WEBGL_VERSION":1} 5D [r, c, d, e, f] => [r, c, f, e, d] FAILED
	Error: Arrays differ: actual[1] = 1, expected[1] = 3.
	    at <Jasmine>
	    at Object.expectArraysClose (src/test_util.ts:95:12 <- src/test_util.js:68:19)
	    at UserContext.<anonymous> (src/ops/transpose_test.ts:89:4 <- src/ops/transpose_test.js:58:21)
	    at <Jasmine>


> Chrome 67.0.3396 (Linux 0.0.0) transpose test-webgl2 {"WEBGL_VERSION":2} 5D [r, c, d, e, f] => [r, c, f, e, d] FAILED
	Error: Arrays differ: actual[1] = 1, expected[1] = 3.
	    at <Jasmine>
[...]

PR #2116, opened 2019-09-29
[tfjs-core] attempt to fix bugs related to sampling and output coordi…

xianrenb:
…nates, matmul, sum, argmax/argmin and decode

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2116)

### DIFF
--- a/tfjs-core/src/backends/webgl/argminmax_gpu.ts
+++ b/tfjs-core/src/backends/webgl/argminmax_gpu.ts
@@ -34,8 +34,8 @@ export class ArgMinMaxProgram implements GPGPUProgram {
     this.outputShape = [batchSize, outSize];
     const compOp = (op === 'max') ? '>' : '<';
     const indexSnippet = firstPass ?
-        'inOffset + i;' :
-        'round(getBestIndicesA(batch, inOffset + i));';
+        'inOffset + i' :
+        'round(getBestIndicesA(batch, inOffset + i))';
 
     this.userCode = `
       void main() {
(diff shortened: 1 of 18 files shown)


xianrenb:
This is my attempt to fix some problems.
My view may be incorrect, but I believe that resultUV from the vertex shader ranges from 0.0 to 1.0, so the boundary pixel may get 1.0, output coordinates should be adjusted as so.
On the other hand, the sampler should not use maximum value of 1.0.

[https://www.khronos.org/opengl/wiki/Rectangle_Texture](https://www.khronos.org/opengl/wiki/Rectangle_Texture)

> When using rectangle samplers, all texture lookup functions automatically use non-normalized texture coordinates. This means that the values of the texture coordinates refer specifically to texels in texel space. The value 12.5 is halfway between the thirteenth and fourteenth texel (remember: 0-base indices).

Therefore, I believe halfCR is not required and I added adjuv() to adjust the value.

xianrenb:
After reading [https://stackoverflow.com/questions/7272579/iphone-opengl-es-2-0-pixel-perfect-textures/7272871#7272871](https://stackoverflow.com/questions/7272579/iphone-opengl-es-2-0-pixel-perfect-textures/7272871#7272871), I think that I was wrong.
"halfCR" should be used.

xianrenb:
Based on the results of previous tests, I think my previous view may not be totally incorrect.
I think there are actually two kinds of WebGL implementations.
One is assuming texture pixel centers being integers, and the other assuming pixel centers being halfway between integers.
Therefore, I updated the code to use vec2(0.25) to adjust instead.

nsthorat:
@xianreb is there an issue that this is fixing or can you explain what you're seeing that this resolves?

xianrenb:
@nsthorat Originally, I found the GPU backend accuracy/performance was not matching CPU backend on my computer. I started analyzing the tfjs-core code and found there may be some potential problems and tried to fix that. I am not sure whether these changes are actually fixing these potential problems or not, but the new code did pass the tests. Maybe some of them are unnecessary or totally irrelevant.

Later, I found that the main cause was the max texture size as mentioned in [https://github.com/tensorflow/tfjs/issues/510](https://github.com/tensorflow/tfjs/issues/510). Either the GPU reported the wrong max texture size, or it did not work correctly when max texture size was used.

jenperson:
This PR is being closed because it hasn't been updated in 3 years.

PR #2118, opened 2019-09-30
webgpu: Implement clip,less ops

qjia7:
FEATURE

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2118)


qjia7:
clip is needed by MobileNet.

tfjs-backend-webgpu/src/kernels/clip_webgpu.ts
```diff
@@ -0,0 +1,53 @@
...
+  variableNames = ['A'];
+  dispatchLayout: {x: number[]};
+  dispatch: [number, number, number];
+  workGroupSize: [number, number, number] = [64, 1, 1];
+
+  constructor(outputShape: number[], minVal: number, maxVal: number) {
+    this.outputShape = outputShape;
+    const size = util.sizeFromShape(this.outputShape);
+    this.dispatchLayout = flatDispatchLayout(this.outputShape);
+    this.dispatch = computeDispatch(
```
annxingyuan: @qjia7 Have you tried profiling clip with different `workGroupSize` values? What about increasing the `workPerThread` (defaults to `1` if not set)? 

You could look at `pad_webgpu` for an example of `workPerThread` > 1. Curious how these parameters potentially tradeoff with each other in terms of performance.

annxingyuan:
@qjia7 Thanks for adding these ops! Happy to merge once we understand the performance implications of `workPerThread` > 1, and how performance changes for different values of `workGroupSize`.

qjia7:
@annxingyuan If we increase the work group size to a proper value, `workPerThread` >1 won't help performance much. I didn't see much difference between `workPerThread` > 1 and `workPerThread` = 1 when workGroupSize = (64, 1, 1). The fluctuation is within the error range. For the new added 'clip' benchmark, workPerThread = 1 is a little better than workPerThread = 4 on my machine. For my understanding, workPerThread won't help much if we don't use shared memory or the parallelism is already good. I think in your previous implementation, seeing `workPerThread` > 1 helped a lot is because the workGroupSize is (1, 1, 1) which results the parallelism is really poor.

But I think support workPerThread is still needed which can simply adjust its value when necessary.

Issue #2130, opened 2019-10-02
Feature request: `tf.spectral.fft2d`


bourdakos1:
A 2-dimensional fft would be helpful for some image processing use-cases. One example would be for object tracking. I can work on adding this op.

gaikwadrahul8:
Hi, @bourdakos1 

Thank you for opening this issue. Since this issue has been open for a long time, the code/debug information for this issue may not be relevant with the current state of the code base.

The TFJs team is constantly improving the framework by fixing bugs and adding new features. We suggest you try the latest TFJs [version](https://github.com/tensorflow/tfjs/releases/tag/tfjs-v4.10.0) with the latest compatible hardware configuration which could potentially resolve the issue. If you are still facing the issue, please create a new GitHub issue with your latest findings, with all the debugging information which could help us investigate.

Please follow the [release notes](https://github.com/tensorflow/tfjs/blob/master/README.md) to stay up to date with the latest developments which are happening in the Tensorflow.js space.

Thank you for your support and cooperation.

PR #2131, opened 2019-10-02
Add ff2d op

bourdakos1:
Add `tf.spectral.fft2d` op.

Usage:
```js
const real = tf.tensor2d([1, 2], [1, 2]);
const imag = tf.tensor2d([1, 2], [1, 2]);
const x = tf.complex(real, imag);

x.fft2d().print();  // tf.spectral.fft2d(x).print();
```

See #2130

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2131)


tfjs-core/src/backends/cpu/backend_cpu.ts
```diff
@@ -3442,6 +3446,59 @@ export class MathBackendCPU implements KernelBackend {
     }
   }
 
+  private fft2dBatch(x: Tensor3D): Tensor3D {
```
Lewuathe: If the input length is the exponent of 2, I think we can use `fftRadix2` which is faster by making use of the divide and conquer approach.

tfjs-core/src/ops/spectral_ops.ts
```diff
@@ -56,6 +56,50 @@ function fft_(input: Tensor): Tensor {
...
+/**
+ * @doc {heading: 'Operations', subheading: 'Spectral', namespace: 'spectral'}
+ */
+function fft2d_<T extends Tensor2D|Tensor3D>(input: T): T {
+  let input3D = input as Tensor3D;
+  let reshapedTo3D = false;
+
+  if (input.rank === 2) {
+    reshapedTo3D = true;
+    input3D = input.as3D(1, input.shape[0], input.shape[1]);
```
Lewuathe: If the input is 2d rank, we can just pass it to `fft`?

bourdakos1: I believe passing it to `fft` would just be the fft across the rows. Where _(I think)_ `fft2d` is the equivalent to fft across the rows, then columns. But definitely `fft(tensor2d) != fft2d(tensor2d)`

tfjs-core/src/backends/cpu/backend_cpu.ts
bourdakos1: yes, that would be faster, I can try to add that :)

tfjs-core/src/ops/spectral_ops.ts
bourdakos1: I looked into it and `fft2d` is equivalent to `transpose(fft(transpose(fft(x))))`

tfjs-core/src/backends/cpu/backend_cpu.ts
bourdakos1: I changed it so that `fft2d` falls back to using `fft` across rows then columns, which should end up using `fftRadix2` if the input length is exponent of 2.

bourdakos1: I removed the use of `fftRadix2`, it's much slower (~x10) than the naïve `fft2d`. I can try to implement an `fft2dRadix2` implementation, but that might take me awhile.

bourdakos1:
@Lewuathe I’m having difficulties implementing a radix2 fft2d, is this a blocker? Is there anything else I should do?

bourdakos1:
@Lewuathe Anything else I need to change?

oeway:
Hi, thanks for working on this. I would like to use the fft2d op and would be super happy to see this PR merged :)

bourdakos1:
@Lewuathe @dsmilkov is there anything else I should do/fix?

bourdakos1:
Should I just close this? 😢

PR #2132, opened 2019-10-02
[tfjs-react-native] add decodeJpeg function

### LINKED ISSUE #2071: [react native] implement custom decodeJpeg

To make handling images in react native easier we should implement a [decodeJpeg](https://www.tensorflow.org/api_docs/python/tf/io/decode_jpeg) utility (similar to [tf.node.decodeJpeg](http://localhost:8080/api_node/1.2.8/#node.decodeJpeg)). One difference from the regular decodeJpeg signature is that we should allow a `ImageResolvedAssetSource` to be passed in as the first parameter. This enables flexibility of doing the byte reading and decoding on the native thread without going across the bridge (though this would be an internal implementation detail). It would also take a byte array if the bytes for the image are already in the js process.

Proposed signature (just the first 3 arguments)

```ts
decodeJpeg(contents: ImageResolvedAssetSource|Uint8Array|ArrayBuffer, channels?:0|1|3, ratio?: 1|2|4|8)
```

kevinvangelder:
Closes #2071

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2132)


kevinvangelder:
@tafsiri Sorry for the delay. I was finally able to get all your feedback implemented and the tests passing.

kevinvangelder:
@tafsiri Lint and tests are now passing.

kevinvangelder:
@tafsiri Updated with requested changes.

PR #2136, opened 2019-10-02
Add compiled proto_pb.js into tfjs-node and test it 

kangyizhang:
1. Compiled api_pb.js from saved_model.proto, same as the proto file used in [tfjs-converter](https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/src/data/api.proto). 

2. Added a test to checking the deserialization of SavedModel binary file. Validating metaGraph tag and signatureDefs.

3. Renamed test_images folder to test_objects and added the SavedModel (times_three_float) in the folder.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2136)


kangyizhang:
Friendly ping.

PR #2141, opened 2019-10-03
Add RN integratation tests (with core) to CI

tafsiri:
This change adds on device RN integration tests to our CI (currently on android).

It is triggered in two conditions:
 - A change to code in tfjs-react-native. This will use the version specified by the package, _however, in this PR it builds against head. This is to deal with the chicken-egg problem of needing changes to core to get these to run. A future PR will set this back to build against the published package_
 - A change to code in tfjs-core. This will build RN against head tfjs-core and run the tests.

There is one step remaining which is not in this PR, and could come in a subsequent PR (or could be added to this one), which is a need for a periodic job to copy pre-built native app resources from google cloud to browserstack.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2141)


tafsiri:
@dsmilkov I figured out the kernel registration failures we discussed earlier. The problem was that in the new scheme kernels are registered under the backend name 'webgl'. However tfjs-react-native registers a new backend name 'rn-webgl'. Solution is to register all webgl kernels under the rn-webgl name as well. Note that packages like tfjs-wechat will need to do the same thing once they upgrade to a version of tfjs that uses the kernel registry. 

This is now ready for re-review and appears to be passing CI reliably.

PR #2159, opened 2019-10-07
[WASM] Add XNNPACK as external dependency

dsmilkov:
Add XNNPACK as external dependency and initialize the library when we initialize the WASM module.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2159)


dsmilkov:
Failure on CI is unrelated -- 7 tests on Android timeout for tfjs-core. Will force merge.

PR #2178, opened 2019-10-09
[tfjs-backend-wasm] Implement prelu by calling XNNPack. Add cc tests

nsthorat:
- Implement's the prelu kernel in prelu.cc
- Calls XNNPack prelu kernel.
- Caches the xnn_operator_t on a map that maps the weights_id to the xnn_operator_t
- Creates a disposal registry mechanism that allows kernel authors like prelu to get called back when a tensor is disposed.
- Use the disposal registry for xnn_operator_t that are mapped from weights_ids.
- Adds cc unit tests to test this logic (no kernels will get correctness tests as they come from JavaScript).
- Adds bazel rules for tfjs_cc_libraries to force static linking at compile time.
- Introduce tfjs::wasm namespace for testing purposes (anything that gets exposed to JavaScript lives under this namespace).
- Add `yarn test-cc` to test the C++ unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2178)


tfjs-backend-wasm/src/cc/prelu.cc
```diff
@@ -0,0 +1,88 @@
...
+#include "src/cc/util.h"
+
+const int kBlockSize = 48;
+
+namespace {
+// The operator cache maps the weights id to the xnn_operator_t instantiated for
+// // this set of weights.
+std::map<int, xnn_operator_t> operator_cache;
+
+void delete_xnn_operator(int weights_id) { operator_cache.erase(weights_id); }
```
Maratyszcza: You should call `xnn_delete_operator`, or you'll have a memory leak.
Alternatively, you may use smart pointers for XNNPACK operators, so they are deleted automatically. See https://github.com/google/XNNPACK/blob/master/test/convolution-operator-tester.h#L548-L549

nsthorat: Done, thanks for the catch! I can't use smart pointers here because the xnn_operator lives on a global map outside the scope of the kernel function.

tfjs-backend-wasm/src/cc/prelu.cc
```diff
@@ -0,0 +1,88 @@
...
+  const TensorInfo x_info = backend::get_tensor_info(x_id);
+  const TensorInfo weights_info = backend::get_tensor_info(weights_id);
+  const TensorInfo out_info = backend::get_tensor_info(out_id);
+
+  const float* x_buf = x_info.buf.f32;
+  const float* weights_buf = weights_info.buf.f32;
+  float* out_buf = out_info.buf.f32;
+
+  xnn_operator_t prelu_op = nullptr;
+  if (operator_cache.count(weights_id) == 0) {
```
Maratyszcza: Would be slightly more efficient to do `operator_cache.find`, then compare to `operator_cache.end()` to detect if the operator is in cache. With current implementation, you typically search the cache twice: once in `operator_cache.count(...)`, and another time in `operator_cache.at(...)`.

tfjs-backend-wasm/src/cc/prelu.cc
```diff
@@ -0,0 +1,88 @@
...
+
+#include "src/cc/backend.h"
+#include "src/cc/util.h"
+
+const int kBlockSize = 48;
+
+namespace {
+// The operator cache maps the weights id to the xnn_operator_t instantiated for
+// // this set of weights.
+std::map<int, xnn_operator_t> operator_cache;
```
Maratyszcza: `std::unordered_map` would be better: `std::map` is typically implemented as a search tree, but `std::unordered_map` is a hash map.

tfjs-backend-wasm/src/cc/prelu.cc
Maratyszcza: You can still use smart pointers, then `operator_cache.erase` would automatically delete the operator.

PR #2236, opened 2019-10-18
Underflow safe norm

### LINKED ISSUE #895: tf.norm and tf.abs(_: complex64) are not underflow-safe.
To get help from the community, check out our [Google group](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs).

#### TensorFlow.js version
v0.13.8

#### Browser version
Any

One of the most important norm properties is the positive definiteness:
`|x|=0 <=> x=0`
The current implementation of `tf.abs(_: complex64)` and `tf.norm` do not satisfy that property for some fringe cases where underflow becomes an issue. `Math.fround(1e-23**2)` rounds down to `0` and therefore the norm would become zero. This can cause issues, e.g. when normalizing a complex number or a vector, or when trying to determine convergence of an optimization.

An underflow-safe implementation can be found [here](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot). In WebGL there is a [length](https://www.khronos.org/files/opengles_shading_language.pdf#page=74) function that can be used both for `tf.abs` as well for `tf.norm` as binary reduction operator.

```ts
const x = tf.tensor1d([1e-23,0]);
x.norm().print();
// Output:
//   Tensor
//       0
```

DirkToewe:
Fixes #895 by making `tf.norm()` under-/overflow-safe.

### DIFF
--- a/tfjs-core/src/ops/norm.ts
+++ b/tfjs-core/src/ops/norm.ts
@@ -15,14 +15,17 @@
  * =============================================================================
  */
 
+import {ENGINE} from '../engine';
 import {Tensor} from '../tensor';
 import {convertToTensor} from '../tensor_util_env';
 import {TensorLike} from '../types';
 import {parseAxisParam} from '../util';
 
 import * as axis_util from './axis_util';
+import {div} from './binary_ops';
+import {where} from './logical_ops';
 import {op} from './operation';
-import {scalar} from './tensor_ops';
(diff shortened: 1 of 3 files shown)


dsmilkov (commented):
I'm no longer working on this. Please request another reviewer. Thanks!

Issue #2011, opened 2019-09-10
support Lgamma and BroadcastTo Ops in order to support probabilistic models  (Poisson distribution)


lukasheinrich:
When trying to covnert a model that uses `tensorflow_probability.distributions.Poisson` I get the error

```
ValueError: Unsupported Ops in the model before optimization
Lgamma, BroadcastTo
```

In order to support probabilistic models using those libraries itwould be great if those Ops could be added

@kratsg @matthewfeickert

https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/lgamma
https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/broadcast-to

dsmilkov:
Hi Lukas,

Can you share a bit about your model/use-case? There are a long tail of ops that TF.js is missing and we are trying to prioritize based on use-case.

lukasheinrich:
hi @dsmilkov,

our specific use case is the computation of probabilistic models for high energy physics models at CERN, but I think being able to compute a Poisson distribution is probably of much broader appeal for any model using tensorflow probability.

DirkToewe:
The rather monstrous [PR#1356](https://github.com/tensorflow/tfjs-core/pull/1356) contained a `broadcastTo` [implementation](https://github.com/tensorflow/tfjs-core/pull/1356/files#diff-b62631f91d28d56ddc898bc41513df9cR106-R156). If You want, I can take it out and submit it as PR to the monorepo.

lukasheinrich:
hi @DirkToewe I think that would be awesome and maybe easier to merge. Do you know anytone working on lgamma?

DirkToewe:
No idea sorry. A `float32/complex64` implementation of lgamma(z) for `Re(z) > 0` can be found in:
```
"Numerical Recipes in C",
2nd edition,
chapter 6, page 213f.
```
It's C code but fairly easy to read and understand. Should be fairly easy to implement using existing TFJS functions as well.

kratsg:
@DirkToewe -- copying here for usefulness

```
#include <math.h>

// Returns the value ln[Γ(xx)] for xx > 0.
float gammln(float xx)
{
//Internal arithmetic will be done in double precision, a nicety that you can omit if five-figure
// accuracy is good enough.
    double x,y,tmp,ser;
    static double cof[6]={76.18009172947146,-86.50532032941677,
                          24.01409824083091,-1.231739572450155,
[...]

PR #2238, opened 2019-10-19
tf.broadcastTo

DirkToewe:
Implements `tf.broadcastTo` as requested in #2011.The implementation uses the backend implementation of `tf.tile` and should therefore be just as fast.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2238)


tfjs-core/src/ops/array_ops.ts
```diff
@@ -26,6 +26,72 @@ import {op} from './operation';
 import {MPRandGauss, RandGamma, UniformRandom} from './rand';
 import {zeros, zerosLike} from './tensor_ops';
 
+/** Broadcast an array to a compatible shape NumPy-style.
+ *  
+ *  The tensor's shape is compared to the broadcast shape from end to beginning.
+ *  Ones are prepended to the tensor's shape until is has the same length as
+ *  the broadcast shape. If input.shape[i]==shape[i], they (i+1)-th axis is
```
annxingyuan: nit: the (i+1)-th

tfjs-core/src/ops/array_ops.ts
```diff
@@ -26,6 +26,72 @@ import {op} from './operation';
...
+  {
+    const newShape =  input.shape.slice();
+    while(newShape.length < shape.length ) {
+          newShape.unshift(1);
+    }
+    input = input.reshape(newShape);
+  }
+
+  const reps: number[] = Array.from(shape);
+  for( let i=shape.length; i-- > 0; )
```
annxingyuan: let i=shape.length; i>0; i--

tfjs-core/src/ops/array_ops.ts
```diff
@@ -26,6 +26,72 @@ import {op} from './operation';
...
+      throw new Error(
+        `broadcastTo(): [${xShape}] cannot be broadcast to [${shape}].`
+      );
+    }
+  }
+
+  const axes = reps.map( ( n,i)  => n > 1 ? i : -1 ).filter( i => i >= 0 );
+
+  if( axes.length === 0 ) {
+    return input as Tensor<R>;
```
annxingyuan: Please clone the result before returning it, like this: `input.clone()`.

DirkToewe: Without `.clone()`, disposing the input would also dispose the return value, right? Didn't even think about that.

tfjs-core/src/ops/array_ops.ts
DirkToewe: Changed it.

tfjs-core/src/ops/array_ops.ts
DirkToewe: Changed.

annxingyuan:
@DirkToewe sorry just one last thing - could you use `expectArraysClose` everywhere, instead of checking for mobile safari? This is what we do in our other tests. After that we'll be good to go :)

DirkToewe:
Okay, I changed it. In [tf.linalg.bandPart](https://github.com/tensorflow/tfjs/pull/2226/files#diff-7eb4288fe3d7c2d1522e5a1bce92e087R134-R136), I've used the same check. If that's okay I'm going to changed it in that commit as well.

annxingyuan:
@DirkToewe that would be great! thanks!

PR #2242, opened 2019-10-21
webgpu: Add conv2dByMatMul

qjia7:
PERF

With this change, the PoseNet demo increases 1 fps.

Meanwhile, this patch unifies the behavior between
dispatchLayout and dispatch thread identification values.
y -> column
x -> row

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2242)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -617,8 +617,37 @@ export class WebGPUBackend extends KernelBackend {
     return this.binaryCompareOp(a, b, binary_op.GREATER_EQUAL);
   }
 
+  private conv2dByMatMul(
+      x: Tensor4D, filter: Tensor4D,
+      convInfo: backend_util.Conv2DInfo): Tensor4D {
+    const xShape = x.shape;
+    const isChannelsLast = convInfo.dataFormat === 'channelsLast';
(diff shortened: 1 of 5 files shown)


qjia7:
I don't have the permission to look at the failures. Not sure what they are. But all the local tests pass and the PoseNet demo works well.

qjia7:
Solved the `yarn lint` error. Please take a look.

annxingyuan:
Thank you @qjia7 ! Could you just explain a bit more what you mean by "unifies the behavior between dispatchLayout and dispatch thread identification values"?

qjia7:
In the matmul algorithm, we directly use globalRow = gl_GlobalInvocationID.y and globalCol = gl_GlobalInvocationId.x. I think globalRow is in the range of [0, outputShape[1]) and globalCol is in the range of [0, outputShape[2]). So the correct mapping between dispatchLayout and outputShape should be dispatchLayout = {x: [2], y: [1], z[0]} so that gl_GlobalInvocationID.x can correspond to outputShape[2] and gl_GlobalInvocationID.y can correspond to OutputShape[1].

If we don't do this change from `this.dispatchLayout = {x: [1], y:[2], z:[0]}` to `this.dispatchLayout = {x: [2], y:[1], z[0]}`, the PoseNet demo won't get correct result.

Issue #2246, opened 2019-10-21
README not rendering on npmjs for tfjs


nsthorat:
See https://npmjs.com/package/@tensorflow/tfjs for 1.2.11.

This is because we don't ship a readme in the tfjs/ folder anymore. Will send a fix.

PR #2247, opened 2019-10-21
Copy tfjs/README.md to README.md during publishing.

nsthorat:
NPM doesn't follow symlinks with `npm pack` so we copy it during publishing and ignore it with .gitignore.

Fixes https://github.com/tensorflow/tfjs/issues/2246

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2247)


tfjs/scripts/build-npm.sh
```diff
@@ -27,5 +27,8 @@ mkdir dist/miniprogram
 cp dist/tf.min.js dist/miniprogram/index.js
 cp dist/tf.min.js.map dist/miniprogram/index.js.map
 
+# Copy the root README over.
+cp ../README.md .
```
dsmilkov: -f to force in case there is already a readme, or do you prefer it to fail?

nsthorat: It won't fail if there's already a readme, cp just overwites.

PR #2251, opened 2019-10-21
Convert all usages of ENV in all packages to env().

nsthorat:
ENV is going to be removed in TensorFlow.js 2.0 because it breaks es2015 module semantics (you shouldn't export a mutable object).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2251)


nsthorat:
Done

PR #2256, opened 2019-10-22
webgpu: Enable rectangle workgroup size for matMul

qjia7:
PERF

With this change, conv2d for Integrated GPU becomes 27ms instead
timeout.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2256)


qjia7:

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -58,6 +58,22 @@ export function computeDispatch(
   ];
 }
 
+export function computeWorkGroupSize(
+    layout: {x: number[], y?: number[], z?: number[]},
+    outputShape: number[]): [number, number, number] {
+  const dim0 = arrayProduct(layout.x.map(d => outputShape[d]));
+  const dim1 = arrayProduct(layout.y.map(d => outputShape[d]));
+  // TODO(jiajia.qin@intel.com): More fine tune based on outputShape.
+  if (dim0 >= 1024 * dim1) {
```
annxingyuan: Can you explain this logic a bit more? How did you arrive at these settings?

tfjs-backend-webgpu/src/kernels/matmul_packed_webgpu.ts
```diff
@@ -20,16 +20,39 @@ import {computeDispatch, tilesFitEvenlyIntoShape} from '../webgpu_util';
...
+    workPerThread: number, workGroupSize: number[]): string {
   return `
     ${matMulHeader}
 
-    const int WorkGroupSize = int(gl_WorkGroupSize.x);  // .x == .y
     const int WorkPerThread = ${workPerThread};
-    const int MatTileSize = WorkGroupSize * WorkPerThread;
+    const int Tile_M = int(gl_WorkGroupSize.y) * WorkPerThread;
+    const int Tile_N = int(gl_WorkGroupSize.x) * WorkPerThread;
+    const int Tile_K = Tile_N;
```
annxingyuan: Can you change these variable names to TileAOuter, TileABOuter, TileInner?

tfjs-backend-webgpu/src/kernels/conv2d_mm_webgpu.ts
```diff
@@ -44,24 +44,21 @@ export class Conv2DMMProgram implements WebGPUProgram {
     util.assert(
```
annxingyuan: Can you assert that x % y is zero, and also that x > y?

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -58,6 +58,22 @@ export function computeDispatch(
...
+    return [16, 4, 1];
+  }
+  if (dim0 * 1024 < dim1) {
+    return [4, 16, 1];
+  }
+
```
annxingyuan: Do these different branches get hit by our test suite?

tfjs-backend-webgpu/src/kernels/conv2d_mm_webgpu.ts
```diff
@@ -31,7 +31,7 @@ export class Conv2DMMProgram implements WebGPUProgram {
   variableNames = ['x', 'W'];
   uniforms = 'ivec2 filterDims, pad, stride;';
   workGroupSize: [number, number, number] = [
-    16, 16,  // must be square (for matmul)
+    16, 16,  // x % y must be zero.
     1
```
annxingyuan: I'm a little bit confused - in your comment you say that x % y must be 0. But doesn't `computeWorkGroupSize` below make x < y sometimes? 

Can you assert that x % y is zero, and also that x > y? Also, maybe this comment belongs in the matMul shader instead?

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -58,6 +58,22 @@ export function computeDispatch(
   ];
 }
 
+export function computeWorkGroupSize(
```
annxingyuan: Is this logic specific to matMul? If so could you rename the method to reflect?

tfjs-backend-webgpu/src/kernels/conv2d_mm_webgpu.ts
```diff
@@ -44,24 +41,22 @@ export class Conv2DMMProgram implements WebGPUProgram {
...
     let matMulSource: string;
     if (workPerThread === 0) {
       elementsPerThread = [1, 1, 1];
       matMulSource = makeMatMulSource();
     } else {
-      elementsPerThread = [workPerThread, workPerThread, 1];
-      matMulSource = makeMatMulPackedSource(workPerThread);
+      elementsPerThread =
+          computeWorkPerThreadForConv2d(this.dispatchLayout, this.outputShape);
+      matMulSource = makeMatMulPackedSource(elementsPerThread);
```
annxingyuan: can't you just pass in this.workGroupSize here, rather than computing elementsPerThread?

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -58,6 +58,30 @@ export function computeDispatch(
   ];
 }
 
+export function computeWorkGroupSizeForConv2d(
+    layout: {x: number[], y?: number[], z?: number[]},
+    outputShape: number[]): [number, number, number] {
+  const dim1 = arrayProduct(layout.y.map(d => outputShape[d]));
+  // TODO(jiajia.qin@intel.com): More fine tune based on outputShape.
+  if (dim1 <= 4) {
```
annxingyuan: Don't we want to additionally check that dim0 is significantly bigger than dim1?

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -58,6 +58,30 @@ export function computeDispatch(
...
+
+  return [16, 16, 1];
+}
+
+export function computeWorkPerThreadForConv2d(
+    layout: {x: number[], y?: number[], z?: number[]},
+    outputShape: number[]): [number, number, number] {
+  const dim1 = arrayProduct(layout.y.map(d => outputShape[d]));
+  // TODO(jiajia.qin@intel.com): More fine tune based on outputShape.
+  if (dim1 <= 4) {
```
annxingyuan: Don't we want to check that dim0 is significantly bigger than dim1? And why do we never want to return square workPerThread (where the first and second values of the returned array match)?

annxingyuan:
Just a few more comments - excited to merge this soon!

tfjs-backend-webgpu/src/webgpu_util.ts
qjia7: I think it's not needed. The conv2d(1, 128, 128, 4) benchmark already proves that it runs very fast when dim0 is significantly bigger than dim1. I think the root cause of the improvement is that dim1 doesn't need 16 threads. Otherwise you will waste at least 12 threads in each work group for nothing (out of bounds). If we set y dimension to 4, the hardware can transfer the other 12 threads to x dimension to do the real work.

tfjs-backend-webgpu/src/webgpu_util.ts
qjia7: The current rule is still not complete. I saw [2, 4, 1] was better than [2, 2, 1] for PoseNet demo on Intel GPU. So I just set it to [2, 4, 1]. Anyway, I add more rules in newest change to include the square workPerThread.
And I think we need to continue to make the rules better. But we can put it in later work to make this patch simple. How do you think?

tfjs-backend-webgpu/src/webgpu_util.ts
qjia7: Even if dim0 is not significantly bigger than dim1, it doesn't matter. We already fully used y dimension. We can add more rules later. Like dim0 <=4, maybe need to do the similar change. And other situations, like if 4< dim0 < 8, 8 may be the best choice for x dimension.

annxingyuan:
Thanks @qjia7 !

PR #2270, opened 2019-10-25
[Core] Fix a bug with fused conv2d

dsmilkov:
- Fix a bug with activation being passed as null. (Happens from the layers side)
- Fix CI settings to reduce flakiness:
  - Lock Chrome to 67.0 on Windows to fix the WebGL problem on latest Chrome
  - Double the jasmine timeout from 5s to 10s to give us breathing room since sometimes tests in Windows take 5.1-5.3 secs

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2270)

### DIFF
--- a/tfjs-core/karma.conf.js
+++ b/tfjs-core/karma.conf.js
@@ -152,7 +152,9 @@ module.exports = function(config) {
       win_10_chrome: {
         base: 'BrowserStack',
         browser: 'chrome',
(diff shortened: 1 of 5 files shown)


caisq:
@dsmilkov Please update PR description

caisq:
@dsmilkov Oh you did already. Never mind.

PR #2280, opened 2019-10-28
[Node] Depend on newest core (1.3.0)

dsmilkov:
Update node to be compatible with the newest core (1.3.0).

Also:
- Add `--grep` functionality when running `yarn test`
- Fix a small bug in the unit tests which fail to delete a temporary directory `tfjs_node_fs_test`

A version bump of node will be in a follow up PR.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2280)


caisq:
cc myself

PR #2283, opened 2019-10-28
[WASM] Add Conv2D that calls into xnn pack.

nsthorat:
Adds Conv2D.

- Calls into xnn pack, caching the operators by arguments to the xnn operator constructor.
- Since the same filter can be used with multiple arguments, we keep 1) a map of filter to array of arguments and 2) a map of arguments to the xnn operator.
- When a filter is disposed, we kill all of the xnn operators associated with it.

Also throws better errors for when kernels aren't defined.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2283)


tfjs-backend-wasm/src/cc/kernels/Conv2D.cc
```diff
@@ -0,0 +1,144 @@
...
+        input_channels /* group_input_channels */,
+        output_channels /* group_output_channels */,
+        input_channels /* input_pixel_stride */,
+        output_channels /* output_pixel_stride */, filter_buf, bias_buf,
+        output_min, output_max, flags, &conv2d_op);
+    if (status != xnn_status_success) {
+      util::warn(
+          "XNN status for xnn_create_convolution2d_nhwc_f32 is not successful. "
+          "Got status %d. Use -c dbg to see XNN logs.",
+          status);
```
Maratyszcza: does `util::warn` throw an exception, or otherwise exit the function? if operation creation failed, it makes no sense to continue

tfjs-backend-wasm/src/cc/kernels/Conv2D.cc
```diff
@@ -0,0 +1,144 @@
...
+      filter_height,   filter_width,   stride_height, stride_width,
+      dilation_height, dilation_width, groups,        input_channels,
+      output_channels, filter_id,      flags};
+
+  auto operator_cache_idx = operator_cache.find(cache_key);
+  if (operator_cache_idx == operator_cache.end()) {
+    float output_min = -std::numeric_limits<float>::infinity();
+    float output_max = std::numeric_limits<float>::infinity();
+
+    const float* bias_buf = nullptr;
```
Maratyszcza: Does TF.js support fused Conv2D+Bias? Adding bias separately is inefficient

PR #2297, opened 2019-10-30
webgpu: add dilated convolution support

### LINKED ISSUE #1639: Support dilated convolutions.

haoyunfeix:
FEATURE Fixes #1639

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2297)


haoyunfeix:
@annxingyuan PTAL

haoyunfeix:
To enable convolution dilated tests, use multiple patterns to match in EXCLUDE_LIST.

BEFORE | EXCLUDE_LIST=['d=2'] |  All dilated tests(convolution dilated and pool dilated) will not run.
------------ | ------------- | ------
AFTER | EXCLUDE_LIST=['pool & d=2'] | Only pool dilated tests will not run.

tfjs-backend-webgpu/src/setup_test.ts
```diff
@@ -58,7 +61,7 @@ const EXCLUDE_LIST: string[] = [
   'c + A',                                // Shader compile fails.
   'int32 * int32',                        // Actual != Expected.
   'conv2dTranspose',                      // DerInput is not Implemented.
-  'd=2',                                  // Dilation is not implemented.
+  'pool & d=2',                           // Pool dilation is not implemented.
```
annxingyuan: Would you mind splitting these into two separate excludes? e.g. 'pool', 'd=2'. I think it will be a little bit more straightforward this way.

haoyunfeix: Yeah, I understand your point, but I don't think we can separate key words into two strings.
Think about cases below, we want to execute 1 and 3, but skip 2.
1. pool, d=1 :heavy_check_mark:
2. pool, d=2 :x:
3. conv2d, d=2 :heavy_check_mark:

If we separate excludes as `'pool'` and `'d=2'`, then all three will not run. What do you think we use `'pool, d=2'` instead?
We can split `'pool, d=2'` with `','` and get both 'pool' and 'd=2' at one time.

annxingyuan: You want to execute 1, 3, but skip 2? Then yes I think excluding `pool, d=2` makes sense. I'm not sure I understand your point about splitting with ',' though, could you clarify?

haoyunfeix: Oh, let me clarify that.
As you know, case description goes like this:
1. xxx pool xxxx, d=1 xxx :heavy_check_mark:
2. xxx pool xxxx, d=2 xxx :x:
3. xxx pool yyyy, d=2 yyy :x:
4. xxx conv2d xxx, d=2 xxx :heavy_check_mark:  

In current source, if we use `'pool, d=2'` in EXCLUDE_LIST, no case above will match, if we use `'pool', 'd=2'`, all four will match. But what we want is to skip 2 and 3, so my solution is to put 'pool' and 'd=2' in an array `arr`, and use `arr.every(str => name.indexOf(str) > -1)` to judge if we should skip current case. 
That means I still write `'pool, d=2'` or `'pool & d=2'`in the EXCLUDE_LIST, but `specFilter` function will split it into array once met ',' or '&' 
And my last question is which separator do you prefer, `const arr = EXCLUDE_LIST[i].split('&');` or `const arr = EXCLUDE_LIST[i].split(',');`

annxingyuan: I see! Sorry for not understanding earlier. I think it's time that we move to this setup: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/setup_test.ts

Let me make that change, then maybe you can merge it into your branch and this problem will go away :)

annxingyuan: Actually, it's going to take a while to move to the new setup (it requires updating the tfjs-core dependency, and a lot of things have changed). Let's go ahead with this as a stop-gap measure. Sorry for the back and forth!

PR #2325, opened 2019-11-04
rely on only the tensorflow cpu pip

pyu10055:
This will fix the issue that oversized tensorflow pip file is causing the cloud build to fail.
And the test seems to be one minute faster.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2325)


dsmilkov:
Yay!

PR #2336, opened 2019-11-05
[node] Add SavedModel execution

kangyizhang:
This is the first implementation of SavedModel execution through TF C API. 

Added more SavedModel for testing. These SavedModels are generated from python code. 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2336)


tfjs-node/src/saved_model.ts
```diff
@@ -278,10 +281,50 @@ export class TFSavedModel implements InferenceModel {
...
-      throw new Error(
-          'predict() of TFSavedModel is not supported yet. ' +
-          'Input node names are ' + this.inputNodeNames.toString() +
-          '. Output node names are ' + this.outputNodeNames.toString() + '.');
+      let inputTensors: Tensor[] = [];
+      if (inputs instanceof Tensor) {
+        inputTensors.push(inputs);
+        return this.backend.runSavedModel(
+            this.sessionId, inputTensors, Object.values(this.inputNodeNames),
+            Object.values(this.outputNodeNames))[0];
```
Pierrci: Hi @kangyizhang, is there a reason why when a single output is provided it returns only the first output, even if there are many? I'm currently experimenting with SavedModel support and because of that I have to pass the inputs as an array to be sure to get all the outputs, even if the model has only one input.

kangyizhang: good catch. This should be fixed. If the output has only one item, it will return one tensor. If the output has multiple nodes, it will return an array. I'll send a PR to fix this and it should be included in the next release.

PR #2344, opened 2019-11-06
[WASM] Transpose the filter of the convolution before calling xnnpack.

nsthorat:
XNNPack expects kernels in the following format:
[output channels, kernel height, kernel width, input channels]

TensorFlow and TensorFlow.js use the following format:
[kernel height, kernel width, input channels, output channels]

This PR transposes the filter when an XNNPack kernel is created. Since XNNPack keeps a copy of the filter we transpose, call xnn pack, and immediately throw out the transposed kernel.

To share transposed I moved the body of Transpose.cc to a separate transpose_impl.cc/h (they must be named differently beyond capitalization or bazel gets confused).

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2344)


nsthorat:
@googlebot I consent.

nsthorat:
Most of the complexity is actually just moving transpose to a shared implementation, however you are right a direct 2d transpose is possible here (we do this internally anyways) but just to be sure I made it 2d.

PR #2356, opened 2019-11-08
[WASM] Add FusedConv2D which only supports fusing bias.

nsthorat:
- Moves the calls to xnn pack conv2d to a shared utility.
- Adds an optional bias. If bias id is -1, we pass a nullptr to xnn pack, otherwise we pass the bias buffer.
- Cache operators from bias and filter ids. If a bias **or** a filter dies, the associated xnn operator dies.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2356)


nsthorat:
The CI will be fixed once this goes in because of the TypeScript version update conflicting with WebGL types (diffing against master but until this goes in there will be a TS version skew between master and this change).

PR #2374, opened 2019-11-11
[WASM] Add DepthwiseConv2dNative

nsthorat:
- Adds DepthwiseConv2dNative (this is the TensorFlow naming) by calling into xnnpack.
- Updates the shared conv2d cache to be exactly the params to xnnpack
- Weights expected by XNNPack are of shape `[filter_height, filter_width, input_channels, multiplier]` which is what TensorFlow also expects. In this case, we don't transpose weights (we have to for conv2d).

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2374)


dsmilkov:
The commits are mine, thus CLA is ok

PR #2386, opened 2019-11-13
for .. in loop to for .. of loop

karikera:
I defined `Array.prototype.something = function(){}`
It makes me error.
I guess `for .. of` loop is better

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2386)


karikera:
@googlebot I signed it!

PR #2388, opened 2019-11-13
fix padding_test.ts & 'for .. in' to 'for .. of'

karikera:
'for .. in' to 'for .. of' for Array

padding_test.ts
fix the test about dataFormat

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2388)


dsmilkov:
I just found there is such a rule. Feel free to enable it in a new PR: https://palantir.github.io/tslint/rules/no-for-in-array/

PR #2402, opened 2019-11-18
webgpu: Fix the delayed mode error

qjia7:
When we used DataStorage class for storing tensor map, below error was
exposed:
Uncaught (in promise) TypeError: Cannot read property 'backend' of
undefined
    at t.moveData (engine.ts:381)
    at t.get (backend.ts:58)
    at WebGPUBackend.maybeReleaseBuffer (backend_webgpu.ts:177)

The root case is that we always called 'this.tensorMap.delete(dataId)'
in disposeData function. However, in delayed mode, only when we really
flush DisposalQueue, we can call tensorMap.delete. Otherwise, the tensor id
is deleted before the buffer value is released and above error will be triggerd.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2402)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -131,7 +131,10 @@ export class WebGPUBackend extends KernelBackend {
   }
 
   flushDisposalQueue() {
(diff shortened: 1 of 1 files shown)


qjia7:

annxingyuan:
@qjia7 Thanks!

PR #2430, opened 2019-11-25
[WIP]webgpu: Don't use Int32Array.from

qjia7:
PERF

Per testing, we find that Int32Array.from is very slow on windows.
If we use 'new Int32Array' to replace 'Int32Array.from', the fromPixels
can speed up 3x-5x on different windows platforms. With this change,
the posenet demo can speedup 2x-3x on plaforms that fromPixels are
bottleneck.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2430)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -939,7 +939,7 @@ export class WebGPUBackend extends KernelBackend {
     const output = this.makeOutputArray(outShape, 'int32');
 
     const info = this.tensorMap.get(output.dataId);
-    info.values = Int32Array.from(pixelArray);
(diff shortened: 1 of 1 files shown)


annxingyuan:
@qjia7 just checking - is this ready for review?

qjia7:
> @qjia7 just checking - is this ready for review?

@annxingyuan I filed a chrome bug https://bugs.chromium.org/p/chromium/issues/detail?id=1028538 to track this issue. It seems that it should be fixed in v8 engine. And Shaobo in our team is doing the copyImageBitmapToTexture so that we don't need to read data back to cpu. In that time, these codes will be removed. The code is ready. But I am not sure if this patch is still needed.

qjia7:
This change was merged with #2451. Close this one.

PR #2451, opened 2019-11-29
webgpu: Add shaderKey as the program member

qjia7:
This change is to reduce the overhead of shaderKey comparison.
Previously, we used a very long string 'userCode' as part
of the shader key which is time consuming when search the shader key.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2451)


qjia7:
@annxingyuan Please have a preview if it's in the right direction. It seems that latest webgpu backend has some regressions. Lots of tests will fail because of SPIRV validation error. I will look at that first.

qjia7:
> Can you make sure `shaderKey` is an optional property and that we default to the user code when shaderKey is undefined?

Please see if the latest change is what you want. I am wondering maybe in future we can change userCode to a function getUserCode(). In this case, we can move a lot logic to getUserCode and don't need to do them every time when we call the program constructor.

>please keep me posted re the SPIRV validation errors

Filed a bug to SPIRV-Tools. See https://github.com/KhronosGroup/SPIRV-Tools/issues/3085

PR #2463, opened 2019-12-02
[converter] Fold conv/depthwiseconv + add + batchnorm

dsmilkov:
PERF

Fold conv/depthwise + add + batchnorm into the conv/depthwise. This case shows up in our models, but not in the tflite model since the Toco converter handles that case.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2463)


dsmilkov:
Thanks. Added test to the converter.

PR #2498, opened 2019-12-05
Remove a few unnecessarily mutable exports from tfjs-core

blickly:
Mutable ESM exports can be difficult to reason about and are often not necessary.

Remove them in a few places that can use `export const` instead.

(At Google internally we forbid them and we are now cleaning up existing usages.)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2498)


dsmilkov:
Thanks!

PR #2513, opened 2019-12-07
Add aarch64 support for tfjs-node

yhwang:
Modify the install.js to support external binaries while installing
npm package. It reads the `custom-binary.json` configuration file to
get URLs for the following resources:
- 'tf-lib': libtensorflow and libtensorflow_framework shared libraries.
  Its value is a URL that points to a `tar.gz` or `zip` and
  contains the shared libaries and header files.
- 'addon': pre-compiled node binding. Its value is a object contains
  three values: `host`, `remote_path` and `package_name`. These three
  values are concatenated into a URL that points to the
  pre-compiled node binding.

If you only provide `tf-lib` and no `addon`. It compiles the node
binding while installing the npm package.

Signed-off-by: Yihong Wang <yh.wang@ibm.com>

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2513)


yhwang:
Hi @kangyizhang, in order to support aarch64, I added architecture into the filename of those pre-compiled resources which are stored in the cloud storage. For example, `CPU-darwin-1.14.0.tar.gz` becomes `CPU-darwin-x86_64-1.14.0.tar.gz`. Please review the naming change and see if it can support more platform and architectures in the future.

And I can help to provide the arm64 binaries for linux platform.

yhwang:
@kangyizhang I made some changes and it detects if the system is supported or not. In arm64 system, it also runs `node-pre-gyp` command with `--build-from-source` option.

yhwang:
In the latest change, `install.js` checks `custom-binary.json` at the same directory and retrieve the custom binaries if external resource URLs are specified. If `custom-binary.json` doesn't exist, then `install.js` retrieves binaries from GCP.

yhwang:
looks good to me. thanks for removing the `try/finally`. I left it there and forgot to remove it.

kangyizhang:
thank you for the PR!

PR #2525, opened 2019-12-10
[webgpu] Fix memory leak in buffer manager

axinging:
key in this.freeBuffers will lead to NOP, so it will not destroy any buffer.
Below sample can be used to verify this.

```
let freeBuffers: Map<string, number[]> = new Map();
let values: number[];

values = [1, 2, 3];
freeBuffers.set('a', values);
values = [4, 5, 6];
freeBuffers.set('b', values);

// The correct way to iterate over map.
for (freeBuffers.forEach((buffers, key) => {
    buffers.forEach(buff => {
    console.log(buff);
  });
}

// The incorrect way to iterate over map.
for (const key in freeBuffers) {
    freeBuffers.get(key).forEach(buff => {
      console.log(buff);
    });
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2525)

### DIFF
--- a/tfjs-backend-webgpu/src/buffer_manager.ts
+++ b/tfjs-backend-webgpu/src/buffer_manager.ts
@@ -93,7 +93,7 @@ export class BufferManager {
   }
 
   dispose() {
-    if (this.freeBuffers == null && this.usedBuffers) {
+    if (this.freeBuffers == null && this.usedBuffers == null) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


axinging:
@annxingyuan , PTAL

axinging:
Hi, @axinging, I add a sample in above comments. PTAL

axinging:
Thanks @annxingyuan ,  this seems like a typo. I fixed it, PTAL.

PR #2532, opened 2019-12-10
[WASM] Use XNNPack for tf.matMul

nsthorat:
For now we only call XNNPack when:
- transpose_a = transpose_b = false.
- batch_size = 1

This is the simplest way to use XNNPack. For the rest, use the old implementation.

In the future we will:
- Support fused activations
- Support transposed bits to xnnpack
- non-1 batch size unrolled into a for loop

I have to leave the office but tomorrow I'll do a micro benchmark (we can't yet do this for a model).

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2532)


nsthorat:
Updated the PR description with benchmarks for posterity.

PR #2538, opened 2019-12-11
Update release notes script to work with the monorepo.

nsthorat:
1.4.0 release notes here: https://github.com/tensorflow/tfjs/releases/tag/tfjs-v1.4.0

This change:
- Moves the script out of tfjs.
- Adds support for the monorepo. This now has to filter commits for the folder name. In the future we can support changelogs for other directories.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2538)


nsthorat:
Merging this because the only tests failing are flaky data tests (will file an issue).

PR #2548, opened 2019-12-12
[data] Update microphone test wait time to remove flaky

kangyizhang:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2548)


tfjs-data/src/iterators/microphone_iterator_test.ts
```diff
@@ -214,11 +216,11 @@ describeBrowserEnvs('MicrophoneIterator', () => {
     // Call iterator.next() every 200 milliseconds, stop after 3 times.
     const interval = setInterval(getTensor, 200);
 
-    // Wait 4 seconds for the intervals to run.
+    // Wait 3 seconds for the intervals to run.
     await new Promise(resolve => {
```
nsthorat: is it possible to write these tests in such a way we don't actually have to wait 3 seconds? like mocking timing information so this is a quicker test?

kangyizhang: Mocking timing information of setInterval is quite tricky. I updated microphone configs in tests to reduce the time to produce a tensor. Each test should only take several milliseconds now.

PR #2557, opened 2019-12-14
Split NonMaxSuppressionV3 and NonMaxSuppresionV5 to two APIs

lina128:
Because in the C implementation, the V3 has 1 output but the V5 has 3 outputs, we have to define different APIs for them. Otherwise, models that uses NonMaxSuppressionV5 may break because subsequent Op may need to access the second or the third output.

Changes made:
1. Add one more API for V5 in backend.ts
2. Add one more API for V5 in image_op.ts
3. Add corresponding implementations for CPU and GPU.
4. Add corresponding tests.

Note: 
1. @kangyizhang identified this issue, because the Node test break, expecting different output shape. After this change, tested Node code locally and it passed.
2. Python API also splits into two APIs (non_max_suppression and non_max_suppression_with_score). tfjs naming follows the Python API naming found here: https://www.tensorflow.org/api_docs/python/tf/image/non_max_suppression
3. This implementation naturally includes V4. Will add V4 API in another PR. V4 will also require a different API because the output is also different.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2557)

### DIFF
--- a/tfjs-backend-wasm/package.json
+++ b/tfjs-backend-wasm/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow/tfjs-backend-wasm",
-  "version": "1.4.0-alpha3",
+  "version": "1.5.0-alpha4",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "jsnext:main": "dist/tf-backend-wasm.esm.js",
@@ -32,7 +32,7 @@
     "path": false
   },
   "peerDependencies": {
-    "@tensorflow/tfjs-core": "link:../tfjs-core"
+    "@tensorflow/tfjs-core": "1.5.0"
   },
   "dependencies": {
     "@types/emscripten": "~0.0.34"
@@ -40,7 +40,7 @@
   "devDependencies": {
     "@bazel/bazel": "^0.28.0",
     "@bazel/buildifier": "0.29.0",
-    "@tensorflow/tfjs-core": "link:../tfjs-core",
+    "@tensorflow/tfjs-core": "1.5.0",
     "@types/jasmine": "~2.8.6",
     "clang-format": "~1.2.4",
     "jasmine": "~3.1.0",
--- a/tfjs-backend-wasm/src/version.ts
+++ b/tfjs-backend-wasm/src/version.ts
@@ -1,5 +1,5 @@
 /** @license See the LICENSE file. */
 
 // This code is auto-generated, do not modify this file!
-const version = '1.4.0-alpha3';
+const version = '1.5.0-alpha4';
 export {version};
--- a/tfjs-backend-wasm/yarn.lock
+++ b/tfjs-backend-wasm/yarn.lock
@@ -73,9 +73,17 @@
   resolved "https://registry.yarnpkg.com/@bazel/hide-bazel-files/-/hide-bazel-files-0.38.3.tgz#e98231d3d360d51860d9c1a7c3345b40dab4cf81"
   integrity sha512-o+dNkfDm3qxWQ8h/04cWuTcjR7qnjZi3pQGv4aklVb16oPWx2jF8BzbkwvWuIkdbOl9VnqYP0vaHzwQVJRRcIA==
 
-"@tensorflow/tfjs-core@link:../tfjs-core":
-  version "0.0.0"
-  uid ""
+"@tensorflow/tfjs-core@1.5.0":
+  version "1.5.0"
+  resolved "https://registry.yarnpkg.com/@tensorflow/tfjs-core/-/tfjs-core-1.5.0.tgz#a011a68882068d91d82410bb71aa0220867286bc"
+  integrity sha512-UcTAoVNmBxJLktGWUNbSq7ZEgBvjD0SWS0QcRBMgs93AeLM3xhlyo8MqECQsvCkuReDZwVHhX2fjMrf2M3R99w==
+  dependencies:
+    "@types/offscreencanvas" "~2019.3.0"
+    "@types/seedrandom" "2.4.27"
+    "@types/webgl-ext" "0.0.30"
+    "@types/webgl2" "0.0.4"
+    node-fetch "~2.1.2"
+    seedrandom "2.4.3"
 
 "@types/emscripten@~0.0.34":
   version "0.0.34"
--- a/tfjs-converter/package.json
+++ b/tfjs-converter/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow/tfjs-converter",
-  "version": "1.4.0",
+  "version": "1.5.0",
   "description": "Tensorflow model converter for javascript",
   "main": "dist/src/index.js",
   "jsnext:main": "dist/tf-converter.esm.js",
@@ -15,10 +15,10 @@
   },
   "license": "Apache-2.0",
   "peerDependencies": {
-    "@tensorflow/tfjs-core": "1.4.0"
+    "@tensorflow/tfjs-core": "1.5.0"
   },
   "devDependencies": {
-    "@tensorflow/tfjs-core": "1.4.0",
+    "@tensorflow/tfjs-core": "1.5.0",
     "@types/deep-equal": "^1.0.1",
     "@types/jasmine": "~2.8.6",
     "@types/long": "~3.0.32",
--- a/tfjs-converter/python/tensorflowjs/version.py
+++ b/tfjs-converter/python/tensorflowjs/version.py
@@ -1,4 +1,4 @@
 # @license See the LICENSE file.
 
 # This code is auto-generated, do not modify this file!
-version = '1.4.0'
+version = '1.5.0'
--- a/tfjs-converter/src/version.ts
+++ b/tfjs-converter/src/version.ts
@@ -1,5 +1,5 @@
 /** @license See the LICENSE file. */
 
 // This code is auto-generated, do not modify this file!
-const version = '1.4.0';
+const version = '1.5.0';
 export {version};
--- a/tfjs-converter/yarn.lock
+++ b/tfjs-converter/yarn.lock
@@ -55,10 +55,10 @@
   resolved "https://registry.yarnpkg.com/@protobufjs/utf8/-/utf8-1.1.0.tgz#a777360b5b39a1a2e5106f8e858f2fd2d060c570"
   integrity sha1-p3c2C1s5oaLlEG+OhY8v0tBgxXA=
 
-"@tensorflow/tfjs-core@1.4.0":
-  version "1.4.0"
-  resolved "https://registry.yarnpkg.com/@tensorflow/tfjs-core/-/tfjs-core-1.4.0.tgz#045c91f93cb3ea77473b664c0d1dbad675d7f046"
-  integrity sha512-a7dHhSsBbtaxp8/1UAeYKrjY1bQxsDy/Uhj57+mTuGLAcL8CDrTOXXZzucCgs5nvQErdscp7Gp/2VCcA1xp6XQ==
+"@tensorflow/tfjs-core@1.5.0":
+  version "1.5.0"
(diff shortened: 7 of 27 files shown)


lina128:
> thanks for fixing this issue.
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2557#-:-LwApvIM2ll_n3Hx1cVG:brjurt9)_ status: 0 of 1 approvals obtained (waiting on @dsmilkov, @lina128, and @pyu10055)
> 
> _[tfjs-core/src/backends/backend.ts, line 600 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwApEke6gJjWkt0w5Ut:-LwApEke6gJjWkt0w5Uu:by1doga) ([raw file](https://github.com/tensorflow/tfjs/blob/558ab3cbccbba423017be7068caea978f6208739/tfjs-core/src/backends/backend.ts#L600)):_
> 
> > ```
> >       boxes: Tensor2D, scores: Tensor1D, maxOutputSize: number,
> >       iouThreshold: number, scoreThreshold?: number,
> >       softNmsSigma?: number): [Tensor1D, Tensor1D, Scalar] {
> > ```
> 
> It might be better to define a interface for the output, otherwise user might not know the meaning of each tensor.
> 
> _[tfjs-core/src/backends/non_max_suppression_impl.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwApgNJBSfrA3Ds8Iar:-LwApgNJBSfrA3Ds8Ias:b-mdpb8l) ([raw file](https://github.com/tensorflow/tfjs/blob/558ab3cbccbba423017be7068caea978f6208739/tfjs-core/src/backends/non_max_suppression_impl.ts#L34)):_
> 
> > ```
> > }
> > 
> > export function nonMaxSuppressionV3(
> > ```
> 
> the method names might be good to be consistent with the external APIs.

Hi Ping, thanks for the review! Changed to use NamedTensorMap for the output. As for method names, I keep it in sync with wasm method name: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/kernels/NonMaxSuppressionV3.ts and TF C++ version: https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/non-max-suppression-v5
I don't have a strong preference on whether the internal kernel name should be consistent with external APIs, as long as we have some consistency within our library. Maybe it makes sense for now, to keep our kernel names to be in sync with C version's kernel names, and keep our external API names to be in sync with Python version's API names? This helps us identify amount of feature parity at each level and help tracing reference.

lina128:
> Really nice work!! One blocking question about the user-facing https://www.tensorflow.org/api_docs/python/tf/image/non_max_suppression_with_scores having 2 outputs instead of 3.
> 
> > Reviewed 6 of 6 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2557#-:-LwEJduP7MQhP7FeDFz_:b-1twexm)_ status: 0 of 1 approvals obtained (waiting on @dsmilkov, @lina128, and @pyu10055)
> 
> _[tfjs-core/src/backends/backend.ts, line 597 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwEHMfdBODK9tRvHcVe:-LwEH3xYALQKdFtmgzP9:b4pvyue) ([raw file](https://github.com/tensorflow/tfjs/blob/558ab3cbccbba423017be7068caea978f6208739/tfjs-core/src/backends/backend.ts#L597)):_
> 
> > ```
> >   }
> > 
> >   nonMaxSuppressionWithScore(
> > ```
> 
> Going forward we decided to add any new kernels in a modular way in order to avoid the ever growing library size. This means that instead of adding a new method to the Backend interface, we add the new kernels in standalone files and call `registerKernel()`. See `src/backends/webgl/square.ts` and `src/backends/cpu/square.ts` for examples of modularized `Square`. You don't have to modularize in this PR, but I think we should modularize before we release new tfjs-core. This way, after we release new tfjs-core, tfjs-node just needs to call registerKernel with its own implementation.
> 
> _[tfjs-core/src/backends/backend.ts, line 600 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwApEke6gJjWkt0w5Ut:-LwEGZ8B2jopfNoQrjN7:bo0zp1g) ([raw file](https://github.com/tensorflow/tfjs/blob/558ab3cbccbba423017be7068caea978f6208739/tfjs-core/src/backends/backend.ts#L600)):_
> 
> Previously, pyu10055 (Ping Yu) wrote…
> Array of tensors is ok here since this is a kernel (internal api) and the inference and backprop infra expects tensor|tensor[] as a result of a kernel, analogous to TF C++ kernels which return array of tensors.
> 
> _[tfjs-core/src/ops/image_ops.ts, line 239 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwEGN5jC_7IJhTiaynL:-LwEGN5jC_7IJhTiaynM:b-veknbh) ([raw file](https://github.com/tensorflow/tfjs/blob/558ab3cbccbba423017be7068caea978f6208739/tfjs-core/src/ops/image_ops.ts#L239)):_
> 
> > ```
> >  *     - A 1D tensor with the corresponding scores for each selected box.
> >  *     - A number representing the number of valid elements in the selected
> >  *       box indices.
> > ```
> 
> The user-facing https://www.tensorflow.org/api_docs/python/tf/image/non_max_suppression_with_scores has 2 outputs instead of 3. Let's match that API at the user-facing level.

Ah, really nice catch! Thank you Daniel! Changed output to 2. The numValidOutputs is only for V4, and for V5, the flag to pad the result to the maxOutputSize is always false, therefore, no need for this output. I made sure the unused tensor is disposed. Also changed to write the new Op using the modular pattern, that design is really smart, it makes it more flexible to name function than the interface method. Also explicitly registering Ops makes it more readable and more flexible to manage Ops. Thank you for letting me know this new pattern!

lina128:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2557)_ status: 0 of 1 approvals obtained (waiting on @dsmilkov, @lina128, and @pyu10055)
> 
> _[tfjs-core/src/backends/backend.ts, line 597 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwEHMfdBODK9tRvHcVe:-LwEZS4BBe9vnJN8fW0X:b-6mj0qi) ([raw file](https://github.com/tensorflow/tfjs/blob/558ab3cbccbba423017be7068caea978f6208739/tfjs-core/src/backends/backend.ts#L597)):_
> 
> Previously, dsmilkov (Daniel Smilkov) wrote…
> I'll lightly push towards doing it in this PR, of course still optional! :) Na, if the way this gets done is confusing let us know and we can sync with you on a GVC!
> 
> _[tfjs-core/src/ops/image_ops.ts, line 239 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwEGN5jC_7IJhTiaynL:-LwE_3X658-OvHVIGRfr:bbstqlv) ([raw file](https://github.com/tensorflow/tfjs/blob/558ab3cbccbba423017be7068caea978f6208739/tfjs-core/src/ops/image_ops.ts#L239)):_
> 
> Previously, dsmilkov (Daniel Smilkov) wrote…
> also I think we should return an object with named return values (an object). It's fine to keep the kernel an array.
> 
> _[tfjs-core/src/ops/image_ops_test.ts, line 251 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwEZp0S2ZxZNxG9hR58:-LwEZp0T6GJetVpGC4VX:bs3trmk) ([raw file](https://github.com/tensorflow/tfjs/blob/558ab3cbccbba423017be7068caea978f6208739/tfjs-core/src/ops/image_ops_test.ts#L251)):_
> 
> > ```
> >       const scoreThreshold = 0;
> >       const softNmsSigma = 0.5;
> >       const indices = await tf.image.nonMaxSuppressionWithScoreAsync(
> > ```
> 
> can you also add a test for memory here? Since we don't wrap async methods in "op" we have to be careful with tensor leaking manually (e.g. make sure tf.memory().numTensors before and after the call only increase tensors by 1, the output)

Hi Nikhil, thank you for your review! Changed to the modular pattern. Also changed the output to use NamedTensorMap. And thanks for asking to test memory leak, I added the test and it did keep unnecessary Tensor because the generic implementation generates three Tensors, whereas V5 only needs two. I disposed the unused Tensor and it passed the memory test.

lina128:
Hi @pyu10055 @dsmilkov @nsthorat , thanks for your great comments! I changed the new op to the modular pattern and the output to named object. Added test for memory leak. Please re-review the PR. Thank you!

lina128:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2557)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @dsmilkov, @lina128, @nsthorat, and @pyu10055)
> 
> _[tfjs-core/src/backends/non_max_suppression_impl.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwApgNJBSfrA3Ds8Iar:-LwKKBOS2vfysqhVVK9E:b-emw2ln) ([raw file](https://github.com/tensorflow/tfjs/blob/558ab3cbccbba423017be7068caea978f6208739/tfjs-core/src/backends/non_max_suppression_impl.ts#L34)):_
> 
> Previously, lina128 (Na Li) wrote…
> this name sounds good, since we are going to register the kernel for these ops, it would be better to use the TF kernel name.
> 
> _[tfjs-core/src/backends/webgl/non_max_suppression_with_score.ts, line 38 at r3](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwKKTT_71KzQH7J7JjM:-LwKKTT_71KzQH7J7JjN:b-qz2twt) ([raw file](https://github.com/tensorflow/tfjs/blob/5d4b34b0e845bb4970a0acd0cd6b752745acea3a/tfjs-core/src/backends/webgl/non_max_suppression_with_score.ts#L38)):_
> 
> > ```
> > registerKernel({
> >   kernelName: 'NonMaxSuppressionWithScore',
> > ```
> 
> It would be better to use the TF op name.

Agreed. Done.

lina128:
> **Really** nice work! I left one small API suggestion. Otherwise LGTM!
> 
> > Reviewed 5 of 10 files at r2, 5 of 5 files at r4.
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2557#-:-LwKrgsRCyE0GDy6GEkJ:b-8zvcod)_ status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128, @nsthorat, and @pyu10055)
> 
> _[tfjs-core/src/ops/image_ops.ts, line 239 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwEGN5jC_7IJhTiaynL:-LwKrEf44PZnemLC2UU_:bccpe83) ([raw file](https://github.com/tensorflow/tfjs/blob/558ab3cbccbba423017be7068caea978f6208739/tfjs-core/src/ops/image_ops.ts#L239)):_
> 
> Previously, lina128 (Na Li) wrote…
> Awesome!
> 
> _[tfjs-core/src/ops/image_ops.ts, line 264 at r4](https://reviewable.io/reviews/tensorflow/tfjs/2557#-LwKrOTjCm6KDQmLJ41X:-LwKrOTjCm6KDQmLJ41Y:b-x6dp1c) ([raw file](https://github.com/tensorflow/tfjs/blob/6dd9afd50fa3050c55e84474a60afe182204dd70/tfjs-core/src/ops/image_ops.ts#L264)):_
> 
> > ```
> >                      attrs) as Tensor[];
> > 
> >   return {selectedIndices: result[0], selectedScores: result[1]};
> > ```
> 
> tiny API suggestion: `indices` and `scores` to make it shorter since "selected" is implicit given what this method does.

Agreed. I used this naming mainly to keep in sync with Python API's naming: https://www.tensorflow.org/api_docs/python/tf/image/non_max_suppression_with_scores?version=stable

PR #2567, opened 2019-12-17
[WASM] add ConcatV2 and BiasAdd op

jamesguoxin:
During our model inference using wasm, there are a few ops missing, after checking the source code, we identify that BiasAdd share the same implementation with Add, as well as for ConcatV2 and Concat. Here we add kernel names BiasAdd and ConcatV2 without changing any kernel functions. For registerBinaryKernel function, it will take three arguments, namely opName, kernelName and supportsBroadcast, to enable the usage that different kernel name using the same kernel function.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2567)


jamesguoxin:
@googlebot  I signed it!

PR #2575, opened 2019-12-17
Replace Object.values to ensure deterministic ordering

Nithanaroy:
`Object.values` for an object in JavaScript may not have a predictable order - https://stackoverflow.com/a/5525820/1585523
It is safer to explicitly specify the column names ourselves to ensure the same order in the CSV file in all browsers

### DIFF
--- a/tfjs-data/src/readers.ts
+++ b/tfjs-data/src/readers.ts
@@ -62,7 +62,7 @@ import {CSVConfig, MicrophoneConfig, WebcamConfig} from './types';
  *       {
  *         // Convert xs(features) and ys(labels) from object form (keyed by
  *         // column name) to array form.
- *         return {xs:Object.values(xs), ys:Object.values(ys)};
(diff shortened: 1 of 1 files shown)


Nithanaroy:
Yup i agree, it is only an example. But ML engineers who are JS beginners may adopt this as a pattern in their work, given it is from a reliable source. 
As an example the same pattern was used in Tensorflow.js course by deeplearning.ai https://www.coursera.org/learn/browser-based-models-tensorflow/lecture/jXk4t/reading-the-data (forward to 2 min 20s)
I mean we all know how challenging it is to debug data processing problems in ML. Just wanted to bring to your attention. Feel free to discard the pull-request :)

> Thanks for the PR. You are correct that `Object.values` does not guarantee the order of returned values. But this code snippet here is just to show how to get dataset from csv file and feed it to a model for training so it doesn't matter. User can do all kinds of customizations through `dataset.map()`.
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2575#-:-Lw_SQdxBkeOf3XEtNgU:b-691rd5)_ status: 0 of 1 approvals obtained (waiting on @kangyizhang)

PR #2592, opened 2019-12-19
Update tfjs-core to 1.5.0.

nsthorat:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2592)


nsthorat:
The only test failure is coming from SoftNMS and tfjs-node so this shouldn't be blocked.

cc @lina128

PR #2598, opened 2019-12-19
[WASM] Add NonMaxSuppressionV5 Op

lina128:
Note: Tested with latest core at e5a0c84c22e0a9e078b4bf34eed938f8a43aac69, need to bump core version.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2598)

### DIFF
--- a/tfjs-backend-wasm/src/cc/BUILD
+++ b/tfjs-backend-wasm/src/cc/BUILD
@@ -105,6 +105,15 @@ tfjs_cc_library(
 )
 
 tfjs_cc_library(
+    name = "non_max_suppression_impl",
+    srcs = ["non_max_suppression_impl.cc"],
+    hdrs = ["non_max_suppression_impl.h"],
+    deps = [
+      ":backend",
+    ],
+)
+
+tfjs_cc_library(
     name = "prelu_impl",
     srcs = ["prelu_impl.cc"],
     hdrs = ["prelu_impl.h"],
@@ -547,6 +556,7 @@ tfjs_cc_library(
     srcs = ["kernels/NonMaxSuppressionV5.cc"],
     deps = [
         ":backend",
+        ":non_max_suppression_impl",
         ":util",
     ],
 )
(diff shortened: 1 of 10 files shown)


lina128:
Hi @nsthorat, thank you for the very detailed review! I changed the selected_indices from int_32* to size_t*, but there's runtime error 'Access memory out of bound'. Addressed other comments otherwise.

lina128:
Hi @dsmilkov, thank you for the great suggestion! Extract logic to shared file, please re-review the PR. Thank you!

lina128:
> > ```
> >   const {pSelectedIndices, selectedSize} =
> >       parseResultStruct(backend, resOffset);
> > ```
> 
> to avoid memleak, call `backend.wasm._free(pSelectedScores);` here since V3 doesn't return the scores vector.

Hi @dsmilkov, given that parseResultStruct has `backend.wasm._free(resOffset)`, do we really need to free pSelectedScores? I thought free(resOffset) will free all the memory allocated for the NonMaxSuppressionResult.

dsmilkov:
Yes we do need another free since `backend.wasm._free(resOffset)` will only remove the 4 bytes allocated for the pointer address of `pSelectedScores`, but it will not remove the array of values that were allocated on the heap that the pointer points to, thus you need explicit removal of that. The reason why you do not need to do the same with `pSelectedIndices` is because we wrap that in a `Tensor` object, which gets globally tracked and later gets disposed by the `tidy` mechanisms when it gets out of scope.

lina128:
> Yes we do need another free since `backend.wasm._free(resOffset)` will only remove the 4 bytes allocated for the pointer address of `pSelectedScores`, but it will not remove the array of values that were allocated on the heap that the pointer points to, thus you need explicit removal of that. The reason why you do not need to do the same with `pSelectedIndices` is because we wrap that in a `Tensor` object, which gets globally tracked and later gets disposed by the `tidy` mechanisms when it gets out of scope.

Thank you for the detailed explanation Daniel! Now I know how it works. :) Done.

dsmilkov:
Thanks for adding a comment on top of that line! Nice work Na!

PR #2568, opened 2019-12-17
webgpu: Add rules to use conv2dWithIm2Col

qjia7:
conv2d with some input shapes is very slow on integrated GPU.
The possible reasons may be that 1) too many complicated calculations
when reading/writing data. 2) the read data is discontinuous in memory.

To some extent, running im2col as a separate shader can solve above issues.
With this change, posenet demo has 1~3 fps improvement on different i-GPU.

Issue: #2553

On i-GPU
Shapes in ResNet   Before   After
[1, 15, 15, 64]:   13.743   4.830
[1, 8, 8, 512]:    8.230    4.711
[1, 17, 17, 256]:  2.462    1.870

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2568)

PR #2620, opened 2019-12-26
webgpu: Optimize conv2d

qjia7:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2620)

With this change, the posenet+resnet demo has 2~3 fps improvement on Intel GPU. 1 fps improvement on AMD GPU.

PERF


qjia7:
The main reason that the previous method is slow is because that we fetch the data as [channel, height, width] sequence not [height, width, channel]. But the data is stored as [height, width, channel] in memory. That results the data fetching is not continuous in memory which is time-consuming.
You can see this change in https://github.com/tensorflow/tfjs/pull/2620/commits/b2860297d50ad0e9499f890e1b4c70c7602c5fa6#diff-d0f5478de3197c2a9075a8a33e7e8905L120  (the calculation of WRow, WCol, channel) to get more details.

tfjs-backend-webgpu/src/kernels/conv2d_mm_webgpu.ts
```diff
@@ -109,41 +113,35 @@ export class Conv2DMMProgram implements WebGPUProgram {
         ${matMulSource}
 
         int batch;
-
+        int dimAOuter = outShape[1] * outShape[2];
+        int dimBOuter = outShape[3];
+        int dimInner = filterDims[0] * filterDims[1] * xShape[3];
```
annxingyuan: Can you splice in the values for `dimAOuter`, `dimBOuter`, and `dimInner` - e.g. `int dimAOuter = ${this.outputShape[1] * this.outputShape[2]}`?

tfjs-backend-webgpu/src/kernels/conv2d_mm_webgpu.ts
```diff
@@ -109,41 +113,35 @@ export class Conv2DMMProgram implements WebGPUProgram {
...
-        float mm_readB(int row, int col) {
-          int r = int(row), c = int(col);
-          int outRow = c / outShape[2];
-          int outCol = c % outShape[2];
+          int outRow = r / outShape[2];
+          int outCol = r % outShape[2];
 
-          int WRow = (r / filterDims[1]) % filterDims[0];
-          int WCol = r % filterDims[1];
+          int WRow = c / (filterDims[1] * xShape[3]);
```
annxingyuan: Can you splice in `${filterDims[1] * xShape[3]}` as a constant here as well?

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -61,8 +61,12 @@ export function computeDispatch(
 export function computeWorkGroupSizeForConv2d(
     layout: {x: number[], y?: number[], z?: number[]},
     outputShape: number[]): [number, number, number] {
+  const dim0 = arrayProduct(layout.x.map(d => outputShape[d]));
   const dim1 = arrayProduct(layout.y.map(d => outputShape[d]));
   // TODO(jiajia.qin@intel.com): More fine tune based on outputShape.
+  if (dim0 <= 4) {
```
annxingyuan: Can you include a brief note explaining how you came up with these dimensions?

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -76,6 +80,9 @@ export function computeWorkPerThreadForConv2d(
   const dim0 = arrayProduct(layout.x.map(d => outputShape[d]));
   const dim1 = arrayProduct(layout.y.map(d => outputShape[d]));
   // TODO(jiajia.qin@intel.com): More fine tune based on outputShape.
+  if (dim0 <= 4) {
```
annxingyuan: Can you include a brief note explaining how you came up with these dimensions?

tfjs-backend-webgpu/src/kernels/conv2d_mm_webgpu.ts
qjia7: I tried to splice in the values. However, it affected the performance seriously. It dropped 2 fps on Intel GPU. I'm not sure the exact reason (Somehow, I think it's related with the location of these variables.). So I prefer we keep it in this patch. I can have more investigations later.

tfjs-backend-webgpu/src/kernels/conv2d_mm_webgpu.ts
qjia7: I understand that making it as constant can save once calculation. But it will results we have variant shaders which will increase the compilation time of shader in the warm up stage. Maybe this overhead can be ignored?

qjia7:
> Just to make sure I understand your comment on #2568 - are you saying that with this change, running the im2Col shader separately is always slower?

Sorry, I missed this one. Theoretically, the time used by these two methods are very close.  However, in my test, im2col is good for Intel GPU, not that friendly for AMD GPU. The change in this method can benefit both. But I can't say im2col is not needed anymore. I find in some situations, im2col is better than the current method for Intel GPU. I need more time to analysis it and understand the hardware. Also it's hard for me to only apply some changes to a specific GPU due to web's limit.

tfjs-backend-webgpu/src/kernels/conv2d_mm_webgpu.ts
annxingyuan: Hm, that's very strange! But investigating later sounds good :)

PR #2628, opened 2019-12-31
webgpu: Increase work per thread for maxpool

qjia7:
PERF

With this change, maxPool[1, 131, 131, 64] has 50%~90% speedup
on different platforms.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2628)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/maxpool_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/maxpool_webgpu.ts
@@ -29,17 +29,20 @@ export class MaxPoolProgram implements WebGPUProgram {
   dispatch: [number, number, number];
   variableNames = ['x'];
   uniforms = 'ivec2 pad, stride, dilation, convDims, filterDims;';
-  workGroupSize: [number, number, number] = [4, 4, 4];
+  // TODO(jiajia.qin@intel.com): Dynamically choose different workGroupSize and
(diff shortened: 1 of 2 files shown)


qjia7:
The reason that I apply workPerThread to outShape[3] is because the data in outShape[3] is tightly stored which can speed up the fetching.

tfjs-backend-webgpu/src/kernels/maxpool_webgpu.ts
```diff
@@ -69,14 +74,36 @@ export class MaxPoolProgram implements WebGPUProgram {
...
+            {
+              break;
             }
           }
-          setOutput(batch, coords[1], coords[2], d, minMaxValue);
         }
       }
     `;
-   this.shaderKey = 'maxpool';
+    this.shaderKey = 'maxpool';
```
annxingyuan: Wouldn't we need to add `this.workPerThread` to the `shaderKey`?

qjia7: Currently, `this.workPerThread` is fixed. So it won't affect the `shaderKey`. But in case  that we forget to change this place in future when we add dynamically to change `this.workPerThread`, it's ok to change it right now. Thanks.
@annxingyuan Please take another look. Thanks.

PR #2680, opened 2020-01-15
Avoid using apply/call due to new ts compiler flag

lina128:
BUG

Avoid using apply/call due to new ts compiler flag.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2680)

### DIFF
--- a/tfjs-converter/src/executor/graph_executor.ts
+++ b/tfjs-converter/src/executor/graph_executor.ts
@@ -45,7 +45,8 @@ export class GraphExecutor {
   set weightMap(weightMap: NamedTensorsMap) {
     const weightIds = Object.keys(weightMap).map(
         key => weightMap[key].map(tensor => tensor.id));
-    this.weightIds = [].concat(...weightIds);
+    // this.weightIds = [].concat(...weightIds);
+    this.weightIds = [].concat.apply([], weightIds);
     this._weightMap = weightMap;
   }
 
--- a/tfjs-converter/src/executor/graph_executor.ts
+++ b/tfjs-converter/src/executor/graph_executor.ts
@@ -45,7 +45,7 @@ export class GraphExecutor {
   set weightMap(weightMap: NamedTensorsMap) {
     const weightIds = Object.keys(weightMap).map(
         key => weightMap[key].map(tensor => tensor.id));
-    this.weightIds = [].concat.apply([], weightIds);
+    this.weightIds = [].concat(...weightIds);
     this._weightMap = weightMap;
   }
 
--- a/tfjs-converter/src/executor/graph_executor.ts
+++ b/tfjs-converter/src/executor/graph_executor.ts
@@ -45,8 +45,7 @@ export class GraphExecutor {
   set weightMap(weightMap: NamedTensorsMap) {
     const weightIds = Object.keys(weightMap).map(
         key => weightMap[key].map(tensor => tensor.id));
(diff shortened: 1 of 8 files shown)


lina128:
> Another thing to try is to see if you can isolate which part of this change is causing the problem (you have 3 files run each change independently).
> 
> Also is this happening for any new change added or is it only this change causing it?
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2680#-:-LzMy1fo0QqBLkByCdAE:bvn0v4f)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)
> 
> _[tfjs-converter/src/executor/graph_executor.ts, line 48 at r2](https://reviewable.io/reviews/tensorflow/tfjs/2680#-LzMxN8F0pFYP15DsjlY:-LzMxN8F0pFYP15DsjlZ:bsxaig0) ([raw file](https://github.com/tensorflow/tfjs/blob/ecf5e758291f712056c428e262706c83421e22de/tfjs-converter/src/executor/graph_executor.ts#L48)):_
> 
> > ```
> >     const weightIds = Object.keys(weightMap).map(
> >         key => weightMap[key].map(tensor => tensor.id));
> >     this.weightIds = [].concat(...weightIds);
> > ```
> 
> can you try doing this with a regular for loop? I wonder if this transpiles to something suboptimal (also this is more readable).
> 
> e.g.
> 
> ```
> const weightIds = [];
> Object.keys(weightMap).forEach(key => {
>   weightMap[key].forEach(tensor => weightIds.push(tensor.id));
> });
> ```
> 
> _[tfjs-converter/src/operations/operation_mapper.ts, line 59 at r2](https://reviewable.io/reviews/tensorflow/tfjs/2680#-LzMxlt8AX-fN25UQixL:-LzMxlt8AX-fN25UQixM:b-mo2nm) ([raw file](https://github.com/tensorflow/tfjs/blob/ecf5e758291f712056c428e262706c83421e22de/tfjs-converter/src/operations/operation_mapper.ts#L59)):_
> 
> > ```
> >       sliceJoin, spectral, transformation
> >     ];
> >     const mappersJson: OpMapper[] = [].concat(...ops.map(op => op.json));
> > ```
> 
> Same here make this a vanilla loop to avoid this magic

Hi @nsthorat , this timeout issue happens on all the Core related PRs, 
https://github.com/tensorflow/tfjs/pull/2696, https://github.com/tensorflow/tfjs/pull/2647. I tried reverting all my changes in one of previous commits, it still times out. Also, another evidence this isn't related to code is, the test can pass in local with the same browser version. Also tests can pass with all other browsers too.

Finally browserNoActivityTimeout to 4 mins works. This explains the behavior. Before, the tests must be able to finish running in 3 mins (previous no activity timeout value). Now after we add more tests, it simply cannot finish in 3 mins. So all the PR timeout consistently, no matter what change it is. There's another feature in karma-browserstack-launcher that periodically poke the browser to make it active: https://github.com/karma-runner/karma-browserstack-launcher/pull/168 We can try this once they publish version 1.6.0.

Regarding for loops, foreach, and spread operator, I think spread operator has the best readability.

nsthorat:
Your call regarding readability but I had to open VSCode to understand that it was flatting a double-nested array (the problem is the combination of concat.apply and the spread operator). A vanilla for loop makes it much clearer, but again your call.

PR #2691, opened 2020-01-22
Update tfjs-react-native to 0.2

tafsiri:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2691)

### DIFF
--- a/tfjs-react-native/src/version.ts
+++ b/tfjs-react-native/src/version.ts
@@ -1,0 +1,5 @@
+/** @license See the LICENSE file. */
+
+// This code is auto-generated, do not modify this file!
+const version = '0.2';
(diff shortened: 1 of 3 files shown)


tfjs-react-native/src/version.ts
```diff
@@ -0,0 +1,5 @@
+/** @license See the LICENSE file. */
+
+// This code is auto-generated, do not modify this file!
+const version = '0.2';
+export {version};
```
dsmilkov: re-export `version` from `index.ts` to be consistent with the other packages

tafsiri: Done!

PR #2701, opened 2020-01-27
improve indexedDB error logging

luucv:
In some cases where the saving of the model to indexedDB failed, the roll-back of the info entry also fails without calling `deleteInfoRequest.onerror`. This results in a not very helpful error message to the developer:

Screenshot 2020-01-27 at 14 07 27

This PR makes the actual error message clear to the developer by returning the error on the roll-back:

Screenshot 2020-01-27 at 14 10 12

It took me some time to figure out why saving my model to indexedDB was failing. I think this will improve debugging when failing to save to the indexedDB.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2701)


luucv:
@googlebot I signed it!

PR #2702, opened 2020-01-27
Improves indexedDB error logging on save

luucv:
In some cases where the saving of the model to indexedDB failed, the roll-back of the info entry also fails without calling deleteInfoRequest.onerror. This results in a not very helpful error message to the developer:

Screenshot 2020-01-27 at 14 10 12

This PR makes the actual error message clear to the developer by returning the error on the roll-back:

Screenshot 2020-01-27 at 14 07 27

It took me some time to figure out why saving my model to indexedDB was failing. I think this will improve debugging when failing to save to the indexedDB.

### DIFF
--- a/tfjs-core/src/io/indexed_db.ts
+++ b/tfjs-core/src/io/indexed_db.ts
@@ -178,7 +178,12 @@ export class BrowserIndexedDB implements IOHandler {
               // If the put-model request fails, roll back the info entry as
               // well.
               infoStore = infoTx.objectStore(INFO_STORE_NAME);
-              const deleteInfoRequest = infoStore.delete(this.modelPath);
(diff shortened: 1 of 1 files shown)


PratsBhatt:
Hello @luucv,

I am facing similar issue. Were you able to solve the problem or were you able to pinpoint the issue?
I will be grateful if you could direct me.

Thanks and Regards,
Prateek Bhatt

luucv:
Hi @PratsBhatt, with similar issue do you mean that the model is too big for the indexedDB? If so, yes, I am working on something which works for bigger models, but it is not optimised yet and will still need some iterations to improve the API/functionalities. You can see the WIP here: https://github.com/luucv/tfjs-indexeddb-actions

PR #2714, opened 2020-01-30
[tfjs-node] Upgrade libtensorflow version

caisq:
1.15.0 --> 2.1.0

Without this update, the latest versions of tfjs-node (Python) tensorflow
depend on different CUDA versions (10.0 vs. 10.1), which causes
confusion and inconvenience to users who use TF in both languages.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2714)

### DIFF
--- a/tfjs-node/scripts/deps-constants.js
+++ b/tfjs-node/scripts/deps-constants.js
@@ -22,7 +22,7 @@ const modulePath =
     module_path_napi.replace('{napi_build_version}', process.versions.napi);
 
 /** Version of the libtensorflow shared library to depend on. */
-const LIBTENSORFLOW_VERSION = '1.15.0';
+const LIBTENSORFLOW_VERSION = '2.1.0';
 
 /** Map the os.arch() to arch string in a file name */
 const ARCH_MAPPING = {


caisq:
@kangyizhang @pyu10055 I'd like to bring your attention to the need to update libtensorflow dependency. See the PR description for details. This PR doesn't pass build yet. But tfjs-node / tfjs-node-gpu need to update to avoid going out of date with tensorflow itself.

yhwang:
@caisq some questions. Based on my understanding, `tfjs-node` and `tfjs-node-gpu` need libtensorflow binaries (shared libs and header files) from here: https://storage.googleapis.com/tensorflow/
I don't see 2.1.0 libraries there yet. I guess that's why CI failed.
And for pre-built node binding, they are stored here:
https://storage.googleapis.com/tf-builds 
I only saw `CPU-linux-1.5.2` there, which means the node-binding will be built on other platforms while installing the `tfjs-node` or `tfjs-node-gpu`. I observed some issues complaining about failures of installation and some seems be related to not able to build the node-binding on the target machine.

Is it possible to provide a list of supported platforms and architectures for `tfjs-node` and `tfjs-node-gpu`? And maybe define a way for the community to help on other platforms and architectures that community may be interested in but google doesn't maintain.

jenperson:
Thank you for your contribution! This PR is being closed because the version of libtensorflow has since been updated.

PR #2728, opened 2020-02-04
[wasm] Add Softmax, Neg, NotEqual.

annxingyuan:
#### Changes

- Add Softmax to `KernelBackend` interface, change softmax op to call kernel directly.
- Added Softmax kernel to WASM backend. Also added neg and notEqual, which are used in certain softmax tests.
- Remove `ffast-math` from wasm build configuration because it causes XNNPACK to produce incorrect outputs.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2728)


tfjs-backend-wasm/src/cc/kernels/Softmax.cc
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    operator_cache.insert({cache_key, softmax_op});
+
+    tfjs::backend::xnn_operator_count++;
+  } else {
+    softmax_op = operator_cache_idx->second;
+  }
+
+  xnn_status status = xnn_setup_softmax_nc_f32(softmax_op, 1, x_buf, out_buf,
```
Maratyszcza: Why do you hard-code batch size 1?

tfjs-backend-wasm/src/cc/kernels/Softmax_test.cc
```diff
@@ -0,0 +1,58 @@
...
+
+  ASSERT_EQ(0, tfjs::backend::num_tensors());
+
+  const size_t x0_id = 1;
+  const size_t x1_id = 2;
+  const size_t x_size = 4;
+  float x_values[x_size] = {1, 2, 2, 2};
+
+  const size_t out_id = 3;
+  const size_t out_size = 8;
```
Maratyszcza: `out_size` > `x_size`!

tfjs-backend-wasm/src/cc/kernels/Softmax.cc
```diff
@@ -0,0 +1,115 @@
...
+  for (size_t i = 0; i < batch; ++i) {
+    for (size_t j = 0; j < channels; ++j) {
+      tfjs::util::log("%f", x_buf[i * channels + j]);
+    }
+  }
+  // prints: 1.0, 2.0, 2.0, 2.0
+
+  xnn_operator_t softmax_op = nullptr;
+
+  OperatorCacheKey cache_key = {channels, batch};
```
nsthorat: don't cache on batch since it's not used in create, maybe this is the problem?

Maratyszcza: it needlessly increase cache size, but shouldn't affect correctness per se

PR #2734, opened 2020-02-07
[converter] Remove use of browserstack

lina128:
Converter tests don't need browser to run. Remove browser brings down the test from ~8min to ~5min.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2734)


nsthorat:
Can you also make sure coverage hasn't decreased by running master against the CPU backend and make sure its the same number (looks like 470: https://pantheon.corp.google.com/cloud-build/builds/b3e40bd6-7734-4b76-98f8-576f3a2a3bb9;step=2?project=learnjs-174218)

lina128:
Hi @nsthorat , thank you for your review! I cleaned up cloudbuild and package.json. Can you re-review the PR? And yes, the converage is the same.

lina128:
> > Reviewed 2 of 3 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2734)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @dsmilkov, @lina128, and @pyu10055)
> 
> _[tfjs-converter/cloudbuild.yml, line 62 at r3](https://reviewable.io/reviews/tensorflow/tfjs/2734#-M0ZN8Yh6i0hoRn6zrpd:-M0ZN8YiDFVL9CrxGs4O:bpfoztn) ([raw file](https://github.com/tensorflow/tfjs/blob/e2f7b43015b837d46de68e35e70cf467a15b46b3/tfjs-converter/cloudbuild.yml#L62)):_
> 
> > ```yaml
> >   entrypoint: 'bash'
> >   args: ['./run-python-tests.sh']
> >   waitFor: ['python-build']
> > ```
> 
> how come do we have to wait for python-build now?

Ah, you are right, I thought the previous step is to build package, it's actually just testing package, so they can run in parallel. Changed back to before.

Issue #2729, opened 2020-02-04
[tfjs-react-native] and managed Expo workflows


robsco-git:
With, `@tensorflow/tfjs-react-native@0.2.1`, `tfjs-react-native` is using `react-native-fs` for the [bundleResourceIO](https://github.com/tensorflow/tfjs/blob/90f52ddb341d5079925af81000edfe845e40774a/tfjs-react-native/src/bundle_resource_io.ts#L21) function. To the best of my knowledge, managed `Expo` workflows do not work with `react-native-fs`, instead, the recommended path is to use [expo-file-system](https://docs.expo.io/versions/latest/sdk/filesystem/). Attempting to run a managed Expo app results in this error at runtime: `TypeError: null is not an object (evaluating 'RNFSManager.RNFSFileTypeRegular')`.

To work around this, I [forked the repo](https://github.com/robsco-git/tfjs/tree/expo-react-native),  and changed the bundle loading code into a stub. I [published this code to NPM](https://www.npmjs.com/package/tfjs-react-native-expo-fix), installed it and ran it successfully in the context of a managed Expo workflow. I managed to get the [example realtime demo](https://github.com/tensorflow/tfjs/blob/master/tfjs-react-native/integration_rn59/components/webcam/realtime_demo.tsx) working.

Based on the [README](https://github.com/tensorflow/tfjs/blob/master/tfjs-react-native/README.md), it appears that managed Expo workflows are within the set of project types supported by `tfjs-react-native`, hence this issue.

tafsiri:
Yep will be looking into an alternative that will work with the managed workflow. Thanks for reporting.

jgough99:
@robsco-git I am having exactly the same issue so I am excited to see you have published a work around. However I am struggling to get tfjs-react-native-expo-fix to work with my project since I am getting many errors. Could you give some advice on how to use this fix please? Does it need to be imported? Do you still need to install the non-fixed tensoflow? Many thanks.

MichaelMcNeil:
@jgough99  same. @robsco-git what's the workflow to run with your fix rn?

tafsiri:
This was autoclosed by a PR, but the release including it should come out shortly.

MichaelMcNeil:
@tafsiri thank you! I was a little mixed up after re-testing when I saw this closed. Looking forward to it!

tafsiri:
[...]

PR #2740, opened 2020-02-10
[tfjs-react-native] Use dynamic import of react-native-fs

tafsiri:
This should resolve https://github.com/tensorflow/tfjs/issues/2729 (once released).

Expo filesystem does not provide the functionality we need so this moves react-native-fs into a dynamic import so that managed expo apps can use the package. Managed expo apps should not hit the code path where this new require statement is used as their assets are always served from the web.

I'm also including an overdue README update.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2740)


tafsiri:
Tested by making a managed expo app and using this version locally.

PR #2747, opened 2020-02-12
[tfjs-react-native] Don't flipHorizontal if using back camera on iOS

robsco-git:
Bugfix: On iOS, using the back camera with `cameraWithTensors` would result is a horizontally mirrored viewfinder render. Android works as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2747)


tafsiri:
published in 0.2.3

PR #2774, opened 2020-02-19
[tfjs-converter] Allow custom shard size for more conversion pairs

pvaneck:
Previously, only the tfjs_layers_model to tfjs_layers_model conversion pair worked with the `weight_shard_size_bytes` flag. This PR allows the flag to be used for any conversion where the output_format is `tfjs_layers_model` or `tfjs_graph_model`.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2774)


pvaneck:
@pyu10055 Thanks for the review. I added support for the shard_size_param in the wizard.

PR #2775, opened 2020-02-19
[tfjs-core] Modularize squaredDiff

tafsiri:
This PR modularizes squaredDifference in a manner that is non-breaking but consistent with our future plans for modularization. 

A few takeaways I got from this:
- We should defer modularizing tfjs-node kernels until we can do a release of core that exports `kernel_names.ts` 
- We need to keep using runKernelFunc to not break backends outside of core that do not use the registry (e.g. webgpu).
- tfjs-node can continue to co-locate kernel registration with kernel definition as we have no need to bundle size reduction in that context (as there is no bundle).
- Its nice having kernel_names.ts be the single source of truth for kernel names, input specs and attr specs. The latter two could be moved into another file, but i don't think we need to do that. 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2775)


tfjs-node/src/kernels/SquaredDifference.ts
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import {KernelConfig, registerKernel} from '@tensorflow/tfjs-core';
+
+// Can only be done once a core release with these exports is done
```
lina128: Maybe add a test filter? Or after test has changed to test against head?

nsthorat:
Nice work!!

tafsiri:
Changes indicated in last comment are not yet on GitHub due to some service issues with github (https://www.githubstatus.com/). Will push once things are stable.

PR #2776, opened 2020-02-19
[node]Test against head.

lina128:
Change tfjs-node package to test against head in local. It depends on union package, which in turn depends on four packages. yalc doesn't propagate the link, it suggests for this use case, just link the packages in the root package.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2776)


dsmilkov:
Talked offline. Would be better to update package.json to link directly to its transitive deps at master so the committed code reflects the testing

PR #2790, opened 2020-02-25
Explicitly replace "import tensorflow" with "tensorflow.compat.v1"

caisq:
- This PR mirror internal CL/297051045 (author: yanhuasun@)

DEV

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2790)

### DIFF
--- a/tfjs-converter/python/tensorflowjs/converters/converter.py
+++ b/tfjs-converter/python/tensorflowjs/converters/converter.py
@@ -27,8 +27,8 @@ import tempfile
 
 import h5py
 import numpy as np
-import tensorflow as tf
-from tensorflow import keras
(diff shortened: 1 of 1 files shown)


dsmilkov:
I think we are already using 2.x: https://github.com/caisq/tfjs-1/blob/converter-compat-v1-import/tfjs-converter/python/requirements.txt#L4

So in that case I think we just need to opt-out from the internal change. Will let @pyu10055 confirm.

pyu10055:
we don't need this change, I have already update our code to use v1 specifically for places that requires that. All other imports are for v2.

pyu10055:
we don't need this change, I have already update our code to use v1 specifically for places that require that. All other imports are for v2.

PR #2793, opened 2020-02-26
Add affected packages analysis.

lina128:
Introduced dependency.json in this PR. After this PR, when calculating diff, CI will also analyze affected packages and run tests for those packages too. dependency.json is manually maintained, if there is change tfjs package dependencies, please update dependency.json as well.

p.s.: Test of this PR will not be able to validate whether this works or not, because the change is in diff.js, which triggers all builds, and skip the added logic. In order to test the logic, in one of the debug commit, I removed diff.js from the whitelist and made some changes to Core, the build shows it works as expected. You can check the log in that build for commit e604026.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2793)


lina128:
> _[scripts/test-util.js, line 31 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2793#-M16h1ks-4OxL2sOjVmc:-M16h1ks-4OxL2sOjVmd:b-fx3yz5) ([raw file](https://github.com/tensorflow/tfjs/blob/e794123aca5047f03dd5d8a4779a28a5adec866e/scripts/test-util.js#L31)):_
> 
> > ```js
> > // Example:
> > //   dependencyGraph = {
> > //     "tfjs-core": ["tfjs-converter", "tfjs", ...],
> > ```
> 
> any reason the .json file doesn't just look like this map?
Hi Nikhil,
Yeah, we can make the .json file look like this and we can avoid compute the dependency graph each time. However, I think have the .json file mirroring the package.json file of each package, allows us to possibly automate the dependency file. Also it makes adding new package to this file easy, just copy paste from package.json of the new package. But either way has some pros, so I don't have strong preference. What do you think? Which pros is more important?

lina128:
> Great stuff. Tiny comments, non-major thus LGTM
> 
> > Reviewed 1 of 3 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2793#-:-M1791FF2PRKzCxcLHKr:b-c5pxxb)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @dsmilkov, @lina128, @nsthorat, and @pyu10055)
> 
> _[scripts/diff.js, line 113 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2793#-M16hx7_7b9Zexe8M20d:-M178NjG8Jv6X2ycqqRZ:bdyis4j) ([raw file](https://github.com/tensorflow/tfjs/blob/e794123aca5047f03dd5d8a4779a28a5adec866e/scripts/diff.js#L113)):_
> 
> Previously, nsthorat (Nikhil Thorat) wrote…
> I think `triggeredBuilds` is already pre-populated with the touched packages, from the code above.
> 
> _[scripts/diff.js, line 115 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2793#-M16hL5jEdAbT0JDWINt:-M178d8VCL_hkm9rCwgi:b-8rwxl8) ([raw file](https://github.com/tensorflow/tfjs/blob/e794123aca5047f03dd5d8a4779a28a5adec866e/scripts/diff.js#L115)):_
> 
> Previously, nsthorat (Nikhil Thorat) wrote…
> +1 no need to write.
You are right. Removed ''.

> _[scripts/diff.js, line 116 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2793#-M177xRJD1xViDFzwyfc:-M177xRJD1xViDFzwyfd:b-l6zdu5) ([raw file](https://github.com/tensorflow/tfjs/blob/e794123aca5047f03dd5d8a4779a28a5adec866e/scripts/diff.js#L116)):_
> 
> > ```js
> >     affectedPackages.forEach(package => {
> >       writeFileSync(join(package, 'diff'), '');
> >       triggeredBuilds.push(package);
> > ```
> 
> push into a separate array, not the one that you are currently traversing (triggeredBuilds.forEach loop above) to avoid looking into the same items multiple times.

Ah, nice catch. Pushed to a separate array.

lina128:
Hi @nsthorat and @dsmilkov , thank you for your review! I changed the dependency file to simpler format as we discussed offline, such as "tfjs-converter": ["tfjs-core"]. It still reflects package.json structure. Also addressed your other comments too, great comments. Thank you both!

lina128:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2793)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128 and @nsthorat)
> 
> _[scripts/diff.js, line 115 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2793#-M16hL5jEdAbT0JDWINt:-M1CGjRo6iiQCR_qx2Bh:b1y4mkj) ([raw file](https://github.com/tensorflow/tfjs/blob/e794123aca5047f03dd5d8a4779a28a5adec866e/scripts/diff.js#L115)):_
> 
> Previously, lina128 (Na Li) wrote…
> I meant why do you need to write the file at all?

Because we rely on whether the dir has a diff file to trigger cloud build, right? https://github.com/tensorflow/tfjs/blob/master/scripts/run-build.sh#L20

lina128:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2793)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @nsthorat, and @pyu10055)
> 
> _[scripts/diff.js, line 117 at r3](https://reviewable.io/reviews/tensorflow/tfjs/2793#-M1CGLCdBlKYfqD1Zkx1:-M1CGLCdBlKYfqD1Zkx2:b-d9n73r) ([raw file](https://github.com/tensorflow/tfjs/blob/132104fbf578ca12738374dc07c6ad8b85a9a08e/scripts/diff.js#L117)):_
> 
> > ```js
> >         computeAffectedPackages(dependencyGraph, triggeredBuild);
> >     affectedPackages.forEach(package => {
> >       writeFileSync(join(package, 'diff'));
> > ```
> 
> package seems to be the name of the packages, with out the directory information, where you are trying to write this file?
> and how would this file used?

After line 72, I'm in the root directory, so the package name is also the dir name. https://github.com/tensorflow/tfjs/blob/master/scripts/diff.js#L71

 I'm trying to write this file under each affected package. So that when the code detects a diff file, it will trigger cloud build: https://github.com/tensorflow/tfjs/blob/master/scripts/run-build.sh

lina128:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2793)_ status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128)
> 
> _[scripts/diff.js, line 115 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2793#-M16hL5jEdAbT0JDWINt:-M1CoV0D0P3iwbB3pIDD:bse2q4t) ([raw file](https://github.com/tensorflow/tfjs/blob/e794123aca5047f03dd5d8a4779a28a5adec866e/scripts/diff.js#L115)):_
> 
> Previously, lina128 (Na Li) wrote…
> ah, right. can we rename this file from 'diff' to something like 'run-ci'? here and above, and maybe make this file called diff.js something like 'find-affected-packages.js'?

Done. Changed diff to run-ci in three places in find-affected-packages.js file. Also add run-ci to gitignore. Changed file name to find-affected-packages, also changed yarn command, and cloudbuild accordingly.

PR #2799, opened 2020-02-26
[tfjs-converter] Add some fixes to converter wizard

pvaneck:
* Fix dryrun arg not working
  - In `main`, argv is a single element list containing a string representing all the args or an empty string for no args, so changed the checks to correctly detect `--dryrun`
* Enable tfjs-layers to tfjs-layers conversion pair
  - This is available in the `tensorflowjs_converter` CLI, but wasn't for the wizard.
* Change conditions for `split_weights_by_layer` argument
  - This arg only works for `Keras H5 | Keras Saved Model` -> `TFJS Layers` conversion paths.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2799)


pvaneck:
@pyu10055 Hey, I encountered a few issues when using the wizard. Added fixes here :thumbsup:

PR #2812, opened 2020-03-03
WIP Move cpu and webgl backends to new folder structure

tafsiri:
This moves the CPU and WebGL backends to a new folder structure in preparation of more extensive modularisation. It also harmonizes the existing modular kernels (Square, NonMaxSuppressionV5, FromPixels) with the our the latest proposal.

A lot of files are touched but this PR is mostly renames.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2812)


dsmilkov:
Ping me once you decide if you are moving the folders now or later and I'll take a pass

tafsiri:
Notes for discussion on this tomorrow. While I would like to move the backends to the top level early. It turns out that it changes the structure of our 'dist' folder in core. I've been able to pretty much track down all our internal uses of this path however what remains is what we put in our package.json main field and whether we consider a path like https://unpkg.com/@tensorflow/tfjs-core@latest/dist/tf-core.js a part of our public api. If we do, then moving the backends out of src/packages to the top level is a breaking change.

tafsiri:
FYI, I updated/clarified the previous comment in case you read these by email.

tafsiri:
Closing, this work was done in a different PR with a different approach (version bump)

PR #2831, opened 2020-03-05
[layers]Test against head.

lina128:


tfjs-layers/package.json
```diff
@@ -39,11 +39,13 @@
   "scripts": {
     "prep": "yarn install && yarn build",
     "build": "tsc",
+    "build-core": "cd ../tfjs-core && yarn && yarn build",
+    "build-core-ci": "cd ../tfjs-core && yarn && yarn build-ci",
     "build-npm": "./scripts/build-npm.sh",
```
dsmilkov: A suggestion: to defend against publishing layers with a linked core package, we should have `build-npm.sh` check that "@tensorflow/tfjs-core": points to a version and not "link:../tfjs-core" . Likewise for the other build-npm.sh scripts in the other packages.

nsthorat: This is a reasonable suggestion for extra safety but in practice this won’t happen because of the release script :)

nsthorat:
One thing I forgot about, you need to do something like this: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/package.json#L23

This script runs every time we build (or do anything for development) to make sure that when developing we're reverting the changes of the version pinning of the release script back to link:../. You should put the link script in top-level of the union package because it can be shared by all packages.

lina128:
> One thing I forgot about, you need to do something like this: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/package.json#L23
> 
> This script runs every time we build (or do anything for development) to make sure that when developing we're reverting the changes of the version pinning of the release script back to `link:../`. You should put the link script in top-level of the union package because it can be shared by all packages.

Exactly, I noticed that too. Created another PR for that: https://github.com/tensorflow/tfjs/pull/2834

PR #2846, opened 2020-03-09
[webgpu] Fix dilation and add stride 2  filter 3x3 same conv2d test case

axinging:
The cases included in this PR may fail when dilation and padding (padding is fixed in another PR: https://github.com/tensorflow/tfjs/pull/2993) are no properly applied. 

Currently there is only one (fused)conv2d test case, "throws when both stride and dilation are greater than 1", whose strides are [2, 1], and filter size is 2x2, with valid padding. And this case can not capture the dilation and padding misuse very well.

This PR adds stride [2,2] and filter 3x3, with same padding. The pattern is originated from handpose model (https://github.com/tensorflow/tfjs-models/tree/master/handpose). Noted, to make these case pass on WebGPU backend, PR https://github.com/tensorflow/tfjs/pull/2993 should also be applied.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2846)


axinging:
@annxingyuan PTAL.
This PR and the padding PR(https://github.com/tensorflow/tfjs/pull/2993) is one of the key reason why the handpose model fail.

PR #2885, opened 2020-03-12
Release and diffing change.

lina128:
Preparing for release process change. This PR includes below changes:
(1) Make diffing logic more generic. It introduces baseBranch, so the script no longer assume diff against master, it will diff against whatever baseBranch cloud build send. This information is ultimately from GitHub.
(2) Add a step in release script, asking which release branch to send the PR to.
(3) Generate a unique local branch for the PR each time. The branches need to be deleted afterwards.
(4) Automate pull-request to send the PR to the release branch, not master.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2885)

### DIFF
--- a/scripts/find-affected-packages.js
+++ b/scripts/find-affected-packages.js
@@ -43,22 +43,19 @@ if (commitSha == null) {
 if (branchName == null) {
   branchName = exec(`git rev-parse --abbrev-ref HEAD`).stdout.trim();
 }
+if (baseBranch == null) {
+  // For Nightly build, baseBranch is null. We use master for Nightly build.
+  baseBranch = 'master';
+}
 console.log('commitSha: ', commitSha);
 console.log('branchName: ', branchName);
 console.log('baseBranch: ', baseBranch);
 
-const toMaster = baseBranch == 'master';
-
-if (toMaster) {
-  console.log('Clone branch master.');
-  // We cannot do --depth=1 here because we need to check out an old merge base.
-  // We cannot do --single-branch here because we need multiple branches.
-  exec(`git clone https://github.com/tensorflow/tfjs ${CLONE_PATH}`);
-} else {
-  console.log(`Clone branch ${baseBranch}`);
-  exec(`git clone -b ${baseBranch} https://github.com/tensorflow/tfjs ${
-      CLONE_PATH}`);
-}
+// We cannot do --depth=1 here because we need to check out an old merge base.
+// We cannot do --single-branch here because we need multiple branches.
+console.log(`Clone branch ${baseBranch}`);
+exec(`git clone -b ${baseBranch} https://github.com/tensorflow/tfjs ${
+    CLONE_PATH}`);
 
 console.log();  // Break up the console for readability.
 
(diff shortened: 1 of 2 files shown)


lina128:
Hi @dsmilkov , thank you for the review! I changed to use branch name + phase as the branch name. Tested in this PR: https://github.com/tensorflow/tfjs/pull/2888
For the sanity check, I intentionally leave it out. Packages like react and vis, may have their own branch naming pattern for release. Also if the branch doesn't exist, it will exit the process, error message is, for example the branch I entered is jfewfw:
warning: Could not find remote branch jfewfw to clone.
fatal: Remote branch jfewfw not found in upstream origin

I think the error message is clear enough. WDYT?

lina128:
Hi @nsthorat, this release branch is for the entire release unit, not single package, so it's possible packages may not have the same numbers. Our best guess is use core's version number, WDYT?

nsthorat:
Right now we’re assuming all patch versions are equal for the branch, is that right? Or are we moving to a minor version base branch with all patch versions on it? In either case that should be infer-able. Let me know where I’m missing something.

lina128:
> Right now we’re assuming all patch versions are equal for the branch, is that right? Or are we moving to a minor version base branch with all patch versions on it? In either case that should be infer-able. Let me know where I’m missing something.

Hi Nikhil, as discussed, we're assuming all patch versions are equal for the branch. The problem is, if we use Core's version number, say 1.7.0, and make a branch 1.7.1, after Core is published, Core's version number is bumped to 1.7.1. Then when releasing other packages, the inferred branch will be 1.7.2, but it should be 1.7.1.

nsthorat:
I don’t quite follow. We’ve decided all of them should be the same. Then only in phase 1 do we make the branch. For phase 2+ we just reuse that branch.

lina128:
> I don’t quite follow. We’ve decided all of them should be the same. Then only in phase 1 do we make the branch. For phase 2+ we just reuse that branch.

Hi Nikhil, I added inferred branch logic, please re-review the PR. This logic has assumption that core will always be published in a release cycle. This assumption should better be backed by release rules. I sent you a release process doc with the new changes. Please review that doc as well. Thank you.

nsthorat:
I will take a look at the doc! I’m on my phone now so as long as that happened LGTM! Small note you don’t have to tie it to core, rather whatever happens to be in Phase 1 (could be something else in the future)!

lina128:
> I will take a look at the doc! I’m on my phone now so as long as that happened LGTM! Small note you don’t have to tie it to core, rather whatever happens to be in Phase 1 (could be something else in the future)!

If we tie it to whatever in Phase 1, we have another assumption, that phase only has one package to publish. Or instead, another assumption, the packages in that phase all have same version numbers. This PR uses Core, as this is closest to reality. WDYT?

nsthorat:
How come that assumption needs to hold?

Example for 1.7.1 if there were 2 packages.
- Phase 1: a) create release branch b1.7.1. b) create dev branch b1.7.1-dev that changes multiple package.json versions, create PR into b1.7.1
- Phase 2...

lina128:
> How come that assumption needs to hold?
> 
> Example for 1.7.1 if there were 2 packages.
> 
> * Phase 1: a) create release branch b1.7.1. b) create dev branch b1.7.1-dev that changes multiple package.json versions, create PR into b1.7.1
> * Phase 2...

If there are two packages in Phase 1, and they have different version numbers, what is the inferred branch name?

nsthorat:
Ah I see the problem. Each release unit then needs a separate level of hierarchy. First, ask the user which release unit. Each release unit then has phases. Does that make sense?

lina128:
> Ah I see the problem. Each release unit then needs a separate level of hierarchy. First, ask the user which release unit. Each release unit then has phases. Does that make sense?

Exactly :)

PR #2912, opened 2020-03-15
core: update the @bazel/bazel to recommended bazellisk

yorkie:
Installing the `@bazel/bazel` via YARN is hardly working because of some download issues, and @bazel/bazelisk could fix the network issues, I guess the failures of using `@bazel/bazel` is because it tries to download almost 300MB when installing, and the later is 6MB.

For more details, see:

- https://github.com/bazelbuild/rules_nodejs/wiki#migrating-to-bazelbazelisk-package.

> To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2912)


yorkie:
See https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run-and-where-to-get-it-from, we are using the official latest Bazel release if we don't have this version file, @dsmilkov do you know which version should we declare here?

yorkie:
@dsmilkov @kangyizhang Added bazel version file, thank you :)

PR #2925, opened 2020-03-18
move inputsToSave and outputsToSave to gradConfig

tafsiri:
This moves the definition of inputsToSave and outputsToSave for gradient computation onto the gradient config. 

runKernelFunc still takes inputsToSave and outputsToSave as fallbacks for ops currently using those directly. However those two params to runKernelFun are effectively deprecated.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2925)

### DIFF
--- a/tfjs-core/src/engine.ts
+++ b/tfjs-core/src/engine.ts
@@ -535,8 +535,8 @@ export class Engine implements TensorTracker, DataMover {
   runKernelFunc<T extends Tensor|Tensor[], I extends NamedTensorMap>(
       forwardFunc: ForwardFunc<T>, inputs: I,
       backwardsFunc?: (dy: T, saved: Tensor[]) => {[P in keyof I]: () => I[P]},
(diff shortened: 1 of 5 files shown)


tafsiri:
@nikhil. Sorry I forgot to hit publish in reviewable before merging. All fixes have been done. Copying my reviewable comments here.

> what does this comment mean? wasm is linked at master so you can update it here if it's something about WASM

Good catch, this was confusing, I updated the comment to be more precise/corrent. Some ops call runKernelFunc and pass inputsToSave and outputsToSave, however it looks like this was done just for ops that were implemented in WASM backend. So this comment was meant to be a breadcrumb to remember why this codepath exists. I think it was done in anticipation of gradient registration but isn't needed once we change those ops.


> != null

Done

> use null instead of undefined here

Done

nsthorat:
Makes sense, thanks for the explanation!

PR #2928, opened 2020-03-19
Create publish-npm.ts

lina128:
Automate publish npm for new release process. Changes in this PR:
1. Created an interactive script for publishing packages to npm. It is based on the ./scripts/publish-npm.sh, with below differences in terms of logic:
    1.a. Takes two parameters, package (e.g. tfjs-core) and release branch (e.g. tfjs_1.7.1). Note, release branch will be inferred if not the first package in the release unit, otherwise need user input.
    1.b. The new script doesn't run ./scripts/make-version.js. This is already ran in release.ts, so I think no need to make-version again during publish. However, I realized that the version bump in package.json's version field and version.ts will only get updated in each release branch, master won't get updated anymore. We should update master in release.ts script.
    1.c. publish-npm.ts will create temporary dir, and checkout the release branch, and publish from there. Whereas publish-npm.sh just uses master in local branch. So this should be an improvement.
    1.d. Overall flow for publish-npm.sh is: Ask release unit -> Ask phase -> Infer release branch and Ask to correct if needed -> Publish for all the packages in the phase from the release branch.

2. Extract some utils used by both release.ts and publish-npm.ts to a util file.

3. Added `yarn publish-npm`

4. In release.ts, added logic to update package.json and version.ts of master branch.

5. Set yarn to use npm registry as a one-off when publish. Some background, shell changes to Yarn's scope and Yarn uses a different registry. So when we do 'npm publish' in the same shell, it will throw "npm ERR! code ENEEDAUTH". We need to set yarn to use npm's registry. There are three ways to do it, (1) set env YARN_REGISTRY to point to npm as a one off in the current shell, and run yarn publish. (2) add a .yarnrc file in the project root directory with a line 'registry https://registry.npmjs.org/'. We adopt (1), because (2) may be a bigger change, involving regenerating yarn.lock files. See discussions of this issue: https://github.com/yarnpkg/yarn/issues/2935

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2928)

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,1 +1,0 @@
-registry "https://registry.npmjs.org/"
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+registry "https://registry.npmjs.org/"
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "scripts": {
     "find-affected-packages": "./scripts/find-affected-packages.js",
     "release": "ts-node ./scripts/release.ts",
+    "publish-npm": "ts-node ./scripts/publish-npm.ts",
     "release-notes": "ts-node ./scripts/release_notes/release_notes.ts",
     "test-release-notes": "ts-node ./scripts/release_notes/run_tests.ts"
   }
--- a/scripts/publish-npm.ts
+++ b/scripts/publish-npm.ts
@@ -125,7 +125,7 @@ async function main() {
 
     console.log(chalk.magenta.bold(`~~~ Publishing ${pkg} to npm ~~~`));
     shell.cd(pkg);
-    $('npm publish');
+    $('YARN_REGISTRY="https://registry.npmjs.org/" yarn publish');
     console.log(`Yay! Published ${pkg} to npm.`);
 
     shell.cd('..');
--- a/scripts/publish-npm.ts
+++ b/scripts/publish-npm.ts
@@ -75,9 +75,9 @@ async function main() {
   const branches =
     Array.from(branchesStr.split(/\n/)).map(line => line.toString().trim());
 
-  // Find the first/latest matching branch, e.g. tfjs_1.7.1
+  // Find the latest matching branch, e.g. tfjs_1.7.1
   // It will not match temprary generated branches such as tfjs_1.7.1_phase0.
-  const exp = `^${name}_([^_]+)$`;
+  const exp = "^" + name + "_([^_]+)$";
   const regObj = new RegExp(exp);
   const maybeBranch = branches.find(branch => branch.match(regObj));
   releaseBranch = await question(`Which branch to publish from
--- a/scripts/publish-npm.ts
+++ b/scripts/publish-npm.ts
@@ -25,10 +25,8 @@
 import * as argparse from 'argparse';
 import chalk from 'chalk';
 import * as mkdirp from 'mkdirp';
-import * as readline from 'readline';
 import * as shell from 'shelljs';
 import {RELEASE_UNITS, question, $, printReleaseUnit} from './release-util';
-import {arrayify} from 'tslint/lib/utils';
 
 const TMP_DIR = '/tmp/tfjs-publish';
 
@@ -99,7 +97,7 @@ async function main() {
   shell.cd(TMP_DIR);
   console.log();
 
-  console.log('~~~ Installing packages ~~~');
+  console.log(chalk.magenta.bold('~~~ Installing packages ~~~'));
   // Yarn in the top-level and in the directory.
   $('yarn');
   shell.cd(pkg);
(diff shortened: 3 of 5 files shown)


lina128:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2928)_ status: 0 of 1 approvals obtained (waiting on @dsmilkov, @lina128, @nsthorat, and @pyu10055)
> 
> _[scripts/publish-npm.ts, line 51 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2928#-M2nBiBB5klbkMLJT6nd:-M2nBiBB5klbkMLJT6ne:b-x5v6of) ([raw file](https://github.com/tensorflow/tfjs/blob/87c56a222f9ebf1c57c47f76db385159779e946d/scripts/publish-npm.ts#L51)):_
> 
> > ```
> >   const releaseUnitStr =
> >     await question('Which release unit (leave empty for 0): ');
> > ```
> 
> can't we infer everything from two pieces of information:
> 
> 1. release unit 2) version
> 
> Then we can look at the branch, read the packages that have versions and not released. Tell the user they are about to publish these packages, and then publish?
> 
> This minimizes the amount that the user has to do.
> 
> _[scripts/publish-npm.ts, line 119 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2928#-M2nBzZtEDjFe1R2vfsc:-M2nBzZuCmPeVrJ_hwj4:bh5uej2) ([raw file](https://github.com/tensorflow/tfjs/blob/87c56a222f9ebf1c57c47f76db385159779e946d/scripts/publish-npm.ts#L119)):_
> 
> > ```
> >   console.log(chalk.magenta.bold('~~~ Publishing to npm ~~~'));
> >   shell.cd(pkg);
> >   $('npm publish');
> > ```
> 
> this might have problems. when you run yarn the authentication is somehow scoped to the yarn command and doesn't read global npm information.
> 
> can you check if there's an npm command to see if you're logged in and print it just to make sure this will work from a yarn script?

Hi Nikhil,
Thank you for the review! As discussed offline, typing version is error prone. And thanks for your suggestion, change to below flow: Ask release unit -> Ask phase -> Infer release branch and Ask to correct if needed -> Publish for all the packages in the phase from the release branch.

And thanks for pointing out the npm issue. I tried with 'npm whoami', yes it will throw error. Adding a .yarnrc config file and specify npm as the registry solves the issue. Tested locally using 'npm whoami' So basically, we need to set yarn to use npm's registry. There are two ways to do it, (1) add a .yarnrc file in the project root directory with a line 'registry https://registry.npmjs.org/'. (2) Run this command $('yarn config set registry "https://registry.npmjs.org/"'). We adopt (1), because it is scoped. We can't use (2) because that command will silently add the line 'registry https://registry.npmsjs.org/' to the user's global ~/.yarnrc. Our community users may have private registry or other registry setup in their environment. So (2) is dangerous. See discussions of this issue: yarnpkg/yarn#2935

I updated the changes, and also this PR's description. Please re-review it. Thank you!

lina128:
> LGTM on my side (one small comment), but wait for Nikhil.
> 
> > Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2928#-:-M2nrZ0x1dkTOVe-FljY:b9drjtm)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @dsmilkov, @lina128, @nsthorat, and @pyu10055)
> 
> _[scripts/publish-npm.ts, line 119 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2928#-M2nBzZtEDjFe1R2vfsc:-M2nqfEo2H8B7Qx-sQfR:b-sucqaj) ([raw file](https://github.com/tensorflow/tfjs/blob/87c56a222f9ebf1c57c47f76db385159779e946d/scripts/publish-npm.ts#L119)):_
> 
> Previously, nsthorat (Nikhil Thorat) wrote…
> +1. This is why we kept `npm publish` in a shell script. Though I'm sure there is a workaround to propagate the global npm settings/key in node.
> 
> _[scripts/release.ts, line 269 at r2](https://reviewable.io/reviews/tensorflow/tfjs/2928#-M2n_QM0-0cHGrMy6xuD:-M2nq6j_011w3bU-JCfq:bhdzpce) ([raw file](https://github.com/tensorflow/tfjs/blob/5bebbdc1bbea073f19b250993f1ba76b3dd12ec7/scripts/release.ts#L269)):_
> 
> Previously, pyu10055 (Ping Yu) wrote…
> I think this is a temporary branch to sending a PR with the version bump for others to review. The destination branch is the same for all phases in a release unit. Na, is that right?
> 
> _[scripts/release.ts, line 303 at r2](https://reviewable.io/reviews/tensorflow/tfjs/2928#-M2nrUM_6EHu2LM-BEjt:-M2nrUM_6EHu2LM-BEju:b-igewan) ([raw file](https://github.com/tensorflow/tfjs/blob/5bebbdc1bbea073f19b250993f1ba76b3dd12ec7/scripts/release.ts#L303)):_
> 
> > ```
> >   console.log(
> >     `Done. FYI, this script does not publish to NPM. ` +
> >     `Please publish by running ./scripts/publish-npm.sh ` +
> > ```
> 
> update this comment to point to publish-npm.ts

Hi Daniel,
Thank you for your review! Updated the release script to point to new command 'yarn publish-npm', that's a nice catch! And thanks for the accurate summarization of the logic of temporary branches and release branch. Regarding npm publish, I finally found a solution, I updated the PR description about this point5. Please re-review the change. Thank you!

PR #2929, opened 2020-03-19
[webgpu] reduce workPerThread size when dispatch equal 1

haoyunfeix:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2929)


haoyunfeix:
@annxingyuan PTAL

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -58,6 +58,19 @@ export function computeDispatch(
...
 }
 
+export function computeWorkPerThreadForMatMul(
+    layout: {x: number[], y?: number[], z?: number[]},
+    outputShape: number[],
+    workGroupSize: [number, number, number],
+    workPerThread: number): number {
+  const dispatch = computeDispatch(layout, outputShape, workGroupSize,
+      [workPerThread, workPerThread, 1]);
+  if (arrayProduct(dispatch) === 1) {
```
qjia7, line 68: Can you add a comment to explain why the WPT should be 1 when only dispatching one work group? And add a test case to show the perf gain?

haoyunfeix: Done. PTAL

qjia7: Hi  Yunfei, your comments confused me. Maybe your want to express that 'If dispaching number is one, it means only one work group is running. For modern GPUs, it supports multiple work groups running in parallel. So there may be some idle hardware threads. In this case, we prefer to reduce the work per thread and improve the thread utilization'. If you think it's better, feel free to use it or make some changes based on it.

tfjs-backend-webgpu/src/kernels/matmul_packed_webgpu.ts
```diff
@@ -136,6 +136,13 @@ export class MatMulPackedProgram implements WebGPUProgram {
     const bShape = transposeB ? [outputShape[0], dimBOuter, dimInner] :
                                 [outputShape[0], dimInner, dimBOuter];
     this.outputShape = outputShape;
+    this.dispatchLayout = {x: [2], y: [1], z: [0]};
+    workPerThread = computeWorkPerThreadForMatMul(
```
annxingyuan, line 140: initialize this variable with `const`

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -58,6 +58,26 @@ export function computeDispatch(
   ];
 }
 
+export function computeWorkPerThreadForMatMul(
```
annxingyuan, line 61: Is this function being used anywhere else? Maybe it would make sense to just unroll this logic within `matmul_packed_webgpu` - otherwise it's a little bit confusing to follow the logic, especially because this function accepts `workPerThread` as an argument while calculating a new `workPerThread`.

tfjs-backend-webgpu/src/webgpu_util.ts
annxingyuan: +1 to @qjia7's comment!

haoyunfeix: Done. PTAL

tfjs-backend-webgpu/src/webgpu_util.ts
haoyunfeix: Done. PTAL

PR #2946, opened 2020-03-23
[core] Modularize Div, DivNoNan.

annxingyuan:
#### Changes
- Modularize `div` op.
- Add modular binary kernel construction helpers.
- Add `div` kernels for WebGL / CPU. Note that WebGL `div` kernel exports an impl function - this will be useful in a follow-up PR for adding `softmax` which invokes `div` as an intermediate kernel.
- Modified `MathBackendWebGL` so that `runWebGLProgram` takes an optional output `TensorInfo`.
- Modularize `divNoNan` op.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2946)


nsthorat:
Wait for @tafsiri :)

tfjs-core/src/backends/webgl/backend_webgl.ts
```diff
@@ -1597,7 +1585,7 @@ export class MathBackendWebGL extends KernelBackend {
     const b = this.exp(a);
     const sumExp = this.sum(b, axes).reshape(expandedShape);
 
-    return this.realDivide(b, sumExp) as T;
+    return b.div(sumExp);
```
lina128: Hi Ann, isn't it going back to the public API? I thought we want to avoid this as much as possible, see this paragraph in the modularization design doc: "When writing a kernel within a backend it is quite reasonable to call other kernels. We want to avoid using the public api to do this. Thus if functionality from another kernel is needed, that functionality should be factored into a separate file (with no side effects) and imported and used directly." Or am I missing something?

tafsiri:
Note I retracted the comment about removing realDivide being a breaking change. I mistook which file that was.

tfjs-core/src/backends/webgl/backend_webgl.ts
lina128: Discussed offline, this is a intermediate step to avoid circular dependency. LGTM.

PR #2949, opened 2020-03-23
Update tfjs-core to 1.7.1.

lina128:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2949)

### DIFF
--- a/tfjs-core/package.json
+++ b/tfjs-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow/tfjs-core",
-  "version": "1.7.0",
+  "version": "1.7.1",
   "description": "Hardware-accelerated JavaScript library for machine intelligence",
   "private": false,
   "main": "dist/index.js",
--- a/tfjs-core/src/version.ts
+++ b/tfjs-core/src/version.ts
@@ -1,5 +1,5 @@
 /** @license See the LICENSE file. */
 
 // This code is auto-generated, do not modify this file!
-const version = '1.7.0';
+const version = '1.7.1';
 export {version};


nsthorat:
this is a duplicate PR, intended?

lina128:
> this is a duplicate PR, intended?

This is updating master branch.

nsthorat:
I thought we weren't updating master with pinned versions?

lina128:
> I thought we weren't updating master with pinned versions?

Updating master is a subset of updating in release branch. In master, we only update the package version field and version.ts, so it's only for informational purpose. We DONOT update dependency link to pinned version. Another choice is we don't update master at all. This doesn't affect development or anything, just makes our repo look like never get updated. We can also just remove version.ts and version field in package.json, but this is more dangerous, I have to check whether any logic depends on them. Which option do you prefer?

lina128:
SGTM

lina128:
> How about doing "version": "master" for all packages at master. This way if we build a bundle from un-released components/packages, tf.version will show us that
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2949#-:-M38B6wt2eepUMSO78iA:bcyq4v2)_ status: 0 of 1 approvals obtained (waiting on @dsmilkov, @nsthorat, and @pyu10055)

I just looked up npm version field, it says it has to be passable by node-semver. But the field itself can be optional. https://docs.npmjs.com/files/package.json

So maybe just omit the field in master branch? My original thought behind still updating master was for anyone who browse the repo to quickly find out what is the latest version. But we can just add this info in read.me.

PR #2952, opened 2020-03-23
Fix bugs in release and publish scripts.

lina128:
Tried the new release scripts, this PR fixes below issues:
1. Change from `yarn publish` to `yarn publish-npm`. Context: yarn publish does not work, it throws error: "Can't answer a question unless a user TTY", same issue is discussed in https://github.com/lerna/lerna/issues/1200. The reason is because yarn publish has an interactive cli, which makes it impossible to use for automation process. Tried the yarn --non-interactive publish method discussed in the thread, doesn't work either. So change to yarn publish-npm, tested locally using npm whoami as a substitute. Yarn 2.x has `yarn npm publish` natively. So once we upgrade to yarn 2.x, we don't need to have publish-npm: "npm publish" in each package.json.
  
2. Combined tag version steps. Context: This script is called from the publish-npm.ts. It didn't throw any errors. But actually, only the first step that creates the tag succeed, the second step that pushes the tag didn't get executed. Combining the two steps works. It is something related to child processes.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2952)

### DIFF
--- a/scripts/publish-npm.ts
+++ b/scripts/publish-npm.ts
@@ -125,7 +125,7 @@ async function main() {
 
     console.log(chalk.magenta.bold(`~~~ Publishing ${pkg} to npm ~~~`));
     shell.cd(pkg);
-    $('YARN_REGISTRY="https://registry.npmjs.org/" yarn publish');
+    $('YARN_REGISTRY="https://registry.npmjs.org/" yarn npm-publish');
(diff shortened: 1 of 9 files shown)


dsmilkov:
Sounds good. Thanks for checking. LGTM

PR #2964, opened 2020-03-25
[tfjs-core] modularize pad op

tafsiri:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2964)


tafsiri:
@annxingyuan I added pad to register_all_chained_ops, so for now we can use it for free. We are keeping this by default for backwards compability in the near future, but will remove it in the further future, which will force those files to import the chain augmentor themselves.

annxingyuan:
@tafsiri oh right forgot about register_all_chained_ops - thanks :)

PR #2955, opened 2020-03-24
[core] Modularize max.

annxingyuan:
#### Changes

- Modularize 'max' (split out op definition and gradient).
- Create modular max kernels for CPU / WebGL backends.
- Split out reduce and reshape utils for WebGL backend.
- Export tfjs-backend-cpu shared namespace.

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2955)

PR #2971, opened 2020-03-26
[core] Modularize transpose op and kernel.

annxingyuan:
This is a prerequisite for https://github.com/tensorflow/tfjs/pull/2955

Something to note is that we can't require the transpose impl functions from `backend_webgl` or `backend_cpu` because that would entail a circular dependency, so for now, when we remove a kernel from the backend, we have to import the op itself if we still want to use it within the backend.

Whenever kernels move out of the backend they can directly use the impl functions from other kernels.

#### Changes
- Modularize transpose op / gradient.
- Modularize transpose kernels for WebGL / CPU backends.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2971)


nsthorat:
Wait for @tafsiri of course.

PR #2762, opened 2020-02-16
Update package.json

etiennedupont:
Solves #2761

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2762)

Issue #2761, opened 2020-02-16
Explicitely make tfjs-core a dependency of tfjs-node


etiennedupont:
#### TensorFlow.js version
"@tensorflow/tfjs-node": "1.5.2"

#### Browser version
node v10.15.3
yarn (pnp) 1.16.0

#### Describe the problem or feature request
"tfjs-node" requires "tfjs-core" in there: https://github.com/tensorflow/tfjs/blob/48ad61389ca1a1fc04425a0ee433e1593b4e4004/tfjs-node/src/nodejs_kernel_backend.ts#L18

So "tfjs-core" should be listed as one of its dependencies there: https://github.com/tensorflow/tfjs/blob/48ad61389ca1a1fc04425a0ee433e1593b4e4004/tfjs-node/package.json#L53

This causes a bug with yarn pnp.

#### Code to reproduce the bug / link to feature request
**package.json**:
```js
{
  "name": "test-tfjs",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "@tensorflow/tfjs-node": "1.5.2",
    "seedrandom":"~2.4.3"
  },
  "devDependencies": {
    "@babel/cli": "7.8.4",
    "@babel/core": "7.8.4",
    "@babel/preset-env": "7.8.4"
  },
  "babel":{
    "presets": ["@babel/preset-env"],
    "plugins": []
  },
  "scripts": {
    "build": "babel src -d lib"
  },
  "installConfig": {
    "pnp": true
  }
}
```

**src/index.js**
```js
import * as tf from '@tensorflow/tfjs-node'

(() => {
    const tensor = tf.tensor([1,1])
    console.log(`tensors = ${tensor}`)
})()
```

Running:
`yarn install && yarn build && yarn node lib`

throws:

```
yarn install v1.16.0
info No lockfile found.
[1/5] 🔍  Resolving packages...
[2/5] 🚚  Fetching packages...
[3/5] 🔗  Linking dependencies...
[5/5] 🔨  Building fresh packages...
success Saved lockfile.
✨  Done in 12.26s.
yarn run v1.16.0
$ babel src -d lib
Successfully compiled 1 file with Babel.
✨  Done in 0.66s.
yarn node v1.16.0
/Users/etiennedupont/Code/test-tfjs/.pnp.js:4636
    throw firstError;
    ^

[...]

PR #2984, opened 2020-03-27
Fix release script.

lina128:
This PR fixes issues found during release. Note, the issues were found right at the spot and were fixed right away. The release versions were correctly published, so the fixes in this PR were tested in the release of 1.7.1. 

This PR has below changes:

1. Update dependency list, because Layers is recently added as a dependency for Node: https://github.com/tensorflow/tfjs/blob/master/tfjs-node/package.json#L50

2. node-gpu needs an additional step to build.

3. wasm needs an additional step to build.

4. The publish script has to use this command to run `YARN_REGISTRY="https://registry.npmjs.org/" yarn publish-npm`. The previous try that adds npm registry information in the last step works on my macbook but not on workstation. On workstation, it is still using yarn registry. Above command works on both macbook and workstation.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2984)

### DIFF
--- a/scripts/publish-npm.ts
+++ b/scripts/publish-npm.ts
@@ -112,6 +112,11 @@ async function main() {
 
     console.log(chalk.magenta.bold(`~~~ Preparing package ${pkg}~~~`));
     console.log(chalk.magenta('~~~ Installing packages ~~~'));
+    // tfjs-node-gpu needs to get some files from tfjs-node.
+    if (pkg === "tfjs-node-gpu") {
+      $('yarn prep-gpu');
+    }
+
     // Yarn above the other checks to make sure yarn doesn't change the lock file.
     $('yarn');
 
--- a/scripts/release-util.ts
+++ b/scripts/release-util.ts
@@ -46,6 +46,9 @@ export const UNION_PHASE: Phase = {
   deps: ['tfjs-core', 'tfjs-layers', 'tfjs-converter', 'tfjs-data']
 };
 
+// We added tfjs-core and tfjs-layers because Node has unit tests that directly
+// use tf.core and tf.layers to test serialization of models. Consider moving
+// the test to tf.layers.
 export const NODE_PHASE: Phase = {
   packages: ['tfjs-node', 'tfjs-node-gpu'],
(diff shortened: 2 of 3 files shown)


lina128:
> Thank you Na for putting this together which makes the release process much more smooth!
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/2984#-:-M3TwNNU726Rs4aiL4mU:bs3datx)_ status: 0 of 1 approvals obtained (waiting on @dsmilkov, @kangyizhang, @lina128, @nsthorat, and @pyu10055)
> 
> _[scripts/release-util.ts, line 51 at r1](https://reviewable.io/reviews/tensorflow/tfjs/2984#-M3Tw2vI2DtW6MDts0J0:-M3Tw2vI2DtW6MDts0J1:bop9r6u) ([raw file](https://github.com/tensorflow/tfjs/blob/fcc2170da2bb97d4b6f6a22b61f64223e37ebf75/scripts/release-util.ts#L51)):_
> 
> > ```
> > export const NODE_PHASE: Phase = {
> >   packages: ['tfjs-node', 'tfjs-node-gpu'],
> >   deps: ['tfjs', 'tfjs-core', 'tfjs-layers'],
> > ```
> 
> I saw that tfjs-layers is added as devDependency in https://github.com/tensorflow/tfjs/pull/2854/files#diff-48e8607fcaefdc233be729e07c18f766R48. Can you provide some context why we did this? tfjs-node depends on the tfjs union package, so it should already have tfjs-layers and tfjs-core installed as dependency.

Hi Kangyi, thanks for the review. We added tfjs-core and tfjs-layers because they were directly used in the code instead of through tfjs, see https://github.com/tensorflow/tfjs/blob/master/tfjs-node/src/io/file_system_test.ts#L19-L20. If they are directly used, they need to be in the dependency list, see this discussion: https://github.com/tensorflow/tfjs/pull/2762 and  https://github.com/tensorflow/tfjs/issues/2761

lina128:
> Node has unit tests that use tf.layers to test serialization of models in node.

Done. Thank you for review, Daniel!

PR #3002, opened 2020-03-31
[core]Modularize addN.

lina128:
Changes in this PR include:
(1) Split AddN out from binary_ops.
(2) Split AddN's gradient out to a separate file.
(3) Add logic to engine to handle array of inputs to save.

Context for (3), the gradient modularization requires a gradConfig for each gradient function in their separate file. This gradConfig has a inputsToSave field, currently the logic is that inputsToSave field will specify tensors to save: https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/engine.ts#L669. But it couldn't handle the case that the inputs have dynamic length of Tensors to save. To deal with this, we can introduce a flag saveAllInputs to gradConfig, which indicates to engine that simply save all the inputs.

An alternative is to change the inputsToSave type from Tensor[] to Array<Tensor|Tensor[]>, but this generates more issues than desired, for example, a lot of the places (logging, profiling, backward functions, etc.) expect the simple form Tensor[], and has logic to access each array element and their shape or dtype, etc. Changing will cause to add more logic to evaluate whether it is an array or a tensor. See this PR for the places to change, and this PR still fails with 10 more places to fix, at which point, I gave up on this solution: https://github.com/tensorflow/tfjs/compare/master...lina128:addn_try?expand=1  

I think the first solution is much simpler and elegant.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3002)


tfjs-core/src/kernel_names.ts
```diff
@@ -18,21 +18,24 @@
 // tslint:disable: variable-name
 // Unfortunately just enabling PascalCase per file (tslint:enable:
 // allow-pascal-case) doesn't work.
-import {NamedTensorInfoMap} from './kernel_registry';
+import {NamedTensorInfoMap, TensorInfo} from './kernel_registry';
 import {PixelData} from './types';
 
+export const AddN = 'AddN';
+export type AddNInputs = TensorInfo[];
```
lina128, line 25: We need to have an exception here, AddNInputs is an array of Tensors. We don't have a fixed number of Tensors, so we cannot use NamedTensorInfoMap. Cast this input as NamedTensorMap is fine, because array is basically an object with index as key.

lina128:
Discussed offline with @tafsiri, the first solution (the one in this PR) works with current code with minimal change. We just need to introduce a flag to gradConfig to handle the case where inputs has a dynamic number of elements to save and we just want to save all inputs. The second solution (in another PR, link in this PR's description) requires logic change in many places, need to understand from Daniel and Nikhil whether this is needed. And what's the original thoughts to handle array of Tensors as object of Tensors with array index as key.

PR #3008, opened 2020-04-01
Move cpu backend out of tfjs-core

tafsiri:
Moves the CPU backend out of tfjs-core into its own package/top level folder.

- tfjs-backend-cpu is a devDependency of core to enable tests that depend on CPU specifically or more generally on having some backend present. This does create a circular dep between packages with respect to test running. But there are some workarounds to make the two libs build (using require over import in some places).
- tfjs-backend-cpu is a now a devDependency of most of the other packages as they relied on CPU backend for running tests.
- This PR allows re-registration of a kernel for a given backend (overwriting the prev kernel config). This is because WebGL backend is still in core and registers kernels as a side effect and core is a dependency of tfjs-backend-cpu. Thus in tests some double registration happens. Once WebGL is out of core the old behaviour could be restored. 
- As far as dependency structure for testing, this PR tries to mirror the status quo. We can tune stuff around when tests get called for what in future.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3008)


tfjs-backend-cpu/src/backend_cpu.ts
```diff
@@ -3795,5 +3794,3 @@ export class MathBackendCPU extends KernelBackend {
     return buffer.toTensor().reshape(shape);
   }
 }
-
-ENGINE.registerBackend('cpu', () => new MathBackendCPU(), 1 /* priority */);
```
annxingyuan, line 3799: Currently I think the WebGL backend assumes there's a registered CPU backend for CPU forwarding, something to be keep in mind when we pull out WebGL.

tfjs-core/src/index.ts
```diff
@@ -101,18 +108,35 @@ export {
   test_util,
   util,
   backend_util,
+  complex_util,
   webgl,
   tensor_util,
   slice_util,
   gather_util,
-  scatter_util
+  scatter_util,
```
annxingyuan, line 116: Curious why you've changed it to export these explicitly rather than through backend_util?

nsthorat:
Amazing work on this Yannick!

tfjs-data/package.json
```diff
@@ -44,9 +45,10 @@
     "build-core": "cd ../tfjs-core && yarn && yarn build",
     "build-core-ci": "cd ../tfjs-core && yarn && yarn build-ci",
     "build-layers": "cd ../tfjs-layers && yarn && yarn build",
+    "build-backend-cpu": "cd ../tfjs-backend-cpu && yarn && yarn build",
```
lina128, line 48: yarn build-ci

tfjs-layers/package.json
```diff
@@ -41,9 +42,10 @@
     "build": "tsc",
     "build-ci": "tsc",
     "build-core": "cd ../tfjs-core && yarn && yarn build",
+    "build-backend-cpu": "cd ../tfjs-backend-cpu && yarn && yarn build",
```
lina128, line 45: yarn build-ci

tfjs-layers/package.json
```diff
@@ -41,9 +42,10 @@
...
     "build-core-ci": "cd ../tfjs-core && yarn && yarn build-ci",
-    "build-deps": "yarn build-core",
-    "build-deps-ci": "yarn build-core-ci",
+    "build-deps": "yarn build-core && yarn build-backend-cpu",
+    "build-deps-ci": "yarn build-core-ci && yarn build-backend-cpu",
```
lina128, line 48: yarn build-backend-cpu-ci

lina128:
Hi Yannick, you probably need to add build-backend-cpu to Node's build-deps too.

tafsiri:
@lina128 Node shouldn't need to depend on backend-cpu to run its tests imo. It provides a backend that tests can be run against. If there are tests in core that specifically rely on CPU they should be constrained via a testEnv. With the tests passing earlier it looks like node does not depend on CPU (which is the expected behaviour imo)

Issue #2987, opened 2020-03-27
[tfjs-react-native] CameraStream wrong orientation in landscape mode.


nicholasio:
To get help from the community, we encourage using Stack Overflow and the [`tensorflow.js`](https://stackoverflow.com/questions/tagged/tensorflow.js) tag.

#### TensorFlow.js version
```
    "@tensorflow/tfjs": "^1.5.2",
    "@tensorflow/tfjs-react-native": "^0.2.3",
```
#### Browser version
This is an issue with tensorflow for react native.
react-native-cli: 2.0.1
react-native: 0.61.5

#### Describe the problem or feature request

The Camera Stream does not work well in landscape mode. The image is streamed/render in the wrong orientation/dimensions.

#### Code to reproduce the bug / link to feature request
In iOS set landscape mode in Xcode and use the TensorCamera as usual:

```jsx
const TensorCamera = cameraWithTensors( ExpoCamera );

const Camera = () => {
	return (<TensorCamera
		style={{ flex: 1} }
		cameraTextureHeight={ 1920 }
		cameraTextureWidth={ 1080 }
		resizeDepth={ 3 }
		onReady={ handleCameraStream }
		autorender={ true }
	/>)
}

nicholasio:
@rthadur @tafsiri I'm willing to contribute with a PR for this one. Is there any guidance you can give on where I should start looking at? After a quick look it seems like we need to rotate the image when in landscape mode. I assume that should be done through WebGL in a custom shader.

Any guidance is appreciated.

tafsiri:
@nicholasio Thanks for being willing to contribute. A couple of starting points to investigate. 

For doing the transform in WebGL the starting point would be the [renderToGLView function](https://github.com/tensorflow/tfjs/blob/a7a07387f66ff46b34641bc0140db55a17e076ca/tfjs-react-native/src/camera/camera.ts#L224), you will see it calls [drawTexture](https://github.com/tensorflow/tfjs/blob/a7a07387f66ff46b34641bc0140db55a17e076ca/tfjs-react-native/src/camera/camera_webgl_util.ts#L154) (which ultimately runs a shader). These would need to support taking a rotation related param. Ideally this is something that can detected outside of the cameraStream component and passed in as a prop.

Another thing that might be worth taking a quick look at, though it may not work, is whether doing a [react native transform](https://reactnative.dev/docs/transforms) on the containing view would work. I'd eliminate this option first as it is quick to try.

[...]

PR #3041, opened 2020-04-06
[camera-stream] Fix for iOS Landscape Mode

nicholasio:
This is a fix for #2987 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3041)


nicholasio:
@googlebot I signed it!

tafsiri:
@nicholasio also it looks like the CLA check failed. You might have different email addresses associated with the commit vs your github?

nicholasio:
@tafsiri Thanks for the review. I think those make sense and I think you're right about swapping width/height params.

I'll address your feedback later today and will trying figure out the CLA check.

nicholasio:
@tafsiri I made a couple of updates to add support for a rotation param instead of "landscape". Regarding android, disregard that, I was able to reproduce the same issue on android.

Regarding the target shape I don't think it's necessary to change, we're rotating the texture but the tensor shape should match whatever the input shape for the model is. 

Still pending the tests, but let me know if there's anything else you wanted to see.

nicholasio:
@googlebot I fixed it

nicholasio:
@tafsiri I notice one thing on Android. if the app starts on landscape mode by default, the bug does not happen and no rotation is needed. However, if the app starts in portrait and then you rotate you phone to landscape mode it bug shows up.

Shouldn't change anything in this PR as it's up to the application choose the rotation they want to use.

tafsiri:
> Regarding the target shape I don't think it's necessary to change, we're rotating the texture but the tensor shape should match whatever the input shape for the model is.

I think this depends on how the user specifies the target shape, so that is what I think the tests can catch and we could add some logic to warn about. When rotating a non square image 90 degrees, the aspect ratio of the target shape will be different from the input shape. Could you add tests for rotating non square images as that is closer to the expected use case.

nicholasio:
@tafsiri Sure, will add a few more tests to cover this scenario

fhenri42:
Hey we are facing the same issue, do you have any news on when this PR will be merged ?

rthadur:
@nicholasio gentle ping for update ?

nicholasio:
I'll try to set some time aside to get this over the finish line, which at this point seems to be only adding a couple more tests. In the meantime we are already using this PR on an app successfully.

> iss

If you want to try this out you can add this `postinstall`script which only overrides the react-native package.

```
    "postinstall": "git clone https://github.com/nicholasio/tfjs.git tfjs-fork && (cd tfjs-fork/tfjs-react-native && git checkout fix/ios-landscape-mode && yarn && yarn build) && cp -rf tfjs-fork/tfjs-react-native/dist node_modules/@tensorflow/tfjs-react-native && rm -rf ./tfjs-fork"
```

Doing this way only requires you to compile the react-native package and not everything.

DanielRHarris:
Hi, any update on this?

dhiogocorrea:
Hello! Any updates ? Thanks . @nicholasio @rthadur

nicholasio:
> Looks like some other folks are also asking about this issue. I will prioritize it and take a closer look this or next week. Will create another PR based on this. Thanks @nicholasio!

Looks like you got this covered! Excited to see this fix started to make it's way into master.

I have a production app that I can help test once you've got a PR open.

PR #3062, opened 2020-04-10
webgpu: Simplify the unary/binary shader conditionally

qjia7:
…ied.

Perf

See ~15% improvement for posenet demo on some platforms.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3062)


qjia7:

tfjs-backend-webgpu/src/kernels/binary_op_webgpu.ts
```diff
@@ -52,16 +52,38 @@ export class BinaryOpProgram implements WebGPUProgram {
   workGroupSize: [number, number, number] = [16, 1, 1];
 
   constructor(op: string, aShape: number[], bShape: number[]) {
-    this.outputShape = backend_util.assertAndGetBroadcastShape(aShape, bShape);
-    const size = util.sizeFromShape(this.outputShape);
+    if (util.arraysEqual(aShape, bShape) &&
```
annxingyuan, line 55: Could you name this constant, here and below, and provide a short explanation of why this threshold is being used?

annxingyuan, line 55: Would you mind rewriting this slightly so that only what's actually different between the two conditions is included in the if/else statement? I think this will shorten the file and make it easy to see what the optimizations are.

For example `this.outputShape = backend_util.assertAndGetBroadcastShape(aShape, bShape);` will still work for both conditions, could you move it above the if/else? Same with `this.dispatchLayout`. For `computeDispatch`, could you call it outside the if/else, and define the last argument of `computeDispatch` within the if/else? Same pattern for `this.userCode`. 

Could you also do this for `unary_up_webgpu` below?

qjia7: Good suggestion.  A quick explanation here: 16 is not a threshold. It's the value of work group size. If the output size  % 16 ===0, it means we don't need to check out of bounds. And we can directly use A[index]/B[index] to get the value.

qjia7: Move most of variables out of if/else and keep `this.userCode`. `this.userCode` is not too complicated. So I prefer not to split part of it out of main body. The advantage is we can clearly see the whole body. Of course, the disadvantage is  `binaryOperation` repeated twice. But it's not a big deal. How do you think?

qjia7:
@annxingyuan Please take another look. Thanks.

PR #3063, opened 2020-04-10
Remove WebGL test matrix from core

tafsiri:
Note core will now test webworker against CPU backend only since we have removed the dependency of core on webgl backend.

I don't think we need to explicitly test webgl backend in webworker because the only thing it was testing that was different was use of OffscreenCanvas. But we have since used OffscreenCanvas by default when available, so all WebGL tests in chrome test that path. @nsthorat @dsmilkov am i missing anything here?

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3063)


tafsiri:
I turned on the nightly configuration for pull requests and corrected a whole bunch of things for the nightly run, and it seems to behaving as expected other than getting starved of browserstack resources. I'm going to turn it back to the default and run the tests, and then we can allow nightly to give us feedback as usual.

tafsiri:
@lina128 this is ready for re-review as its changed a few things. I noticed that some of the builds weren't actually running correctly, and that nightly flags weren't being passed for the new folders.

tafsiri:
Going to force merge this after manually looking at all logs from CI. All tests are passing other than test-snippets in WASM. The source of WASM failures is an issue downloading the following file. Will follow up via email on ideas for this.

"[---------Error: Downloading URL 'https://storage.googleapis.com/webassembly/emscripten-releases-builds/linux/40f3caabcef7b52bdde63d3883462414d7a25bec/wasm-binaries.tbz2': [Errno 104] Connection reset by peer"

PR #3076, opened 2020-04-13
Add retry to download emsdk due to flaky download.

lina128:
Context: downloading emsdk is pretty flaky recently. One observation is manually clicking on retry usually just works. One guess is this is due to unstable network. If so, retry can reduce the occurrences of this error. This PR automates retry. 

Long term, consider using docker image with emsdk preinstalled, which can completely solve the problem, suggested by @tafsiri.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3076)

### DIFF
--- a/tfjs-backend-wasm/scripts/build-ci.sh
+++ b/tfjs-backend-wasm/scripts/build-ci.sh
@@ -16,8 +16,16 @@
 
 set -e
 
-# Install emsdk
-git clone --depth=1 --single-branch https://github.com/emscripten-core/emsdk.git
+# Install emsdk with up to 1 retry.
+for i in $(seq 0 1)
+do
+  # Wait for 15 seconds then retry.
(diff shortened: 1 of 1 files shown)


tafsiri:
@lina128 is this the part that is failing? My impression from the logs that it was something in `./emsdk install 1.39.1` that fails given the url connection that gets closed goes to gcp as opposed to github? Should a retry be placed around the emsdk install?

tafsiri:
the line i'm thinking of line 32, I think that should probably be in the retry block.

lina128:
> @lina128 is this the part that is failing? My impression from the logs that it was something in `./emsdk install 1.39.1` that fails given the url connection that gets closed goes to gcp as opposed to github? Should a retry be placed around the emsdk install?

Ah, you are right. I will send another PR now.

lina128:
This PR adds retry to emsdk isntall: https://github.com/tensorflow/tfjs/pull/3078 @tafsiri

PR #3108, opened 2020-04-20
[wasm] Enable compatibility with emscripten 1.39.11, which in turn enables pthread support.

annxingyuan:
Without this change, our tests fail with emsdk 1.39.11 because `unordered_map.at` errors out if the key is not found. 

This change just enables compatibility with emsdk 1.39.11, but doesn't yet change our CI to use 1.39.11. I will make that change once we turn on pthread support. 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3108)

### DIFF
--- a/tfjs-backend-wasm/src/cc/batch_mat_mul_impl.cc
+++ b/tfjs-backend-wasm/src/cc/batch_mat_mul_impl.cc
@@ -206,9 +206,11 @@ void slow_batch_matmul(const size_t a_id, const size_t* a_shape_ptr,
   float* out_buf = out_info.f32_write();
 
   const float* bias_buf = nullptr;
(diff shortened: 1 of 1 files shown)


annxingyuan:
I'm not sure - I read over the release notes https://emscripten.org/docs/introducing_emscripten/release_notes.html but I can't find anything that points to it. Although it did mention "The default c++ version is no longer fixed at c++03.  We now fall back to  clang's default which is currently c++14." - so maybe this has to do with a change in c++.

dsmilkov:
Good point, but it's probably not related it since we lock our version to c++11: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/.bazelrc#L12

PR #3139, opened 2020-04-22
fix nightly error on isMobile

pyu10055:
isNavigatorDefined method cannot be mocked from outside of the package.
created a helper method to allow setting the value.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3139)


tfjs-core/src/device_util.ts
```diff
@@ -1,4 +1,5 @@
 /**
+import { isNavigatorDefined } from './device_util';
```
lina128, line 2: Remove?

PR #3194, opened 2020-05-01
Add exported symbols back to webgl backend

tafsiri:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3194)


scripts/publish-npm.ts
```diff
@@ -150,7 +150,7 @@ async function main() {
     shell.cd(pkg);
     const otp =
         await question(`Enter one-time password from your authenticator: `);
-    $(`npm publish --otp=${otp}`);
+    $(`YARN_REGISTRY="https://registry.npmjs.org/" npm publish --otp=${otp}`);
```
lina128, line 153: This doesn't work. I have tried this before. The env variable in the console effectively controls the env of the immediately following yarn command. However, shelljs doesn't seem to be able to override that env. So it has to be in the console. Somehow, weirdly, adding it in shelljs works on my macbook. But it doesn't work in my workstation. Having the env set in console works cross platforms.

lina128: We can add this though, because it works in some platform. But we should advise people to also set the env in the console for now. If shellsjs can fix this problem in the future, then we can drop the outside part.

tafsiri: good to know, interestingly it did work (i think) on my mac. I'll test futher and see if it works without this line.

tafsiri:
@lina128 this PR is still in progress, but i'm curious if you think the change to removing cpu as a dep on core when doing the release pr makes sense. To me it makes sense and resolves the circular dependency issue from a release process. I'm going to test it more thoroughly at the next RC release.

tafsiri:
FYI @annxingyuan this PR overalaps a little with yours in that it removes the dist import of transposeImpl from CPU. This was needed to sync to g3. If your max PR gets in before this one, I'll revert the change i've made here and sync with yours.

scripts/publish-npm.ts
lina128: It works on my mac too. It doesn't work on gLinux.

lina128:
> @lina128 this PR is still in progress, but i'm curious if you think the change to removing cpu as a dep on core when doing the release pr makes sense. To me it makes sense and resolves the circular dependency issue from a release process. I'm going to test it more thoroughly at the next RC release.

Yeah, I saw that. It does solve the release process issue. So that's a plus. But it does feel a little bit weird to have symlink in released version. If we move webworker test to cpu, and exclude ops/ and gradients/ tests in core, will we be able to remove cpu as a devDep? For local test, we can revise the test script to also cd into cpu package and trigger cpu's test. WDYT?

PR #3208, opened 2020-05-05
Nightly test fixes

tafsiri:
This PR (incorporating Na's change which made the webgl runs sequential in nightly) passes nightly.

It primarily fixes the webworker build, but also disables the cloud benchmark integration test, which seems to use an out of date approach to building packages, (yalc link, no build-deps etc). If this was in fact working since we moved to `link://` style dependencies I can put this back, else it may have to wait until we refactor those integration tests more fully.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3208)

### DIFF
--- a/GALLERY.md
+++ b/GALLERY.md
@@ -6,6 +6,7 @@ Please reach out to us if you'd like to put your project on the list.
 
 ## Demos & Applications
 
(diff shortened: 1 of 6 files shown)


tafsiri:
@lina128 Ready for re-review, this passes CI even though it doesn't look like it, if you click into the failing test target, I've run it and it passes. I just can't run them all at the same time without something disconnecting.

PR #3255, opened 2020-05-12
Update tfjs to 2.0.0-rc.4.

tafsiri:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3255)


tafsiri:
@tafsiri investigate why tfjs-node fails when union is has pinned version deps. (doesn't fail on master).

PR #3280, opened 2020-05-15
add back arm support

kangyizhang:
arm32 support was accidentally removed in https://github.com/tensorflow/tfjs/commit/63235661d1c1f26ff5b6853375a59213fe99d9a5#diff-943b21d25f7c485a9de071e9a1d1c18bL80. This PR add it back. 

Please note it is arm32 with TF C version 1.14. We do not support arm64 and TF C version 1.15 yet because we do not have the device to test.

@yhwang please take a look to make sure that it does not break support for external binaries.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3280)


yhwang:
@kangyizhang Thanks for adding armv7 back. The custom binary support should be good after your change. One question, do you plan to upgrade the tf shared lib to v1.15?

PR #3297, opened 2020-05-20
replace static async storage library with storage provider

hayzem:
to be able to use other storage libraries.

### DIFF
--- a/tfjs-react-native/src/async_storage_io.ts
+++ b/tfjs-react-native/src/async_storage_io.ts
@@ -15,7 +15,6 @@
  * =============================================================================
  */
 
-import {AsyncStorageStatic} from '@react-native-community/async-storage';
 import {io} from '@tensorflow/tfjs-core';
 import {fromByteArray, toByteArray} from 'base64-js';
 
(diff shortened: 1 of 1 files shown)


hayzem:
@googlebot I signed it!

tafsiri:
@hayzem Could you give sample code of how you use this with a different async storage library? I'd like to think about wehther that should just be a different IOHandler

hayzem:
Sure;

For example with expo's react-native fork:

```
import { AsyncStorage } from 'react-native';
import { AsyncStorageHandler, StorageProviderInterface } from '@tensorflow/tfjs-react-native'

CustomAsyncStorage implements StorageProviderInterface {
  // tslint:disable-next-line:no-any
  protected storage: any;
  constructor() {
    this.storage = AsyncStorage
  }
  getItemAsync(key: string): Promise<string> {
    return this.storage.getItem(key);
  }
  setItemAsync(key: string, value:string): Promise<unknown> {
    return this.storage.setItem(key);
  }
  removeItemAsync(key: string): Promise<unknown> {
    return this.storage.removeItem(key);
  }
}


const model = require('../assets/model/model.json');

// Use the bundleResorceIO IOHandler to load the model
const model = await tf.loadLayersModel( AsyncStorageHandler(model, CustomAsyncStorage));

```

If you think it is better to have another IOHandler for this, I will do so. I thought about it and didn't want to repeat itself. I also realized that the interface is not exportable. I'm going to fix it.

PR #3320, opened 2020-05-26
webgpu: Use shared memory for binary op

qjia7:
PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3320)


qjia7:
See ~6% improvement for resnet model on some platforms.

The optimization is mainly for broadcast ops. For example, aShape=[1, 29, 29, 256], bShape=[256]. In this case, we need to access each data of B 29x29 times. It will be better to put B in shared memory to reduce global memory access times.

qjia7:
@kainino0x Please take another look.
Sorry that I don't know how to reply your comments in the review system. It seems that I don't have permission to comment. I think all comments are done. Thanks.

qjia7:
> Did you prefer not to split the class now (maybe too much duplication)? I think it would make sense to do now (but don't know how simple it would be).

I left a TODO there. I agree this class needs refactor. I will consider how to expand the shared memory to more situations. For example, it can apply shared memory if aShape.length > bShape.length or bShape.length > aShape.lengh. Currently, this is only applied to one shape's length is 1 and the other is larger than 1. So I prefer to leave it to another PR.

PR #3321, opened 2020-05-26
webgpu: Remove uniform buffer when not needed

qjia7:
PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3321)


qjia7:
This change will slightly improve the perf of resnet model.

qjia7:
> nit: Consider inverting from `noUniform` to `needsUniform`
> nit: Consider describing what's actually in the uniform if it's simple to do so. (Is it the shapes?`needsShapesUniform`?)

Yes. It's the shape. Change it to `needsShapesUniforms`. Previously, I used `noUniform` was because in this case I could make it optional and I didn't need to change many other programs.

qjia7:
Please take another look. Not sure why the bots fail.

kainino0x:
> Change it to `needsShapesUniforms`. Previously, I used `noUniform` was because in this case I could make it optional and I didn't need to change many other programs.

Ah, I didn't realize this. Sorry that required a change to so many files.

If you think it would be better to keep it as an optional `noShapeUniforms` to avoid affecting so many files, please feel free (I'm neutral).

PR #3325, opened 2020-05-26
Update tfjs-layers, tfjs-converter to 2.0.0, 2.0.0.

tafsiri:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3325)

### DIFF
--- a/tfjs-converter/package.json
+++ b/tfjs-converter/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow/tfjs-converter",
-  "version": "0.0.0",
+  "version": "2.0.0",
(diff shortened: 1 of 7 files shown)


tafsiri:
A could of node tests fails in CI because tfjs-core is resolved in two different ways leading to two IORouter singletons, which leads to two distict registries for save and load handlers for models. This happens because of a mix of link:// dependencies and pinned versions and should resolve when all dependencies are pinned.

lina128:
Hi Yannick, thanks for investigating this. Given that the nightly build and e2e pass, this should be fine once all the dependency changed to pinned. LGTM!

Issue #3221, opened 2020-05-07
float16 quantization


JesseFarebro:
#### Describe the problem or feature request

It would be useful to support quantizing to float16 over the wire and then dequantizing back to float32 on the client.
I'm working with a model where the 2-byte affine quantization in TFJS performs very poorly but using float16 quantization has a negligible impact on performance.

pyu10055:
Thanks, that is very a good suggestion.

JesseFarebro:
I'll probably end up putting together a PR for this in the next week or so as I'm going to end up needing this.

pyu10055:
@JesseFarebro that would be great, thank you!

JesseFarebro:
I'm just finishing this work up but I'm now debating what's the best way to incorporate quantization in the converter CLI. What I have right now is:
```
tensorflowjs_converter --quantization_dtype float16 ...
tensorflowjs_converter --quantization_dtype int8 ...
tensorflowjs_converter --quantization_dtype int16 ...
```

The problem with this is it doesn't allow for using `float16` for weights with type `float32` and `int8|int16` for weights with type `*int*`. Perhaps making this even more configurable could be advantageous, for example, only quantizing certain layers could provide improved accuracy while still reducing the model size. Perhaps flags like:

```
tensorflowjs_converter --quantize_float32 block1/*=float16
                       --quantize_float32 block2/*=uint16
                       --quantize_int32 output/*=uint8
```
while also allowing for:
```
tensorflowjs_converter --quantize_float32 float16
                       --quantize_int32 uint16
```

We could also maintain backwards compatibility by still allowing for `--quantization_bytes` being the default quantization if none of the `--quantize_*` flags are present. If the `--quantize_*` flags are present it would override the value of `--quantization_bytes` for the specific dtype or layer specified. e.g.,

```
tensorflowjs_converter --quantization_bytes 1
                       --quantize_float32 block1/*=float16
```
would quantize every weight outside of `block1/*` to uint8 and every float32 weight inside `block1/*` to float16.

I'm not happy with how the block specific flags look but I just want to get a discussion going before I start refining this.

PR #3340, opened 2020-05-28
float16 quantization + granular quantization

JesseFarebro:
As discussed in #3221 this enables support for quantizing to float16 over the wire. Some models perform better with float16 quantization versus the current integer-based affine scaling method.

I also needed the ability so support a form of granular quantization where I want to quantize only certain values. This PR also includes these changes to the converter, no changes are required to the core in order to support this.

To recap the major changes are:

## Converter

`--quantization_bytes` is no longer a flag in the converter. It didn't make sense as we support two methods for 2 bytes. I added three new flags: `--quantize_float16`, `--quantize_uint8`, `--quantize_uint16`. These flags can be supplied on their own or you can specify specific weight names or groups of weights using a wildcard expansion. For example,

```
--quantize_float16=conv/*/weights
--quantize_uint16=conv/*/biases
--quantize_uint8
```
Would quantize weights that match the prefix `conv/` to float16, all biases with the prefix `conv/` using 2 byte affine quantization, and all remaining weights with 1 byte affine quantization.

## Weight Decoding

The only other addition revolves around decoding the float16 values in Javascript to float32 values. JS doesn't support any form of `Float16Array` at this time so I manually decode the weights using a table-based lookup method that is referenced from http://www.fox-toolkit.org/ftp/fasthalffloatconversion.pdf 

## Testing

I tested this myself with my own models and was able to deterministically reproduce my own results from Python TF. I wrote some tests for the float16 decoding edge cases and a couple of basic decoding test cases.
All the converter test cases are passing and I wrote a couple more tests for the new quantization scheme.

I haven't really tested the converter wizard yet, it doesn't run locally for me (`tf.app` is not supported?) and I haven't gotten around to building the pip package to test it out. I think the wizard quantization flow is kind of weird and will probably require some modifications.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3340)


JesseFarebro:
Hi @pyu10055,

Thanks for taking a look, I resolved most of the issues you outlined. I need to resolve a couple more things before it's ready to be looked at again. More specifically, I need a better way of resolving the case where you want to catch all other weights with a specific quantization value, i.e., 
```
--quantize_float16=conv/*/weights
--quantize_uint16=conv/*/biases
--quantize_uint8
```

won't catch all remaining variables and quantize to `uint8`. I do want to support this though. The way I invasion this is if you supply two patterns that resolve to the same weight it will fail. But if you supply a pattern that matches a weight __and__ you supply a fallthrough like `--quantize_uint8` (without any pattern supplied) it won't fail and will quantize all remaining weights with affine `uint8`.

It won't take much work but I do want to provide sufficient tests for this mapping scheme. I'll let you know when it's ready to be looked at again.

JesseFarebro:
Thanks for the careful reviews @pyu10055. Do you have any idea what the timeframe for a new release will look like?

PR #3349, opened 2020-05-29
webgpu: Fix the deprecated type warnings

qjia7:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3349)


qjia7:
@kainino0x Not sure if these warnings will affect the performance. So I prefer to fix these warnings.

But there is still one warning left:
GPUBuffer.setSubData is deprecated: use createBufferMapped(with copyBufferToBuffer if needed)(but note the design/spec of this API is still in flux).

It's a little complicated to use createBufferMapped in currrent design. In TFJS, it uses `BufferManager` to manage all allocated buffers. When we want to use a buffer, we need to call `bufferManager.acquireBuffer` to get a buffer (It can re-use allocated buffer). However, if we change it to createBufferMapped, it will return `[GPUBuffer, ArrayBuffer]`. Should we save `[GPUBuffer, ArrayBuffer]` for BufferManager? Except of setSubData,  getBufferData also uses acquireBuffer. But it   uses copyBufferToBuffer to copy data from a storage buffer to a readback buffer.  Any idea for this warning?

kainino0x:
The warnings don't affect performance, but it is necessary to fix them. We will be removing the deprecated items from Chrome in the next few days.

An alternative to setSubData has landed in the spec: writeBuffer. I don't think we've implemented that yet so I'll try to remember we need to not remove setSubData yet.

qjia7:
@kainino0x Thanks. I think we can wait `writeBuffer`. I don't have permission to land PR. Feel free to land it if you can.
Or add @annxingyuan to do it.

PR #3399, opened 2020-06-06
Initial check in of the tfjs inference api.

lina128:
Initial check in of the tfjs inference api.

This PR has below changes:
1. Provide a nodejs application that exposes a limited set of tfjs functionalities through cmd.
2. Script to generate tfjs-inference binary.
3. An example python script to invoke the nodejs binary in subprocess.
4. js tests and python tests.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3399)


nsthorat:
Hey, I didn't realize we were checking this into GitHub. Is this is synced from google3?

If it's here, is it possible to develop in GitHub and sync back to Google3 instead? That way we can keep it in third_party/ over there.

PR #3407, opened 2020-06-08
Modularise atan2, floorDiv, mod and mul

tafsiri:
This modularizes the remaining binary ops, atan2, floorDiv, mod and mul.

One note about floorDiv: in doing this I noticed we had lost test coverage, i restored an old test for forward pass, however gradient calc didn't have a test. So I didn't remove the chaining api from FloorDiv grad. A followup PR will add tests for the gradient and then we can more safely make edits to that code.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3407)


annxingyuan:
@tafsiri oops - sorry I missed your note in the PR description :)

tafsiri:
@annxingyuan i did have to rename Mul.[ts|cc] to Multiply.[ts|cc], however this is a good thing as we did set the convention that the filename should match the kernel name.

Issue #3114, opened 2020-04-20
[WASM] Error: 'oneHot' not yet implemented or not found in the registry 


syonfox:
Hey it looks like oneHot is not implemented in the wasm Backend yet.

```
backend.ts:665 Uncaught (in promise) Error: 'oneHot' not yet implemented or not found in the registry. Did you forget to import the kernel?
    at Fa (backend.ts:665)
    at BackendWasm.t.oneHot (backend.ts:586)
```

#### Versions
```
tfjs: "1.7.1"
tfjs-converter: "1.7.1"
tfjs-core: "1.7.1"
tfjs-data: "1.7.1"
tfjs-layers: "1.7.1"

tf.wasm.version_wasm
"1.7.1"
```

Browser: ff 68.6.0esr

dsmilkov:
Hi @syonfox,

Can you share some info about the ML model that you need this for, which helps us better prioritize?

syonfox:
Sorry about the slow response.  I was using onehot in my median filter function  I tried to lay it out in a simmaler style to existing tensor flow function hope this helps. 
 If you wish to look at a usage I used it in a project here.  This is by no means a minimal example since lots of other stuff is happening.   The relivante code is pasted for the most part. 
http://syon.ca/cmpt414/pcv/src/pcv.html
the code is all in the http://syon.ca/cmpt414/pcv/src/js/main.js and http://syon.ca/cmpt414/pcv/src/js/medianFilter.js


Filter where one hot is used
```
function medianFilter(x, filter, strides, pad) {
    //make Kernal
    //todo allow for filter as array or number
    let filterSize = filter ** 2;
    let locs = tf.range(0, filterSize, filterSize );
    //makes a bunc of arrays each one reprensentin one of the valuesin the median window ie 2x2 filter i in chanle and 4 out chanles
    let f = tf.oneHot(tf.range(0,filterSize,1, 'int32'), filterSize).reshape([filter, filter, 1, filterSize]);
    let y = tf.conv2d(x,f,strides,pad);
    let m_idx = Math.floor(filterSize/2)+1;

    let top = tf.topk(y, m_idx, true);
    let median = tf.slice(top.values, [0,0,m_idx-1], [-1,-1,1] );

    return median;
}
```

Example Usage of Function
```
        res = res.reshape([imSize.x, imSize.y, 1])
        res = medianFilter(res, 3, [1, 1], 'same');
```

oukigen:
[...]

PR #3409, opened 2020-06-08
[WASM] Add OneHot op

pvaneck:
This adds the OneHot operation to the WebAssembly backend.

Closes: https://github.com/tensorflow/tfjs/issues/3114

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3409)

### DIFF
--- a/tfjs-backend-wasm/src/kernels/OneHot.ts
+++ b/tfjs-backend-wasm/src/kernels/OneHot.ts
@@ -1,0 +1,58 @@
+/**
+ * @license
+ * Copyright 2020 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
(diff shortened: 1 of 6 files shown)


tfjs-backend-wasm/src/kernels/OneHot.ts
```diff
@@ -0,0 +1,58 @@
...
+  const indicesData = backend.dataIdMap.get(indices.dataId);
+  const indicesId = indicesData.id;
+
+  wasmOneHot(indicesId, depth, onValue, offValue, outId);
+
+  return out;
+}
+
+registerKernel({
+  kernelName: 'OneHot',
```
annxingyuan, line 54: Import `OneHot` from core as well and use here.

tfjs-core/src/ops/one_hot.ts
```diff
@@ -56,17 +56,16 @@ function oneHot_(
...
   };
 
   const inputs: OneHotInputs = {indices: $indices};
   const attrs: OneHotAttrs = {depth, onValue, offValue};
 
-  return ENGINE.runKernelFunc(
+  const res = ENGINE.runKernelFunc(
       forward, inputs as unknown as NamedTensorMap, null /* grad */, OneHot,
       attrs as unknown as NamedAttrMap);
+  return reshape(res, outShape);
```
annxingyuan, line 68: We would like things like reshape to happen within the kernel as well. Could you add logic to reshape in the WASM OneHot typescript implementation?

tfjs-backend-wasm/src/kernels/OneHot.ts
pvaneck: Done!

tfjs-core/src/ops/one_hot.ts
pvaneck: I modified the one hot op in core a bit so that flatten is moved into the forward function. To get the correct output shape in the WASM op, I need to pass in the original indices tensor (i.e. not flattened).  I hope this is fine.

pvaneck:
@annxingyuan Thanks for the review! Please take another look when you have a chance.

PR #3436, opened 2020-06-12
fix file:// URLs on Windows

strager:
BUG In Node.js on Windows, file: URLs have three slashes before the
drive letter. For example:

    file:///c:/Users/strager/[snip]/nsfwjs/model.json

When converting a file: URL to a file path, NodeFileSystem's router
incorrectly preserves one leading slash. This results in a
drive-relative path, which refers to a different file than intended.
This leads to errors down the road:

    Error: Path C:\C:\Users\strager\[snip]\nsfwjs\model.json does not exist: loading failed
        at C:\Users\strager\[snip]\node_modules\@tensorflow\tfjs-node\dist\io\file_system.js:67:23

Use Node.js' built-in functions for dealing with file: URLs instead of
hackily parsing them ourselves.

On POSIX systems, this commit should not change behavior.

### DIFF
--- a/tfjs-node/src/io/file_system.ts
+++ b/tfjs-node/src/io/file_system.ts
@@ -18,6 +18,7 @@
 import * as tf from '@tensorflow/tfjs';
 import * as fs from 'fs';
 import {dirname, join, resolve} from 'path';
+import {fileURLToPath} from 'url';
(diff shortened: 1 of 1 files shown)


strager:
Ping! I still use this patch and I think it should be merged.

PR #3452, opened 2020-06-15
Modularize stack, unstack, reshape, expandDims, squeeze

annxingyuan:
- Not removing reshape from `Tensor` interface to avoid circular dependency.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3452)


annxingyuan:
@lina128 Hmm, I guess I thought we were strictly following that, I think sometimes their input tensors are named `x` as well. I don't think we need to go back and change anything we've already modularized, but maybe we should use the name `tensor` going forward if it appears in ops.pbtxt? Or maybe we were avoiding using `tensor` due to name conflicts? Do you have thoughts on this @tafsiri ?

tafsiri:
lets chat on gvc. personally i don't think it matters too much but we should pick a convention. I our assumptions about how well things could align at the beginning might have changed. i'll ping you on chat.

PR #3459, opened 2020-06-16
[WASM] Update to latest XNNPACK and add Windows benchmarks.

annxingyuan:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3459)


tfjs-backend-wasm/README.md
```diff
@@ -133,6 +134,7 @@ the device).
 | iPhone XS                    | 21.4 | 10.5  | 176.9    | N/A         |
 | Pixel 3                      | 40.7 | 31.8  | 535.2    | N/A         |
 | Desktop Linux                | 12.6 | 12.7  | 249.5    | N/A         |
+| Desktop Windows              | 7.5  | 7.1   | 270.9    | 16.2        |
```
Maratyszcza, line 137: This is strange, SIMD is supposed to be faster than WAsm MVP. Worth to report to V8 devs, probably there's a mix of SSE/AVX instructions in the codegen, cc @ngzhian

annxingyuan: Apologies - I switched the columns by mistake! SIMD is indeed faster - thanks for catching this :)

By the way I plan to update the Pixel3 benchmarks as soon - I'm using Browserstack to test as I don't have access to a real device, so I need to wait for their devices to update to Chrome 84+ so the feature test will work.

PR #3468, opened 2020-06-17
[core] Warn the user to import the CPU backend for better performance if CPU forwarding fails.

annxingyuan:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3468)


annxingyuan:
@pyu10055 I think that's how it will work eventually when we generate minimal builds according to a model's op manifest.

PR #3517, opened 2020-06-26
Modularize argMax, argMin, min, mean and sum

tafsiri:
Also move norm and max tests into their own files.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3517)


tafsiri:
Modularizes the rest of the reduction ops and makes the max kernels handle the keepDims attr across backends. 

It also fixes a bug that was present in the max kernel in wasm where if an internal transpose was used the input tensor would be changed after the op completed. I added a test for that to max and min, but a general thing to watch out for is calling `.set` on the typedarray for a models inputs.

PR #3525, opened 2020-06-29
webgpu: Fix the binding buffer size error

qjia7:
BUG

In latest chrome canary, many below errors are printed when running
'yarn test'
'Binding sizes too small for bind group 0.'

The original code used util.bytesPerElement(tensor.dtype) which will
treat bool as 1 byte. However, for GPU, bool should be treated as 4 bytes.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3525)


qjia7:

PR #3532, opened 2020-06-30
webgpu: Use setOutput instead of directly assignment

qjia7:
BUG

The inputs data and output data types may be different. So we should use
setOutput instead of direclty assignement to process the different type casting.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3532)


qjia7:

PR #3534, opened 2020-06-30
webgpu: Use uniform buffer to store small tensor

qjia7:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3534)


qjia7:
@kainino0x @annxingyuan, in last sync, we talked about using uniform buffer to store small size tensor. But it seems that there is no obvious perf improvement. Please check if I am using the right way to do it.

PS: 
Uniform buffer follows std140 rules. So  we can't directly use 'float[]' to store the tensor data which needs paddings. So we only support below uniform blocks which don't need paddings:
1.
{
  float B;
}
2.
{
  vec2 B;
}
3.
{
  vec3 B;
}
4
{
  vec4 B [size/4];
}

PR #3538, opened 2020-06-30
Move all the remaining chaining api outside of tensor.ts

tafsiri:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3538)


tafsiri:
This is a large scale change that moves all the the remaining chaining api ops out of tensor.ts. Primarily done using code gen with hand editing after. Its a lot of files but very uniform type of change. 

I went down this path (automation) for a couple of reasons:
 1. Speed up modularization workflow (3 less steps!)
 2. Make OpHandler and Tensor as slim as possible with respect to ops as we look ahead towards the work in the backends. In parallel i'm trying to define what precisely we want to keep as part of the tensor type which will have implications on the backend interface and possibly the TensorInfo object backends use. So moving these out is a step in clarifying what we want those interfaces to be. More to come on this in a doc after the holidays.
 3. I realized I wanted to move things like the strict* versions of ops out of tensor.ts so that it can be refactor independent of the actual removal of those functions from public api.

PR #3552, opened 2020-07-04
Add DEBUG_QUERY flag

axinging:
Fix: https://github.com/tensorflow/tfjs/issues/3551

This DEBUG_QUERY flag is used for getting performance data of each operators, so no read back and numerical error check.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3552)

### DIFF
--- a/tfjs-core/src/engine.ts
+++ b/tfjs-core/src/engine.ts
@@ -615,7 +615,9 @@ export class Engine implements TensorTracker, DataMover {
     // Stop recording to a tape when running a kernel.
     this.scopedRun(
         () => this.state.kernelDepth++, () => this.state.kernelDepth--, () => {
-          if (!this.ENV.getBool('DEBUG')) {
+          const debugFlag =
+              this.ENV.getBool('DEBUG') || this.ENV.getBool('DEBUG_QUERY');
+          if (!debugFlag) {
             outputs = kernelFunc();
           } else {
             outputs = this.profiler.profileKernel(
--- a/tfjs-core/src/flags.ts
+++ b/tfjs-core/src/flags.ts
@@ -33,6 +33,14 @@ ENV.registerFlag('DEBUG', () => false, debugValue => {
   }
 });
 
+/** Whether to enable debug query mode. */
+ENV.registerFlag('DEBUG_QUERY', () => false, debugValue => {
+  if (debugValue) {
+    console.warn('Debugging mode is ON. This significantly impacts ' +
+        'performance on WebGL but small impacts on WebGPU.');
+  }
+});
+
 /** Whether we are in a browser (as versus, say, node.js) environment. */
 ENV.registerFlag('IS_BROWSER', () => device_util.isBrowser());
 
--- a/tfjs-core/src/flags_test.ts
+++ b/tfjs-core/src/flags_test.ts
@@ -42,6 +42,30 @@ describe('DEBUG', () => {
   });
 });
 
+describe('DEBUG_QUERY', () => {
+  beforeEach(() => {
+    tf.env().reset();
+    spyOn(console, 'warn').and.callFake((msg: string) => {});
+  });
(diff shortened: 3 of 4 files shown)


axinging:
@annxingyuan, PTAL

annxingyuan:
Thanks @axinging ! However I don't think this addresses the root cause of https://github.com/tensorflow/tfjs/issues/3551. Also I think it's necessary to wait for `data()` in order to get accurate profiling information. Without calling `data()` on the output of a GPU program, we can't know that it has finished executing. 

However it's possible that checking the output for errors distorts the profile - have you observed this?

axinging:
@annxingyuan  @kainino0x 

### Instructions about reproducing how WEBGLQUery distort the profiling:
1.  Merge the change in this PR: https://github.com/tensorflow/tfjs/pull/3552
2.  Put below file inside tfjs\tfjs-backend-webgl\src\benchmark_models_test.ts, set DEBUG_QUERY(Only WebGLQuery and logs) and DEBUG(WebGLQUery, ErrorCheck, ) properly(Do not turn both as true).
3.  tfjs-backend-webgl/package.json, add a line:  "benchmark": "karma start --grep=benchmark",
4.  and run "yarn benchmark --grep=resnet50 --no-single-run"

Expectation: the time logged by console.log(times) should be similar when DEBUG_QUERY is on or off.

```
// tfjs\tfjs-backend-webgl\src\benchmark_models_test.ts
import * as tfc from '@tensorflow/tfjs-converter';
import * as tf from '@tensorflow/tfjs-core';
import {ALL_ENVS, describeWithFlags} from '@tensorflow/tfjs-core/dist/jasmine_util';
import * as profiler from './profiler';

describeWithFlags('Models benchmarks', ALL_ENVS, () => {
  beforeAll(() => {
    jasmine.DEFAULT_TIMEOUT_INTERVAL = 1000000;
  });

  async function benchmark(
      benchmarkFn: () => tf.Tensor | tf.Tensor[] |
          {[name: string]: tf.Tensor}): Promise<number> {
    const start = performance.now();


    let oldLog: any;
    let kernels: any = [];
    const logOutside = false;
    if (logOutside) oldLog = console.log;
    if (logOutside) profiler.startLog(kernels, oldLog);

    const result = benchmarkFn();

    if (result instanceof Array) {
      await result[0].data();
    } else if (result instanceof tf.Tensor) {
      const tmp = result;
      await result.data();
      tmp.dispose();
    } else {
      await result;
    }
    if (logOutside) await profiler.logKernelTime(kernels, oldLog);
    return performance.now() - start;
  }

  it('resnet50', async () => {
    const url =
        'https://storage.googleapis.com/tfjs-models/savedmodel/posenet/resnet50/quant2/model-stride32.json';
    const model = await tfc.loadGraphModel(url);
    const bench = () => model.predict(input);
    await benchmark(bench);
    input.dispose();
    const times = [];
    const trials = 50;
    // tf.ENV.set('DEBUG_QUERY', true);
    // tf.ENV.set('DEBUG', false);
    for (let t = 0; t < trials; ++t) {
      const bench = () => model.predict(input);
      const time = await benchmark(bench);
      input.dispose();
      times.push(time);
    }

    console.log(times);
    const mean = times.reduce((a, b) => a + b, 0) / trials;
    const min = Math.min(...times);
    const fmt = (n: number) => n.toFixed(3);
    console.log(`Mean time: ${fmt(mean)} ms -> ${fmt(mean / 1)} / rep`);
    console.log(`Min time: ${fmt(min)} ms -> ${fmt(min / 1)} / rep`);
  });
});

```

kainino0x:
@kenrussell We discussed this issue at the Intel/TF.js meeting, with disjoint timer query (or maybe some timing code in TF.js) causing significant slowness as measured from the CPU side. Something we'll probably want to discuss at tomorrow's meeting.

PR #3569, opened 2020-07-07
Modularize sign, ceil, floor, neg

tafsiri:
Modularize sign, ceil, floor, neg.

Also improves tooling for speeding up modularization:
 - scripts/extract_op.ts will move ops from one file into their own file, and handle updating the old file as well as exports in ops.ts
 - scripts/touch_modular_ops_files.ts emits a more complete kernel template and also supports generating files for multiple ops and kernels in one call. 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3569)


tafsiri:
@annxingyuan yep, if def helped speed it up. I also saw noticed that it is useful to make local modifications to templates when necessary, for example the gradient of the unary ops have a particular shared pattern, so as i do the rest i'll adapt the template to that pattern (but not commit that change).

PR #3589, opened 2020-07-11
Support set tunable flags in TFJS environment

Linchenn:
This PR does the following:
* write a wrapper of `tf.env().setFlags()` with a tunable flag list, to constrain flag setting.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3589)


nsthorat:
Fly-by review here, we shouldn't be adding setFlags as a public API. This was purposeful so we wouldn't modify flags mid-program as they are supposed to be lazy, read-only (their values are cached and a function of each other).

Linchenn:
Thanks Nikhil! @nsthorat Yes, flags normally should not be modified by users, also because the flags are automatically set to the best supported for a given environment.

Actually, when benchmarking models, users may want the performance data with different flag combinations on their local devices. So we need a solution to provide different flag combinations for a given environment.

As for the lazy loading problem: Since there are dependencies between flags, only modifying an independent flag without modifying its dependents will cause inconsistency. As a result, we excluded flags depending on others from the `TUNABLE_FLAG_TYPE_MAP` and clean the cache when modifying flags (in `tf.env().setFlags(flagConfig)`), so the dependent flags will be added based on the new flags.

What about we use this method only internally and will not add this function into TFJS API doc? Or do you have suggestions here?

nsthorat:
Thanks for the explanation, was just giving you FYI on thought process! Not documenting sounds good.

In a follow up we could also enforce 2 stages of flags: setting and getting. If you try to set after the getting phase we throw an error.

We could roll this out with TF.js 3.0 as well since it could be a breaking change.

cc @tafsiri

Linchenn:
Enforced two stages idea sounds cool, thanks!

tafsiri:
@nsthorat @Linchenn also flying by in response to being tagged. Unless i'm missing something this PR doesn't touch core and thus doesn't have any api impact (private or not), it just adds a utility method to the benchmarking tool?

nsthorat:
Added you for thoughts on the 2-stage + TF.js 3.0 idea (not about the API).

tafsiri:
Meta comment about flags, if we do make any changes to flag functionality in core we should have a at least a lightweight design doc for this. We've had enough discussions about flags over the years that having some written artifact about the design goals/constraints for flags would be good imo :).

Linchenn:
Yes, currently the benchmarking tool has not been exposed as APIs. Thanks Yannick!

PR #3594, opened 2020-07-12
Add ESM builds

munrocket:
**Motivation**

Making lazy load with ES modules easier to use, right now we need to use packages like [fetch-inject](https://www.npmjs.com/package/fetch-inject) that add `js` into head of the page. Dynamic imports with ES modules supported by [%89](https://caniuse.com/#feat=es6-module-dynamic-import) of browsers globally and node 12 without flags.
```js
import('https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm@2.0.1/dist/tf-backend-wasm.esm.js')
  .then(wasmModule => console.log('module loaded'))
  .catch(err => console.log('something wrong'))
```

**Solution**

Just add ES module version into rollup as it already added into tfjs-backend-wasm and tfjs-models/*.

/cc @tafsiri

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3594)

### DIFF
--- a/tfjs-core/src/platforms/platform_browser.ts
+++ b/tfjs-core/src/platforms/platform_browser.ts
@@ -30,7 +30,7 @@ export class PlatformBrowser implements Platform {
   private textEncoder: TextEncoder;
 
   // For setTimeoutCustom
-  private messageName = 'setTimeoutCustom';
+  private readonly messageName = 'setTimeoutCustom';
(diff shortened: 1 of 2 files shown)


munrocket:
🤔 tfjs-converter need tfjs-core, but it works locally for me, when after tfjs-core I build tfjs-converter.

PR #3597, opened 2020-07-13
Modularize fused ops

tafsiri:
This splits out fused ops into their own files.

It doesn't make them modular because they use a pattern that doesn't fit in with how other kernels are implemented and will need some offline discussion. However I wanted to at least save state on moving the code around. 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3597)


tafsiri:
The underlying issue was that these kernels have an optional input tensor (bias) that is sometimes null. They also use a non-ideal way to pass tensors from forward to backward pass, they close over the reference rather than call save([]) as you can't pass null to save. This is also not compatible with new modular gradients, and in my opinion isn't something we should generally do (nullable parameters). In the backward pass an extra result would be added with bias is not null.

Talked with @annxingyuan offline and here are two options I see for this:

1. _Split the fused kernels into two versions_, one that takes a bias and one that doesn't. The op can still take an optional bias argument, but it will use that to decide which kernel to call (which will have a fixed/non-variable interface). 
2. _Do not support gradients for these kernels_. My understanding from talking to Ann is that these fused op calls are generated by the converter for fast inference. We also only support gradients for fused ops using the linear activation (1 of 5 options). We also don't document these ops publicly so they shouldn't appear in arbitrary code that is used for training. If we go down this route we can still keep the optional input tensor in the kernel definition (though this is still exceptional compared to other kernels) as we don't need to worry about keeping a potentially nullable argument for the backward pass. If we don't really see a need for gradients for fused ops this is a simple approach that allows for more dynamic kernels in future.

Thoughts?

tafsiri:
Update: After discussion offline, I tried out option 2, but found out we do used the gradients for these in layers. So went back to the custom op approach and ultimately found an approach that worked (and that doesn't require actually changing customGrad or engine. The summary of the approach is to return different customOps depending on which params are passed in. 

Notes for future consideration: Because CustomGradFunc expects a comma separate list of inputs and then a saveFunc at the end, optional parameters cannot be added (because the last param is non optional). Future refactor should consider separating input tensors into an array argument separate, but that changes the api of customOp.

tafsiri:
One final important note, I had to add a dy.reshape inside the grad func for fused matMul due to the reshape it does at the end of the custom op. 

But this is now ready for review @annxingyuan

PR #3627, opened 2020-07-20
[wasm] Add support for pthreads.

annxingyuan:
This fixes https://github.com/tensorflow/tfjs/issues/3718

### Changes
- adds support for multithreading in the WASM backend.
- deprecates `setWasmPath` in favor of `setWasmPaths` which allows users to specify location of multiple WASM binaries.
- generates Emscripten modules for vanilla / SIMD in standalone mode so the same module can be used for both vanilla and SIMD.

### Bundle sizes:
`master`
- WASM plain
    - js bundle size: 200k
    - WASM binary size: 155k
- WASM SIMD
    - js bundle size: 200k
    - WASM binary size: 213k

`threads_wasm`
- WASM plain
	- js bundle size: 251k
	- WASM binary size: 156k
- WASM SIMD
	- js bundle size: 251k
	- WASM binary size: 214k
- WASM threaded SIMD
	- js bundle size: 251k
	- WASM binary size: 230k

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3627)


annxingyuan:
Hey @Maratyszcza - just wanted to add you as an FYI - no need to fully review :)

tfjs-backend-wasm/src/backend_wasm.ts
```diff
@@ -220,8 +239,18 @@ export async function init(): Promise<{wasm: BackendWasmModule}> {
...
-    const wasm = simdSupported ? wasmFactorySimd(factoryConfig) :
-                                 wasmFactory(factoryConfig);
+    let wasm: BackendWasmModule;
+    if (threadsSupported && simdSupported) {
+      wasm = wasmFactoryThreadedSimd(factoryConfig);
+      wasm.mainScriptUrlOrBlob = new Blob(
+          [`var _scriptDir = undefined; var WasmBackendModuleSimd = ` +
+           wasmFactoryThreadedSimd.toString()],
+          {type: 'text/javascript'});
+      // } else if (simdSupported) {
```
Maratyszcza, line 249: I think you still need this `else if` branch. E.g. Chrome Android supports only SIMD, but not threads

tfjs-backend-wasm/src/cc/backend.cc
```diff
@@ -48,6 +48,7 @@ TensorInfo &get_tensor_info_out(const size_t tensor_id) {
 }
 
 size_t xnn_operator_count = 0;
+pthreadpool *threadpool = pthreadpool_create(4);
```
Maratyszcza, line 51: It is best to not hardcode the number of threads, but get it dynamically from `emscripten_num_logical_cores`. I suggest to use `std::min(emscripten_num_logical_cores(), 4)`.

PR #3647, opened 2020-07-24
webgpu: Use constant value instead of shapes uniforms

qjia7:
PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3647)


qjia7:

Currently, upstream has some test cases failed due to the change https://dawn-review.googlesource.com/c/dawn/+/24481/6/src/common/Constants.h#42. In those shaders, we used more than 4 storage buffers. It seems that we should use uniform buffer or constant value instead of storage buffer to reduce the binding storage buffer count.

PR #3694, opened 2020-07-30
[wasm] Use single Emscripten JS module.

annxingyuan:
This change generates WASM binaries in standalone mode which moves work out of the emscripten-generated JS modules into the binary. In standalone mode, the JS modules for vanilla and SIMD are the same, so we can remove the SIMD module.

Here is how the file sizes compare:

`master`:
tfjs-backend-wasm.js: 27kb
tfjs-backend-wasm-simd.js: 27kb
tfjs-backend-wasm.wasm: 155kb
tfjs-backend-wasm-simd.wasm: 213kb

`wasm_standalone`:
tfjs-backend-wasm.js: 27kb
tfjs-backend-wasm.wasm: 156kb
tfjs-backend-wasm-simd.wasm: 214kb

This means we save 27kb in the JS bundle, but the binaries are significantly larger. 

Performance is unaffected.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3694)


annxingyuan:
FYI @dsmilkov @nsthorat

annxingyuan:
Hi @lina128 - good question - I meant that the inference time is unaffected. I'm not totally sure this is the right change to make - but I wanted to get the team's thoughts. The benefit of this change is to reduce the size of the JS bundle. As we try to support more experimental WASM features, we have to generate separate binaries for each permutation of features, and in some cases separate JS modules as well. In the case of SIMD, generating with emscripten's `STANDALONE=1` option is enough to remove the differences in the JS modules. This necessarily makes the binaries larger, but the upside is that the JS will be faster to load.

For context, this came up because we're trying to add threading support to WASM (https://github.com/tensorflow/tfjs/pull/3627), which does require a separate JS module, so I'm trying to manage the overall bundle size.

tafsiri:
@annxingyuan Maybe to help us with giving feedback, could you group the files downloaded by the end user by 'execution/usage path'. i.e. which pairs/groups of things will a user download for a given setup? e.g. something like

#### master
 - WASM plain
   - js bundle size (total)
   - wasm module size (total)
 - WASM with SIMD support
   - js bundle size (total)
   - wasm module size (total)

#### wasm_standalone
 - WASM plain
   - js bundle size (total)
   - wasm module size (total)
 - WASM with SIMD support 
   - js bundle size (total)
   - wasm module size (total)

I'm not sure if those are the right combinations and maybe there is a better way to do this but I'm currently not sure what to compare against what? It seems like we want to compare some sum of things in the old way to a sum of things in the new way. As it is the total kb in the new branch is more then double the total kb in the first one.

annxingyuan:
Hi Yannick,

Yes, here are the sizes:

`master`
- WASM plain
    - js bundle size (tf-backend-wasm.js): 191k
    - WASM binary size: 52k

- WASM SIMD
    - js bundle size: 191k
    - WASM binary size: 60k

`wasm_standalone`
- WASM plain: 
    - js bundle size: 163k
    - WASM binary size: 156k

- WASM SIMD
    - js bundle size: 163k
    - WASM binary size: 214k

You're right - with this change the total kb downloaded by the user is nearly 2x master, although the initial download of the JS bundle will be slightly faster.

dsmilkov:
For non-simd, looks like this triples the wasm binary size (156k over 52k) which is a significant increase. Given how important bundle sizes are in the JS world, we should maintain the same .wasm binary size for users that are not using simd or threading (which is almost all users of WASM today).

dsmilkov:
A more general question about the benefits of this change.
It seems that the user will still make 2 XHR requests (.js and .wasm) to use TF.js (in either SIMD and non-SIMD configuration). The only difference is that some of the files overlap between the non-SIMD and SIMD configurations, but it is still making 2 requests, is that right?

annxingyuan:
@dsmilkov No, the user only makes one XHR request for the binary in both `master` and `wasm_standalone`. The differences are that with `wasm_standalone` (1) the emscripten module that gets bundled into tf-backend-wasm.js works for both vanilla and SIMD thereby reducing the JS bundle size and (2) the WASM binaries are much larger.

dsmilkov:
Ah, I guess I should have clarified. By 2 requests I meant:
- 1 request to fetch the main js library (the entire js bundle).
- 1 request to fetch the wasm binary.

For today's users of WASM, this PR makes the main js library smaller by 27kb, but .wasm large by 104kb. Thus I see it as a neg-negative change (2 xhr requests, more kb).

nsthorat:
Seeing this PR fly by, meta question: why is it that the .wasm binary is larger in both cases?

That seems counter-intuitive, though this solution generally seems like the right one.

annxingyuan:
@dsmilkov yes, this increases the overall bytes that the user has to download. I've been trying to experiment with different flags for reducing the binary, but haven't made significant progress. Will keep you posted. 

@nsthorat in standalone mode a lot of work that the JS is doing (e.g. webassembly table creation) happens in the binary instead, so it is larger.

tafsiri:
@annxingyuan sure, we would want to add the new method in 2.x and deprecate the old one and then we can remove the old one in 3.x

annxingyuan:
Closing in favor of https://github.com/tensorflow/tfjs/pull/3627

Issue #3637, opened 2020-07-22
Unknown layer: ConvLSTM2D


ZissisT:
Hello, 

I get the following error when trying to import a model in tensorflow.js (v2.0.1)

Uncaught (in promise) Error: Unknown layer: ConvLSTM2D. This may be due to one of the following reasons:
1. The layer is defined in Python, in which case it needs to be ported to TensorFlow.js or your JavaScript code.
2. The custom layer is defined in JavaScript, but is not registered properly with tf.serialization.registerClass().

Is this layer going to be supported in general (it is supported by tensorflow python api already, https://www.tensorflow.org/api_docs/python/tf/keras/layers/ConvLSTM2D ) ?

Thank you

rthadur:
@ZissisT there is not plan to support this in near future, do you wish to contribute ?

ZissisT:
@rthadur I wish I could, but I am new in this area :/ I need some more reading before I can contribute to this one. Thank you for your response though :)

dikatok:
@rthadur may I contribute to implement the layer?

@ZissisT would you mind sharing the model or any examples you are converting from? Just for testing purposes later

ZissisT:
@dikatok  I tried training a model described in this article https://thebinarynotes.com/video-classification-keras-convlstm/ (whole code and model are included there) and my goal was first to predict those videos using tfjs (using my trained model). Then as a second step instead of providing videos as input I would input frames from webcam and try to predict motion sequence (e.g. me executing a Golf Swing). So I trained the model in colab, converted it for tfjs and tried loading it but of course since it is not supported I got the error described in my first post.

beniroquai:
I also had the this message earlier today. Recently, the tensorflow team ported it to the tflite framework. Would be very neat to have it in the tfjs framework too. The link can be found [here](https://github.com/tensorflow/tensorflow/issues/38220). 

The model I would like to export to tfjs looks like this: 


````
from tensorflow import keras
from tensorflow.keras import layers

import tensorflow as tf


def SOFI_seq(Ntime=1, Nbatch=1, Nx=100, Ny=100, features=1, Nchannel=1, upsample=2):
    # create our model here
    kernel_size = 3
    nfilters_lstm = 4
    
    model = keras.Sequential(
    [
     layers.Reshape(target_shape=(Nbatch,Nx//upsample,Ny//upsample,Ntime), name='reshape_1d_2d'),
[...]

PR #3702, opened 2020-07-31
Support for ConvLSTM2D

dikatok:
Support for ConvLSTM2D.

Fixes https://github.com/tensorflow/tfjs/issues/3637.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3702)


dikatok:
Origin issue https://github.com/tensorflow/tfjs/issues/3637

dikatok:
@rthadur is there anyone else to notify about this new feature?

dikatok:
@dsmilkov @pyu10055 hi guys, is it possible to run specific unit tests?

tfjs-layers/src/initializers.ts
```diff
@@ -545,9 +545,8 @@ export class Orthogonal extends Initializer {
 
   apply(shape: Shape, dtype?: DataType): Tensor {
     return tidy(() => {
-      if (shape.length !== 2) {
-        throw new NotImplementedError(
-            'The Orthogonal Initializer does not support non-2D shapes yet.');
+      if (shape.length < 2) {
```
dikatok, line 548: I modified this check since the default recurrent initializer in ConvLSTM2DCell is orthogonal which needs to support 4D. 

This update seems to be inline too with Python implementation (link below). 

Currently, I have only run unit tests in layer repo. Do you have any concerns?

https://github.com/tensorflow/tensorflow/blob/6016e8009ff8d9b3afa5ae5bafd052ed51d22f74/tensorflow/python/ops/init_ops.py#L571

dikatok:
@dsmilkov @pyu10055 I encountered a compatibility issue between BatchNorm and ConvLSTM2D where the BatchNorm does not support tensor rank past 4. 

Is this limited rank support kernel implementation related?

dikatok:
@pyu10055 would you mind taking a look at this? 

I didn't expect this to be so large 😭

dikatok:
@pyu10055 ping

caisq:
@dikatok Sorry for the delay. I'm not reviewing this PR and will get back to you with a review by end of day today.

dikatok:
@caisq 

```
class ConvRNN2D extends RNN {
```
How verbose would this docstring need to be?
<br /><br />
```
if (this.cell.dropoutMask != null) {
        tfc.dispose(this.cell.dropoutMask);
        this.cell.dropoutMask = null;
      }
```
Actually, I was following the way it is structured in recurrent cells, not quite sure about the why either since I could not find any from the commit changes.
<br /><br />
```
generateDropoutMask
```
Should I update the one in recurrent.ts with mine? I think it looks cleaner.

dikatok:
@caisq Done!

Thanks for the quick feedback loop, by the way, I thought the PR would be dangling for quite a while.

Also, would you mind taking a look at my previous comments up there and see if you have any suggestions?

caisq:
Looks like there are lint errors. 

> ERROR: (no-unnecessary-type-assertion) /workspace/tfjs-layers/src/layers/convolutional_recurrent.ts[160, 35]: This assertion is unnecessary since it does not change the type of the expression.

@dikatok Can you run `yarn lint` locally and fix all the errors?

caisq:
@dikatok The code snippet in the comments should use `tf.*` instead of `tfc.*`. This is why the CI is failing.

```js
const filters = 3;
const kernelSize = 3;

const sequenceLength = 1;
const size = 5;
const channels = 3;

const inputShape = [sequenceLength, size, size, channels];
const input = tfc.ones(inputShape);

const cell = tf.layers.convLstm2dCell({filters, kernelSize});

cell.build(input.shape);

const outputSize = size - kernelSize + 1;
const outShape = [sequenceLength, outputSize, outputSize, filters];

const initialH = tfc.zeros(outShape);
const initialC = tfc.zeros(outShape);

const [o, h, c] = cell.call([x, initialH, initialC], {});
```

dikatok:
@caisq ah okay, will fix it later

dikatok:
@caisq can this jsdoc assertion performed locally? or it is only run in CI?

caisq:
@dikatok The snippet-testing code is here:
https://github.com/tensorflow/tfjs/blob/master/tfjs-layers/scripts/test_snippets.ts

I'm not sure about the correct command to invoke it. But you can try figuring it out or maybe @pyu10055 can tell you more. It runs on CI for sure.

caisq:
Actually, running `yarn test-snippet` in the tfjs-layers folder should be the way. See:
https://github.com/tensorflow/tfjs/blob/master/tfjs-layers/package.json#L65

dikatok:
Ah my bad, it should be fixed now.

About these issues I encountered earlier, should they be handled in separate issues/PR in the future?

- batchNormalization if returnSequences: true Error: batchNormalization is not implemented for array of rank 5 yet
- running ConvLSTM2DCell via RNN or StackedRNN, although this should be discouraged
- goBackwards: true Error: WebGL backend: Reverse of rank-5 tensor is not yet supported
- gather (webgl) Error: Gather for rank 5 is not yet supported

caisq:
@pyu10055 PTAL? I think PR is ready to merge.

dikatok:
@caisq BTW, is it acceptable for outside contributors like me to do code refactoring? I kind of want to refactor the `recurrent.ts` file to make it easier to digest. Probably will start with arguments inheritance and the unnecessary getters.

caisq:
@dikatok I think it is acceptable. I'd suggest doing it as a separate PR from this one, though.

PR #3710, opened 2020-08-01
Use custom docker for tfjs-wasm.

lina128:
This PR has below changes:
1. Created a custom docker for tfjs-wasm CI. 
2. Changed the wasm cloudbuild.yml
3. Fix the lint step issue. Before, the step uses bitnami/node image, which doesn't have python 2, so lint step simply skip to run. If we run lint in local, it actually failed. This PR fixes the lint issue, and also changes the image so that lint actually runs.
4. Removed installing emsdk.

Result:
1. The build step reduces from 359s -> 80s, which is 4x+ faster.
2. Making the build stable, by eliminating the flakiness problem during downloading and installing emsdk.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3710)

### DIFF
--- a/dockers/wasm-docker/Dockerfile
+++ b/dockers/wasm-docker/Dockerfile
@@ -1,0 +1,5 @@
+ # Official emsdk docker: https://hub.docker.com/r/emscripten/emsdk
+ FROM emscripten/emsdk:1.39.15
+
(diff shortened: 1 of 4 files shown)


annxingyuan:
Thank you @lina128 !! You are too fast for me :) Thank you for being so on top of everything.

PR #3712, opened 2020-08-01
Optimise defaultComparator function

piyushverma955:
Optimise defaultComparator function.


piyushverma955:
@googlebot I fixed it

PR #3706, opened 2020-07-31
Remove chained ops usage from core

tafsiri:
To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3706)

PR #3726, opened 2020-08-04
Add TF-Op to tfjs-op mapper to converter

tafsiri:
This PR introduces a tool to map from TF-Ops (aka TF-kernels) used in the converter/model.json to the corresponding tfjs op the converter calls. It does this by parsing the executor code.

All TF-ops that do not map into a tfjs-op are listed in a test file to check that only those ops have empty mappings. This includes ones we expect (like TensorListGetItem) as well as a few that should map to ops but don't currently because they used chained ops (e.g. Placeholder). A followup PR will remove chained op calls from converter (similar to #3706). With that the list of unmapped ops will shrink.

This tool is run on build/test and the result is checked into the repo (metadata/kernel2op.json) so that we can have a mapping specific to each commit (and more importantly each release) of tfjs.

We need this mapping to generate custom op imports for the converter based on a specific model.json.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3726)


tafsiri:
@pyu10055 we need this file to be in the npm package so that the build tool can read it when parsing model.json files. We don't want the end user to need to check out the repo.

PR #3737, opened 2020-08-06
Define a list of browsers for the BrowserStack-benchmark tool

Linchenn:
We selected 61 combinations between os, os_version, browser, browser_version, device. (Windows 4, OS X 3, android 33, ios 21).

If you want to use new rules to filter out the list of combinations, or, you want to get the full configuration for a certain combination, you can use the following tool:
``` bash
git clone https://github.com/tensorflow/tfjs.git
cd tfjs/e2e/benchmarks/browserstack-benchmark
git checkout select-BrowserStack-browsers
yarn install

export BROWSERSTACK_USERNAME=YOUR_USERNAME
export BROWSERSTACK_ACCESS_KEY=YOUR_ACCESS_KEY
```
- If you want to use new rules, you can override the [selectBrowsers](https://github.com/tensorflow/tfjs/blob/select-BrowserStack-browsers/e2e/benchmarks/browserstack-benchmark/app.js#L83) function with your rules. (the `availableBrowsers` is an array of all combinations.) Then run `node app.js`, then `browser_list.json` is the updated combination list.
- If you want to get the full configuration for a certain combination, you can run `node app.js` and open http://localhost:8001/ (the webpage includes all available combinations), you can see things like:
Screen Shot 2020-08-05 at 8 37 17 PM

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3737)

### DIFF
--- a/e2e/benchmarks/browserstack-benchmark/browser_list.json
+++ b/e2e/benchmarks/browserstack-benchmark/browser_list.json
@@ -1,0 +1,490 @@
+[
+  {
+    "os": "Windows",
+    "os_version": "10",
+    "browser": "chrome",
+    "device": null,
+    "browser_version": "84.0",
+    "real_mobile": null
+  },
+  {
+    "os": "Windows",
+    "os_version": "10",
+    "browser": "ie",
+    "device": null,
+    "browser_version": "11.0",
+    "real_mobile": null
+  },
+  {
+    "os": "Windows",
+    "os_version": "10",
+    "browser": "edge",
+    "device": null,
+    "browser_version": "84.0",
+    "real_mobile": null
+  },
+  {
+    "os": "Windows",
+    "os_version": "10",
+    "browser": "firefox",
+    "device": null,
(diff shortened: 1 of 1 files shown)


e2e/benchmarks/browserstack-benchmark/browser_list.json
```diff
@@ -0,0 +1,490 @@
...
+    "os_version": "Catalina",
+    "browser": "firefox",
+    "device": null,
+    "browser_version": "79.0",
+    "real_mobile": null
+  },
+  {
+    "os": "ios",
+    "os_version": "13",
+    "browser": "iphone",
```
pyu10055, line 61: can the browser type be specified for iphone/ipad or android? maybe as safari or chrome?

Linchenn: No, currently it is not tunable for mobile devices. (for android, the browser is Chrome; for ios, the browser is Safari)

PR #3133, opened 2020-04-22
Modularize div

haoyunfeix:
To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3133)

PR #3740, opened 2020-08-06
webgpu: Optimize depthwise_conv2d

qjia7:
PERF

This patch also adds the mobilenet as the benchmark. See 15%~30%
improve with different platforms.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3740)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -917,6 +917,10 @@ export class WebGPUBackend extends KernelBackend {
     return this.binaryOp(a, b, binary_op.MUL);
   }
 
+  realDivide(a: Tensor, b: Tensor): Tensor {
+    return this.binaryOp(a, b, binary_op.DIV);
+  }
+
   floorDiv(a: Tensor, b: Tensor): Tensor {
     return this.binaryOp(a, b, binary_op.INT_DIV);
   }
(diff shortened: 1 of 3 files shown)


tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -917,6 +917,10 @@ export class WebGPUBackend extends KernelBackend {
     return this.binaryOp(a, b, binary_op.MUL);
   }
 
+  realDivide(a: Tensor, b: Tensor): Tensor {
```
qjia7, line 920: I don't know why #3133 doesn't work. If I don't add it, the mobilenet model will report 'realDivide' is not implemented error.

annxingyuan: hmmmm that is strange... it should work. I wonder if an isolated div benchmark would throw the same error? or maybe the benchmark script doesn't properly build dependencies (this requires a relatively recent version of tfjs-core)?

qjia7: I think the tfjs-core is already the latest version since you use `link: ../tfjs-core` in `package.json`.  If we switch the backend to webgl in `setup_test.ts`, it can also work well (it seems that there is no `realDivide` in backend_webgl.ts. ). So I am not sure what's the problem.

qjia7:
> Thanks @qjia7 :) are there many other places where we are unnecessarily using a 3D dispatch?

Yes, I am revisiting all supported shaders to see whether we can apply the similar optimization to them. At least,  I think the `pool2d` can do it.  Will submit a separate PR for them. Thanks.

PR #3756, opened 2020-08-09
padding empty tensor fails on node backend

### LINKED ISSUE #3757: Tensor.pad does not work with empty tensor on node backend
To get help from the community, we encourage using Stack Overflow and the [`tensorflow.js`](https://stackoverflow.com/questions/tagged/tensorflow.js) tag.

#### TensorFlow.js version

2.1.0

#### Browser version

Node 14.4.0

Would expect an empty tensor to get padded like any other, it instead throws an exception. I wrote this as a work around and in my tests while writing this I wrote some tests to expect that tensorflowjs throws an error. Lo-and-behold, when I switched backends, my test failed because tfjs didn't actually throw an error, it behaved exactly as my shim did. So I figured I'd share what I learned in case it is something we want to fix.

```
export function pad(x, paddings, constantValue = 0) {
  return tf.tidy(() => {
    if (x.shape.flat().every((n) => n > 0)) {
      return tf.pad(x, paddings, constantValue);
    } else {
      return tf.fill(
        tf.tensor(paddings).sum(-1).add(tf.tensor(x.shape)).arraySync(),
        constantValue
      );
    }
  });
}
```

https://github.com/tensorflow/tfjs/pull/3756

brysgo:
Fixes #3757 issuesI may come back and try and fix this, but this simple pad test throws an exception on the node backend and passes with tfjs.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3756)

### DIFF
--- a/tfjs-core/src/ops/pad_test.ts
+++ b/tfjs-core/src/ops/pad_test.ts
@@ -271,4 +271,11 @@ describeWithFlags('pad', ALL_ENVS, () => {
     // 0, 0, 0
     expectArraysClose(await res.data(), [0, 0, 0, 0, 1, 0, 0, 2, 0, 0, 0, 0]);
   });
+
+  it("pads an empty tensor", async () => {
+    await expect(async () => {
+      const padded = tf.pad(tf.ones([0, 3]), [[5, 6]], 12);
+      expectArraysClose(await padded.data(), await tf.fill([11, 3], 12).data());
+    }).not.toThrow();
+  });
 });


tfjs-core/src/ops/pad_test.ts
```diff
@@ -271,4 +271,17 @@ describeWithFlags('pad', ALL_ENVS, () => {
...
+    expect(
+      tf.pad(
+        tf.ones([0, 3]),
+        [
+          [5, 6],
+          [0, 0]
+        ],
+        12
+      )
+     ).toEqualTensor(tf.fill([11, 3], 12));
```
brysgo, line 285: Oops, I left the matcher I made in 🤦

brysgo:
___
*[tfjs-core/src/ops/pad_test.ts, line 285 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3756#-MEFy1F0cIg84edk2Bht-r1-285:-MEG-dPNC_ngph0d-gYO:b-mju3ki) ([raw file](https://github.com/tensorflow/tfjs/blob/73806c11b73e3d627a50c827ec5bcb9ecc92c2f0/tfjs-core/src/ops/pad_test.ts#L285)):*
<details><summary><i>Previously, brysgo (Bryan Goldstein) wrote…</i></summary><blockquote>

Oops, I left the matcher I made in 🤦 
</blockquote></details>

Fixed (test is still intended to fail as a demonstration of inconsistent behavior between back-ends)

brysgo:
@pyu10055 - didn't feel confident enough to PR my fix since I haven't contributed to TF before 😄 

I can take a stab at it though

brysgo:
I still haven't fixed it, but I pushed a fix for the test which was broken because I wrote it in the GitHub UI.

brysgo:
Closing this because while the behavior is different between backends, the documentation clearly states that the length of the paddings should be equal to the tensor rank, and I derped that one with my example.

PR #3783, opened 2020-08-13
[webgpu] Cache programs in map to reuse them

haoyunfeix:
FEATURE

* store WebGPUBinary in program instance

* remove shaderKey from program

* use pure numbers as key to reduce lookup time

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3783)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -347,6 +347,15 @@ export class WebGPUBackend extends KernelBackend {
   }
 
   // This hash algorithm came from https://source.chromium.org/chromium/chromium/src/+/master:third_party/dawn/src/common/HashUtils.h;l=47;bpv=1;bpt=0
+  /**
+   * Computes unique hash value for each WebGPU program. When hash value is
+   * the same to the other's which means they could reuse the same
+   * bindGroupLayout and pipeline.
+   *
+   * @param input Variables that affect the shaders, including program name,
+   *     input tensors shape, input tensors dtype, program construction
+   *     parameters.
+   */
   private getHash<R extends Rank>(input: Array<boolean|number|ShapeMap[R]|
       number[]>) {
     let hash = 0, i, l;
@@ -368,7 +377,8 @@ export class WebGPUBackend extends KernelBackend {
           view.setFloat64(0, next);
           const i0 = view.getInt32(0);
           const i1 = view.getInt32(1);
-          hash ^= (hash << 6) + (hash >> 2) + i0 + i1 + offset;
+          hash ^= (hash << 6) + (hash >> 2) + i0 + offset;
+          hash ^= (hash << 6) + (hash >> 2) + i1 + offset;
         }
       } else if (typeof(next) === 'boolean'){
         if (next){
@@ -385,26 +395,8 @@ export class WebGPUBackend extends KernelBackend {
   }
 
   private getTensorDType(inputs: TensorInfo[]): number[] {
-    function mapReduceTypeToNum (dtype: string) {
-      if (dtype === 'float32') {
-        return 0x01;
-      }
-      if (dtype === 'string') {
-        return 0x01;
-      }
-      if (dtype === 'int32') {
-        return 0x01;
-      }
-      if (dtype === 'bool') {
-        return 0x01;
-      }
-      if (dtype === 'complex64') {
-        return 0x01;
-      }
-      return 0x00;
-    }
     const dTypes = inputs.map(input => this.tensorMap.get(input.dataId).dtype);
-    const result = dTypes.map(dtype => mapReduceTypeToNum(dtype));
+    const result = dTypes.map(dtype => webgpu_util.hashForType[dtype]);
     return result;
   }
 
@@ -660,7 +652,7 @@ export class WebGPUBackend extends KernelBackend {
   }
 
   private binaryOp(a: Tensor, b: Tensor, op: string): Tensor {
-    const key = this.getHash([binary_op.getOpNum(op), a.shape, b.shape,
+    const key = this.getHash([binary_op.hashForOp[op], a.shape, b.shape,
         this.getTensorDType([a, b])]);
     const program = this.getAndSaveProgram(key, () => {
       return binary_op.getBinaryProgram(op, a.shape, b.shape);
@@ -688,7 +680,8 @@ export class WebGPUBackend extends KernelBackend {
   }
 
   private binaryCompareOp(a: Tensor, b: Tensor, op: string): Tensor {
-    const key = this.getHash([binary_op.getOpNum(op), a.shape, b.shape,
+    const key = this.getHash([webgpu_util.BINARYOPPROGRAM,
+        binary_op.hashForOp[op], a.shape, b.shape,
         this.getTensorDType([a, b])]);
     const program = this.getAndSaveProgram(key, () => {
         return new BinaryOpProgram(op, a.shape, b.shape);
@@ -881,7 +874,7 @@ export class WebGPUBackend extends KernelBackend {
       return binary_op.PRELU;
     }
     throw new Error(`Activation ${
-        activation} has not been implemented for the WebGL backend.`);
+        activation} has not been implemented for the WebGPU backend.`);
   }
 
   fusedConv2d(
@@ -905,7 +898,7 @@ export class WebGPUBackend extends KernelBackend {
       const key = this.getHash([webgpu_util.CONV2DNAIVEPROGRAM,
           input.shape, filter.shape, this.getTensorDType([input, filter]),
           convInfo.outShape, hasBias,
-          webgpu_util.mapActivationToNum(activation),
+          webgpu_util.hashForActivation[activation],
           hasPreluActivationWeights]);
       program = this.getAndSaveProgram(key, () => {
           return new Conv2DNaiveProgram(
@@ -916,7 +909,7 @@ export class WebGPUBackend extends KernelBackend {
           this.getTensorDType([input, filter]),
           convInfo.filterHeight, convInfo.filterWidth, convInfo.inChannels,
           convInfo.inShape, workPerThread, hasBias,
-          webgpu_util.mapActivationToNum(activation),
+          webgpu_util.hashForActivation[activation],
           hasPreluActivationWeights]);
       program = this.getAndSaveProgram(key, () => {
           return new Conv2DMMProgram(
@@ -947,7 +940,7 @@ export class WebGPUBackend extends KernelBackend {
       Tensor {
     const key = this.getHash([webgpu_util.ARGMINMAXPROGRAM, x.shape, 
         this.getTensorDType([x]), axis,
-        webgpu_util.mapReduceTypeToNum(reduceType)]);
+        webgpu_util.hashForReduce[reduceType]]);
     const program = this.getAndSaveProgram(key, () => {
         return new ArgMinMaxProgram(x.shape, axis, reduceType);
     });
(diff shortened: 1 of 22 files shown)


tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -340,12 +338,83 @@ export class WebGPUBackend extends KernelBackend {
...
 
+  private getAndSaveProgram(
+      key: number, getProgram: () => webgpu_program.WebGPUProgram) {
+    if (!this.programCache.has(key)) {
+      this.programCache.set(key, getProgram());
+    }
+    return this.programCache.get(key);
+  }
+
+  // This hash algorithm came from https://source.chromium.org/chromium/chromium/src/+/master:third_party/dawn/src/common/HashUtils.h;l=47;bpv=1;bpt=0
```
qjia7, line 349: Please also add a comment to describe `getHash`'s purpose, and what are the components of its input so that the developer can easily find the unique hash inputs for each program when there are program added.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -592,7 +688,11 @@ export class WebGPUBackend extends KernelBackend {
   }
 
   private binaryCompareOp(a: Tensor, b: Tensor, op: string): Tensor {
-    const program = new BinaryOpProgram(op, a.shape, b.shape);
+    const key = this.getHash([binary_op.getOpNum(op), a.shape, b.shape,
```
qjia7, line 691: Can you let binaryCompareOp to reuse `binaryOp` so that we only need to add key in `binaryOp`? Maybe change the prototype of binaryOp to `binaryOp(a: Tensor, b: Tensor, op: string, compareOp?:bool)`

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -936,20 +1113,32 @@ export class WebGPUBackend extends KernelBackend {
     if (this.shouldExecuteOnCPU([x])) {
       return this.cpuBackend.neg(x);
     }
-    const program = new UnaryOpProgram(x.shape, unary_op.NEG);
+    const key = this.getHash([webgpu_util.UNARYOPPROGRAM, x.shape,
```
qjia7, line 1116: Please use a helper function `UnaryOp` like `BinaryOp` did to reduce the key definiation.

tfjs-backend-webgpu/src/kernels/binary_ops.ts
```diff
@@ -43,6 +43,22 @@ const CHECK_NAN_SNIPPET = `
...
+  if (op === MUL) {return 0x0b03;}
+  if (op === DIV) {return 0x0b04;}
+  if (op === GREATER) {return 0x0b05;}
+  if (op === GREATER_EQUAL) {return 0x0b06;}
+  if (op === LESS) {return 0x0b07;}
+  if (op === LESS_EQUAL) {return 0x0b08;}
+  if (op === SQUARED_DIFFERENCE) {return 0x0b09;}
+  if (op === INT_DIV) {return 0x0b0a;}
+  if (op === PRELU) {return 0x0b0b;}
+  if (op === MAX) {return 0x0b0c;}
```
qjia7, line 58: Add `throw new Error('Unsupported binary op ${op}');`?

tfjs-backend-webgpu/src/kernels/webgpu_program.ts
```diff
@@ -92,10 +93,7 @@ export const compileProgram =
     };
 
 export function makeShaderKey<R extends Rank>(
-    program: WebGPUProgram, shapes: Array<ShapeMap[R]>,
-    types: string[]): string {
-  const key = (program.workGroupSize ? program.workGroupSize.join(',') : '') +
-      shapes.join(',') + types.join(',') + program.variableNames.join(',') +
-      (program.shaderKey ? program.shaderKey : program.userCode);
+    program: WebGPUProgram, types: string[]): string {
```
qjia7, line 96: `makeShaderKey` can be removed?

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -340,12 +338,83 @@ export class WebGPUBackend extends KernelBackend {
...
+  private getAndSaveProgram(
+      key: number, getProgram: () => webgpu_program.WebGPUProgram) {
+    if (!this.programCache.has(key)) {
+      this.programCache.set(key, getProgram());
+    }
+    return this.programCache.get(key);
+  }
+
+  // This hash algorithm came from https://source.chromium.org/chromium/chromium/src/+/master:third_party/dawn/src/common/HashUtils.h;l=47;bpv=1;bpt=0
+  private getHash<R extends Rank>(input: Array<boolean|number|ShapeMap[R]|
```
kainino0x, line 350: Using a hash function in this way is a bit hazardous. This hash function is pretty good, but we can't be totally confident that collisions won't occur. However if we can't find a better solution then it will probably be okay.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -340,12 +338,83 @@ export class WebGPUBackend extends KernelBackend {
...
+      } else if (typeof(next) === 'number'){
+        if (Number.isInteger(next)){
+          hash ^= (hash << 6) + (hash >> 2) + next + offset;
+        } else {
+          const buffer = new ArrayBuffer(8);
+          const view = new DataView(buffer);
+          view.setFloat64(0, next);
+          const i0 = view.getInt32(0);
+          const i1 = view.getInt32(1);
+          hash ^= (hash << 6) + (hash >> 2) + i0 + i1 + offset;
```
kainino0x, line 371: Adding the two numbers together like this is collision prone. You should hash-combine each one separately.

tfjs-backend-webgpu/src/backend_webgpu.ts
kainino0x, line 349: I'm not confident that porting this hash-combine function from C++ to JS is a good idea, because JS numbers behave so differently (in particular, `+` is a f64 operation, not a 2's-complement integer operation). Maybe there's an existing JS hash-combine function you can find?

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -102,7 +101,7 @@ export class WebGPUBackend extends KernelBackend {
   commandQueue: GPUCommandEncoder[];
 
   private commandQueueOwnedIds = new WeakSet<DataId>();
-  private binaryCache: {[key: string]: WebGPUBinary};
+  private programCache: Map<number, webgpu_program.WebGPUProgram> = new Map();
```
kainino0x, line 104: Another (very possibly bad) idea to avoid the need for a hash is to make a big nested table of Maps such that the lookup would do `programCache.get(x).get(y).get(z)`. No idea how well this would perform. The depth could be reduced a little bit by combining some numbers into a single key (e.g. `width * 0x4000000 + height` if you know width is &lt; 2<sup>26</sup>).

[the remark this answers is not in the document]
haoyunfeix: **I like this wonderful idea, and I tried to make it work on posenet demo. (code [here](https://github.com/haoyunfeix/tfjs/compare/cache_programs...haoyunfeix:exp_bigint))
But result shows BigInt doesn't perform well, because the arithmetic operations involving BigInt are SLOW.**

getBigInt

> The floating-point arithmetic is implemented in hardware, Hence it is faster. The Arbitrary-precision arithmetic is entirely implemented in software hence slower. --[tektutorialshub](https://www.tektutorialshub.com/typescript/typescript-bigint-vs-number/)

haoyunfeix:
> I have some comments/ideas inline including some possible alternate implementations that I'm not sure would be any good.
> 
> Do you have any preliminary performance results from this showing an improvement?
> 
> Also, do you know whether it was the string construction or object lookup that was previously slow?

Thanks for the review!
This change will bring about 13~18% improvement for CPU time.
Here are some profiling data for this patch (ResNet50 model with 1000 frames)
- CPU total time: 24042ms down to 20933ms
- LookUp time: 410ms down to 20ms.
- Key construction time: I guess saved about 1000ms, I didn't collect the time in each programs.

 Target | Before | After
------------ | ------------ | -------------
CPU Execution Time | origin-24042 | cache-20933
LookUp Time | origin-lookup | cache-lookup
Key Construction Time | origin-key1 origin-key3 | cache-key

kainino0x:
Awesome data, thanks!

tfjs-backend-webgpu/src/backend_webgpu.ts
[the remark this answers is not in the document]
kainino0x: Hm, I would kind of expect BigInt could perform a little better. If `BigInt(0)` is really slow that's a red flag. BigInts theoretically need allocations (since they can be arbitrarily sized), but up to at least 64 bits ought to be implemented without an allocation (similar to std::string, which can hold short strings internally without any allocations).

That said, fundamentally BigInt and string are the same - they're both an arbitrary length vector of bytes and both could do that internalization optimization. So I'm not too surprised that the performance is similar.

Aside: The simplification that bigint arithmetic is "implemented in software" is kind of misleading. Yes, a floating point operation only requires one instruction, but bigint arithmetic is implemented on top of hardware instructions too, there are just more instructions required. I guess the dealbreaker is that too many instructions are required.

tfjs-backend-webgpu/src/backend_webgpu.ts
haoyunfeix: updated.

tfjs-backend-webgpu/src/backend_webgpu.ts
haoyunfeix: done.

tfjs-backend-webgpu/src/kernels/binary_ops.ts
haoyunfeix: updated.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -340,12 +338,83 @@ export class WebGPUBackend extends KernelBackend {
...
+      }
+      if (dtype === 'int32') {
+        return 0x01;
+      }
+      if (dtype === 'bool') {
+        return 0x01;
+      }
+      if (dtype === 'complex64') {
+        return 0x01;
+      }
```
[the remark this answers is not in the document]
haoyunfeix: updated.

tfjs-backend-webgpu/src/kernels/webgpu_program.ts
haoyunfeix: Done.

tfjs-backend-webgpu/src/kernels/binary_ops.ts
```diff
@@ -43,6 +43,22 @@ const CHECK_NAN_SNIPPET = `
 export const MAX = CHECK_NAN_SNIPPET + `
   return max(a, b);
 `;
+export function getOpNum(op: string): number {
```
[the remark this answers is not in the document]
haoyunfeix: Done. use dictionary instead.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -340,12 +338,83 @@ export class WebGPUBackend extends KernelBackend {
...
+    return hash;
+  }
+
+  private getTensorDType(inputs: TensorInfo[]): number[] {
+    function mapReduceTypeToNum (dtype: string) {
+      if (dtype === 'float32') {
+        return 0x01;
+      }
+      if (dtype === 'string') {
+        return 0x01;
```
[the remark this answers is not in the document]
haoyunfeix: updated.

tfjs-backend-webgpu/src/backend_webgpu.ts
haoyunfeix: done.

tfjs-backend-webgpu/src/backend_webgpu.ts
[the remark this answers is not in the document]
haoyunfeix: @kainino0x I found `BigInt(0)` will take more time than `0n` , e.g. `0 + 0` vs `0n + 0n` vs `BigInt(0) + BigInt(0)` => `3ms` vs `5ms` vs `100ms` in 1 million loops. Bottleneck maybe at BigInt()? I'll continue investigate this and try to make `0n` works on posenet demo to see if we could get some expected data.

PR #3819, opened 2020-08-18
Release and bundle tests

tafsiri:
- Add "npm install" to release e2e tests. This is in addition to yarn as we have observed different behaviour.
- Add e2e tests that test the script tag version of the libray for the union package.
- Disable tree shaking when building the union package bundles.

### DIFF
--- a/e2e/cloudbuild.yml
+++ b/e2e/cloudbuild.yml
@@ -20,7 +20,10 @@ steps:
   entrypoint: 'yarn'
   id: 'test'
   args: ['test-ci']
-  env: ['BROWSERSTACK_USERNAME=deeplearnjs1', 'NIGHTLY=$_NIGHTLY']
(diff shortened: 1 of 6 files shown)


tafsiri:
Note: currently forcing nightly build for this PR since that is where the new tests run. Will remove once its all passing nightly. I've also confirmed locally that I could replicate the bug with the missing OpHandler methods and the script tag tests catch that. I will also try and replicate that here against CI (but that change will be reverted before merging).

tafsiri:
Confirmed that this opHandler bug that was encountered is caught by the tests in CI during nightly tests for the prebuilt bundle (c7074d0). Temporary code forcing nightly builds has been removed and this is ready for final review. 

@pyu10055 Could you take a look. Na caught me up on all the work you did last week, thanks for finding the bug! Let me know if there are any other cases that wouldn't be covered.

PR #3829, opened 2020-08-19
[wasm] Exclude node from the threaded version. 

annxingyuan:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3829)

### DIFF
--- a/tfjs-backend-wasm/package.json
+++ b/tfjs-backend-wasm/package.json
@@ -36,7 +36,9 @@
   },
   "browser": {
     "fs": false,
-    "path": false
+    "path": false,
+    "worker_threads": false,
+    "perf_hooks": false
   },
   "peerDependencies": {
     "@tensorflow/tfjs-core": "link:../tfjs-core"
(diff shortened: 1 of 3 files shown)


annxingyuan:
Hey @tafsiri - I know I mentioned yesterday that I would try to implement a quick fix for adding Node support for threaded WASM yesterday, but after looking into it a bit more I think it makes more sense to exclude Node for now (more details in this issue: https://github.com/tensorflow/tfjs/issues/3830). Let me know if you have any questions!

tfjs-backend-wasm/package.json
```diff
@@ -36,7 +36,9 @@
   },
   "browser": {
     "fs": false,
-    "path": false
+    "path": false,
+    "worker_threads": false,
```
JesseFarebro, line 40: It would be nice to have this change upstream, without specifying these sinks bundlers get confused and will throw errors even for client-side packages. Right now I have to manually patch this for my deployments.

tafsiri: Could you give us more info on this? When you say manually patch, do you mean you need to add these to your bundler configuration explicitly (i.e. these are not picked up automatically). What kind of set up are you using?

JesseFarebro: You can see the issue by just using the WASM webpack starter. Update the tfjs version to `2.3` and try running the dev server. You'll see in the console that it fails on finding both `worker_threads` and `perf_hooks`.

By patch, I mean adding `worker_threads` and `perf_hooks` to the tfjs-wasm package.json. It doesn't seem to work if you add it to the root package.json.

PR #3832, opened 2020-08-19
[tfjs-node]Support cnpm registry

SirM2z:
Support taobao npm mirror.

When tfjs-node is installed, it will download files from storage.googleapis.com.

In China, the [cnpm](https://developer.aliyun.com/mirror/NPM) tool will be used to solve network problems, and it will proxy the above-mentioned link to the mirror site. 

However, when some people only use the parameter (`--registry=https://registry.npm.taobao.org`) to install, they cannot proxy.

This pr is to solve this problem: only use the link installed by the parameter (`--registry=https://registry.npm.taobao.org`)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3832)


SirM2z:
Pass the url through the environment variable `STORE_BASE_HOST=https://cdn.npm.taobao.org/dist/`

```
const BASE_HOST = process.env.STORE_BASE_HOST;
```
But for the developer, he does not know this url: https://cdn.npm.taobao.org/dist/.
So is it possible to tell the developer this information in the installation document

pyu10055 (commented):
> Pass the url through the environment variable `STORE_BASE_HOST=https://cdn.npm.taobao.org/dist/`
> 
> ```
> const BASE_HOST = process.env.STORE_BASE_HOST;
> ```
> 
> But for the developer, he does not know this url: https://cdn.npm.taobao.org/dist/.
> So is it possible to tell the developer this information in the installation document

Thanks, Is this change pushed?

SirM2z:
> > Pass the url through the environment variable `STORE_BASE_HOST=https://cdn.npm.taobao.org/dist/`
> > ```
> > const BASE_HOST = process.env.STORE_BASE_HOST;
> > ```
> > 
> > 
> > But for the developer, he does not know this url: https://cdn.npm.taobao.org/dist/.
> > So is it possible to tell the developer this information in the installation document
> 
> Thanks, Is this change pushed?

already pushed

pyu10055 (approved):

joebnb:
should i use it like `npm config set STORE_BASE_HOST URL` ? 

could give a example url of path corresponding

joebnb (commented):
could change it before tfjs-node publish？

joebnb:
it's seems already published,i track this from a old third node modules😰 pity for that

PR #3848, opened 2020-08-24
make flags dependency in io namespace explicit

tafsiri:
- Make dependencies on IS_BROWSER flag being registered explicit. 
- Also move registration of `BrowserLocalStorage.URL_SCHEME` and `BrowserIndexedDB.URL_SCHEME` from the top-level of thier corresponding IO handlers to the platform_browser file. This co-locates side effects that happen in a browser environment and removes global side effects from the IO handler files themselves.

This fixes a bug we say in syncing to g3 where sometimes flags.ts would get loaded after io/indexed_db.ts (for examples). Since the dependency between the two is implicit rather than explicit. This makes that link explicit which imo is a good pattern and makes it easier to move around other imports as we separate side-effectful code from non side-effectful code.

cc @dsmilkov (this externalizes (and makes more complete) the fix in my last sync cl).

A bigger question is whether we want to replicate this pattern in all modules that depend on some flag being evaluatable. In many cases we have been able to rely on load order as defined in our imports. And once all the side effects and registries have been disentangled from code that we want to tree shake,  relying on load order might be sufficient. But I suspect making the dependency explicit is beneficial for documentation/debugging) purposes, and arguably more correct.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3848)

### DIFF
--- a/tfjs-core/src/io/browser_files.ts
+++ b/tfjs-core/src/io/browser_files.ts
@@ -20,6 +20,7 @@
  * user-selected files in browser.
  */
 
+import '../flags';
 import {env} from '../environment';
 
 import {basename, concatenateArrayBuffers, getModelArtifactsInfoForJSON} from './io_utils';
--- a/tfjs-core/src/io/indexed_db.ts
+++ b/tfjs-core/src/io/indexed_db.ts
@@ -15,10 +15,11 @@
  * =============================================================================
  */
 
+import '../flags';
+
 import {env} from '../environment';
 
 import {getModelArtifactsInfoForJSON} from './io_utils';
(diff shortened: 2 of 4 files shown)


tafsiri:
@pyu10055 for http it does call fetch through a method (`env().platform.fetch()`) rather than the flag mechanism, so I think that is sufficiently explicit for resolution/specifying load order. There would also be a clear exception if not platform is registered. The flags issue caused a very subtle problem where our tests would pass because by the time they run the flag has been registered but the code might throw an exception at load time depending on how the files got loaded.

PR #3854, opened 2020-08-26
model.json support in custom bundle tool & converter

tafsiri:
This PR adds support for creating custom bundles of tfjs that use tfjs-converter to execute a model. There are a lot of files in the diff but many are a change of pattern repeated across different files. I'll try and give high level context of the problem and solution here. But you can see how this is used in [this sample application using blazeface](https://github.com/tafsiri/tfjs-treeshake/tree/master/blazeface). 

### Primary challenge:
We need to determine which ops the converter will call for a given model and make sure only those ops get added to the final bundle. However the converter imports many more ops as part of the op executors that drive model inference.

### High level solution:
In the converter's [op executors](https://github.com/tensorflow/tfjs/tree/master/tfjs-converter/src/operations/executors), rather than doing:

```js
import * as tfc from '@tensorflow/tfjs-core';
tfc.max(...);
```

We do

```js
import * as tfOps from '@tensorflow/tfjs-core/dist/ops/ops_for_converter';
tfOps.max(...)
```

By default tfOps exports all ops in core. But in a custom build a bundler can [alias that module](https://github.com/tafsiri/tfjs-treeshake/blob/master/blazeface/rollup.config.js#L27) to a [generated module containing only the ops](https://github.com/tafsiri/tfjs-treeshake/blob/master/blazeface/blazeface_tfjs/custom_ops_for_converter.js) used by the model. Thus tfOps becomes smaller and the final bundle is smaller.

Other symbols can still be imported from `@tensorflow/tfjs-core` and will go through the regular tree-shaking done by bundlers.

Pretty much everything else is in support of this and other aspects of supporting model.json in custom builds.

### Aliasing modules

This change also fleshes out the workflow creating custom builds on the user end and in particular is geared towards support module aliasing features in bundlers. 

This allows users to switch from regular to custom builds by changing their build files rather than their source code. For example the aliasing defined in this [rollup file](https://github.com/tafsiri/tfjs-treeshake/blob/master/blazeface/rollup.config.js) allow [this source file](https://github.com/tafsiri/tfjs-treeshake/blob/master/blazeface/blazeface_demo.js) to keep importing from `@tensorflow/tfjs` but automatically get re-routed to custom build files. 

This will generally be a better developer experience as it makes switching from full to custom builds easily as feature development is done (a user could even build full and slim bundles and even test them in the same setup).

### Breakdown of changes

- **[tfjs-converter]** Convert `import * as tfc from '@tensorflow/tfjs-core';` -> `import * as tfOps from '@tensorflow/tfjs-core/dist/ops/ops_for_converter';` in all executors and their corresponding tests. Rollup file has also been updated to correctly create a script tag bundle.
- **[tfjs-core]** add `src/ops/ops_for_converter.ts` that exports everything from ops.
- **[tfjs-core]** Move namespaced ops like `image.resizeBilinear` from `src/ops/resize_bilinear` to `src/ops/image/resize_bilinear`. i.e. We make the namespace match the path. This allows us to easily map from a call like tfOps.image.resizeBilinear to a path on disk in order to generate a [custom ops_for_converter](https://github.com/tafsiri/tfjs-treeshake/blob/master/blazeface/blazeface_tfjs/custom_ops_for_converter.js) that correctly recreates the namespace. The diffs for these files come purely from updating import paths that they reference and that reference them. Fused op op files have been renamed to match the path/naming convention. No public exports have changed.
- **[tfjs] cli tool** Add support for parsing model.json files and generating a custom ops_for_converter.js files.
- **[tfjs] cli tool** Generate a custom 'tfjs-core' module in addition to the previously generated custom 'tfjs' module. This avoids circular dependencies after aliasing because our libraries internally import @tensorflow/tfjs-core
- **[tfjs, e2e]** karma.conf files have been updated to match webgl and layers karma.conf files.
- **[tfjs-wasm]** name of exported function from DepthwiseConv2dNativeConfig changed to correctly match the kernel name (2D => 2d).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3854)


tafsiri:
Here is a zip file with script tag builds (and a sample index.html) from this PR, cc @pyu10055 
[tfjs-script-tag-tester.zip](https://github.com/tensorflow/tfjs/files/5138727/tfjs-script-tag-tester.zip)

PR #3868, opened 2020-09-01
Modularize cast, complex, concat, add, sub, multiply, notEqual, slice, fft, ifft.

lina128:
This PR has below changes:
1. It constitutes a minimal set of kernels need to modularize together due to Complex kernel modularization.
2. Changed Complex underlying tensors to tensorInfos.
3. Created/Refactored utility functions for binary kernels.
4. Moved complex related memory usage test to cpu and webgl backend, as they are backend specific now.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868)


lina128 (commented):
Hi Yannick, thank you for the detailed review! Please see my reply below. Happy to discuss in next meeting :)

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868#-:-MGKyWOYDV73jmPLIGls:b-n9ayzu)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 56 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEKgJtCJNuE1gDIGpC:-MGJv2Zo8JtTYowvAo-Y:b8r22i1) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/backend_cpu.ts#L56)):*

should we rename this to complexTensorData or complexTensorInfo
</blockquote></details>

vote for complexTensorInfo. I just realized that this is not handled when moving data from backend to backend. I expect when moving data, we move the underlying complexTensor too, but looking at our code, it seems we don't handle it at all. Or am I missing something?
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 139 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEKZ3vAOzhFim2QZm1:-MGJvklD4tZy3cVm61vl:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/backend_cpu.ts#L139)):*

nit: remove this newline and the one right below.
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 1211 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEMuov-dmRv8PfJ8IK:-MGJw7mQ0mI_706PTae_:b-thfoi8) ([raw file](https://github.com/tensorflow/tfjs/blob/98435f4f70e6afeec00cd3bc91a46e404aba4b1b/tfjs-backend-cpu/src/backend_cpu.ts#L1211)):*

this may still need to be a method on backend (it is like a utility/internal implementation detail—that seems to be used only by cast?). I see in this PR you have converted it to a kernel. Is there a corresponding tensorflow kernel?
</blockquote></details>

There's not a corresponding tensorflow kernel. And it is only used by cast. I can move it back to backend or move it to a util so that it is treeshakable.
___
*[tfjs-backend-cpu/src/kernels/Add.ts, line 20 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEO4OD7Bo2chyfuoEL:-MGK0cm40aMkdj7kbW2R:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Add.ts#L20)):*

naming suggestion for these utilities (that basically mirrors the old names)

createBinaryKernelImpl -> broadcastedBinaryKernelSimple
createBinaryKernelComplexSupportImpl = broadcastedBinaryKernelComplex
createBinaryKernelComplexSupportKernelFunc -> binaryKernelFunc (and make complexImpl an optional parameter). This will allow you to do warnings about whether a kernel supports complex params in one place.
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/kernels/Cast.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEUKt73A3swT1bX05K:-MGK12KJ3WkbBhwKF2BV:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Cast.ts#L34)):*

Suggested comment line above this block. // casting to complex64
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/kernels/Cast.ts, line 58 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEUFcJ9tUhe0G-mbAn:-MGKo8_aATr7cHjdFRhh:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Cast.ts#L58)):*

Suggested comment line above this block. // casting from complex64
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/kernels/Cast.ts, line 59 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGETyU44RxNJ6AcYDjO:-MGK1Sf8AcW3dNs3KC30:b-scba6e) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Cast.ts#L59)):*

Can we ever cast from complex64 to anything else without encoding loss? I ask this because I wanted to suggest moving this up to come right after the block on after the block on casting to complex64. Reading hasEncodingLoss suggests that if you cast from complex64 to bool you wouldn't get encoding loss, but i don't think that is true. Am I missing something?
</blockquote></details>

Good point. Moved up. I think it will always be encoding loss from complex64 to other types.
___
*[tfjs-backend-cpu/src/kernels/Cast.ts, line 76 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEWMGC-SwoEhZZXAsi:-MGK1mCPB5G4SBC3xrbI:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Cast.ts#L76)):*

nit: remove this newline
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/kernels/Complex.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEWk9zBS4vYa44cdLC:-MGK2ABf7KUKuAiPIcDz:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Complex.ts#L34)):*

tensors => tensorInfos
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 65 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEXGq93pI8iPej1zDt:-MGKU7jb2j0jfy18qeew:b-c04iqw) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Complex_test.ts#L65)):*

could you add an assertion for expected numDataBuffers at this point.
</blockquote></details>

Done. Changed to test numDataIds.
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 90 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEXNE48BCW-tRCl49e:-MGKUG8tD-DpPmxIlIDg:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Complex_test.ts#L90)):*

could you add an assertion for expected numDataBuffers at this point.
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 93 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEXbXi82PFUzA8g_i2:-MGKUL1G0itbEWgMHF8Y:b-abk5me) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Complex_test.ts#L93)):*

is this covered by a core test? If not i think this could be in core since it does not look at anything backend specific.
</blockquote></details>

Added numDataIds assertions.
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 108 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEYUo86R_3CXefnp-l:-MGKURpg9WCKbv9OqlMt:b-32ypvn) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Complex_test.ts#L108)):*

is this comment correct? should it be complex still exists?
</blockquote></details>

Nice catch. Removed.
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 141 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEYj6EDLjFnl9IXNX5:-MGKUzvp1_1GRmkiWI5l:b-gpi5ta) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Complex_test.ts#L141)):*

could you split what you have below into another test (something like "dispose after reshape"), it would repeat the same operations as above but without the previous 'expects'
</blockquote></details>

What's the reason? The test case follows the same pattern as other test in this file. 
___
*[tfjs-backend-cpu/src/kernels/Reshape.ts, line 30 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEaLTd6-pN2yPr2oMP:-MGKVcYL7pITP5kPH_rb:b-hukn99) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Reshape.ts#L30)):*

While i don't think these assertions are terribly expensive, we may want to actually measure the impact in a real model since we generally want reshape to be as cheap as possible. This could be a todo to look into with assertions on and off.  We could also put this one behind a debug flag if it is costly.
</blockquote></details>

Added TODO.
___
*[tfjs-backend-cpu/src/kernels/Slice.ts, line 39 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEaqOz7XONcHqO-xky:-MGKVxNL9OQ9xsBNF2Ks:b-xfspli) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Slice.ts#L39)):*

nit: seems like a lot of extra newlines in this block and the next. :)
</blockquote></details>

Removed newlines.
___
*[tfjs-backend-cpu/src/utils/kernel_utils.ts, line 18 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEdaDW94r5VRgchJh7:-MGKW3Bh72Pz0cMPkknA:b-mo5gfb) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/utils/kernel_utils.ts#L18)):*

meta comment about the naming of this file, should this be renamed to binary_kernel_utils.ts to discourage this from getting infinitely long as it gains support functions for various kinds of kernels?
</blockquote></details>

I'm thinking to refactor this file in another PR to make this PR not getting even bigger. Agree that this should be binary_util.ts
___
*[tfjs-backend-cpu/src/utils/kernel_utils.ts, line 25 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEcaxP6dssoBpM1qhO:-MGKWGdw5pzPK158NKAo:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/utils/kernel_utils.ts#L25)):*

These types (their names) confuse me somewhat. Maybe if this were called BinaryOperation/BinaryOperationComplex and the one below SimpleBinaryKernel or something it would help with the the overloading issue between mathematical operations and our ops/kernels.

I also had some naming suggestions for the utility functions below.
</blockquote></details>

Done.
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 31 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEe-lu8xddY5Gb25Ho:-MGKWJFG5P2-_Pvbz59g:b-ce5dl2) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-wasm/src/kernels/Reshape.ts#L31)):*

similar concern here about keeping wasm super cheap. Any particular reason or situation for adding this?
</blockquote></details>

We move the value transformation from reshape.ts op to kernels, these are moved along with them. So they are not getting more expensive than before. If we remove these checks, then there is no where this will be checked.
___
*[tfjs-core/src/engine_test.ts, line 548 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEeWhI2ZCX9nqO_hFh:-MGKvOiw7y4jCVkabjsW:b-gk48gc) ([raw file](https://github.com/tensorflow/tfjs/blob/98435f4f70e6afeec00cd3bc91a46e404aba4b1b/tfjs-core/src/engine_test.ts#L548)):*

did this move? curious about this deletion. seems like a test for core coordinating across multiple backends.
</blockquote></details>

Ah my bad, I wanted to move this to e2e but forgot. Just moved to e2e backend_tests.ts. Keep the test here is kind of more complicated to set up now, because we also have to register kernels for these fake backends. The main thing to test here is backend switch but not registering kernels, so I just move to e2e, the backend_tests.ts is all about testing switching backends. WDYT?
___
*[tfjs-core/src/ops/reshape.ts, line 63 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEfDO44xufWSPZAMFj:-MGKwJ8W1OIaZv7HMAA6:b-ufbbry) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-core/src/ops/reshape.ts#L63)):*

I think these assertions are ones we want to keep at the op level.
</blockquote></details>

These have transformation, not just assertion. If not implemented in kernel, other kernels calling into Reshape kernel will fail (e.g. when the shape has -1).
___
*[tfjs-core/src/ops/slice.ts, line 72 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEfLhz5RkYArPjpkXt:-MGKw_oi2UmB4uSCDFej:bqfo3og) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-core/src/ops/slice.ts#L72)):*

same as above should be kept at op level as part of input validation.
</blockquote></details>

These are not validating raw input, these are validating computed value.
___
*[tfjs-layers/src/layers/core_test.ts, line 118 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEfxJK6NHOd4NZJt6b:-MGKwoXZDl4qNo5yJGiT:b-qdehia) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-layers/src/layers/core_test.ts#L118)):*

is this related?
</blockquote></details>

Yes, after slice get modularized, this will alert kernels not registered without this change. After wrapping in it(){}, it starts working again. We probably should change it to describeMathCPUAndGPU to make the dependency clearer.
___
*[tfjs-layers/src/layers/normalization_test.ts, line 672 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEg-isE9B_cN3XfBxS:-MGKx7gc-DGx6tLV_kF8:bw77io8) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-layers/src/layers/normalization_test.ts#L672)):*

is this related?
</blockquote></details>

Yes, the test below needs to wait for backend to be ready and kernels to be registered.

annxingyuan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[e2e/integration_tests/backends_test.ts, line 130 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhbdLh3f7VjiWWj9MI:-MGhbdLh3f7VjiWWj9MJ:bczjeqt) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/e2e/integration_tests/backends_test.ts#L130)):*
> ```XML
> 
>     tfc.setBackend('webgl');
>     // This scalar lives in cpu2.
> ```

should this be "this scalar lives in webgl"?
___
*[e2e/integration_tests/backends_test.ts, line 143 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhc77x7eGyU3GdT0hT:-MGhc77x7eGyU3GdT0hU:b-2av68r) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/e2e/integration_tests/backends_test.ts#L143)):*
> ```XML
>     expect(tfc.memory().numTensors).toBe(numTensors + 2);
>     expect(tfc.findBackend('webgl').numDataIds()).toBe(webglNumDataIds + 2);
>     expect(tfc.findBackend('cpu').numDataIds()).toBe(cpuNumDataIds);
> ```

i actually thought that both backends would have one more dataId than before :) could you briefly explain the expectations for cpu / webgl? 
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 98 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhifmX0xuiH5LIJAzH:-MGhifmX0xuiH5LIJAzI:bqihly) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/backend_cpu.ts#L98)):*
> ```XML
>   }
> 
>   makeTensorInfo(
> ```

would it make sense for the interface here to just be shape and dtype, to match webgl?
___
*[tfjs-backend-cpu/src/kernels/Complex.ts, line 38 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhf4roE3y9UmhX1P14:-MGhf4rpCgw4Ic1bqjhO:b-qyef3l) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex.ts#L38)):*
> ```XML
>   // underlying tensorData will be disposed.
>   complex.complexTensors = {
>     real: backend.makeTensorInfo(realVals, real.shape, 'float32'),
> ```

so with your change, rather than sharing a dataId from the input real/imag (cloning), we copy the input values into the new tensorinfos - is that right? if so, could you help me understand why it's necessary to create a copy rather than sharing? sorry if i'm missing something we've already talked about!
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 71 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhfzo7EkBJNWsk1PEe:-MGhfzo7EkBJNWsk1PEf:b-q56jw) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex_test.ts#L71)):*
> ```XML
>     // They should not increase tensor count, only complex tensor does.
>     // 3 new tensorData is created for complex, real and imag.
>     expect(tf.memory().numTensors).toBe(numTensors + 1);
> ```

for my own understanding, did this behavior change? without your changes, would numTensors have increased by 3 here?
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 193 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhpuKp8jjjdp0YHP33:-MGhpuKp8jjjdp0YHP34:b-7s6p3c) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex_test.ts#L193)):*
> ```XML
>     const b = a.clone();
> 
>     // 1 new tensor from the reshape.
> ```

clone
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 195 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhq0746X4RRiRnqa9O:-MGhq0746X4RRiRnqa9P:b7l1zcd) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex_test.ts#L195)):*
> ```XML
>     // 1 new tensor from the reshape.
>     expect(tf.memory().numTensors).toBe(memoryBefore.numTensors + 2);
>     // No new tensor data.
> ```

this comment should be updated
___
*[tfjs-backend-cpu/src/kernels/Imag.ts, line 32 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhmiJ_59GEpder2P92:-MGhmiJ_59GEpder2P93:b-9oajkh) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Imag.ts#L32)):*
> ```XML
>   // When complex tensor is disposed, its underlying parts will be disposed too.
>   // Make new tensor out of the imag value of the complex. This makes sure the
>   // value is still accessible even if complex tensor is disposed.
> ```

just to make sure i'm understanding your comment correctly - do we not get this behavior today with clone?
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 37 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhnlu84XS0hD_TCS1X:-MGhnlu995aqs_pJ6muV:b-ye2u8f) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-wasm/src/kernels/Reshape.ts#L37)):*
> ```XML
>       xSize === util.sizeFromShape($shape),
>       () => `new shape: ${$shape}, old shape: ${x.shape}. New shape and old ` +
>           `shape must have the same number of elements.`);
> ```

The new shape (${$shape}}) has ${util.sizefromShape($shape)} elements and the old shape (${x.shape})  has ${xSize} elements. The new shape and old shape must have the same number of elements. 
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 24 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhplcdEAC9CB1IEnU5:-MGhplcdEAC9CB1IEnU6:benim3x) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-webgl/src/kernels/Complex_test.ts#L24)):*
> ```XML
> // tslint:disable-next-line: no-imports-from-dist
> import {describeWithFlags, ALL_ENVS} from '@tensorflow/tfjs-core/dist/jasmine_util';
> ```

nit: it would be nice if we could compare these tests line by line across webgl / cpu
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 52 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhpVXy3Z_TKeAymr8U:-MGhpVXy3Z_TKeAymr8V:bfz4ar1) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-webgl/src/kernels/Complex_test.ts#L52)):*
> ```XML
>     expect(tf.memory().numDataBuffers).toBe(numBuffers + 3);
>     numTensors = tf.memory().numTensors;
>     numBuffers = tf.memory().numDataBuffers;
> ```

iiuc, it looks like a consequence of using tensorinfos instead of tensors for complex tensors on the cpu is that we lose the ability to share data buckets for complex's real / imag components, rather we need to copy them into new tensorinfos? accordingly, do you expect these assertions to change after complex is modularized in webgl?

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[e2e/integration_tests/backends_test.ts, line 130 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhbdLh3f7VjiWWj9MI:-MGifiRUBelLLzKvzA3b:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/e2e/integration_tests/backends_test.ts#L130)):*

should this be "this scalar lives in webgl"?
</blockquote></details>

Done.
___
*[e2e/integration_tests/backends_test.ts, line 143 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhc77x7eGyU3GdT0hT:-MGiw-Q9E8Na-MRRvOKH:b-i1uwg4) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/e2e/integration_tests/backends_test.ts#L143)):*

i actually thought that both backends would have one more dataId than before :) could you briefly explain the expectations for cpu / webgl? 
</blockquote></details>

Added more assertions. The result is disposed after engine.endScope().
___
*[tfjs-backend-cpu/src/kernels/Complex.ts, line 38 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhf4roE3y9UmhX1P14:-MGix6nC3dIrSMWsinGP:b-p049px) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex.ts#L38)):*

so with your change, rather than sharing a dataId from the input real/imag (cloning), we copy the input values into the new tensorinfos - is that right? if so, could you help me understand why it's necessary to create a copy rather than sharing? sorry if i'm missing something we've already talked about!
</blockquote></details>

Because we can no longer 'keep' the underlying tensor, which means if complex is disposed, the underlying parts gets disposed too, imagine these two scenarios: (1) real and imag input tensors are disposed (2) extract the real part from complex and dispose complex, in both scenarios, we want to keep the real part. So we need to copy the value into new tensorinfo when creating complex and also copy the value into new tensorinfo in real kernel.
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 71 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhfzo7EkBJNWsk1PEe:-MGiy0Mt2G-WrrC8MSGI:b-vohbb) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex_test.ts#L71)):*

for my own understanding, did this behavior change? without your changes, would numTensors have increased by 3 here?
</blockquote></details>

Yes. After the change, only the complex tensor is exposed, the underlying parts are tensorData wrapped as tensorInfos, they are not exposed or can be used immediately.
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 193 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhpuKp8jjjdp0YHP33:-MGiyPmD7WowZ5lczqcC:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex_test.ts#L193)):*

clone
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 195 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhq0746X4RRiRnqa9O:-MGiyabICRuJ_229vgaO:b-4pme32) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex_test.ts#L195)):*

this comment should be updated
</blockquote></details>

Why? No new tensor data is created.

lina128 (commented):
Hi @annxingyuan and @tafsiri , thank you for your detailed review, I did another pass. Please take a look. Thank you!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868#-:-MGp3tQX0ONUCLnoJk8c:b93gk6r)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 56 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEKgJtCJNuE1gDIGpC:-MGoiKj6-O2GHJAJl2iS:bkzqg3) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/backend_cpu.ts#L56)):*

vote for complexTensorInfo. I just realized that this is not handled when moving data from backend to backend. I expect when moving data, we move the underlying complexTensor too, but looking at our code, it seems we don't handle it at all. Or am I missing something?
</blockquote></details>

Changed to complexTensorInfo
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 1211 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEMuov-dmRv8PfJ8IK:-MGp-yH-DREHRxlLySEb:b9bhdll) ([raw file](https://github.com/tensorflow/tfjs/blob/98435f4f70e6afeec00cd3bc91a46e404aba4b1b/tfjs-backend-cpu/src/backend_cpu.ts#L1211)):*

There's not a corresponding tensorflow kernel. And it is only used by cast. I can move it back to backend or move it to a util so that it is treeshakable.
</blockquote></details>

Changed to a util.
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 98 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhifmX0xuiH5LIJAzH:-MGjaxox56HYlBAhQfkz:b-tlc5uj) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/backend_cpu.ts#L98)):*

would it make sense for the interface here to just be shape and dtype, to match webgl?
</blockquote></details>

Hi Ann, I changed the function name to makeTensorInfoData to make it less ambiguous. Actually curious why webgl's makeTensorInfo is implemented this way, i.e. write to the backend without value and then read out the tensorData and then assign the value. This adds two additional line of code comparing to makeTensorInfoData here. Is there a case like it needs to know the id before computing the value?
___
*[tfjs-backend-cpu/src/kernels/Cast.ts, line 40 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEQgw_EVHgwfZMio8v:-MGp0HZe2mvz-n47LjZw:bybgnv8) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Cast.ts#L40)):*

you could use makeTensorInfo here right? Would the downside of that be that it is effectively the implementation of zeros? on the other hand it could pretty much stay that way (it doesn't have to import the zeros kernel func). I'm not necessarily opposed to this approach (and then replacing it with the kernelFunc call) but wanted to bring it up for discussion.
</blockquote></details>

Whether we directly use the implementation of zeros here or call zeros kernel has the same effect. I want to use kernel where-ever possible, because I think this force us to think how a large computation can be represented by a set of smaller computations. I remember that Ping mentioned something like eventually any computation can be represented by a minimum set of kernels, so if there is a day when we want to represent any logic by that set, this helps that, but it's a long stretch :)
___
*[tfjs-backend-cpu/src/kernels/Cast.ts, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEVirE1-1bppHA6X3O:-MGp2Cpj8Hh9dvpb12wr:b-adywoj) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Cast.ts#L69)):*

this is what i suspect should be backend.int() or just makeTensorInfo with the existing vals and an int dtype. 
</blockquote></details>

Changed int to a cast util.
___
*[tfjs-backend-cpu/src/kernels/Imag.ts, line 32 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhmiJ_59GEpder2P92:-MGjcCwT2ptZOANhjOzb:b-7epxmy) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Imag.ts#L32)):*

just to make sure i'm understanding your comment correctly - do we not get this behavior today with clone?
</blockquote></details>

Discussed offline, today we use clone and keep, in the new world, we copy the value and create new data bucket.
___
*[tfjs-backend-cpu/src/kernels/Int.ts, line 17 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGE_3Df-rT_2iIe1zwP:-MGp2QwS2pgmH5FjuGyO:bss4a2b) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Int.ts#L17)):*

See comments above about whether we need this kernel.
</blockquote></details>

Removed.
___
*[tfjs-backend-cpu/src/kernels/Reshape.ts, line 30 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEaLTd6-pN2yPr2oMP:-MGp2ZpV9uGu9Nq5ulGd:b-vh1ncv) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Reshape.ts#L30)):*

Added TODO.
</blockquote></details>

Discussed offline, transformation validation should happen in kernels.
___
*[tfjs-backend-cpu/src/utils/fft_utils.ts, line 290 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEbrt13C5-IhAi6aVF:-MGp2xAI9yq9C4irp-ds:b-h0jcuy) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/utils/fft_utils.ts#L290)):*

optional: You could make disposeIntermediateTensorInfo take an array of TensorInfo as well.
</blockquote></details>

Got it. I want to leave it as is for now, and then after modularization, we can decide what's the best way to address something like this. My thinking is, if it happens rather often, we may consider automatic disposal, if it only happens once or twice, then optimize it or not doesn't make much difference.
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 37 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhnlu84XS0hD_TCS1X:-MGjdS6tAohMB47celMr:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-wasm/src/kernels/Reshape.ts#L37)):*

The new shape (${$shape}}) has ${util.sizefromShape($shape)} elements and the old shape (${x.shape})  has ${xSize} elements. The new shape and old shape must have the same number of elements. 
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 52 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhpVXy3Z_TKeAymr8U:-MGjdcxp3WhFgN19p-6Q:b-deucxy) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-webgl/src/kernels/Complex_test.ts#L52)):*

iiuc, it looks like a consequence of using tensorinfos instead of tensors for complex tensors on the cpu is that we lose the ability to share data buckets for complex's real / imag components, rather we need to copy them into new tensorinfos? accordingly, do you expect these assertions to change after complex is modularized in webgl?
</blockquote></details>

Discussed offline, depending on webgl implementation, the test may be different. Having the same implementation as cpu is the easiest when come to disposing, but it does require the underlying data to be copied to new tensorinfos. 
___
*[tfjs-core/src/kernel_names.ts, line 361 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEedYX55hJMtQkzggi:-MGp3j5U9p_n2ZGB80wK:bss4a2b) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-core/src/kernel_names.ts#L361)):*

See earlier comments about Int
</blockquote></details>

Removed.
___
*[tfjs-core/src/ops/slice_util.ts, line 18 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEfnGl6DjM8wlfgNBH:-MGjeIeb9SlvUfb1IQLo:b-ydb2be) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-core/src/ops/slice_util.ts#L18)):*

To the degree that these are op utils they should take tensors and not tensorInfos. Probably related to some of the input validation changes. so we should prob discuss offline.
</blockquote></details>

This is asserting transformed value, after today's discussion, we agreed that transformed value assertion should happen in kernel.

tafsiri (changes requested):
Thanks Na! Looking good, left a few comments, but mostly about naming :). 

> Reviewed 5 of 13 files at r2, 3 of 4 files at r3, 12 of 12 files at r4.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868#-:-MGse7IEDWMGZVI4nemg:bajt6kr)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 98 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhifmX0xuiH5LIJAzH:-MGsYmLX2CDI_tzxgKZ3:b-ipo0w7) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/backend_cpu.ts#L98)):*

Hi Ann, I changed the function name to makeTensorInfoData to make it less ambiguous. Actually curious why webgl's makeTensorInfo is implemented this way, i.e. write to the backend without value and then read out the tensorData and then assign the value. This adds two additional line of code comparing to makeTensorInfoData here. Is there a case like it needs to know the id before computing the value?
</blockquote></details>

to chime in i found the previous name less ambiguous :). The new name makes me wonder if there is makeTensorInfoWithout data. But i'll let ya'll discuss and see what convention makes most sense.
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 141 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEYj6EDLjFnl9IXNX5:-MGsXPSZ80h7QT9gqa0I:bvdlt91) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Complex_test.ts#L141)):*

What's the reason? The test case follows the same pattern as other test in this file. 
</blockquote></details>

Mostly granularity of tests, having fewer expectations per test makes it easier to track down potential source of bugs imo. In this case when reading this test it seems like the first half asserts that reshape works, and the bottom parts asserts that disposing a reshaped complex tensor work. The second half seems to be more of a test of disposing rather than reshaping.

I made a similar comment about the clone test below. I think reading the name of the test makes me think it should just focus on that particular operation. 

What do you think?
___
*[tfjs-backend-cpu/src/kernels/Imag.ts, line 32 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhmiJ_59GEpder2P92:-MGsZkSo7xfnw5j8Tckm:b-bfis91) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Imag.ts#L32)):*

Discussed offline, today we use clone and keep, in the new world, we copy the value and create new data bucket.
</blockquote></details>

Question/thought for future consideration (and to reify my understanding), once we have complex/real/imag implemented in _all_ the backends, we _might_ be able to special case complex tensor handling to not force delete underlying tensorinfos from core and thus enable memory sharing?  (I just took a look at the node backend and it's going to be interesting when accounting for mem management there...). My understanding is that with the need to support both approaches (where core has more direct interactions with backend data via disposeData) we need to copy. 

No need to respond to this unless my understanding is terribly wrong :), just wanted to document this as a something we might be able to revisit if we want to (though I also think it's an okay strategy for cpu backend as is). 
___
*[tfjs-backend-cpu/src/utils/fft_utils.ts, line 290 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEbrt13C5-IhAi6aVF:-MGscas-72s62I3mJeRH:b-1izhep) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/utils/fft_utils.ts#L290)):*

Got it. I want to leave it as is for now, and then after modularization, we can decide what's the best way to address something like this. My thinking is, if it happens rather often, we may consider automatic disposal, if it only happens once or twice, then optimize it or not doesn't make much difference.
</blockquote></details>

+1. tbh i think its probably okay if kernels are a bit more verbose/low level. less abstraction to worry about when thinking about performance.
___
*[tfjs-backend-cpu/src/utils/kernel_utils.ts, line 18 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEdaDW94r5VRgchJh7:-MGscnFyDYReyOhX-wdL:bua2mqt) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/utils/kernel_utils.ts#L18)):*

I'm thinking to refactor this file in another PR to make this PR not getting even bigger. Agree that this should be binary_util.ts
</blockquote></details>

+1 for future PR. 
___
*[tfjs-backend-cpu/src/utils/kernel_utils.ts, line 25 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEcaxP6dssoBpM1qhO:-MGsczz4DKfXTM37NIcm:bi86nwb) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/utils/kernel_utils.ts#L25)):*

Done.
</blockquote></details>

I was also trying to suggest 

SimpleBinaryOp => BinaryOperation
SimpleBinaryOpImpl => SimpleBinaryKernelImpl

ComplexBinaryOp => ComplexBinaryOperation
ComplexBinaryOpImpl => ComplexBinaryKernelImpl

What do you think?
___
*[tfjs-backend-cpu/src/utils/cast_utils.ts, line 22 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsS4QP4s-m3rxYEcNU:-MGsS4QP4s-m3rxYEcNV:bim738y) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/tfjs-backend-cpu/src/utils/cast_utils.ts#L22)):*
> ```XML
> import {assertNotComplex} from '../cpu_util';
> 
> export function int(args: {inputs: {x: TensorInfo}, backend: MathBackendCPU}):
> ```

Simplify this signature to `(x: TensorInfo, backend: MathBackendCPU)`.

I would also call this method something like makeInt32TensorInfo()
___
*[tfjs-backend-cpu/src/utils/cast_utils.ts, line 27 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsUUXd-0gHbMi48Jme:-MGsUUXd-0gHbMi48Jmf:b-oqzyad) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/tfjs-backend-cpu/src/utils/cast_utils.ts#L27)):*
> ```XML
>   const {x} = inputs;
> 
>   assertNotComplex(x, 'int');
> ```

will this be necessary? given the logic already in cast?

A higher level question, would any other kernel use this method directly rather than calling cast? If not do we need to split this out from the cast kernel since its not a particularly long function.
___
*[tfjs-backend-cpu/src/utils/cast_utils.ts, line 33 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsTiLS2iA7fs2e3GCY:-MGsTiLS2iA7fs2e3GCZ:bjt18za) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/tfjs-backend-cpu/src/utils/cast_utils.ts#L33)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>

> ```
>   const resultValues = new Int32Array(util.sizeFromShape(x.shape));
>   for (let i = 0; i < values.length; ++i) {
>     resultValues[i] = values[i];
>   }
> ```
</details>


Replace this with `Int32Array.from(values)`?

annxingyuan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[e2e/integration_tests/backends_test.ts, line 159 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsjOqP5na4O6YTjZO4:-MGsjOqQB4jICAUvx_Gz:b-yoiovj) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/e2e/integration_tests/backends_test.ts#L159)):*
> ```XML
> 
>   // tslint:disable-next-line: ban
>   xit('can move complex tensor from cpu to webgl.', async () => {
> ```

does moving between backends not work? will we be able to re-enable these tests in the future?
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 98 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhifmX0xuiH5LIJAzH:-MGskTxo1_J3Ojk2wJaI:b-ep29w7) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/backend_cpu.ts#L98)):*

to chime in i found the previous name less ambiguous :). The new name makes me wonder if there is makeTensorInfoWithout data. But i'll let ya'll discuss and see what convention makes most sense.
</blockquote></details>

I think it's because webgl kernels operate on outputs that have been registered with `texData` (the backend data registry), but when we create these outputs we don't have the values yet. 

makeTensorInfoWithData works for me, or maybe you could do makeTensorInfo but the values could be the final optional argument?
___
*[tfjs-backend-cpu/src/kernels/Complex.ts, line 38 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhf4roE3y9UmhX1P14:-MGsn7fnAypMBg4EkeUS:bo3zir7) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex.ts#L38)):*

Because we can no longer 'keep' the underlying tensor, which means if complex is disposed, the underlying parts gets disposed too, imagine these two scenarios: (1) real and imag input tensors are disposed (2) extract the real part from complex and dispose complex, in both scenarios, we want to keep the real part. So we need to copy the value into new tensorinfo when creating complex and also copy the value into new tensorinfo in real kernel.
</blockquote></details>

makes sense thanks for the explanation!
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 195 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhq0746X4RRiRnqa9O:-MGsnHph53PWInkiXZRP:b-mcud69) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex_test.ts#L195)):*

Why? No new tensor data is created.
</blockquote></details>

doesn't the assertion below say that there are three new dataIds? maybe i'm not sure what tensor data refers to in your comment.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 52 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhpVXy3Z_TKeAymr8U:-MGso-bRC9QrjhIBqBT1:b-gn6a6l) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-webgl/src/kernels/Complex_test.ts#L52)):*

Discussed offline, depending on webgl implementation, the test may be different. Having the same implementation as cpu is the easiest when come to disposing, but it does require the underlying data to be copied to new tensorinfos. 
</blockquote></details>

makes sense!

tafsiri (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[e2e/integration_tests/backends_test.ts, line 159 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsjOqP5na4O6YTjZO4:-MGt09ZN9Sp6UAHunUTm:b-3w6zgt) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/e2e/integration_tests/backends_test.ts#L159)):*

does moving between backends not work? will we be able to re-enable these tests in the future?
</blockquote></details>

I had suggested this be skipped for now in this PR. My understanding is that this would fail on master (but was discovered while making these changes for CPU). Is that correct Na?

But yeah the idea was that we would have a marker that we have a skipped test to work on fixing.

lina128 (commented):
Thank you again Yannick and Ann! Addressed the comment.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868#-:-MGtOPMd0ZiF1TcfvshS:b-lec0lt)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[e2e/integration_tests/backends_test.ts, line 159 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsjOqP5na4O6YTjZO4:-MGtAHUf2QzchJ4JTNz2:b-k8lwvp) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/e2e/integration_tests/backends_test.ts#L159)):*

I had suggested this be skipped for now in this PR. My understanding is that this would fail on master (but was discovered while making these changes for CPU). Is that correct Na?

But yeah the idea was that we would have a marker that we have a skipped test to work on fixing.
</blockquote></details>

Yes, this would fail on master. Right now, webgl doesn't allow moving complex64 data. And moving data to cpu logic doesn't work due to readSync merge values. As Yannick suggested, I added this pair of tests. After webgl modularization, we should expect this be solved.
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 98 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhifmX0xuiH5LIJAzH:-MGtAvQC3UNKFp_tBwRZ:b3ab0gg) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/backend_cpu.ts#L98)):*

I think it's because webgl kernels operate on outputs that have been registered with `texData` (the backend data registry), but when we create these outputs we don't have the values yet. 

makeTensorInfoWithData works for me, or maybe you could do makeTensorInfo but the values could be the final optional argument?
</blockquote></details>

Ok, I will change it back to makeTensorInfo and make value optional. Thank you both for suggestion!
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 141 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEYj6EDLjFnl9IXNX5:-MGtFUF75skb6C414ixF:b6765t6) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Complex_test.ts#L141)):*

Mostly granularity of tests, having fewer expectations per test makes it easier to track down potential source of bugs imo. In this case when reading this test it seems like the first half asserts that reshape works, and the bottom parts asserts that disposing a reshaped complex tensor work. The second half seems to be more of a test of disposing rather than reshaping.

I made a similar comment about the clone test below. I think reading the name of the test makes me think it should just focus on that particular operation. 

What do you think?
</blockquote></details>

The pattern I see in these memory test all start from some state (num of tensors, num of dataIds), then do something and assert along the way (with a set of assertions), at the end end with the same state as before. We can go extreme to create a test per assertion, which conforms with test principle, but we ended up having several tests with pretty much the same body, so maintenance becomes an issue. For debugging, it shouldn't be an issue, the failure will tell which assertion fail. Based on these two points, I prefer to keep the test this way instead of splitting. WDYT?
___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 195 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhq0746X4RRiRnqa9O:-MGtH45J2xD-JyafYAlY:b-uiyvx0) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Complex_test.ts#L195)):*

doesn't the assertion below say that there are three new dataIds? maybe i'm not sure what tensor data refers to in your comment.
</blockquote></details>

Ah, I see the confusion. The numDataIdsBefore is the initial num of dataIds. The last assertion (3 lines above) also asserts numDataIdsBefore + 3, so this assertion just shows there is no change since last assertion.
___
*[tfjs-backend-cpu/src/kernels/Imag.ts, line 32 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhmiJ_59GEpder2P92:-MGtHb8P6wjiw2d7nFkn:b-e5qfhs) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-cpu/src/kernels/Imag.ts#L32)):*

Question/thought for future consideration (and to reify my understanding), once we have complex/real/imag implemented in _all_ the backends, we _might_ be able to special case complex tensor handling to not force delete underlying tensorinfos from core and thus enable memory sharing?  (I just took a look at the node backend and it's going to be interesting when accounting for mem management there...). My understanding is that with the need to support both approaches (where core has more direct interactions with backend data via disposeData) we need to copy. 

No need to respond to this unless my understanding is terribly wrong :), just wanted to document this as a something we might be able to revisit if we want to (though I also think it's an okay strategy for cpu backend as is). 
</blockquote></details>

Sounds good, once we modularized all backends, we can revisit. Basically, we need a solution to fulfill these two scenarios: (1) real and imag input tensors are disposed (2) extract the real part from complex and dispose complex. 
___
*[tfjs-backend-cpu/src/utils/kernel_utils.ts, line 25 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEcaxP6dssoBpM1qhO:-MGtNTKl3bVQ6coI80cl:bdp7a4a) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/utils/kernel_utils.ts#L25)):*

I was also trying to suggest 

SimpleBinaryOp => BinaryOperation
SimpleBinaryOpImpl => SimpleBinaryKernelImpl

ComplexBinaryOp => ComplexBinaryOperation
ComplexBinaryOpImpl => ComplexBinaryKernelImpl

What do you think?

</blockquote></details>

Done. Except the first one I changed to SimpleBinaryOperation, for consistency.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 24 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGhplcdEAC9CB1IEnU5:-MGtNcXo8DsBZ4WxknuG:b-54snbz) ([raw file](https://github.com/tensorflow/tfjs/blob/271252255d08df43e4ced67d0232cc4a742cca96/tfjs-backend-webgl/src/kernels/Complex_test.ts#L24)):*

nit: it would be nice if we could compare these tests line by line across webgl / cpu
</blockquote></details>

Acknowledged.
___
*[tfjs-layers/src/layers/core_test.ts, line 118 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEfxJK6NHOd4NZJt6b:-MGtNxyqDa0ah0pj7uZ3:b-vxnqaj) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-layers/src/layers/core_test.ts#L118)):*

Yes, after slice get modularized, this will alert kernels not registered without this change. After wrapping in it(){}, it starts working again. We probably should change it to describeMathCPUAndGPU to make the dependency clearer.
</blockquote></details>

This describe is already in describeMathCPUAndGPU, so no need to change that part. We still have to wrap the test in it(){}, otherwise it cannot find the kernel. Something I think relate to the order of executing side effect vs. test runner execution order.
___
*[tfjs-backend-cpu/src/utils/cast_utils.ts, line 22 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsS4QP4s-m3rxYEcNU:-MGtM50y8hHDqEIUn6ww:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/tfjs-backend-cpu/src/utils/cast_utils.ts#L22)):*

Simplify this signature to `(x: TensorInfo, backend: MathBackendCPU)`.

I would also call this method something like makeInt32TensorInfo()
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/utils/cast_utils.ts, line 27 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsUUXd-0gHbMi48Jme:-MGtMD7Z6g7w-VjMG9Uw:b-hxdoeo) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/tfjs-backend-cpu/src/utils/cast_utils.ts#L27)):*

will this be necessary? given the logic already in cast?

A higher level question, would any other kernel use this method directly rather than calling cast? If not do we need to split this out from the cast kernel since its not a particularly long function.
</blockquote></details>

No other places use it right now, so removed the assertion. If the use case expand in future, we can move out at that time.
___
*[tfjs-backend-cpu/src/utils/cast_utils.ts, line 33 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsTiLS2iA7fs2e3GCY:-MGtMZk863R1ge4kqZAf:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/tfjs-backend-cpu/src/utils/cast_utils.ts#L33)):*

> ```
>   const resultValues = new Int32Array(util.sizeFromShape(x.shape));
>   for (let i = 0; i < values.length; ++i) {
>     resultValues[i] = values[i];
>   }
> ```

Replace this with `Int32Array.from(values)`?
</blockquote></details>

Done.

tafsiri (approved):
LGTM from my end! Two suggestions though for you to consider.

> Reviewed 10 of 10 files at r5.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868#-:-MGxmH236dzUrvAiUFJt:b12te7n)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 141 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEYj6EDLjFnl9IXNX5:-MGxlTCpDUJkvDFAfQCp:b-3l2jvf) ([raw file](https://github.com/tensorflow/tfjs/blob/4af61bd8db929795b8959e79c5eaf0f2de64fa12/tfjs-backend-cpu/src/kernels/Complex_test.ts#L141)):*

The pattern I see in these memory test all start from some state (num of tensors, num of dataIds), then do something and assert along the way (with a set of assertions), at the end end with the same state as before. We can go extreme to create a test per assertion, which conforms with test principle, but we ended up having several tests with pretty much the same body, so maintenance becomes an issue. For debugging, it shouldn't be an issue, the failure will tell which assertion fail. Based on these two points, I prefer to keep the test this way instead of splitting. WDYT?
</blockquote></details>

Sure, it's definitely somewhat subjective what the ideal granularity is. So +1 to your preference.
___
*[tfjs-core/src/engine_test.ts, line 548 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEeWhI2ZCX9nqO_hFh:-MGxjTDz85s1hgnxkZd0:b-yfc3um) ([raw file](https://github.com/tensorflow/tfjs/blob/98435f4f70e6afeec00cd3bc91a46e404aba4b1b/tfjs-core/src/engine_test.ts#L548)):*

Ah my bad, I wanted to move this to e2e but forgot. Just moved to e2e backend_tests.ts. Keep the test here is kind of more complicated to set up now, because we also have to register kernels for these fake backends. The main thing to test here is backend switch but not registering kernels, so I just move to e2e, the backend_tests.ts is all about testing switching backends. WDYT?
</blockquote></details>

I do think its good for a basic version of this test to live here, for reference here is a code snippet for how we re-register all kernels to a new backend in tfjs-react-native

```
// Register all the webgl kernels on the rn-webgl backend
    const kernels = tf.getKernelsForBackend('webgl');
    kernels.forEach(kernelConfig => {
      const newKernelConfig =
          Object.assign({}, kernelConfig, {backendName: 'rn-webgl'});
      tf.registerKernel(newKernelConfig);
    });
```

In this case you can copy from 'cpu' to 'cpu1' and 'cpu2'.

Using fakeBackends means there isn't a dependency on any webgl specific behaviour (or bugs), it just focuses on move semantics and keeps it close to CPU.

So I would strongly recommend it (and this can be in addition to moving data between specific real backends in the e2e tests).
___
*[tfjs-backend-cpu/src/utils/cast_utils.ts, line 27 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsUUXd-0gHbMi48Jme:-MGxiXAH5iHT1V-weblk:b-k1exr8) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/tfjs-backend-cpu/src/utils/cast_utils.ts#L27)):*

No other places use it right now, so removed the assertion. If the use case expand in future, we can move out at that time.
</blockquote></details>

Sorry by "do we need to split this out" i meant to ask, why split this out from cast at this time since nothing else is using it.  Non blocking (but recommended), but I think you could just include this function (without exporting it), or just inline the functionality in cast. Right now it opens up the potential for using it in other functions whereas you may want to force them to go through cast.

tafsiri (commented):
And by suggestions I mean its okay if you don't take them :).

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868#-:-MGxmPLH077dP7q20Isa:b1u4v24)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

tafsiri:
Also, great work on this PR @lina128!!

annxingyuan (approved):

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3868)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-core/src/engine_test.ts, line 548 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGEeWhI2ZCX9nqO_hFh:-MHCekiG4Vr_dLI8MjFA:b-eyb036) ([raw file](https://github.com/tensorflow/tfjs/blob/98435f4f70e6afeec00cd3bc91a46e404aba4b1b/tfjs-core/src/engine_test.ts#L548)):*

I do think its good for a basic version of this test to live here, for reference here is a code snippet for how we re-register all kernels to a new backend in tfjs-react-native

```
// Register all the webgl kernels on the rn-webgl backend
    const kernels = tf.getKernelsForBackend('webgl');
    kernels.forEach(kernelConfig => {
      const newKernelConfig =
          Object.assign({}, kernelConfig, {backendName: 'rn-webgl'});
      tf.registerKernel(newKernelConfig);
    });
```

In this case you can copy from 'cpu' to 'cpu1' and 'cpu2'.

Using fakeBackends means there isn't a dependency on any webgl specific behaviour (or bugs), it just focuses on move semantics and keeps it close to CPU.

So I would strongly recommend it (and this can be in addition to moving data between specific real backends in the e2e tests).
</blockquote></details>

Make sense. Added the test back, only test result, not memory.
___
*[tfjs-backend-cpu/src/utils/cast_utils.ts, line 27 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3868#-MGsUUXd-0gHbMi48Jme:-MHCcFCH2Civj_Ir26k7:b-u3lm14) ([raw file](https://github.com/tensorflow/tfjs/blob/28aa12cda9027523bb65548a01a1ed884fad1545/tfjs-backend-cpu/src/utils/cast_utils.ts#L27)):*

Sorry by "do we need to split this out" i meant to ask, why split this out from cast at this time since nothing else is using it.  Non blocking (but recommended), but I think you could just include this function (without exporting it), or just inline the functionality in cast. Right now it opens up the potential for using it in other functions whereas you may want to force them to go through cast.
</blockquote></details>

Ah, got it. Removed this util, just inline the code.

PR #3891, opened 2020-09-08
Update package.json

### LINKED ISSUE #3858: tfjs-automl/demo/object_detection gives no predictions
#### TensorFlow.js version
Not sure... is that the same as the version of tfjs-core that is used?
If I check the tfjs-automl package.json, it looks like it is using tfjs-core 1.2.8, but that seems strange as tfjs is at 2.x right?
Then again if I check the commits it looks like tfjs-automl was never upgraded to 2.x?

#### Browser version
Does not seem relevant. Latest Chrome.

Predictions always return an empty array.

The reason I cloned this repo and tried the object_detection demo is that this was exactly what I was seeing on my own model... Simply no predictions are returned, just an empty array. So I tried to replace my model with a 'known good' one from Google itself, but still the same result. So, thinking I might have some error in my code somewhere, I decided to try the official example in this repo, giving the same result. Empty array...

##### `yarn`
```sh
C:\ws\tfjs\tfjs-automl\demo\object_detection>yarn
yarn install v1.12.3
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.9: The platform "win32" is incompatible with this module.
info "fsevents@1.2.9" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
[4/4] Building fresh packages...
Done in 34.04s.
```
##### `yarn watch`
```sh
C:\ws\tfjs\tfjs-automl\demo\object_detection>yarn watch
yarn run v1.12.3
$ cross-env NODE_ENV=development parcel index.html --no-hmr --open
Server running at http://localhost:1234
| Building index.html...Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`

WARNING: We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely change in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.

You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the following commands:

  npm install --save core-js@2    npm install --save core-js@3
  yarn add core-js@2              yarn add core-js@3

/ Building index.js...Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`

WARNING: We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely change in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.

You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the following commands:

  npm install --save core-js@2    npm install --save core-js@3
  yarn add core-js@2              yarn add core-js@3

/ Building index.js...Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`

WARNING: We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely change in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.

You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the following commands:

  npm install --save core-js@2    npm install --save core-js@3
  yarn add core-js@2              yarn add core-js@3

√  Built in 13.89s.
```

##### Open your browser
Open your browser at http://localhost:1234

##### Empty array is returned
After you open the page, it takes a few seconds in which inference is running. After that, it *should* print a JSON with the returned results below the image and draw some boxes on top of the image containing the detected objects, but instead it only prints `[]` and draws no boxes at all.

### LINKED ISSUE #3885: tfjs-react-native: Error ‘func is not a function’
#### TensorFlow.js version
"@tensorflow/tfjs": "^2.3.0"
"@tensorflow/tfjs-react-native": "^0.3.0"

Hi. I'm trying to run tfjs-react-native project [realtime_demo](https://github.com/tensorflow/tfjs/blob/master/tfjs-react-native/integration_rn59/components/webcam/realtime_demo.tsx) with my android smartphone, but I'm stuck with TensorCamera.

My project directory is [here](https://github.com/piecemakerz/tfjs-example), and if you need more files to inspect error just ask. The error I’m getting is this.

error2

When I debugged my project, I found out that when **TensorCamera** triggers onReady and calls **handleImageTensorReady**, the first argument of handleImageTensorReady - which is **images** - are not passed properly thus undefined. That's why my code gets stuck when `const imageTensor = images.next().value;` is called.

I'm using bare workflow with react-native-unimodules attached manually, but just to check I also tried using [expo-managed-example](https://github.com/tafsiri/tfjs-expo-managed-example). It throws the same `func is not a function` error.

What is causing this problem? Is it because the testing device I'm using too old to run tensorflow modules? Plz help.

#### ps1.

I'm sure this error is not related to loading tensorflow models (blazeface, posenet) because when I comment out all model-related codes from the demo project, I still get same `func is not a function` error. 

I'm sure TensorCamera is the problem because when I changed TensorCamera to ordinary Camera module from "expo-camera", no error occurred.

#### ps2.

I've borrowed Galazy S20 from one of my friends and test the project, and I got the some error. The problem was not with smartphone itself.

rthadur:
Fixes #3885 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3891)

### DIFF
--- a/tfjs-react-native/integration_rn59/package.json
+++ b/tfjs-react-native/integration_rn59/package.json
@@ -21,7 +21,7 @@
     "@tensorflow-models/blazeface": "^0.0.2",
     "@tensorflow-models/mobilenet": "^2.0.4",
     "@tensorflow-models/posenet": "^2.2.1",
-    "@tensorflow/tfjs": "2.0.0",
+    "@tensorflow/tfjs": "2.3.0",
     "@tensorflow/tfjs-react-native": "0.3.0",
     "expo-camera": "^7.0.0",
     "expo-gl": "^7.0.0",


tafsiri:
I don't see how this change would fix the issue, my understanding is that the problem might be in our release. The tests fail so i think it demonstrates that we have a regression in 2.3.0. We are working to get another release out to fix this.

rthadur:
Sure will close the PR , thank you

PR #3892, opened 2020-09-08
increase test timeouts

tafsiri:
Increases timeouts to try and resolve some of the issues/timeouts we are seeing in nightly tests.

From my observation either e2e or wasm fails (sometimes webgl). But there aren't error messages from tests.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3892)

### DIFF
--- a/e2e/benchmarks/browserstack-benchmark/app_test.js
+++ b/e2e/benchmarks/browserstack-benchmark/app_test.js
@@ -50,7 +50,7 @@ describe('benchmark multiple browsers', () => {
   const browsers = {};
 
   beforeAll(() => {
-    jasmine.DEFAULT_TIMEOUT_INTERVAL = 100000;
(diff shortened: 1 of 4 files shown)


lina128 (approved):
Thank you Yannick. Could you also check in Browserstack session when the fail begins? In the past, it can fail early on, in those cases, it is not timeout issue but connection issue. Because we set reconnect tolerance to 0, so once disconnect, we fail the whole test without retry. If that's the case, you can try to change the reconnect to a higher number and see if the situation improves. But of course, let's try one thing at a time. These intermittent issues are hard to reproduce, I know...

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3892#-:-MGj_e0112JtR1rmhsSr:bettru3)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

PR #3897, opened 2020-09-09
Distinguish ops from kernels in tf.profile

tafsiri:
8fPPbEdzvwb9jpL

This makes it easier to process the list of kernels for a custom build since ops implemented with customGrad don't have true backend kernels.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3897)

### DIFF
--- a/tfjs-core/src/ops/operation.ts
+++ b/tfjs-core/src/ops/operation.ts
@@ -16,6 +16,8 @@
  */
 import {ENGINE} from '../engine';
 
+export const OP_SCOPE_SUFFIX = '__op';
+
 /**
  * Used for wrapping functions that perform math operations on
  * Tensors. The function will be wrapped in a named scope that cleans all
@@ -38,9 +40,8 @@ export function op<T extends Function>(f: {[name: string]: T}): T {
     opName = opName.substring(0, opName.length - 1);
   }
 
-  // an an __op suffix to distinguish ops from kernels in tf.profile
-
-  opName = opName + '__op';
+  // add an __op suffix to distinguish ops from kernels in tf.profile
+  opName = opName + OP_SCOPE_SUFFIX;
 
   // tslint:disable-next-line:no-any
   const f2 = (...args: any[]) => {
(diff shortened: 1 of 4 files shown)


tafsiri:
cc @nsthorat in case there is any corner case or other use of the scope names as we had them before.

annxingyuan (approved):
Just to make sure I understand - in the example described in the PR description, the model calls the `FusedConv2D` kernel and calls `fusedConv2D` through custom grad?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3897#-:-MGo4Dcv9_ge1Fuglzn_:bm9jr00)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan and @tafsiri)

___
*[tfjs/tools/custom_bundle/cli.ts, line 107 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3897#-MGo3NCzEX5El3BL3HPr:-MGo3NCzEX5El3BL3HPs:b3dnbqh) ([raw file](https://github.com/tensorflow/tfjs/blob/6755bfd5232754fcd11ead7f17fb1918d4a03c0d/tfjs/tools/custom_bundle/cli.ts#L107)):*
> ```XML
>   // kernels so we need to filter them out.
>   function isNotCustomOp(kernelName: string) {
>     return !kernelName.endsWith('__op');
> ```

could we maybe pull out `__op` into a variable and reference it elsewhere?
___
*[tfjs-core/src/ops/operation.ts, line 41 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3897#-MGo3Hwb8_Rl8c1_6Ugr:-MGo3Hwb8_Rl8c1_6Ugs:b-pngnao) ([raw file](https://github.com/tensorflow/tfjs/blob/6755bfd5232754fcd11ead7f17fb1918d4a03c0d/tfjs-core/src/ops/operation.ts#L41)):*
> ```XML
>   }
> 
>   // an an __op suffix to distinguish ops from kernels in tf.profile
> ```

Add an __op

tafsiri:
@annxingyuan fusedConv2d the op is called which starts a scope with the op name (this name won't be replaced by a kernel name when you enter the first engine.runKernelFunc). Later on this op will call another engine.runKernelFunc with FusedConv2d which will also get added to tf.profile (it will start a new scope).

For reference above are the kernels captured for blazeface by tf.profile (sorry I should have put that in the original discussion).

PR #3939, opened 2020-09-16
WIP add cached build deps script

tafsiri:
With change measurement 1: 25m 16s
Without change measurement 1: 26m 29s
With change measurement 2: 32m 16s

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3939)


pyu10055 (commented):
Thanks for looking into the testing speed, can you explain the time measurements? is it getting slower with changes or I read it wrong?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3939#-:-MJCrN6C8YQ6qRG34Php:b-5j0j2n)* status: 0 of 1 approvals obtained

lina128 (commented):
I thought this is not ready for review yet, maybe Yannick is still experimenting?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3939#-:-MJDGfHk-rsB1RSnaQmf:b-n2tdzn)* status: 0 of 1 approvals obtained (waiting on @lina128)

tafsiri:
Yeah this was an experiment that I wanted to try out (and had briefly discussed with Na). Was trying to understand if we can save some built time by reusing 'build-deps' outputs from previous steps. From current results its not having any effect. I left this open for myself to dig a bit deeper when I get a chance/new ideas. My current suspicion is that within a cloudbuild.yml we don't re-build the same deps as often, but across all our cloudbuild.ymls we do. however they don't share the same disk (or there is a race condition in the code I have to check if something is already built). In any case this is def not ready for review/merging.

tafsiri:
But to explain the results they imply no significant change in build time with or without the change. I applied the change, unapplied it and then applied it again and recorded times for each.

PR #3940, opened 2020-09-17
webgpu: Fix the mapAsync error

qjia7:
BUG

This change fixes below error:
Uncaught (in promise) DOMException: Unknown error in mapAsync.

Since f is an async function, we should use 'await f()' to make sure the promise has finished.
Otherwise, the execution sequence of function in and out of the body will mess and generate weird result.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3940)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu_test.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu_test.ts
@@ -252,8 +252,8 @@ describeWebGPU('backend webgpu', () => {
       tf.setBackend('webgpu');
       const d = tf.add(a, b);
       await d.data();
+      expect(() => d.dataSync()).not.toThrow();
     };
-
-    expect(f).not.toThrow();
+    await f();
   });
 });


tfjs-backend-webgpu/src/backend_webgpu_test.ts
```diff
@@ -252,8 +252,8 @@ describeWebGPU('backend webgpu', () => {
       tf.setBackend('webgpu');
       const d = tf.add(a, b);
       await d.data();
+      expect(() => d.dataSync()).not.toThrow();
     };
-
-    expect(f).not.toThrow();
```
qjia7, line 257: Not sure for async function, how should we use expect.not.toThrow. So I changed the logic here and move it to the body of async function.

qjia7:

annxingyuan (approved):
Thanks @qjia7 !

PR #3956, opened 2020-09-22
[webgl] Modularize cast, complex, concat, add, sub, multiply, notEqual, slice, fft, ifft.

annxingyuan:
#### Changes
- Export `impl` functions in CPU binary kernels so they can be used in WebGL. 
- Introduce `kept` property of WebGL texture data that prevents it from being disposed. 
- Properly clean WebGL canvas contexts in between tests to prevent flaky failures.
- Modularize cast, complex, concat, add, sub, multiply, notEqual, slice in the WebGL backend.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956)


lina128 (commented):
This looks great, Ann! I left some questions, but overall LGTM. If we all agree with the kept approach, I can make the same change in cpu too.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956#-:-MIW0akEBjOIO39AcoqX:bppp4yj)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[e2e/benchmarks/model_config.js, line 223 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVq0Rb96rrqEUhuouc:-MIVq0Rb96rrqEUhuoud:bxk3fui) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/e2e/benchmarks/model_config.js#L223)):*
> ```JavaScript
>     }
>   },
>   'speech-commands': {
> ```

Is this related to this PR?
___
*[tfjs-backend-cpu/src/kernels/Concat.ts, line 53 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVqyqlBzIaRp4jU_yM:-MIVqyqlBzIaRp4jU_yN:b-4kkit8) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-cpu/src/kernels/Concat.ts#L53)):*
> ```XML
>     const imags = $inputs.map((t) => imag({inputs: {input: t}, backend}));
> 
>     const realConcated = concat({inputs: reals, backend, attrs: {axis: $axis}});
> ```

Nice catch!!!
___
*[tfjs-backend-cpu/src/utils/kernel_utils.ts, line 38 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVrVgS-bcLrPM2e3a-:-MIVrVgS-bcLrPM2e3a0:bjaqtj1) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-cpu/src/utils/kernel_utils.ts#L38)):*
> ```XML
>  */
> export function binaryKernelFunc(
>     name: string, binaryKernelImpl: SimpleBinaryKernelImpl,
> ```

sugg: simpleImpl, complexImpl
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 627 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVz31XAXnE5atensqw:-MIVz31XAXnE5atensqx:b8a3w3n) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/backend_webgl.ts#L627)):*
> ```XML
>     const {complexTensorInfos} = this.texData.get(dataId);
>       this.texData.get(complexTensorInfos.real.dataId).kept = false;
> ```

If I create 2 complex tensors (complex1 and complex2) from the same real tensor (real). Then I delete complex1, then I delete real, then I think complex2 don't have real anymore, right? Should the kept field be a int field, which record how many other data depend on it?
___
*[tfjs-backend-webgl/src/tex_util.ts, line 81 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVt5Tp2uJ8cQzjtZKS:-MIVt5Tp2uJ8cQzjtZKT:b-678wt2) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/tex_util.ts#L81)):*
> ```XML
>   texture?: WebGLTexture;
>   // For complex numbers, the real and imaginary parts are stored as their own
>   // individual tensors, with a parent joining the two with the
> ```

individual tensorInfos
___
*[tfjs-backend-webgl/src/kernel_utils/int.ts, line 25 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVtQpx295wY8IFB5Xa:-MIVtQpx295wY8IFB5Xb:bazjkr7) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/int.ts#L25)):*
> ```XML
> const TO_INT = `return float(int(x));`;
> 
> export function int(input: TensorInfo, backend: MathBackendWebGL): TensorInfo {
> ```

For cpu, Yannick suggests not using int as a kernel but a util, because there's no equivalent int kernel in TF c++ api. @tafsiri please confirm.
___
*[tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts, line 66 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVvokD-UQl0UDbw5dR:-MIVvokD-UQl0UDbw5dS:bkhofam) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts#L66)):*
> ```XML
>  */
> export function binaryKernelFunc({
>   opSnippet,
> ```

Curious why do we define inputs as named parameters here? I thought most code (except kernels and layers) in our codebase define inputs as positional parameters, like binaryKernelFunc(opSnippet, packedOpSnippet, checkOutOfBounds = false, ...).
___
*[tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts, line 73 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVtwn2-dtZhWXXaS6g:-MIVtwn2-dtZhWXXaS6h:b-y2iqq2) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts#L73)):*
> ```XML
>   dtype
> }: {
>   opSnippet: string,
> ```

Can we define the type maybe type BinaryKernelFuncInput?
___
*[tfjs-backend-webgl/src/kernel_utils/shared.ts, line 23 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVvW6E0NE55iRIf7CD:-MIVvW6E0NE55iRIf7CE:b-2wzplx) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/shared.ts#L23)):*
> ```XML
> import * as shared from '@tensorflow/tfjs-backend-cpu/dist/shared';
> // tslint:disable-next-line: no-imports-from-dist
> import {SimpleBinaryKernelImpl} from '@tensorflow/tfjs-backend-cpu/dist/shared';
> ```

Curious, why do we need to import from dist?
___
*[tfjs-backend-webgl/src/kernels/Cast.ts, line 63 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIW-TQvBwY4Unnu8IID:-MIW-TQw5zlj09vMNTlx:bvovnd0) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Cast.ts#L63)):*
> ```XML
> 
>   // Casting from complex64
>   if (x.dtype === 'complex64') {
> ```

In the cpu version, Yannick suggests moving this condition above the hasEncodingLoss condition, because any code behind it means there will be some encoding loss. @tafsiri please confirm.
___
*[tfjs-backend-webgl/src/kernels/Complex.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVxVdiCr8tteQSdnSp:-MIVxVdj4RWFv-1EuBuw:b-4ff6de) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex.ts#L34)):*
> ```XML
>   realData.kept = true;
>   const realTensorInfo:
>       TensorInfo = {dataId: real.dataId, shape: real.shape, dtype: 'float32'};
> ```

Why not use identity?
___
*[tfjs-backend-webgl/src/kernels/Imag.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIW0RAb-H2z2GLzzdPG:-MIW0RAc2OrHZbcD7mx6:bhasvmd) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Imag.ts#L28)):*
> ```XML
> 
>   const inputData = backend.texData.get(input.dataId);
>   backend.incRef(inputData.complexTensorInfos.imag.dataId);
> ```

Consider use identity.

jinjingforever (commented):
> Reviewed 6 of 42 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, and @tafsiri)

*[a discussion](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIWFlbs4IMRTVUTfHKl:-MIWFlbt32vurXFtEw5A:b-h2iitc) (no related file):*
Great PR! Thanks Ann very much. LGTM

---
___
*[tfjs-backend-cpu/src/utils/kernel_utils.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIWDI7742eh5zCYNdWe:-MIWDI7742eh5zCYNdWf:b-az9ixx) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-cpu/src/utils/kernel_utils.ts#L36)):*
> ```XML
>  *     result has the same dtype as the first input. This is mainly used in
>  *     comparison kernels, such as Equal, Less, Greater, etc.
>  */
> ```

Update comments to match the new params

tafsiri (changes requested):
Great work, Fantastic to see this! I had a few requests for changes/clarifications/requests for extra tests. I'll hold off on giving my thoughts on the kept approach in cpu for now :). I'm still generally a bit curious about the mechanism for the kept bit being cleared, and also as Na asked, what happens if you dispose a complex tensor but you already called real()/imag() on it and want to hold on to that tensor?

It also looks like there are some e2e tests for moving complex tensors that may not be passing atm. 

> Reviewed 42 of 42 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956#-:-MIXEKG20nSfc4r5F_2v:bzffnyt)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 357 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIX0LbL70bD9bPs1coR:-MIX0LbM-qhfS_y9bC84:b6s1m4x) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/backend_webgl.ts#L357)):*
> ```XML
>     if (dtype === 'complex64') {
>       const realValues =
>           this.getValuesFromTexture(complexTensorInfos.real.dataId);
> ```

should this call this.readSync on these dataIds? (similar to the this.read call below) in are they are packed/sliced.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 620 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIX1Ixg2fwVFrppYMhx:-MIX1Ixg2fwVFrppYMhy:bc4wzyt) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/backend_webgl.ts#L620)):*
> ```XML
> 
>     if (this.texData.get(dataId).kept) {
>       this.texData.get(dataId).refCount--;
> ```

could this ever become negative?
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 627 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVz31XAXnE5atensqw:-MIX0fkB5G5VcyItvIia:b-sjzlta) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/backend_webgl.ts#L627)):*

If I create 2 complex tensors (complex1 and complex2) from the same real tensor (real). Then I delete complex1, then I delete real, then I think complex2 don't have real anymore, right? Should the kept field be a int field, which record how many other data depend on it?
</blockquote></details>

this is good one to write up as a test.
___
*[tfjs-backend-webgl/src/kernel_utils/int.ts, line 25 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVtQpx295wY8IFB5Xa:-MIX30l7Cc6pqfSxX5AY:b-uc6f70) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/int.ts#L25)):*

For cpu, Yannick suggests not using int as a kernel but a util, because there's no equivalent int kernel in TF c++ api. @tafsiri please confirm.
</blockquote></details>

+1. But i think is is a util already.
___
*[tfjs-backend-webgl/src/kernel_utils/shared.ts, line 23 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVvW6E0NE55iRIf7CD:-MIX4-cxDMPgLLM67jgZ:bf33c4v) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/shared.ts#L23)):*

Curious, why do we need to import from dist?
</blockquote></details>

I think its because in the kernel_funcs_utils.ts file above, in order to templatize cpu forwarding of binary kernels this type is imported. It does produce a somewhat tight coupling between the binary kernels in cpu to those in webgl. But I think that is the reason (Ann feel free to clarify). Na is this a reasonable constraint to put on binary kernels in CPU?

Or were you more specifically asking about importing from dist as a path? This is to not trigger auto registration of CPU kernels.
___
*[tfjs-backend-webgl/src/kernels/Cast.ts, line 63 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIW-TQvBwY4Unnu8IID:-MIX668R7lXdXh_N3djK:b-72ocxn) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Cast.ts#L63)):*

In the cpu version, Yannick suggests moving this condition above the hasEncodingLoss condition, because any code behind it means there will be some encoding loss. @tafsiri please confirm.
</blockquote></details>

+1. It was mainly to keep the logic close to the other block for handling complex tensors as when casting from a complex tensor to a non complex tensor you will always have encoding loss (in this case we throw out the imaginary part). 
___
*[tfjs-backend-webgl/src/kernels/Complex.ts, line 22 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIX8UzT4w0dF5-OiP6F:-MIX8UzT4w0dF5-OiP6G:b-mketub) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex.ts#L22)):*
> ```XML
> import {MathBackendWebGL} from '../backend_webgl';
> 
> export function complex(
> ```

given the (ahem) complexity of implementing complex would this be a good place for a high level description of how complex is implemented in webgl? It could shed some light on memory/texture concerns that suggest the importance of sharing memory as well as the general expected lifecycle of complex tensors (and their sub parts) and how they relate to stuff in the engine like disposal. I think it would be good to have that written out (somewhere) while its still in our heads.
___
*[tfjs-backend-webgl/src/kernels/Complex.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVxVdiCr8tteQSdnSp:-MIX9D2W1_yJrhQ8VK0d:bqrcr5i) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex.ts#L34)):*

Why not use identity?
</blockquote></details>

+1 (and same below)
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 35 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIX9Yz87hI4h4LpWaIY:-MIX9Yz87hI4h4LpWaIZ:bp7zoll) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L35)):*
> ```XML
>     const imag1 = tf.tensor1d([2]);
> 
>     // 2 new Tensors: real1, imag1, complex1, and two internal clones.
> ```

the rest of the comment still describes 5. also add info about expected dataids
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 42 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXA0q17dnUm3NE4p8h:-MIXA0q17dnUm3NE4p8i:b-3psc31) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L42)):*
> ```XML
> 
>     const complex1 = tf.complex(real1, imag1);
>     // 1 new Tensor and 1 new TensorData for complex, real and imag.
> ```

could you remind me of the relation between tensordata, tensorinfo and dataId?

Would you also want to assert here that you are still have the same number of bytes in GPU memory (or same number of textures)?  (i'm guessing it should be the same as before the complex was created or am i mistaken?).
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 51 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXAhF4-snJwSnU2qtP:-MIXAhF4-snJwSnU2qtQ:b-67ssct) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L51)):*
> ```XML
>     const imag2 = tf.tensor1d([4]);
> 
>     // 2 new Tensors: real2, imag2.
> ```

add tensordata info to stay consistent 
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXAyrXE2OP-C45vQd_:-MIXAyrXE2OP-C45vQda:b5c4tz2) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L69)):*
> ```XML
>     // A complex tensor is created, which is composed of real and imag parts.
>     // They should not increase tensor count, only complex tensor does.
>     // 3 new tensorData is created for complex, real and imag.
> ```

do you want to add an assertion about underlying memory as well?
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 113 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXD2GJ4tGETd_Y1F1M:-MIXD2GJ4tGETd_Y1F1N:b32fz6k) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L113)):*
> ```XML
>   });
> 
>   it('tf.complex disposing underlying tensors', async () => {
> ```

adding a similar test but where the real and imaginary parts are created from the complex tensor and then disposed would also be helpful.

(a parallel version could be added to core in complex_ops that disposes the real and imag and check that isDisposed is true on the tensor, and that tf.memory has had the appropriate reduction). 
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 186 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXBaiEDkXW_D89IA1e:-MIXBaiEDkXW_D89IA1f:b-1n1qjs) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L186)):*
> ```XML
>     // 1 new tensor, the complex64 tensor
>     expect(tf.memory().numTensors).toBe(memoryBefore.numTensors + 1);
>     // 1 new tensor and 2 underlying tensors for real and imag.
> ```

should this be tensorData? (or whatever we pick)
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 194 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXBgQLBAX0_Q-4KF05:-MIXBgQLBAX0_Q-4KF06:b59g0eh) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L194)):*
> ```XML
>     expect(tf.memory().numTensors).toBe(memoryBefore.numTensors + 2);
>     // No new tensor data.
>     expect(tf.engine().backend.numDataIds()).toBe(numDataIdsBefore + 3);
> ```

Af first glance this doesn't seem to match the comment above. I think its because here you don't use the style you do above or re assigning the counts after each assertion so that you can assert the specific increase. Could you use that style here?
___
*[tfjs-backend-webgl/src/kernels/Imag.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIW0RAb-H2z2GLzzdPG:-MIXDt996rKtebOfd749:b32v57) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Imag.ts#L28)):*

Consider use identity.
</blockquote></details>

+1
___
*[tfjs-backend-webgl/src/kernels/Real.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXE7nL8vvNRSb683VW:-MIXE7nL8vvNRSb683VX:b-m2jtqc) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Real.ts#L28)):*
> ```XML
> 
>   const inputData = backend.texData.get(input.dataId);
>   backend.incRef(inputData.complexTensorInfos.real.dataId);
> ```

use identity

annxingyuan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[e2e/benchmarks/model_config.js, line 223 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVq0Rb96rrqEUhuouc:-MI_jJelBJWR6wYs5bPn:bqv66i1) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/e2e/benchmarks/model_config.js#L223)):*

Is this related to this PR?
</blockquote></details>

I added it in order to benchmark FFT given my changes to complex. I think it would be nice to be able to easily benchmark speech-commands in the future as well so I added it here.
___
*[tfjs-backend-cpu/src/utils/kernel_utils.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIWDI7742eh5zCYNdWe:-MI_jx2Q-kXz5YY4QvN4:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-cpu/src/utils/kernel_utils.ts#L36)):*

Update comments to match the new params
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/utils/kernel_utils.ts, line 38 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVrVgS-bcLrPM2e3a-:-MI_kA9z06qxlCiBvVsn:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-cpu/src/utils/kernel_utils.ts#L38)):*

sugg: simpleImpl, complexImpl
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 357 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIX0LbL70bD9bPs1coR:-MI_lvVU2bnTEAc9x56X:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/backend_webgl.ts#L357)):*

should this call this.readSync on these dataIds? (similar to the this.read call below) in are they are packed/sliced.
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 620 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIX1Ixg2fwVFrppYMhx:-MI_m_jjDhScugO5B2eO:b-zaa0iy) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/backend_webgl.ts#L620)):*

could this ever become negative?
</blockquote></details>

Yes, if you keep calling dispose on a tensor that is `kept`. What do you think would be the best way to handle this case?
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 627 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVz31XAXnE5atensqw:-MIdtNc8A3YLKHhw71yC:b-5a3mfe) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/backend_webgl.ts#L627)):*

this is good one to write up as a test.
</blockquote></details>

Done

Thanks for catching that :)
___
*[tfjs-backend-webgl/src/tex_util.ts, line 81 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVt5Tp2uJ8cQzjtZKS:-MIdtX94505YEJW01Kih:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/tex_util.ts#L81)):*

individual tensorInfos
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernel_utils/int.ts, line 25 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVtQpx295wY8IFB5Xa:-MIdtd6nAJ0Bc1OsOXcQ:biskj8j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/int.ts#L25)):*

+1. But i think is is a util already.
</blockquote></details>

Yes, this is a util.
___
*[tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts, line 66 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVvokD-UQl0UDbw5dR:-MIdtiVI3JgSaxxnd_kj:b-cm09tz) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts#L66)):*

Curious why do we define inputs as named parameters here? I thought most code (except kernels and layers) in our codebase define inputs as positional parameters, like binaryKernelFunc(opSnippet, packedOpSnippet, checkOutOfBounds = false, ...).
</blockquote></details>

I think in a few places we have switched from positional to named parameters when the list becomes too long (e.g. fused kernels) - but we haven't been consistent :) I find this a little bit more readable.
___
*[tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts, line 73 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVtwn2-dtZhWXXaS6g:-MIduTDW-G6dNQAXb70Q:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts#L73)):*

Can we define the type maybe type BinaryKernelFuncInput?
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Cast.ts, line 63 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIW-TQvBwY4Unnu8IID:-MIdukjG4ZH9wh2rr8NP:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Cast.ts#L63)):*

+1. It was mainly to keep the logic close to the other block for handling complex tensors as when casting from a complex tensor to a non complex tensor you will always have encoding loss (in this case we throw out the imaginary part). 
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex.ts, line 22 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIX8UzT4w0dF5-OiP6F:-MIdyuju6X4nLWMPR48N:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex.ts#L22)):*

given the (ahem) complexity of implementing complex would this be a good place for a high level description of how complex is implemented in webgl? It could shed some light on memory/texture concerns that suggest the importance of sharing memory as well as the general expected lifecycle of complex tensors (and their sub parts) and how they relate to stuff in the engine like disposal. I think it would be good to have that written out (somewhere) while its still in our heads.
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVxVdiCr8tteQSdnSp:-MIdzyDlB7Q7uoZ-BIR6:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex.ts#L34)):*

+1 (and same below)
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 35 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIX9Yz87hI4h4LpWaIY:-MIsacWe7kl1ec55oKKZ:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L35)):*

the rest of the comment still describes 5. also add info about expected dataids
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 42 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXA0q17dnUm3NE4p8h:-MIsipTq7BKxAmlWuORt:b-e3j8ak) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L42)):*

could you remind me of the relation between tensordata, tensorinfo and dataId?

Would you also want to assert here that you are still have the same number of bytes in GPU memory (or same number of textures)?  (i'm guessing it should be the same as before the complex was created or am i mistaken?).
</blockquote></details>

Done

TensorInfo is an object with tensor metadata (shape, dtype) - we use TensorInfo's to represent tensors in kernel backends. TensorData on the other hand is an object containing the data behind the TensorInfo, as well as information about that data, e.g. in WebGL properties of the physical GPU texture holding that data. dataId is a property of the TensorInfo that identifies its corresponding TensorData.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 51 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXAhF4-snJwSnU2qtP:-MIsjYx63rqe8LMKl-V-:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L51)):*

add tensordata info to stay consistent 
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXAyrXE2OP-C45vQd_:-MIsnvtf4F8OVQXfDhAL:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L69)):*

do you want to add an assertion about underlying memory as well?
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 113 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXD2GJ4tGETd_Y1F1M:-MIstK0q5eSrO0XTdfZy:b-cq5544) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L113)):*

adding a similar test but where the real and imaginary parts are created from the complex tensor and then disposed would also be helpful.

(a parallel version could be added to core in complex_ops that disposes the real and imag and check that isDisposed is true on the tensor, and that tf.memory has had the appropriate reduction). 
</blockquote></details>

Done

I feel like we've been adding complex tensor memory related tests in the backends rather than in core, so I've added this to cpu but not to core.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 186 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXBaiEDkXW_D89IA1e:-MIsuo_sAAJuqaEAHQyi:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L186)):*

should this be tensorData? (or whatever we pick)
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 194 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXBgQLBAX0_Q-4KF05:-MIswXAv1kojaCYepN8x:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L194)):*

Af first glance this doesn't seem to match the comment above. I think its because here you don't use the style you do above or re assigning the counts after each assertion so that you can assert the specific increase. Could you use that style here?
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Imag.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIW0RAb-H2z2GLzzdPG:-MIe0Cwd82RoCJ4oWpFH:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Imag.ts#L28)):*

+1
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Real.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXE7nL8vvNRSb683VW:-MIe0DqG3lq9y5YrTamz:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Real.ts#L28)):*

use identity
</blockquote></details>

Done

tafsiri (changes requested):
This is really coming along nicely. Left a few comments as I did have some trouble following some of the tests. It also looks like there are still some failures on CI. 

> Reviewed 14 of 15 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956#-:-MIyfjJ_AcD1hRQw-bnP:bzbpw8o)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 143 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyURQrBlW6epq30v07:-MIyURQrBlW6epq30v08:b-sgqy6c) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-cpu/src/kernels/Complex_test.ts#L143)):*
> ```XML
>     // 1 fewer tensor, 1 fewer data buckets.
>     expect(tf.memory().numTensors).toBe(numTensors - 1);
>     expect(tf.engine().backend.numDataIds()).toBe(numDataIds - 3);
> ```

this doesn't match the comment above? should be 3 fewer data buckets?
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 620 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIX1Ixg2fwVFrppYMhx:-MIyUy9UEj3nluq0qS5U:b7v9s8g) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/backend_webgl.ts#L620)):*

Yes, if you keep calling dispose on a tensor that is `kept`. What do you think would be the best way to handle this case?
</blockquote></details>

hmm actually it looks like it may be handled at the tensor level. the dispose method on a tensor will set  isDisposedInternal to true and make future disposes a no-op. So maybe its not an issue.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 629 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyXpLT7bXfM0iCVvUf:-MIyXpLU0lyW_9Pbrx3K:b-kl4n8p) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/backend_webgl.ts#L629)):*
> ```XML
>     }
> 
>     if (this.texData.get(dataId).keptRefCount > 0) {
> ```

I'm trying to understand what this block is supposed to do overall. Mind adding a comment as to where we are in the overall logic? My current understanding is that this is is true only for real and imag parts of tensors. Is it correct to say we only do this here because their refcount wont be updated in some kernel? I'm also guessing this condition is false when the complex part has been disposed first?
___
*[tfjs-backend-webgl/src/tex_util.ts, line 90 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyWD2d1CtovP79iZUa:-MIyWD2eC7qAYuPQSSU2:b-wiv1vq) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/tex_util.ts#L90)):*
> ```XML
> 
>   refCount: number;
>   // Number of references to prevent disposal of the texture data.
> ```

I think we need more docs here to make clear the difference in intent between refCount and keptRefCount. As is this comment is true for this line and the one above.

Is it fair to say these only exist on tensorInfos that will be in complexTensorInfos.real and complexTensorInfos.imag for a complexTensorInfo? If so could we make that more explicit here, maybe some kind of object like complexTensorPartRefCount?: { keptRefCount: number}.

coming back to this after reading a bit more code, is it fair to call this complexParentRefCount as in this counts the number of complex tensors pointing to a real/imag part?
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 186 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXBaiEDkXW_D89IA1e:-MIyf1Ug56uFy8o5wrhT:b-f80rs7) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L186)):*

Done
</blockquote></details>

don't think the change propagated
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 43 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyb6Sh0m6rhRpiA6Id:-MIyb6Sh0m6rhRpiA6Ie:b-z4c218) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L43)):*
> ```XML
>     // 2 new Tensors: real1, imag1, and two data buckets created.
>     expect(tf.memory().numTensors).toBe(numTensors + 2);
>     expect(tf.engine().backend.numDataIds()).toBe(numDataIds + 2);
> ```

could you adda numbytes assertion here like the one you have below.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 51 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIybj5r8TkEoAOBG4ON:-MIybj5r8TkEoAOBG4OO:b-co1p07) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L51)):*
> ```XML
>     // 1 new complex Tensor and 1 new data bucket created.
>     expect(tf.memory().numTensors).toBe(numTensors + 1);
>     expect(tf.memory().numBytes).toBe(numBytes + BYTES_PER_COMPLEX_ELEMENT);
> ```

i think this threw me off becuase numbytes isn't reset after real1 and imag1 is created. With the assertion requested, it would nice if this assertion asserted there are no new bytes. Unless i am misunderstanding how the memory allocation works.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 71 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyca4aD2_W3pQVcTzU:-MIyca4b0f1DJqbeXQMJ:b-5hkd76) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L71)):*
> ```XML
>     const complex2 = tf.complex(real2, imag2);
> 
>     // 1 new Tensor and 1 new TensorData.
> ```

nit: i think we should stick to one term for tensorinfo/tensordata/data bucket/dataId
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 73 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIycnVq56bBmNvkFcBX:-MIycnVq56bBmNvkFcBY:b-1zf3zs) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L73)):*
> ```XML
>     // 1 new Tensor and 1 new TensorData.
>     expect(tf.memory().numTensors).toBe(numTensors + 1);
>     // numBytes stays the same because it is determined by tensor creation at
> ```

shouldn't this say it is **not** determined by tensor creation at the engine level.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 93 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyd7qn8EpVYDH0hYhG:-MIyd7qo7j0urhJYxUHp:b-vrcp9b) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L93)):*
> ```XML
>     // Two new 1x1 textures are created to compute the sum of real / imag
>     // components, respectively. No new textures are allocated for the inputs
>     // because they are beneath the uniform upload threshold.
> ```

will this change when we set the threshold to zero via flags in test runs?
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 107 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIydMTX1Ah1HjEmf9kW:-MIydMTX1Ah1HjEmf9kX:b-18uf1u) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L107)):*
> ```XML
>     // A new tensor is created. No new tensorData is created.
>     expect(tf.memory().numTensors).toBe(numTensors + 1);
>     expect(tf.memory().numBytes).toBe(numBytes + 4);
> ```

why does numbytes increase here?  if numDataIds doesn't and no new tensorData's are made?
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 119 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIydlXyCTL5RZarcilH:-MIydlXyCTL5RZarcilI:b4h9dio) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L119)):*
> ```XML
>     // A new tensor is created. No new tensorData is created.
>     expect(tf.memory().numTensors).toBe(numTensors + 1);
>     expect(tf.memory().numBytes).toBe(numBytes + 4);
> ```

same question as above. i may not have a good understanding of the relationship between numDataIds and numBytes so feel free to set up some GVC time if there is something i'm just fundamentally missing.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 171 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIye9Uq4LooFqqsZJpP:-MIye9Uq4LooFqqsZJpQ:bse687e) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L171)):*
> ```XML
>     expect(tf.memory().numTensors).toBe(numTensors - 1);
>     expect(tf.engine().backend.numDataIds()).toBe(numDataIds - 1);
> ```

could you assert here that the real and imag parts are not disposed and have the values you expect.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 249 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyedY69ztRxMFHagKq:-MIyedY69ztRxMFHagKr:b2fal3s) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L249)):*
> ```XML
>     expect(tf.memory().numTensors).toBe(memoryBefore.numTensors);
>     // The underlying memory should now be released.
>     expect(tf.engine().backend.numDataIds()).toBe(numDataIdsBefore);
> ```

Generaly question. is numDataIds a good proxy for memory used in webgl backend? Is this different for complex and non complex tensors. I'm basically wondering if there is value in asserting stuff about numBytesInGPU instead.

annxingyuan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/Complex_test.ts, line 143 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyURQrBlW6epq30v07:-MIzTiJ3DeVuzAPxpGzU:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-cpu/src/kernels/Complex_test.ts#L143)):*

this doesn't match the comment above? should be 3 fewer data buckets?
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 629 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyXpLT7bXfM0iCVvUf:-MIzV7_W9hGAeYTQOOJo:b-f61dl) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/backend_webgl.ts#L629)):*

I'm trying to understand what this block is supposed to do overall. Mind adding a comment as to where we are in the overall logic? My current understanding is that this is is true only for real and imag parts of tensors. Is it correct to say we only do this here because their refcount wont be updated in some kernel? I'm also guessing this condition is false when the complex part has been disposed first?
</blockquote></details>

Done

Yes it happens that only complex tensors are making use of the keptRefCount feature, so this will only be triggered in case we are trying to dispose the real / imag components of a complex tensor. This condition is false if all the complex tensors that use a certain tensor as one of their components have been disposed (because as part of disposing a complex tensor, we decrement the `keptRefCount`s of their underlying components).
___
*[tfjs-backend-webgl/src/tex_util.ts, line 90 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyWD2d1CtovP79iZUa:-MIzb74J2Cvz7Gsv84ju:bxk6pua) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/tex_util.ts#L90)):*

I think we need more docs here to make clear the difference in intent between refCount and keptRefCount. As is this comment is true for this line and the one above.

Is it fair to say these only exist on tensorInfos that will be in complexTensorInfos.real and complexTensorInfos.imag for a complexTensorInfo? If so could we make that more explicit here, maybe some kind of object like complexTensorPartRefCount?: { keptRefCount: number}.

coming back to this after reading a bit more code, is it fair to call this complexParentRefCount as in this counts the number of complex tensors pointing to a real/imag part?
</blockquote></details>

Done

I added some comments that explain how this is different from refCount. I think I would prefer to keep the name general as I can imagine other ops needing this type of logic in the future - does that seem fair? It's meant to mimic the `keep` logic at the engine level, which is general.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 186 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIXBaiEDkXW_D89IA1e:-MIzrs4uDssoTeKNqq5e:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernels/Complex_test.ts#L186)):*

don't think the change propagated
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 43 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyb6Sh0m6rhRpiA6Id:-MIzi-cbCmpusv8zfc5d:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L43)):*

could you adda numbytes assertion here like the one you have below.
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 51 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIybj5r8TkEoAOBG4ON:-MIzi1nJB-s9K3NsjD2M:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L51)):*

i think this threw me off becuase numbytes isn't reset after real1 and imag1 is created. With the assertion requested, it would nice if this assertion asserted there are no new bytes. Unless i am misunderstanding how the memory allocation works.
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 71 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyca4aD2_W3pQVcTzU:-MIziBD7Epba38ZW18e4:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L71)):*

nit: i think we should stick to one term for tensorinfo/tensordata/data bucket/dataId
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 73 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIycnVq56bBmNvkFcBX:-MIzjG_y69Z1bXe5uE7Y:bshjciu) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L73)):*

shouldn't this say it is **not** determined by tensor creation at the engine level.
</blockquote></details>

It is determined by tensor creation at the engine level, but we do not increment memory for complex tensors: https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/engine.ts#L798
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 93 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyd7qn8EpVYDH0hYhG:-MIzkr_j0LwvSdjQ4L2m:bbt6wgy) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L93)):*

will this change when we set the threshold to zero via flags in test runs?
</blockquote></details>

Done

nice catch!
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 107 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIydMTX1Ah1HjEmf9kW:-MIzmi5_1LhuV11i7-b2:b-pf5zad) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L107)):*

why does numbytes increase here?  if numDataIds doesn't and no new tensorData's are made?
</blockquote></details>

numBytes is tracked at the engine level. when we call `tf.real(result)`, it creates a new tensor out of the real component of `result`, incrementing numBytes. Before this point the engine had no awareness of the real component of result, because it is created entirely within the webgl backend as a `TensorInfo`.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 119 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIydlXyCTL5RZarcilH:-MIznAIO4ZxPuNgFT1OL:bw5vrb1) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L119)):*

same question as above. i may not have a good understanding of the relationship between numDataIds and numBytes so feel free to set up some GVC time if there is something i'm just fundamentally missing.
</blockquote></details>

Let me know if my explanation above is still confusing!
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 171 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIye9Uq4LooFqqsZJpP:-MIznlc55NitvAORa8h6:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L171)):*

could you assert here that the real and imag parts are not disposed and have the values you expect.
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 249 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyedY69ztRxMFHagKq:-MIzoLQJEij69_Uk7DJj:b-z1r7sl) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L249)):*

Generaly question. is numDataIds a good proxy for memory used in webgl backend? Is this different for complex and non complex tensors. I'm basically wondering if there is value in asserting stuff about numBytesInGPU instead.
</blockquote></details>

numDataIds is a good proxy for webgl memory usage. I'm less keen on asserting stuff about numBytesInGPU because it's a little finicky - depends on a bunch of flag settings and really gets into webgl internals.

lina128 (approved):
This is great, thank you Ann! I also want to discuss whether cpu should introduce keptRefCount in next meeting.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956#-:-MJ3-dCUExQwqw-mCSot:bikg0ie)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[e2e/benchmarks/model_config.js, line 223 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVq0Rb96rrqEUhuouc:-MJ2qjP4BADpqJEA0Cx9:bcr5s2o) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/e2e/benchmarks/model_config.js#L223)):*

I added it in order to benchmark FFT given my changes to complex. I think it would be nice to be able to easily benchmark speech-commands in the future as well so I added it here.
</blockquote></details>

Got it, thanks!
___
*[tfjs-backend-webgl/src/kernel_utils/int.ts, line 25 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVtQpx295wY8IFB5Xa:-MJ2tLyQ5eQlpXE3jL5p:b-3yi8ld) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/int.ts#L25)):*

Yes, this is a util.
</blockquote></details>

Ah, you are right.
___
*[tfjs-backend-webgl/src/kernel_utils/shared.ts, line 23 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIVvW6E0NE55iRIf7CD:-MJ2tyrs5OmgjHKu3HlR:bugw5i5) ([raw file](https://github.com/tensorflow/tfjs/blob/2f1ead285d8afc5b4ffc8177b689bf962a8b625f/tfjs-backend-webgl/src/kernel_utils/shared.ts#L23)):*

I think its because in the kernel_funcs_utils.ts file above, in order to templatize cpu forwarding of binary kernels this type is imported. It does produce a somewhat tight coupling between the binary kernels in cpu to those in webgl. But I think that is the reason (Ann feel free to clarify). Na is this a reasonable constraint to put on binary kernels in CPU?

Or were you more specifically asking about importing from dist as a path? This is to not trigger auto registration of CPU kernels.
</blockquote></details>

Thank you for the detailed explanation, Yannick! I was more specifically asking about importing from dist.

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-webgl/src/kernels/Concat_impl.ts, line 49 at r3](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MJ42TUN1-xS0be1HZ-P:-MJ42TUN1-xS0be1HZ-Q:b-rp4wly) ([raw file](https://github.com/tensorflow/tfjs/blob/2966e905fe10bfca65818c9aeb8a2ed1b275c97e/tfjs-backend-webgl/src/kernels/Concat_impl.ts#L49)):*
> ```XML
>     return result;
>   }
> ```

Looks like in the non-modularized version, [cpu forwarding](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/backend_webgl.ts#L806) was used in concat, but it is not here. Just curious about this. Thanks!

tafsiri (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-webgl/src/tex_util.ts, line 90 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyWD2d1CtovP79iZUa:-MJ8tcjN-R5QZG9uacDq:b-ejz78v) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/tex_util.ts#L90)):*

Done

I added some comments that explain how this is different from refCount. I think I would prefer to keep the name general as I can imagine other ops needing this type of logic in the future - does that seem fair? It's meant to mimic the `keep` logic at the engine level, which is general.
</blockquote></details>

Just going to reply to this before reviewing the rest. Are there any situations you can think of where you would use this rather than regular refCount other than a 'container' type relationship like the one that exists between complex and its components? 

I don't get the feeling that this is all that general, but maybe there is a use case I am missing? I overall think you want to keep this as hidden as possible when not dealing with complex tensor components (I had briefly considered suggesting you subclass TensorInfo for real and imag but held off from that). I just worry that this code will be difficult to reason about in a few months and that it might easily be mistaken by someone modularizing a kernel as the appropriate ref count to increment rather than the one above.

What do you think?

tafsiri (approved):
> Reviewed 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

*[a discussion](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MJAKBgP6tsSTie8fBX7:-MJAKBgP6tsSTie8fBX8:b503og1) (no related file):*
Going to LGTM this since we are pretty much down to naming. I still don't think the notion of a 'kept' TensorInfo in the backend is all that common. IMO we want to 'keep' tensors but incref/decref tensorInfos. Complex/real/imag seem be be a special case of linking because complex cannot exist without a real/imag part. Other than optimizers & gradients/tape, I think complex may have been the only place we were using keep internally.

Final call is yours and since others don't mind I'm not going to block on this. 

Great work!

---
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 107 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIydMTX1Ah1HjEmf9kW:-MJAG9WK0lYiq6cMkbkc:bjk8iq1) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L107)):*

numBytes is tracked at the engine level. when we call `tf.real(result)`, it creates a new tensor out of the real component of `result`, incrementing numBytes. Before this point the engine had no awareness of the real component of result, because it is created entirely within the webgl backend as a `TensorInfo`.
</blockquote></details>

ah kk. we should probably revisit the tracking of numBytes in future (it seems more accurately knowable by the backend) and in current form makes reading the test quite confusing.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 119 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIydlXyCTL5RZarcilH:-MJAGVEQERbB1_srnOzs:b-ossir) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L119)):*

Let me know if my explanation above is still confusing!
</blockquote></details>

The explanation above did help a bit (though do take a look at my interpretation) but I think I found the ambiguity in 'data bucket' confusing.

When I read

`// A new tensor is created. No new data buckets created.`

I expected numBytes to not increase as there are no new 'data buckets'.  But

`expect(tf.memory().numBytes).toBe(numBytes + 4);`

suggests that there are 4 bytes of data that are new (and presumably they would live in a bucket?).  It is hard to tell from reading the test what I should expect to be asserted. Given your explanation above about numBytes only representing bytes visible to the engine it makes sense.

annxingyuan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3956)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-webgl/src/tex_util.ts, line 90 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIyWD2d1CtovP79iZUa:-MJBjR1uCXo0FlwG_hc8:b-r5nlfz) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/tex_util.ts#L90)):*

Just going to reply to this before reviewing the rest. Are there any situations you can think of where you would use this rather than regular refCount other than a 'container' type relationship like the one that exists between complex and its components? 

I don't get the feeling that this is all that general, but maybe there is a use case I am missing? I overall think you want to keep this as hidden as possible when not dealing with complex tensor components (I had briefly considered suggesting you subclass TensorInfo for real and imag but held off from that). I just worry that this code will be difficult to reason about in a few months and that it might easily be mistaken by someone modularizing a kernel as the appropriate ref count to increment rather than the one above.

What do you think?
</blockquote></details>

Done

That makes sense - renamed to complexParentRefCount.
___
*[tfjs-backend-webgl/src/kernels/Complex_test.ts, line 107 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MIydMTX1Ah1HjEmf9kW:-MJBkDxIAldeICqrJ5P6:b1mhigi) ([raw file](https://github.com/tensorflow/tfjs/blob/c9d870ad9fb83a887230cdd377708a3f982cc66a/tfjs-backend-webgl/src/kernels/Complex_test.ts#L107)):*

ah kk. we should probably revisit the tracking of numBytes in future (it seems more accurately knowable by the backend) and in current form makes reading the test quite confusing.
</blockquote></details>

agreed :)
___
*[tfjs-backend-webgl/src/kernels/Concat_impl.ts, line 49 at r3](https://reviewable.io/reviews/tensorflow/tfjs/3956#-MJ42TUN1-xS0be1HZ-P:-MJBk_gTDHiDxvQjTvoJ:bz7orod) ([raw file](https://github.com/tensorflow/tfjs/blob/2966e905fe10bfca65818c9aeb8a2ed1b275c97e/tfjs-backend-webgl/src/kernels/Concat_impl.ts#L49)):*

Looks like in the non-modularized version, [cpu forwarding](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/backend_webgl.ts#L806) was used in concat, but it is not here. Just curious about this. Thanks! 
</blockquote></details>

The purpose was just to reduce the scope of this PR :)

PR #3958, opened 2020-09-22
[converter]Support table initializer in converter.

lina128:
**Background:**
For saved model graphs that have hashtable or lookuptable operations, the underlying hashtable needs to be initialized before being used. From graph architecture, the initialization subgraph is detached from the inference subgraph, so that initialization doesn't happen in every inference. To support hashtable ops in TF.js, we made below changes:
  1. Extract the initialization subgraph and store its topology in `modelInitializer` field in model.json. See this PR.
  2. Load the initialization subgraph and execute it during model load. See https://github.com/tensorflow/tfjs/pull/3975
  3. Add hashtable ops implementation and add hashtableMap in model. Only dispose the hashtables when model is disposed. (To resume this PR: https://github.com/tensorflow/tfjs/pull/3576)

**Changes in this PR:**
1. Get `table_initializer` from `meta_graph` `collection_def['table_initializer']`.
2. Use the nodes in `table_initializer` as outputs to extract a initializer subgraph and freeze the graph using `convert_variables_to_constants.py` and grappler.
3. Extract the Const ops' values in the initialization graph to weights file.
4. Created a `modelInitializer` field in model.json. Store the initialization graph's topology in the `topology` subfield. Store the outputs in the `outputs` subfield.

**Notes: **
  1. This implementation only handles table initializer. We intentionally don't generalize to other initializers until we know the use case.
  2. This solution only solves saved model in 1.x style. For 2.x style, we need more investigation how to support table initializer.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3958)


pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3958)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[tfjs-converter/python/tensorflowjs/converters/tf_saved_model_conversion_v2.py, line 374 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3958#-MIvKQ3b6IRbj_L_4AeW:-MIvKQ3b6IRbj_L_4AeX:bft23z5) ([raw file](https://github.com/tensorflow/tfjs/blob/1c08bf494a81130d33663d5cb8d92cb64108dfaf/tfjs-converter/python/tensorflowjs/converters/tf_saved_model_conversion_v2.py#L374)):*
> ```Python
>     A freezed and optimized graph.
>     Nullable. A freezed and optimized initializer graph.
>     Nullable. A list of output node names of initializer.
> ```

no output node name as part of the returns, remove?

PR #3960, opened 2020-09-22
Modularize BatchNorm in cpu and webgl backend

jinjingforever:
For cpu backend, looks like I don't really need to do all the reshaping since the algorithm can operate on the underlying data directly. I only need to calculate (with conversion) the output shape. Please correct me if I am wrong.

For webgl backend, I used the Reshape kernelfunc.

Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3960)

### DIFF
--- a/tfjs-backend-cpu/src/backend_cpu.ts
+++ b/tfjs-backend-cpu/src/backend_cpu.ts
@@ -2306,50 +2306,6 @@ export class MathBackendCPU extends KernelBackend {
     return tf.tensor4d(output, x.shape, x.dtype);
   }
 
-  batchNorm(
-      x: Tensor4D, mean: Tensor4D|Tensor1D, variance: Tensor4D|Tensor1D,
-      offset?: Tensor4D|Tensor1D, scale?: Tensor4D|Tensor1D,
-      varianceEpsilon?: number): Tensor4D {
-    assertNotComplex([x, mean, variance, scale, offset], 'batchNorm');
-
-    const xVals = this.readSync(x.dataId) as TypedArray;
-    const mVals = this.readSync(mean.dataId) as TypedArray;
-    const varVals = this.readSync(variance.dataId) as TypedArray;
-    const sVals = scale ? this.readSync(scale.dataId) as TypedArray :
-                          new Float32Array([1]);
-    const offVals = offset ? this.readSync(offset.dataId) as TypedArray :
-                             new Float32Array([0]);
-    const outVals = new Float32Array(xVals.length);
-
-    const offValsLength = offVals.length;
-    const sValsLength = sVals.length;
-    const varValsLength = varVals.length;
-    const mValsLength = mVals.length;
-
-    let offi = 0;
-    let mi = 0;
-    let si = 0;
-    let vi = 0;
-    for (let i = 0; i < xVals.length; ++i) {
-      outVals[i] = offVals[offi++] +
-          (xVals[i] - mVals[mi++]) * sVals[si++] /
-              Math.sqrt(varVals[vi++] + varianceEpsilon);
-      if (offi >= offValsLength) {
-        offi = 0;
-      }
-      if (mi >= mValsLength) {
-        mi = 0;
-      }
-      if (si >= sValsLength) {
-        si = 0;
-      }
-      if (vi >= varValsLength) {
-        vi = 0;
-      }
-    }
-    return tf.tensor4d(outVals, x.shape);
-  }
-
   localResponseNormalization4D(
       x: Tensor4D, depthRadius: number, bias: number, alpha: number,
       beta: number): Tensor4D {
--- a/tfjs-backend-cpu/src/kernels/BatchNorm.ts
+++ b/tfjs-backend-cpu/src/kernels/BatchNorm.ts
@@ -26,12 +26,7 @@ export function batchNormKernelFunc(args: {
   attrs: FusedBatchNormAttrs
 }): TensorInfo {
   const {inputs, backend, attrs} = args;
-  let {varianceEpsilon} = attrs;
   const {x, scale, offset, mean, variance} = inputs;
-
-  if (varianceEpsilon == null) {
-    varianceEpsilon = 0.001;
-  }
 
   util.assert(
       mean.shape.length === variance.shape.length,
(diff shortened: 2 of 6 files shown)


lina128 (commented):
Hi Jing, for BatchNorm, I think the input will expect a 4D tensor, to align with the TF api: https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/fused-batch-norm In this case, we transform tensor to 4D in ops and pass in the transformed tensor. So in the kernel, we can assume x is 4D. Other than that, LGTM.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3960#-:-MHw3EMiC0hqglIufngJ:b-hb4z2b)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/BatchNorm.ts, line 87 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3960#-MHw3BQO7ZwrynQQtNCV:-MHw3BQO7ZwrynQQtNCW:b-lqhh0s) ([raw file](https://github.com/tensorflow/tfjs/blob/154233420c49e4f3ad34fa78074fb729a48f392e/tfjs-backend-cpu/src/kernels/BatchNorm.ts#L87)):*
> ```XML
>     }
>   }
>   return backend.makeTensorInfo(get4DShapeFromX(x), x.dtype, outVals);
> ```

I think this can just be x.shape.
___
*[tfjs-backend-webgl/src/kernels/BatchNorm.ts, line 31 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3960#-MHw3ust-3owvq7CaG2q:-MHw3ust-3owvq7CaG2r:b-s7hzns) ([raw file](https://github.com/tensorflow/tfjs/blob/154233420c49e4f3ad34fa78074fb729a48f392e/tfjs-backend-webgl/src/kernels/BatchNorm.ts#L31)):*
> ```XML
>   backend: MathBackendWebGL,
>   attrs: FusedBatchNormAttrs
> }) => TensorInfo | TensorInfo[] = ({inputs, backend, attrs}) => {
> ```

Just TensorInfo.
___
*[tfjs-backend-webgl/src/kernels/BatchNorm.ts, line 53 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3960#-MHw45VcAXjSKOgOcyFq:-MHw45VcAXjSKOgOcyFr:b-cyqdho) ([raw file](https://github.com/tensorflow/tfjs/blob/154233420c49e4f3ad34fa78074fb729a48f392e/tfjs-backend-webgl/src/kernels/BatchNorm.ts#L53)):*
> ```XML
> 
>   const $x: TensorInfo = xAs4D(x, backend);
>   const $mean = as1DOr4D(mean, backend);
> ```

These transformation will be done at op level, because the TF api specifies the kernel as 4D, so in the kernel, we can assume these have the right shape.

jinjingforever (commented):
Thanks! One question: I remember you mentioned that kernelfunc can also be called directly from other places (instead of going through op)? In those cases, would it be the caller's responsibility to transform the inputs?  

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3960#-:-MHw7pEj3Sm1Mnn98CCD:b9ihil3)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, and @tafsiri)

jinjingforever (commented):
Thanks Na! Removed reshape related code.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3960#-:-MHxFaP--DmbBNeTVEYW:b54b386)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/BatchNorm.ts, line 87 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3960#-MHw3BQO7ZwrynQQtNCV:-MHxBDkT1L9dk-imUqNf:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/154233420c49e4f3ad34fa78074fb729a48f392e/tfjs-backend-cpu/src/kernels/BatchNorm.ts#L87)):*

I think this can just be x.shape.
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/kernels/BatchNorm.ts, line 31 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3960#-MHw3ust-3owvq7CaG2q:-MHxBKab0FYp6TcA03Fk:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/154233420c49e4f3ad34fa78074fb729a48f392e/tfjs-backend-webgl/src/kernels/BatchNorm.ts#L31)):*

Just TensorInfo.
</blockquote></details>

Done
___
*[tfjs-backend-webgl/src/kernels/BatchNorm.ts, line 53 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3960#-MHw45VcAXjSKOgOcyFq:-MHxFRd9Ca79WTDR1owC:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/154233420c49e4f3ad34fa78074fb729a48f392e/tfjs-backend-webgl/src/kernels/BatchNorm.ts#L53)):*

These transformation will be done at op level, because the TF api specifies the kernel as 4D, so in the kernel, we can assume these have the right shape.
</blockquote></details>

Done

lina128 (commented):
Hi Jing, so in general, op should only contain validation, any transformation should happen in kernels. With the exception that if the TF api specifies a tensor to be certain shape, and if our op has a more relaxed requirement, then we would transform the tensor into the specified shape and pass that to the kernel. So when in doubt, check the TF api. There is the risk that other kernels may call this kernel and is not aware of this specification because it is not guarded by code in any way but rather just in documentation. We'll see whether we encounter this case in actual use. If so, we can change the type definition in kernel_names to be more specifically. For example, right now, it is just extend NamedTensorInfo. We can change it to be less generic, like 4DTensorInfo. I also shared two docs with you today, which has documented some of these decisions during the modularization.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3960#-:-MHxHIi8C8HFhhLZTG9H:b-p34ve0)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

lina128 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3960#-:-MHxIo4N1XCIFpcvd0ba:b-jbsf8z)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

PR #3965, opened 2020-09-24
update the tf deps and moved PyInquirer to extra_requires

pyu10055:
When integrating with TFX OSS, some pip dependencies problem was found, namely:
1. we are depending on tensorflow-cpu, which conflict with TFX dependencies of Tensorflow
2. we are using PyInquirer which uses an old prompt-toolkit that conflict with one of TFX dependencies

Since TFX does not need the wizard and tfjs converter should work with TensorFlow pip as well, we decide to make following changes:

1. Update the tensorflow-cpu dependencies to tensorflow, since this is mainly a size consideration. 
2. use the extra_require as following:
```
    extras_require = {
      "wizard": ["PyInquirer==1.0.3"]
    }
```
when users need to install tensorflowjs pip, they need to do 
```bash
pip install tensorflowjs[wizard]
```

Update the READMEs, and will print error message if the PyInquirer is not installed when users run tensorflow_wizard command.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3965)

### DIFF
--- a/tfjs-converter/README.md
+++ b/tfjs-converter/README.md
@@ -48,8 +48,9 @@ virtualenv --no-site-packages venv
 
 __1. Install the TensorFlow.js pip package:__
 
+Install the library with interactive CLI:
 ```bash
- pip install tensorflowjs
+ pip install tensorflowjs[wizard]
 ```
 
 __2. Run the converter script provided by the pip package:__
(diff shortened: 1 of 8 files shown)


lina128 (approved):
LGTM!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3965#-:-MHyWgKAAFLNWKH7NLAf:b-85i648)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

PR #3969, opened 2020-09-24
Log error caught when creating WebGL context

jameshfisher:
My day today: I spent a _long_ time debugging the error `Initialization of backend webgl failed` on a particular machine. Ultimately, the cause was really obscure: Chrome was returning `null` from `canvas.getContext` when passed [this `failIfMajorPerformanceCaveat` flag](https://github.com/tensorflow/tfjs/blob/c823c1f944fcf97c8070fd75036da8f4bf14c660/tfjs-backend-webgl/src/canvas_util.ts#L27). Another line in TensorFlow assumes that `getContext` cannot return null, and so throws an error `TypeError: Cannot read property 'isContextLost' of null`. This is another bug. But that error was then silently swallowed at the line here.

On restarting the machine, Chrome mysteriously stopped behaving this way. But that underlying issue is a story for another time.

The point is, silently swallowing all thrown errors is a pretty bad practice. To go further, unless we actually _expect_ errors thrown here, I think this `try`/`catch` should be removed - if it fails unexpectedly, just let it fail. If we do expect errors thrown here, that should at least be documented in a comment here.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3969)

### DIFF
--- a/tfjs-backend-webgl/src/webgl_util.ts
+++ b/tfjs-backend-webgl/src/webgl_util.ts
@@ -534,6 +534,7 @@ export function isWebGLVersionEnabled(webGLVersion: 1|2) {
       return true;
     }
   } catch (e) {
+    console.log("Error when getting WebGL context: ", e);
     return false;
   }
   return false;
--- a/tfjs-backend-webgl/src/webgl_util.ts
+++ b/tfjs-backend-webgl/src/webgl_util.ts
@@ -534,7 +534,7 @@ export function isWebGLVersionEnabled(webGLVersion: 1|2) {
       return true;
     }
   } catch (e) {
-    console.log("Error when getting WebGL context: ", e);
+    console.log('Error when getting WebGL context: ', e);
     return false;
   }
   return false;


annxingyuan (approved):

jameshfisher:
(I'll follow up with a separate PR fixing the `TypeError: Cannot read property 'isContextLost' of null` error that I found was being silently swallowed)

jameshfisher:
@annxingyuan thanks for the +1 - I see the PR failed [at this URL](https://console.cloud.google.com/cloud-build/builds/ff99cf1c-9656-4d19-8d24-2c90c28ed0dc?project=834911136599), however that page just says "You do not have sufficient permissions to view this page" :-(

annxingyuan:
Hi @jameshfisher - in order to see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

Looks like you figured out the issue :)

jameshfisher:
Thanks - I found that text in my second PR description; I think I mindlessly deleted it the first time!

PR #3973, opened 2020-09-24
Modularize MaxPool, MaxPoolBackprop, AvgPool, AvgPoolBackprop cpu kernels. 

jinjingforever:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3973)


tafsiri (changes requested):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3973)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 57 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MI5HAsyEtJHD3KbrNy7:-MI5HAsyEtJHD3KbrNy8:bh9xt9t) ([raw file](https://github.com/tensorflow/tfjs/blob/8ff632b09172ece93421fbba550291ff3b2411e2/tfjs-backend-cpu/src/kernels/AvgPool.ts#L57)):*
> ```XML
>   if (convInfo.filterWidth === 1 && convInfo.filterHeight === 1 &&
>       util.arraysEqual(convInfo.inShape, convInfo.outShape)) {
>     res = {...x};
> ```

why this over res = x?
___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 58 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MI5HG0w8XdEsA9R4Uxn:-MI5HG0w8XdEsA9R4Uxo:bsripbm) ([raw file](https://github.com/tensorflow/tfjs/blob/8ff632b09172ece93421fbba550291ff3b2411e2/tfjs-backend-cpu/src/kernels/AvgPool.ts#L58)):*
> ```XML
>       util.arraysEqual(convInfo.inShape, convInfo.outShape)) {
>     res = {...x};
>     backend.incRef(x.dataId);
> ```

should this (both these lines) go through identity instead? I think keeping the incRef calls in fewer places might be better to make it easier to change memory management strategies. 
___
*[tfjs-backend-cpu/src/kernels/MaxPool.ts, line 56 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MI5ISDRB59r4_xHb0Qv:-MI5ISDRB59r4_xHb0Qw:bp5naog) ([raw file](https://github.com/tensorflow/tfjs/blob/8ff632b09172ece93421fbba550291ff3b2411e2/tfjs-backend-cpu/src/kernels/MaxPool.ts#L56)):*
> ```XML
>   if (convInfo.filterWidth === 1 && convInfo.filterHeight === 1 &&
>       util.arraysEqual(convInfo.inShape, convInfo.outShape)) {
>     res = {...x};
> ```

same comment as above on using identity here

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3973)* status: 0 of 1 approvals obtained (waiting on @lina128 and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 57 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MI5HAsyEtJHD3KbrNy7:-MI5TZSH9kYCl0gcTids:b-5j9s4z) ([raw file](https://github.com/tensorflow/tfjs/blob/8ff632b09172ece93421fbba550291ff3b2411e2/tfjs-backend-cpu/src/kernels/AvgPool.ts#L57)):*

why this over res = x?
</blockquote></details>

Replaced with identity
___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 58 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MI5HG0w8XdEsA9R4Uxn:-MI5TUVKELlFjQ_Fn6D6:b-simsv7) ([raw file](https://github.com/tensorflow/tfjs/blob/8ff632b09172ece93421fbba550291ff3b2411e2/tfjs-backend-cpu/src/kernels/AvgPool.ts#L58)):*

should this (both these lines) go through identity instead? I think keeping the incRef calls in fewer places might be better to make it easier to change memory management strategies. 
</blockquote></details>

Ah, good point. I forgot about identity. Updated.
___
*[tfjs-backend-cpu/src/kernels/MaxPool.ts, line 56 at r1](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MI5ISDRB59r4_xHb0Qv:-MI5TbX31P1rcpV_3BI8:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/8ff632b09172ece93421fbba550291ff3b2411e2/tfjs-backend-cpu/src/kernels/MaxPool.ts#L56)):*

same comment as above on using identity here
</blockquote></details>

Done

lina128 (changes requested):
Hi Jing, pool and conv operations are some of the most complicated, thank you for starting this! I left some comment below.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3973#-:-MICBEyy6SC11xVhcoiI:bdh3oog)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 40 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC3n6s5IKIt3jc7o0d:-MIC3n6s5IKIt3jc7o0e:bnbcwbs) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPool.ts#L40)):*
> ```XML
> 
>   const xRank = x.shape.length;
>   util.assert(
> ```

This assertion is at op level, I think we agreed to not repeat it at kernel level because x is not transformed. Is that right, @tafsiri ?
___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 45 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC48Va0tWHx6dj94Cs:-MIC48Va0tWHx6dj94Ct:b5nk2vy) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPool.ts#L45)):*
> ```XML
>           x.shape.length}}.`);
>   if (dimRoundingMode != null) {
>     util.assert(
> ```

This assertion is also at op level, and no transformation of this attribute, so no need to assert at kernel level. Otherwise these assertions always run twice if call starts from op level, which has cost. We had the discussion of whether it is safe to do so because call may start from other kernels, which may not have any checking. I think the consensus was, if there's no transformation in this kernel, we don't do the assertion, and kernels basically trust other kernels to pass in the right values. You may ask: Why we need to do the assertion at op level? Because op is public api, the data is user-provided, so we need to validate. Why we sometimes do assertion at kernel level if we trust kernel implementation? Most of them are for convenience, actually we should use unit tests to test all the possible cases instead of relying on assertion and bug report. We can consider converting them into unit tests once we have bandwidth. Jing, sorry I know this is sort of confusing about where to do transformation and where to assert. Above is basically what we discovered and agreed upon along the way. Hope it makes sense. :) Adding @tafsiri to double confirm above summary.
___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 62 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC7iE86H9YneNDjdgm:-MIC7iE86H9YneNDjdgn:b2mi7bq) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPool.ts#L62)):*
> ```XML
>     const xValues = backend.data.get(x.dataId).values as TypedArray;
>     const strides = util.computeStrides(x.shape);
>     const buffer = pool(xValues, x.shape, x.dtype, strides, convInfo, 'avg');
> ```

This uses op buffer and returns a tensorBuffer. Can we create a new util, say poolV2 and avoid using buffer and only return TypedArray? Alternatively, you can just revise pool and modularize all the affected kernels in this PR. Either way is fine by me, but I probably will prefer to not creating poolV2, because these versions generally align with tf kernel names' name.
___
*[tfjs-backend-cpu/src/kernels/AvgPoolBackprop.ts, line 35 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC9d0Y9oeg6Nigkf1v:-MIC9d0Y9oeg6Nigkf1w:b-j9h2vb) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPoolBackprop.ts#L35)):*
> ```XML
>   const inputRank = input.shape.length;
>   const dyRank = dy.shape.length;
>   util.assert(
> ```

No need to check rank here, this is type checked in Avgpool_grad.ts. Also if avgPool only work with 4D, it's not likely the backward propagation can get anything not 4D. So we can save an assertion here.
___
*[tfjs-backend-cpu/src/kernels/AvgPoolBackprop.ts, line 58 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MICAF5F0eN-m8hlc_ma:-MICAF5G726H62pVWkpS:bzbcxm9) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPoolBackprop.ts#L58)):*
> ```XML
>   const padTop = effectiveFilterHeight - 1 - convInfo.padInfo.top;
>   const dx =
>       buffer<Rank.R4>(x.shape as [number, number, number, number], 'float32');
> ```

This uses op buffer again. You can use create a typedArray for dx.
___
*[tfjs-backend-cpu/src/kernels/AvgPoolBackprop.ts, line 63 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MICAVbV0-PCVacPqcKK:-MICAVbV0-PCVacPqcKL:b2nreh6) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPoolBackprop.ts#L63)):*
> ```XML
> 
>   const dyData = backend.data.get(dy.dataId).values as Float32Array;
>   const dyBuf = buffer<Rank.R4>(
> ```

Same as above.
___
*[tfjs-backend-cpu/src/kernels/MaxPool.ts, line 35 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MICAcAs5wOBv_ZEiul4:-MICAcAs5wOBv_ZEiul5:b-zg8y64) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/MaxPool.ts#L35)):*
> ```XML
> 
>   const xRank = x.shape.length;
>   util.assert(
> ```

Same as above. No need for assertion.
___
*[tfjs-backend-cpu/src/kernels/MaxPool.ts, line 61 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MICAlVB106Q9NK-DuFb:-MICAlVB106Q9NK-DuFc:b-6k55j8) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/MaxPool.ts#L61)):*
> ```XML
>     const xValues = backend.data.get(x.dataId).values as TypedArray;
>     const strides = util.computeStrides(x.shape);
>     const buffer = pool(xValues, x.shape, x.dtype, strides, convInfo, 'max');
> ```

Same as above about changing pool implementation.
___
*[tfjs-backend-cpu/src/kernels/MaxPoolBackprop.ts, line 36 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MICB03c15MXxS9mPYhp:-MICB03c15MXxS9mPYhq:baqjcd) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/MaxPoolBackprop.ts#L36)):*
> ```XML
>   const inputRank = input.shape.length;
>   const dyRank = dy.shape.length;
>   util.assert(
> ```

Same, no need to assert.
___
*[tfjs-backend-cpu/src/kernels/MaxPoolBackprop.ts, line 60 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MICB5jX9jDe6sGW9Dn2:-MICB5jX9jDe6sGW9Dn3:b-dd5guu) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/MaxPoolBackprop.ts#L60)):*
> ```XML
>       1 /* dilations */, pad, dimRoundingMode);
>   const xValues = backend.data.get(x.dataId).values as TypedArray;
>   const maxPosBuf = buffer(
> ```

Same as above, remove buffer.
___
*[tfjs-backend-cpu/src/kernels/MaxPoolBackprop.ts, line 72 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MICB8g86OQdtNZ4ko3H:-MICB8g86OQdtNZ4ko3I:b-dd5guu) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/MaxPoolBackprop.ts#L72)):*
> ```XML
>   const padTop = effectiveFilterHeight - 1 - convInfo.padInfo.top;
>   const dx =
>       buffer<Rank.R4>(x.shape as [number, number, number, number], 'float32');
> ```

Same as above, remove buffer.
___
*[tfjs-backend-cpu/src/kernels/MaxPoolBackprop.ts, line 75 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MICBAhR1D14DfdibG9r:-MICBAhR1D14DfdibG9s:b-dd5guu) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/MaxPoolBackprop.ts#L75)):*
> ```XML
> 
>   const dyData = backend.data.get(dy.dataId).values as Float32Array;
>   const dyBuf = buffer<Rank.R4>(
> ```

Same as above, remove buffer.

tafsiri (commented):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3973)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 40 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC3n6s5IKIt3jc7o0d:-MIK39gU-uEgNthmFeYu:b32v57) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPool.ts#L40)):*

This assertion is at op level, I think we agreed to not repeat it at kernel level because x is not transformed. Is that right, @tafsiri ?
</blockquote></details>

+1
___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 45 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC48Va0tWHx6dj94Cs:-MIK3D6v3gllcvk1tLSF:bb6dct9) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPool.ts#L45)):*

This assertion is also at op level, and no transformation of this attribute, so no need to assert at kernel level. Otherwise these assertions always run twice if call starts from op level, which has cost. We had the discussion of whether it is safe to do so because call may start from other kernels, which may not have any checking. I think the consensus was, if there's no transformation in this kernel, we don't do the assertion, and kernels basically trust other kernels to pass in the right values. You may ask: Why we need to do the assertion at op level? Because op is public api, the data is user-provided, so we need to validate. Why we sometimes do assertion at kernel level if we trust kernel implementation? Most of them are for convenience, actually we should use unit tests to test all the possible cases instead of relying on assertion and bug report. We can consider converting them into unit tests once we have bandwidth. Jing, sorry I know this is sort of confusing about where to do transformation and where to assert. Above is basically what we discovered and agreed upon along the way. Hope it makes sense. :) Adding @tafsiri to double confirm above summary.
</blockquote></details>

+1. 
___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 62 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC7iE86H9YneNDjdgm:-MIK3aiiCO-RWOy6-VSG:bhhsrgb) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPool.ts#L62)):*

This uses op buffer and returns a tensorBuffer. Can we create a new util, say poolV2 and avoid using buffer and only return TypedArray? Alternatively, you can just revise pool and modularize all the affected kernels in this PR. Either way is fine by me, but I probably will prefer to not creating poolV2, because these versions generally align with tf kernel names' name.
</blockquote></details>

A general comment about buffer (relevant here and below). I would say it is safe to use buffer because it is not really an op in the tensorflow definition of op (it involves no tensors). It is just a wrapper around the constructor for TensorBuffer. 

TensorBuffer gives a nice interface for indexing into tensors, the logic for that would need to be reimplemented if you just have a typed array. So in terms of not using buffer(), i would ignore the 'kernels should not call ops' rule. However if there are other reasons to not use it we should discuss those, but I don't think buffer being an 'op' is a blocker in these cases.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3973)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 62 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC7iE86H9YneNDjdgm:-MIKcILC3kWhQIhg7ac0:bu7i25p) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPool.ts#L62)):*

A general comment about buffer (relevant here and below). I would say it is safe to use buffer because it is not really an op in the tensorflow definition of op (it involves no tensors). It is just a wrapper around the constructor for TensorBuffer. 

TensorBuffer gives a nice interface for indexing into tensors, the logic for that would need to be reimplemented if you just have a typed array. So in terms of not using buffer(), i would ignore the 'kernels should not call ops' rule. However if there are other reasons to not use it we should discuss those, but I don't think buffer being an 'op' is a blocker in these cases. 
</blockquote></details>

Make sense.

jinjingforever (commented):
Thanks Na and Yannick! I removed the unnecessary asserts and kept the "buffer" calls as Yannick suggested. PTAL. Thanks!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3973#-:-MIKlqVd7ugHCkpy9VMQ:b-65jn77)* status: 0 of 1 approvals obtained (waiting on @lina128 and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 40 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC3n6s5IKIt3jc7o0d:-MIKgdLI2uhbNXbnRbjj:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPool.ts#L40)):*

+1
</blockquote></details>

Done
___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 45 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC48Va0tWHx6dj94Cs:-MIKgfUR98VqOXXFKH2q:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPool.ts#L45)):*

+1. 
</blockquote></details>

Done
___
*[tfjs-backend-cpu/src/kernels/AvgPool.ts, line 62 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC7iE86H9YneNDjdgm:-MIKghl19ftUA9TfTAE-:b-tttsqh) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPool.ts#L62)):*

Make sense.
</blockquote></details>

It makes sense. Thanks! Keep it as it is for now.
___
*[tfjs-backend-cpu/src/kernels/AvgPoolBackprop.ts, line 35 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MIC9d0Y9oeg6Nigkf1v:-MIKhcweDgXENQpfteKJ:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/AvgPoolBackprop.ts#L35)):*

No need to check rank here, this is type checked in Avgpool_grad.ts. Also if avgPool only work with 4D, it's not likely the backward propagation can get anything not 4D. So we can save an assertion here.
</blockquote></details>

Done
___
*[tfjs-backend-cpu/src/kernels/MaxPool.ts, line 35 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MICAcAs5wOBv_ZEiul4:-MIKljWSDzUJOv7I_nM8:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/MaxPool.ts#L35)):*

Same as above. No need for assertion.
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/kernels/MaxPoolBackprop.ts, line 36 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3973#-MICB03c15MXxS9mPYhp:-MIKhY9z5sUEMnPfw6h0:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/88d0220cd72c8e471712aae68419c78bd22c9946/tfjs-backend-cpu/src/kernels/MaxPoolBackprop.ts#L36)):*

Same, no need to assert.
</blockquote></details>

Done

lina128 (approved):
Thanks Jing!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3973#-:-MIKqSxeDg1-59NWFt9v:b772o2f)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128 and @tafsiri)

tafsiri (approved):
> Reviewed 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3973)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128 and @tafsiri)

Issue #3864, opened 2020-08-31
WASM -> TypeError: Cannot read property 'cwrap' of undefined


sbstcano:
Hi,
I'm trying to use wasm with facemesh but keep getting this when trying to set wasm as backend.
I'm using version 2.3.0. And it's an Angular 9 project.

```
import * as tf from '@tensorflow/tfjs-core'
import * as facemesh from '@tensorflow-models/facemesh'
import {setWasmPath} from "@tensorflow/tfjs-backend-wasm"

setWasmPath(`https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm@2.3.0/dist/tfjs-backend-wasm.wasm`)

tf.setBackend('wasm')
```

```
TypeError: Cannot read property 'cwrap' of undefined
    at Object.setupFunc (tf-backend-wasm.esm.js:17)
    at engine.js:161
    at Array.forEach (<anonymous>)
    at Engine.setupRegisteredKernels (engine.js:159)
    at Engine.setBackend (engine.js:152)
    at Module.setBackend (globals.js:258)
    at Module../src/app/enrollment/enrollment.component.ts (enrollment.component.ts:20)
    at __webpack_require__ (bootstrap:79)
    at Module../src/app/app-routing.module.ts (main.js:450)
    at __webpack_require__ (bootstrap:79)
```

deel77:
I think I have identified main problem. Angular is overwriting `Promise.resolve()` function with its own wrapping function in order to use `ZoneAwarePromise` as a return type in order to automatically trigger change detection after promise finishes.

And TF-Core is using following code to test if created backend is Promise (which is Promise type only for WASM backend, as to my knowledge):
```
// Test if the factory returns a promise.
if (Promise.resolve(backend) === backend) {
```
see https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/engine.ts#L294

and this type of test does not return true as return type of `Promise.resolve(backend)` is not same `Promise` but `ZoneAwarePromise` wrapper of original `Promise`.

I have tried to run `tf.setBackend('wasm')` inside `ngZone.runOutsideAngular()` but it does not help. I would try something more.

deel77:
With changes proposed in pull request #3993 I was able to make it work in Angular 9 and Angular 10 project ... still having problems with SIMD version but basic WASM backend is working now.

PR #3993, opened 2020-09-30
fix Promise type backend init in Angular

deel77:
This fixes https://github.com/tensorflow/tfjs/issues/3864

As specified in #3864 Angular is currently not working with WASM backend due to usage of custom `Promise.resolve()` implementation which is always wrapping object to `ZoneAwarePromise` (even if it already was).

Therefore check for Promise type backend using `Promise.resolve(backend) === backend` is always returning `false`.

It is only manifesting with WASM backend in Angular as both WebGL and CPU are not returned as Promise from factory.

This is fix to this problem, which is keeping previous functionality unchanged.

Optionally it can be solved by always wrapping all backends into `Promise.resolve()` and handling all backends in async way, but it would mean changing more function signature and all usages could be changed to handle only async way.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3993)


annxingyuan:
Hi @deel77 - thank you so much for this fix! Could you include a comment in the code explaining that this is a more liberal check and accounts for custom `Promise` implementations?

deel77:
Hi @annxingyuan, I have expanded comment to explain the change

annxingyuan (approved):
@deel77 awesome - thank you!

PR #3995, opened 2020-09-30
Fix webgl CPU forwarding for ops that have been modularized in cpu backend

jinjingforever:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995)


jinjingforever:
Hi Ann,

Here is a draft PR where I attempt to fix the problem for one simple op "floor" as an example. This op has the exact problem you described because "floor" has been modularized in cpu (so ```this.cpuBackend.floor``` call would fail).

Please take a look and see if this PR is heading towards the right direction:) The basic idea is: for each cpu forwarding call from webgl (```this.cpuBackend.xxx()```), we need to check if the corresponding op has been modularized or not. If so (an error would be thrown), we need to call the shared cpu implementation instead.

Once you think this fix is good, I will continue working on this PR to apply the same logic to all the other places.

Thank you very much!

annxingyuan:
Thank you @jinjingforever ! This is pretty much exactly what I was thinking :) The only feedback I have is - I think elsewhere in our codebase we have used `typeof func === 'function'` to check whether functions are defined, rather than using `try... catch`. Also, since we're manually modularizing each kernel, maybe we could just call the modularized CPU impl function where necessary, rather than going through a generic check for each kernel.

Would also be awesome if you could add a few e2e tests that explicitly test CPU forwarding!

jinjingforever (commented):
Thank you Ann for the feedback! Really appreciate it. I will continue working on this and will let you know when the fix is ready.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995#-:-MIZteGa-msBCs9Y_xbG:b-86t3la)* status: 0 of 1 approvals obtained (waiting on @annxingyuan)

tafsiri:
+1 to just calling the modularized cpu impl where available. However in general we should wrap the other (non modularized) cpu forwarding calls in a try-catch block before falling back to webgl implementations. Using `typeof func === 'function'` may not work because there would usually be a function definition inherited from the base class (that throws an exception)

jinjingforever:
Some notes about this PR:

- I had to pull some changes (mainly the changes in kernel_utils.ts and its callers) from Ann's PR (#3956) to make things work with minimum conflicts.

- This PR also touches some webgl kernels that will be modularized (and fixed) in Ann's PR (e.g. multiply, concat). Depending which PR is merged first, one of us could just safely delete the code in backend_webgl.ts to resolve conflicts.

- Last week, I added a bunch of e2e tests covering all kernels that directly do cpu forwarding. As you guys said, it is probably not necessary, but I will just leave them there for now:)

- I tested locally with cpu forwarding turned-on (thanks to Yannick's draft PR) and things seem to be passing.

annxingyuan (commented):
Thank you Jing! Do we want to also make the changes that we discussed from Yannick's draft PR? (https://github.com/tensorflow/tfjs/pull/4031)

I think the changes are:
- Change existing environments in `backend_webgl_test_registry.ts` to not explicitly set `WEBGL_CPU_FORWARD`
- Add a third test environment to `backend_webgl_test_registry.ts`that explicitly sets `WEBGL_CPU_FORWARD` to `true`
- Add another run-browserstack invocation in `tfjs-backend-webgl/scripts/test-ci.sh` nightly to explicitly enable `WEBGL_CPU_FORWARD`
- Disable `IS_TEST` check in the backend_webgl.ts `shouldExecuteOnCPU` function

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995#-:-MJ3Y15cE0g0bi9HHC08:bik9sa1)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, @pyu10055, and @tafsiri)

tafsiri (changes requested):
Thanks for making this PR.  Just to confirm, this doesn't cover all the uses of CPU forwarding in the webgl backend right? (for example I was expecting to see `return this.cpuBackend.greater(a, b);` wrapped in a try catch).

Also could you merge the branch from my PR into this branch so we can see the output on CI and iterate on the config we want for testing locally and CI right in the context of this PR?

With regard to try-catch blocks I think we have a number of options:
 - The commits on my branch should catch unwrapped calls into the cpu backend (e.g. `return this.cpuBackend.greater(a, b);`) so in theory we do not need to wrap them. In fact leaving them unwrapped forces us to do the new style CPU forwarding in the same PR as modularizing a CPU kernel.
 - Alternatively we can wrap them so that even if they do fail webgl falls back onto a webgl implementation, this would cause a performance regression in those cases but will always allow the webgl backend to keep executing.
- We can combine the two ideas above by re-throwing the exception in the catch block in tests (but in prod allowing fallback to webgl). The advantage of this is that even if something slips through our tests an op wouldn't fail on webgl. Here is a code snippet illustrating that.

backend_webgl greater()
```ts
greater(a: Tensor, b: Tensor): Tensor {
    if (this.shouldExecuteOnCPU([a, b])) {
      try {
        return this.cpuBackend.greater(a, b);
      } catch() {
         if env().getBool('IS_TEST') {
            throw new Exception('CPU forwarding failed')
         }
         // else we just fall back to webgl implementation.
      }
    }

    if (env().getBool('WEBGL_PACK_BINARY_OPERATIONS')) {
      return this.packedBinaryOp(a, b, binaryop_packed_gpu.GREATER, 'bool');
    }

    const program = new BinaryOpProgram(binaryop_gpu.GREATER, a.shape, b.shape);
    return this.compileAndRun(program, [a, b], 'bool');
  }
```

Thoughts? I lean towards the third option.

> Reviewed 4 of 8 files at r1, 18 of 18 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995#-:-MJ3cVkP39Yxd20GXP_k:b2a9tau)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, @pyu10055, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/Ceil.ts, line 24 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3aYtw95pk5RjXvGpG:-MJ3aYtx7yBW6G6mX5ob:b-p57rut) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-cpu/src/kernels/Ceil.ts#L24)):*
> ```XML
> 
> export const ceilImpl = createSimpleUnaryImpl((xi) => Math.ceil(xi));
> export const ceilKernelFunc = unaryKernelFuncFromImpl(Ceil, ceilImpl);
> ```

we should stick to one convention for naming these exported kernel funcs. In existing code they do not have the kernelFunc suffix. I think we should stick to that so that on importing these in other cpu kernels we do have the kernelFunc suffix present (or alternatively we could have the suffix on all of them). 
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 734 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3b6cn8Wyv_baWhTBg:-MJ3b6co5ycJF7tNNh7Y:b-i64svq) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L734)):*
> ```XML
>     if (this.shouldExecuteOnCPU([x])) {
>       try {
>         return this.cpuBackend.slice(x, begin, size);
> ```

We shouldn't need to have this and the code in the catch block below. We can just use the code in the catch block.

For kernels that aren't modularized in cpu we should try catch. 
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 820 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3bVLM46UnFE75k5N_:-MJ3bVLM46UnFE75k5Na:b-6uxgoj) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L820)):*
> ```XML
>     // PR #3956 will modularize concat without this.
>     // if (this.shouldExecuteOnCPU(tensors)) {
>     //   return this.cpuBackend.concat(tensors, axis);
> ```

I think there should be a try catch block here. If it fails we skip the early return and let webgl proceed as if there was no cpu forwarding.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 950 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3bf0o2lys-y41_2Zy:-MJ3bf0o2lys-y41_2Zz:bj75fvf) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L950)):*
> ```XML
>     if (this.shouldExecuteOnCPU([a, b])) {
>       try {
>         return this.cpuBackend.multiply(a, b);
> ```

same as above, if we can import multipleImplCPU we should not even try calling cpuBackend.multiply.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 1732 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3cs7vDRgXuV6ndXZU:-MJ3cs7vDRgXuV6ndXZV:b9ltg07) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L1732)):*
> ```XML
>         return this.cpuBackend.rsqrt(x);
>       } catch (e) {
>         const outValues = rsqrtImplCPU(
> ```

should a utility function be made for these? there seem to be a lot of similar calls.

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @pyu10055)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 1658 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3igCuAtNI9bc9pGXJ:-MJ3igCuAtNI9bc9pGXK:b7f1vbz) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L1658)):*
> ```XML
> 
>   expm1<T extends Tensor>(x: T): T {
>     if (this.shouldExecuteOnCPU([x])) {
> ```

can we create a method something like
return cpuExecuteOrFallback(executeFunc, fallbackFunc)?
the executeFunc will be
() => this.cpuBackend.expm1(x)
the fallbackFunc will be
()  => {
        const outValues = expm1ImplCPU(
            this.texData.get(x.dataId).values as TypedArray, x.dtype);
        const outInfo = this.makeTensorInfo(x.shape, x.dtype);
        this.texData.get(outInfo.dataId).values = outValues;
        return engine().makeTensorFromDataId(
                   outInfo.dataId, outInfo.shape, outInfo.dtype) as T;
      }

lina128 (commented):
Regarding test env in test registry, why can't we keep it. And then just add one line to test-ci.sh, `yarn run-browserstack --browsers=bs_chrome_mac --testEnv webgl2 --flags '{"WEBGL_CPU_FORWARD": true}'`

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995#-:-MJ3l5Dp7CPoBO6HWdID:b-yxefzd)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 734 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3b6cn8Wyv_baWhTBg:-MJ3jUBGCnYED041leUk:b32v57) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L734)):*

We shouldn't need to have this and the code in the catch block below. We can just use the code in the catch block.

For kernels that aren't modularized in cpu we should try catch. 
</blockquote></details>

+1

tafsiri:
Chatted with @jinjingforever offline. I think we want to pick between the registerTestEnv approach and the explicit settings of flags approach. If we do the following (as Ann suggests)

```
registerTestEnv({
  name: 'webgl1',
  backendName: 'webgl',
  flags: {
    'WEBGL_VERSION': 1,
    'WEBGL_CPU_FORWARD': false,
    'WEBGL_SIZE_UPLOAD_UNIFORM': 0
  },
  isDataSync: true
});

registerTestEnv({
  name: 'webgl2',
  backendName: 'webgl',
  flags: {
    'WEBGL_VERSION': 2,
    'WEBGL_CPU_FORWARD': false,
    'WEBGL_SIZE_UPLOAD_UNIFORM': 0
  },
  isDataSync: true
});

registerTestEnv({
  name: 'webgl2',
  backendName: 'webgl',
  flags: {
    'WEBGL_VERSION': 2,
    'WEBGL_CPU_FORWARD': true,
    'WEBGL_SIZE_UPLOAD_UNIFORM': 0
  },
  isDataSync: true
});
```

Then we do not need to add a line with an explicit setting of `'WEBGL_CPU_FORWARD': true`, this is because when no flags are passed the tests will run in all the test_envs listed above. I'm fine with that as its super safe and give immediate feedback (at the cost of longer test runs—but the functionality being tested seems pretty key). 

However I do think there may be places where we have been running the same config twice, looking at our current test-ci for webgl

```
if [ "$NIGHTLY" = true ]; then
  yarn run-browserstack --browsers=bs_safari_mac,bs_ios_11 --testEnv webgl1 --flags '{"WEBGL_CPU_FORWARD": false, "WEBGL_SIZE_UPLOAD_UNIFORM": 0}'
  yarn run-browserstack --browsers=bs_firefox_mac,bs_chrome_mac
  yarn run-browserstack --browsers=bs_chrome_mac,win_10_chrome,bs_android_9 --testEnv webgl2 --flags '{"WEBGL_CPU_FORWARD": false, "WEBGL_SIZE_UPLOAD_UNIFORM": 0}'
  
  yarn run-browserstack --browsers=bs_chrome_mac --testEnv webgl2 --flags '{"WEBGL_PACK": false}'
else
  yarn run-browserstack --browsers=bs_chrome_mac
fi
```

To my eyes, in the nightly build the bs_chrome_mac on the second run and bs_chrome_mac on the third line run under the same conditions. I'm not 100% sure of that so we don't have to resolve it in this PR. But overall being able to easily trace where to set things to get the conditions needed is useful, which is why i was initially in favor of primarily using one of these mechanisms. But I've come around to the approach described above.

pyu10055 (commented):
> registerTestEnv({
>   name: 'webgl2',
>   backendName: 'webgl',
>   flags: {
>     'WEBGL_VERSION': 2,
>     'WEBGL_CPU_FORWARD': false,
>     'WEBGL_SIZE_UPLOAD_UNIFORM': 0
>   },
>   isDataSync: true
> });
> 
> registerTestEnv({
>   name: 'webgl2',
>   backendName: 'webgl',
>   flags: {
>     'WEBGL_VERSION': 2,
>     'WEBGL_CPU_FORWARD': true,
>     'WEBGL_SIZE_UPLOAD_UNIFORM': 0
>   },
>   isDataSync: true
> });

There two webgl2 registered with different WEBGL_CPU_FORWARD flag values?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995#-:-MJ3w6E05aqLYx3lYZno:b-392369)* status: 0 of 1 approvals obtained (waiting on @jinjingforever and @tafsiri)

annxingyuan (commented):
Actually let's do as Na suggested and leave the backend_webgl test registry alone, but add another call to run-browserstack in the nightly test-ci for WebGL that explicitly turns CPU forwarding on.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995#-:-MJ3ewfUAc9gGtWpVOr-:bj6okl)* status: 0 of 1 approvals obtained (waiting on @jinjingforever and @tafsiri)

jinjingforever (commented):
Thanks for all the comments! PTAL

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995#-:-MJ46rSTDZ37SDGj2QJk:bb25mzd)* status: 0 of 1 approvals obtained (waiting on @pyu10055 and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/Ceil.ts, line 24 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3aYtw95pk5RjXvGpG:-MJ46xhGCqE-Trw_Z3t3:b82ir7z) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-cpu/src/kernels/Ceil.ts#L24)):*

we should stick to one convention for naming these exported kernel funcs. In existing code they do not have the kernelFunc suffix. I think we should stick to that so that on importing these in other cpu kernels we do have the kernelFunc suffix present (or alternatively we could have the suffix on all of them). 
</blockquote></details>

Good point. I will go through them in another PR to make them consistent.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 734 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3b6cn8Wyv_baWhTBg:-MJ47F0sCeBKW_H8sYqd:b-hi2kgu) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L734)):*

+1
</blockquote></details>

Ah, of course. Good point. Updated (here and other places)
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 820 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3bVLM46UnFE75k5N_:-MJ47Ksl8Y3ugrmmtPly:bjmiwrz) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L820)):*

I think there should be a try catch block here. If it fails we skip the early return and let webgl proceed as if there was no cpu forwarding.
</blockquote></details>

For now, I just removed the cpu-forwarding logic, since Ann's PR doesn't have this after concat is modularized. Will sync with her about this offline.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 950 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3bf0o2lys-y41_2Zy:-MJ47Zrq0qti-QkHv_Ew:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L950)):*

same as above, if we can import multipleImplCPU we should not even try calling cpuBackend.multiply.
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 1658 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3igCuAtNI9bc9pGXJ:-MJ47m8W6nD7mCmijdYS:bl3apo) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L1658)):*

can we create a method something like
return cpuExecuteOrFallback(executeFunc, fallbackFunc)?
the executeFunc will be
() => this.cpuBackend.expm1(x)
the fallbackFunc will be
()  => {
        const outValues = expm1ImplCPU(
            this.texData.get(x.dataId).values as TypedArray, x.dtype);
        const outInfo = this.makeTensorInfo(x.shape, x.dtype);
        this.texData.get(outInfo.dataId).values = outValues;
        return engine().makeTensorFromDataId(
                   outInfo.dataId, outInfo.shape, outInfo.dtype) as T;
      }

</blockquote></details>

I added/refactored some helper methods combining various comments above.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 1732 at r2](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ3cs7vDRgXuV6ndXZU:-MJ47s9b8e9gkFj5nHN8:b-akoeej) ([raw file](https://github.com/tensorflow/tfjs/blob/47e8350d97089468a658127f989caa02dd37dffe/tfjs-backend-webgl/src/backend_webgl.ts#L1732)):*

should a utility function be made for these? there seem to be a lot of similar calls. 
</blockquote></details>

Done.  (I didn't make it super generic due to some edge cases).

lina128 (commented):
Thank you Jing for the fast iteration! Is test condition added to test-ci?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995#-:-MJ4AyF4ErPv6XRqd1Sj:b-x5ph8k)* status: 0 of 1 approvals obtained (waiting on @pyu10055 and @tafsiri)

lina128 (approved):
LGTM!

annxingyuan (approved):
Thanks so much for the quick fix Jing! Just left a small suggestion.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995#-:-MJ4WKy0A21nYj_2E6oV:beep3qp)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @jinjingforever, @pyu10055, and @tafsiri)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 2420 at r3](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ4VsfCEbit6lrL4izK:-MJ4VsfCEbit6lrL4izL:bg0r4u8) ([raw file](https://github.com/tensorflow/tfjs/blob/a466d13fc8dbf47cac489fd7acf607383b2bbb5d/tfjs-backend-webgl/src/backend_webgl.ts#L2420)):*
> ```XML
>   private makeOutput<T extends Tensor>(
>       shape: number[], dtype: DataType, values?: BackendValues): T {
>     const {dataId} = this.makeTensorInfo(shape, dtype);
> ```

if you pass in values to makeTensorInfo you shoudn't need the if block below

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @annxingyuan, @pyu10055, and @tafsiri)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 2420 at r3](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ4VsfCEbit6lrL4izK:-MJ4Xz_l39D8RURAK2yt:b2rl78t) ([raw file](https://github.com/tensorflow/tfjs/blob/a466d13fc8dbf47cac489fd7acf607383b2bbb5d/tfjs-backend-webgl/src/backend_webgl.ts#L2420)):*

if you pass in values to makeTensorInfo you shoudn't need the if block below
</blockquote></details>

Nice! Done. Thank you!

tafsiri (approved):
Great! (left one suggestion about testing cpu forwarding on PRs as well as nightly)

> Reviewed 2 of 3 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995#-:-MJ4buax5KHLbzdpUbsE:baavqgs)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, and @pyu10055)

___
*[tfjs-backend-webgl/scripts/test-ci.sh, line 26 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ4arzC8v72JPMwMuw4:-MJ4arzDA0GgvBUWX1ea:b-4crluz) ([raw file](https://github.com/tensorflow/tfjs/blob/2dcd7e0a5fa98874f6219d01f5861e2835fb35d9/tfjs-backend-webgl/scripts/test-ci.sh#L26)):*
> ```Shell
>   yarn run-browserstack --browsers=bs_chrome_mac --testEnv webgl2 --flags '{"WEBGL_CPU_FORWARD": true}'
> else
>   yarn run-browserstack --browsers=bs_chrome_mac
> ```

Should we add the line added above to here so that we can catch these bugs like this on PRs and not just nightly?

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @pyu10055)

___
*[tfjs-backend-webgl/scripts/test-ci.sh, line 26 at r4](https://reviewable.io/reviews/tensorflow/tfjs/3995#-MJ4arzC8v72JPMwMuw4:-MJ4eoc-8Kt7y9PfrHe4:b-jk53ar) ([raw file](https://github.com/tensorflow/tfjs/blob/2dcd7e0a5fa98874f6219d01f5861e2835fb35d9/tfjs-backend-webgl/scripts/test-ci.sh#L26)):*

Should we add the line added above to here so that we can catch these bugs like this on PRs and not just nightly?
</blockquote></details>

Thanks! I asked @lina128 about this and she said we probably don't have enough resources on browserstack to run more tests for PRs.

pyu10055 (approved):
> Reviewed 4 of 8 files at r1, 17 of 18 files at r2, 2 of 3 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/3995)* status: :shipit: complete! 4 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @pyu10055)

PR #4016, opened 2020-10-05
Avoid keeping pointer to temporary strings

estyrke:
substr() creates a temporary string which will be deallocated, leaving the c_str() pointer pointing to garbage.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4016)

### DIFF
--- a/tfjs-node/binding/tfjs_backend.cc
+++ b/tfjs-node/binding/tfjs_backend.cc
@@ -1134,12 +1134,12 @@ napi_value TFJSBackend::RunSavedModel(napi_env env,
     std::string name(input_op_name_array[i]);
     int index = name.find(":");
     std::string input_op_name = name.substr(0, index);
-    const char *input_op_index = name.substr(index + 1).c_str();
+    std::string input_op_index = name.substr(index + 1);
     int input_tensor_index;
-    if (strlen(input_op_index) == 0) {
+    if (input_op_index.length() == 0) {
       input_tensor_index = 0;
     } else {
-      input_tensor_index = atoi(input_op_index);
+      input_tensor_index = atoi(input_op_index.c_str());
     }
 
     // Add input op into input ops list.
@@ -1162,12 +1162,12 @@ napi_value TFJSBackend::RunSavedModel(napi_env env,
     std::string name(output_op_name_array[i]);
     int index = name.find(":");
     std::string output_op_name = name.substr(0, index);
-    const char *output_op_index = name.substr(index + 1).c_str();
+    std::string output_op_index = name.substr(index + 1);
     int output_tensor_index;
-    if (strlen(output_op_index) == 0) {
+    if (output_op_index.length() == 0) {
       output_tensor_index = 0;
     } else {
-      output_tensor_index = atoi(output_op_index);
+      output_tensor_index = atoi(output_op_index.c_str());
     }
 
     TF_Operation *output_op = TF_GraphOperationByName(


pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4016)* status: 0 of 1 approvals obtained (waiting on @estyrke)

___
*[tfjs-node/binding/tfjs_backend.cc, line 1142 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4016#-MItdRwS3gm9HOLWcn4G:-MItdRwS3gm9HOLWcn4H:b8a0tqh) ([raw file](https://github.com/tensorflow/tfjs/blob/6177e7174b688a93fcbd74e09b52e3ce546f7346/tfjs-node/binding/tfjs_backend.cc#L1142)):*
> ```cpp
>       input_tensor_index = 0;
>     } else {
>       input_tensor_index = atoi(input_op_index.c_str());
> ```

Is the concern that the temporary string was not disposed? If so here the temporary string is created?

estyrke:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4016)_ status: 0 of 1 approvals obtained (waiting on @estyrke)
> 
> _[tfjs-node/binding/tfjs_backend.cc, line 1142 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4016#-MItdRwS3gm9HOLWcn4G:-MItdRwS3gm9HOLWcn4H:b8a0tqh) ([raw file](https://github.com/tensorflow/tfjs/blob/6177e7174b688a93fcbd74e09b52e3ce546f7346/tfjs-node/binding/tfjs_backend.cc#L1142)):_
> 
> > ```c++
> >       input_tensor_index = 0;
> >     } else {
> >       input_tensor_index = atoi(input_op_index.c_str());
> > ```
> 
> Is the concern that the temporary string was not disposed? If so here the temporary string is created?

Not exactly. The problem is on this line: 

```c++
const char *input_op_index = name.substr(index + 1).c_str();
```

The expression `name.substr(index + 1)` creates a temporary variable whose scope ends at the end of the "full-expression" (essentially the end of this line). The `c_str()` call then takes the address of the string pointed to by this temporary variable. Thus, after this line, the pointer in `input_op_index` points to an already deallocated memory area. When the pointer is then used on the line you pointed out, the result is undefined. In most cases, the memory contents may still be there and no problem is observed. But if the deallocated memory area has already been cleared or reused, the atoi (or the strlen before it) will fail. This is what we observed on our compiler (VS 2019) - the strlen returns 0 and the output_tensor_index (in our case) will always be zero no matter which output we try to select.

In general, it is risky to hold on to the value returned by c_str, since it is an unmanged C pointer which will be undefined if used after its backing std::string has been deallocated - It is better to only call it "temporarily" when it is needed, such as within the atoi call here.

estyrke:
VC2019 prints a warning to this effect, unfortunately I don't have a windows computer at hand now so I can't quote it.

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4016)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @estyrke)

___
*[tfjs-node/binding/tfjs_backend.cc, line 1142 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4016#-MItdRwS3gm9HOLWcn4G:-MIyJIMFEidKqpF4nO7o:bmvddru) ([raw file](https://github.com/tensorflow/tfjs/blob/6177e7174b688a93fcbd74e09b52e3ce546f7346/tfjs-node/binding/tfjs_backend.cc#L1142)):*
<details><summary><i>Previously, estyrke (Emil Styrke) wrote…</i></summary><blockquote>

Not exactly. The problem is on this line: 

```c++
const char *input_op_index = name.substr(index + 1).c_str();
```

The expression `name.substr(index + 1)` creates a temporary variable whose scope ends at the end of the "full-expression" (essentially the end of this line). The `c_str()` call then takes the address of the string pointed to by this temporary variable. Thus, after this line, the pointer in `input_op_index` points to an already deallocated memory area. When the pointer is then used on the line you pointed out, the result is undefined. In most cases, the memory contents may still be there and no problem is observed. But if the deallocated memory area has already been cleared or reused, the atoi (or the strlen before it) will fail. This is what we observed on our compiler (VS 2019) - the strlen returns 0 and the output_tensor_index (in our case) will always be zero no matter which output we try to select.

In general, it is risky to hold on to the value returned by c_str, since it is an unmanged C pointer which will be undefined if used after its backing std::string has been deallocated - It is better to only call it "temporarily" when it is needed, such as within the atoi call here. 
</blockquote></details>

got it, thanks for the explanation.

PR #4024, opened 2020-10-05
Link to custom bundle binary directly in e2e tests

tafsiri:
Should fix errors related to npx running the binary.

The behaviour in CI is different from what I have locally (i cannot reproduce the error in CI locally). I'm logging the versino of node used to I can investigate differences more closely but for now i'm just directly invoking the binary rather than relying on npx to set the PATH to unblock releasing.

I'm also making sure we are testing fresh custom modules each time by deleting the old ones before a custom build is done.

This is currently going through e2e with NIGHTLY=true, so even if it passes PR check's i'll wait until that passes.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4024)

### DIFF
--- a/CONTRIBUTING_MISSING_OP.md
+++ b/CONTRIBUTING_MISSING_OP.md
@@ -1,0 +1,75 @@
+## How to Contribute a New op to TF.js Repository
+
+### Step 1. Add the new op to tfjs-core ops directory.
+**Prerequisite**
+1. The op should already be supported in [TF](https://www.tensorflow.org/api_docs/python/tf/all_symbols) api.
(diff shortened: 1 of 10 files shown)


lina128 (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4024#-:-MIusXcG3rVC9r9pmEit:bb6ywmx)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

pyu10055 (approved):

PR #4037, opened 2020-10-08
Python converter bugfix

patlevin:
## Motivation

The Python converter code supports writing boolean weights, but not reading them. This causes problems when reading TFJS models from Python.

Furthermore, the Python converter fails to read scalar string weights. Numeric scalar weights are read just fine, but string scalars cause an error.

## Implementation

`np.bool` has been added to the list of supported input dtypes.

The deserialization code for string arrays has been fixed so the size calculation always returns a valid `int`.

## Details

Scalars are stored in the weight manifest with an empty shape ([]), which is technically correct to reflect the nature of a scalar.
The string deserialization uses `np.prod()` to calculate the total number of items. Unfortunately, `np.prod()` returns `1.0` when given an empty iterable (see  [numpy docs](https://numpy.org/doc/stable/reference/generated/numpy.prod.html) under _Notes_).

This value cannot be used with `range()` later, which leads to an error in situations where a weight manifest contains string scalars (e.g. model [SSD OpenImages v4](https://tfhub.dev/google/openimages_v4/ssd/mobilenet_v2/1)).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4037)


pyu10055 (approved):
Thank you for the great contribution!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4037#-:-MJ9mhsI0wCJRUCi-D7N:b1l8mm4)* status: :shipit: complete! 1 of 1 approvals obtained

PR #4042, opened 2020-10-08
e2e custom bundling for dense models.

tafsiri:
- Adds a e2e test for a CUJ covering creating a custom bundle with a Dense model that was trained with tfjs_layers and converted with tfjs_converter (these two steps are not part of the test as they are covered by other tests). The model is the created here https://github.com/tensorflow/tfjs-examples/tree/master/getting-started.
- This e2e test also covers a locally hosted model (vs the blazeface one which loads a model hosted on GCP)
- Adds a code path to test custom bundling with rollup

- Modularizes CPU kernels necessary for dense models. BatchMatMul, _FusedBatchMatMul, relu, relu6 & prelu

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4042)


lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4042)* status: 0 of 1 approvals obtained (waiting on @lina128 and @tafsiri)

___
*[e2e/custom_bundle/dense_model/package.json, line 16 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8OQBGCB9LhGMAGJTb:-MJ8OQBGCB9LhGMAGJTc:b-1feov2) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/e2e/custom_bundle/dense_model/package.json#L16)):*
> ```JSON
>   "dependencies": {
>     "@tensorflow/tfjs": "link:../../../tfjs"
>   },
> ```

Hi Yannick, for release flow, do we want to test it against pinned version?
___
*[e2e/scripts/run-custom-builds.sh, line 48 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8PmSs9DAncJy_bj0A:-MJ8PmSs9DAncJy_bj0B:ba03n0m) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/e2e/scripts/run-custom-builds.sh#L48)):*
> ```Shell
> yarn webpack:full
> yarn webpack:custom
> yarn rollup:full
> ```

Curious, does rollup work with blazeface now? If not, what's the difference between dnn app and blazeface app? 
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8R8vO6RnM_JcP1fHL:-MJ8R8vO6RnM_JcP1fHM:baa9n5r) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/backend_cpu.ts#L34)):*
> ```XML
>  * @deprecated remove once all fused kernels are modularized.
>  *
>  * Use fused_utils.applyActication instead.
> ```

applyActivation
___
*[tfjs-backend-cpu/src/kernels/_FusedMatMul.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8SrxFDNt5fRNPMrkl:-MJ8SrxFDNt5fRNPMrkm:b2ff7gp) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/kernels/_FusedMatMul.ts#L34)):*
> ```XML
>   const {a, b, bias, preluActivationWeights} = inputs;
>   const {transposeA, transposeB, activation} = attrs;
> ```

There's a lot of transformation in mat_mul.ts, do they need to be moved here?
___
*[tfjs-backend-cpu/src/kernels/BatchMatMul.ts, line 31 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8RdD27TQfMXCsyaJI:-MJ8RdD27TQfMXCsyaJJ:b1lrarf) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/kernels/BatchMatMul.ts#L31)):*
> ```XML
>   const {a, b} = inputs;
>   const {transposeA, transposeB} = attrs;
> ```

There's a lot of transformation code in mat_mul op, we need to move those to here as well.

tafsiri (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4042)* status: 0 of 1 approvals obtained (waiting on @lina128)

___
*[e2e/custom_bundle/dense_model/package.json, line 16 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8OQBGCB9LhGMAGJTb:-MJ8Zemc4CbooUb1GVCK:bfg8tm5) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/e2e/custom_bundle/dense_model/package.json#L16)):*

Hi Yannick, for release flow, do we want to test it against pinned version?
</blockquote></details>

I'm not worried about that since we always release from a commit that passed nightly so have a specific commit that we know has passed these tests.  How would it work to run against a pinned version during release (i'm assuming the version hasn't yet been published to npm).

Also do these tests get run during release flow?
___
*[e2e/scripts/run-custom-builds.sh, line 48 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8PmSs9DAncJy_bj0A:-MJ8VgaZCNisIxw-OJy1:byomcvx) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/e2e/scripts/run-custom-builds.sh#L48)):*

Curious, does rollup work with blazeface now? If not, what's the difference between dnn app and blazeface app? 
</blockquote></details>

Main difference is that I can't get wasm to work with rollup (not even for custom builds in particular—i've just never seen or been able to get a wasm based app to build with rollup).
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8R8vO6RnM_JcP1fHL:-MJ8W0v83gm0ol-KSjno:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/backend_cpu.ts#L34)):*

applyActivation
</blockquote></details>

Done
___
*[tfjs-backend-cpu/src/kernels/_FusedMatMul.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8SrxFDNt5fRNPMrkl:-MJ8ZInUDg4UUMczW4Xv:bic4h6e) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/kernels/_FusedMatMul.ts#L34)):*

There's a lot of transformation in mat_mul.ts, do they need to be moved here?
</blockquote></details>

Since this is our own kernel we can pretty much define the kernel interface however we want. Currently the kernels were written to only operate on 3d tensors (and any transformation needs to be done at the op level). I am fine keeping that interface. But see my comment below on mat_mal. 
___
*[tfjs-backend-cpu/src/kernels/BatchMatMul.ts, line 31 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8RdD27TQfMXCsyaJI:-MJ8Z15U-ifeO976mLFj:brh1b6i) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/kernels/BatchMatMul.ts#L31)):*

There's a lot of transformation code in mat_mul op, we need to move those to here as well.
</blockquote></details>

Ordinarily we wouldn't because it is outside the forwardFunc, indicating that the kernel itself only takes 3d tensors. However when i look at https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/batch-mat-mul it looks like the kernel should be able to handle any rank above 2. Maybe we just forgot to move that code inside the forward func when modularizing it?

I can move it now and then update this function.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4042)* status: 0 of 1 approvals obtained (waiting on @lina128 and @tafsiri)

___
*[e2e/custom_bundle/dense_model/package.json, line 16 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8OQBGCB9LhGMAGJTb:-MJ8dcq662-_KdBFhnsu:b-5nbrjg) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/e2e/custom_bundle/dense_model/package.json#L16)):*

I'm not worried about that since we always release from a commit that passed nightly so have a specific commit that we know has passed these tests.  How would it work to run against a pinned version during release (i'm assuming the version hasn't yet been published to npm).

Also do these tests get run during release flow?
</blockquote></details>

Yes, the tests (kicked off from test-ci.sh) get run during release flow. For release flow, before running test-ci.sh, the packages will be published locally to verdaccio, and e2e/package.json will get changed to pinned version. This will capture any bug in bundle building and publishing.
___
*[e2e/scripts/run-custom-builds.sh, line 48 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8PmSs9DAncJy_bj0A:-MJ8f9EO3WD_ejeJlmRo:b-qon1w1) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/e2e/scripts/run-custom-builds.sh#L48)):*

Main difference is that I can't get wasm to work with rollup (not even for custom builds in particular—i've just never seen or been able to get a wasm based app to build with rollup).
</blockquote></details>

Got it. Thanks for the explanation!
___
*[tfjs-backend-cpu/src/kernels/BatchMatMul.ts, line 31 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8RdD27TQfMXCsyaJI:-MJ8fWSRAh3ezGVO94We:b9pab1h) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/kernels/BatchMatMul.ts#L31)):*

Ordinarily we wouldn't because it is outside the forwardFunc, indicating that the kernel itself only takes 3d tensors. However when i look at https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/batch-mat-mul it looks like the kernel should be able to handle any rank above 2. Maybe we just forgot to move that code inside the forward func when modularizing it?

I can move it now and then update this function.
</blockquote></details>

I remember at some point, we said we would stopped moving the code inside forward func and will do it while modularizing kernel. Anyway, I would vote for move it here to prevent future discrepancy. It's not just 3d tensors, but also a and b type match, bias, and preluActivationWeights. If you look at fft_utils, there's a todo to use matmul once it supports complex64, so there can be other kernels calling it.

tafsiri (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4042)* status: 0 of 1 approvals obtained (waiting on @lina128 and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/BatchMatMul.ts, line 31 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8RdD27TQfMXCsyaJI:-MJ8rtpfEDMfMxZYfulK:b-7qga58) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/kernels/BatchMatMul.ts#L31)):*

I remember at some point, we said we would stopped moving the code inside forward func and will do it while modularizing kernel. Anyway, I would vote for move it here to prevent future discrepancy. It's not just 3d tensors, but also a and b type match, bias, and preluActivationWeights. If you look at fft_utils, there's a todo to use matmul once it supports complex64, so there can be other kernels calling it.
</blockquote></details>

Hmm, I don't remember that decision... I would say we want to make sure to do it before modularizing kernels to make sure implementations in different backends conform to the same interface. If you happen to remember skipping this for others please flag and we can update them early.

Agreed for this one we should make it conform to the expected (more flexible interface). Doing that now.

tafsiri (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4042)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/_FusedMatMul.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8SrxFDNt5fRNPMrkl:-MJ90n2920q0in09xVhG:bx6gkjc) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/kernels/_FusedMatMul.ts#L34)):*

Since this is our own kernel we can pretty much define the kernel interface however we want. Currently the kernels were written to only operate on 3d tensors (and any transformation needs to be done at the op level). I am fine keeping that interface. But see my comment below on mat_mal. 
</blockquote></details>

@annxingyuan do you have any thoughts on what you want the accepted tensor rank for the _FusedMatMul to be? See the comment below for the update we are making to BatchMatMul to make it conform with tf. But we have flexibility with this kernel. We probably discussed this while modularizing the op. But if you feel like it would be useful for the kernel to directly handle ranks other than 3 nows a good time to fix that.

tafsiri:
The gradient has been updated to handle non 3d tensors

Notes after diving down many a rabbit hole:

A rather strange bug manifests when moving the tensor transformation in mat_mul into the forward func. 

- The following test fails if you reshape the output inside the forward func: https://github.com/tensorflow/tfjs/pull/4042/files#diff-cc708fab06d0a3dc8632cff94b269533
  - This is actually a test of debug mode. IMO the profiler fails to execute the kernel correctly
  - This only happens in WebGL2 (works just fine in the webgl1 testEnv)
  - It *might* be because the forwardFunc has a mix of modularized and unmodularized kernels including a reshape that increments the backend refcounter. This is speculation, but is driven by seeing that in debug mode, while trying to profile the forwardFunc, an error is thrown claiming a there is a memory leak in BatchMatMul on webgl. This may prevent the profile function from finishing which leaves a value uninitialized causing the test to fail (the profiler throughs a null pointer exception)
  - I've currently skipped it in this PR to demonstrate that all the other tests pass (including the webgl tests for matMul itself)

I was trying not to modularize webgl batchmatmul in the same PR to keep the PR size down, but that may be the most practical option to see if we can bring that test back.

annxingyuan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4042)* status: 0 of 1 approvals obtained (waiting on @lina128 and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/_FusedMatMul.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8SrxFDNt5fRNPMrkl:-MJBoNBX19PC_-nFSe0k:b-w9tn1s) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/kernels/_FusedMatMul.ts#L34)):*

@annxingyuan do you have any thoughts on what you want the accepted tensor rank for the _FusedMatMul to be? See the comment below for the update we are making to BatchMatMul to make it conform with tf. But we have flexibility with this kernel. We probably discussed this while modularizing the op. But if you feel like it would be useful for the kernel to directly handle ranks other than 3 nows a good time to fix that.
</blockquote></details>

If batchMatMul is able to handle any rank tensor it would be nice if fusedMatMul worked the same way - I think it's the least surprising if the two kernels have the same expectations for the tensors being multiplied, since conceptually fusedMatMul wraps batchMatMul. But I'm also fine with keeping the existing interface :)

annxingyuan (commented):
In WebGL when we read an output texture we "decode" it by copying its values densely to a new texture in order to ultimately download the smallest possible texture from the GPU (e.g., a 2x2 texture with 4 values is copied to a 1x1 texture with all 4 values stuffed into the RGBA channels). This means the WebGL backend temporarily registers an extra dataId during decoding. So this test fails because in WebGL2 we read from the GPU asynchronously and the profiler catches the WebGL backend before it has had a chance to clean up after its decoder. This errors out because in debug mode, with the `IS_TEST` flag enabled we check each kernel for memory leaks. I think it's OK to leave this test out for now because users should never have `IS_TEST` enabled. I created a tracking issue here: https://github.com/tensorflow/tfjs/issues/4050 I'll work on a fix as a follow-up to this PR. 

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4042#-:-MJD1tof-49PzsWpkY1G:bld6v8o)* status: 0 of 1 approvals obtained (waiting on @lina128 and @tafsiri)

tafsiri:
Thanks for the info @annxingyuan. I'm also going to hold off of changing the fused mat_mul kernel interface in this PR. We can do that as an enhancement in future PRs. 

In that case this ready for re-review.

lina128 (approved):
Hi Yannick, this is great,  just left one comment about input transformation, otherwise LGTM.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4042#-:-MJDI0TG0-rwClzh67h3:b-ha0h9q)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-wasm/src/kernels/BatchMatMul.ts, line 51 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJDHta24iw0-CumW9Lh:-MJDHta24iw0-CumW9Li:b-y9m1cx) ([raw file](https://github.com/tensorflow/tfjs/blob/e29f2b8101b35d6a743a9b5429203f2f500a6bb8/tfjs-backend-wasm/src/kernels/BatchMatMul.ts#L51)):*
> ```XML
>   const {a, b} = inputs;
>   const {transposeA, transposeB} = attrs;
> ```

const [$a, $b] = makeTypesMatch(a, b);
___
*[tfjs-core/src/ops/mat_mul.ts, line 51 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJDHm7dCB7nek8FCwh8:-MJDHm7dCB7nek8FCwh9:b-dwdh6x) ([raw file](https://github.com/tensorflow/tfjs/blob/e29f2b8101b35d6a743a9b5429203f2f500a6bb8/tfjs-core/src/ops/mat_mul.ts#L51)):*
> ```XML
>   let $a = convertToTensor(a, 'a', 'matMul');
>   let $b = convertToTensor(b, 'b', 'matMul');
>   [$a, $b] = makeTypesMatch($a, $b);
> ```

Should this be moved to forwardFunc too?

annxingyuan (approved):

tafsiri (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4042)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128 and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/BatchMatMul.ts, line 31 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJ8RdD27TQfMXCsyaJI:-MJDUSYy7WGNXzzb0z8m:bb6un1j) ([raw file](https://github.com/tensorflow/tfjs/blob/58d63f7999156887541ea26491d1560679b6465a/tfjs-backend-cpu/src/kernels/BatchMatMul.ts#L31)):*

Hmm, I don't remember that decision... I would say we want to make sure to do it before modularizing kernels to make sure implementations in different backends conform to the same interface. If you happen to remember skipping this for others please flag and we can update them early.

Agreed for this one we should make it conform to the expected (more flexible interface). Doing that now.
</blockquote></details>

Done
___
*[tfjs-backend-wasm/src/kernels/BatchMatMul.ts, line 51 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJDHta24iw0-CumW9Lh:-MJDULZ1EQEG-LgKNl-Q:b-bg1rl4) ([raw file](https://github.com/tensorflow/tfjs/blob/e29f2b8101b35d6a743a9b5429203f2f500a6bb8/tfjs-backend-wasm/src/kernels/BatchMatMul.ts#L51)):*

const [$a, $b] = makeTypesMatch(a, b);
</blockquote></details>

see comment below about leaving this transform in the op.
___
*[tfjs-core/src/ops/mat_mul.ts, line 51 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4042#-MJDHm7dCB7nek8FCwh8:-MJDU7464SCaQluvYXEy:bccwzrv) ([raw file](https://github.com/tensorflow/tfjs/blob/e29f2b8101b35d6a743a9b5429203f2f500a6bb8/tfjs-core/src/ops/mat_mul.ts#L51)):*

Should this be moved to forwardFunc too?
</blockquote></details>

I left it here to match what I saw in ops.pbtxt for BatchMatMul, there it constrains the types to be the same, so I think we should have this transform as part of the op and not part of the kernels.

PR #4065, opened 2020-10-13
use webgl1 testEnv flags as registered

tafsiri:
This should make our nightly output less noisy. We should note that when we pass flags into karma, they will replace all the flags (it does not do a merge). In this case we replace flags with the values already defined in the test env, but did not also include the webgl version, leaving it unspecified. This change removes the unnecessary flag customization. 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4065)


annxingyuan (approved):
Niiiiiiice thank you yannick :)

tafsiri:
NP Ann. Merged!

Issue #3661, opened 2020-07-27
Modify fromPixel in webgpu backend to async function


shaoboyan:
We want to do this modification to enhance CPU/GPU parallel execution.

shaoboyan:
@qjia7

axinging:
Thanks @rthadur, we are still working on this, will update this soon.

rthadur:
related PR https://github.com/tensorflow/tfjs/pull/4375 has been merged , closing this request.Thank you

PR #4090, opened 2020-10-19
Add fromPixelsAsync

axinging:
Fix: https://github.com/tensorflow/tfjs/issues/3661

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4090)

### DIFF
--- a/tfjs-core/src/ops/browser_async.ts
+++ b/tfjs-core/src/ops/browser_async.ts
@@ -1,0 +1,29 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


tfjs-core/src/ops/browser_async.ts
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import { Tensor3D } from '../tensor';
+import { PixelData } from '../types';
+import { fromPixels } from './browser';
```
qjia7, line 20: Shouldn't you add a `fromPixelsAsync` in `browser` and import here?

qjia7: I take another look. Maybe the best way is to directly add `export const fromPixelsAsync = op({fromPixelsAsync_});` in browser.ts. Then `browser_async.ts` can be removed.

qjia7:
It's no needed anymore since #4375 is merged.

PR #4096, opened 2020-10-19
Modularize Fill.

lina128:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4096)


tafsiri (changes requested):
Left a few comments on ideas to reduce casting in this and subsequent modularizations for kernels that support string dtype.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4096#-:-MK1rhzsATuf6-xxx6Du:b50dc0i)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 87 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4096#-MK1qfnpBz11Z3HhpHhs:-MK1qfnpBz11Z3HhpHht:bt49q3i) ([raw file](https://github.com/tensorflow/tfjs/blob/f69888f2e34469914e998f8c2c31723b852e8709/tfjs-backend-cpu/src/backend_cpu.ts#L87)):*
> ```XML
>   makeTensorInfo(
>       shape: number[], dtype: DataType,
>       values?: backend_util.BackendValues): TensorInfo {
> ```

the type for values here should probably be something like `DataTypeMap<D>` where dtype is also of type `D` and `D` extends `DataType` (D would be a type parameter for the function). (related to comments below on getting rid of the casting).
___
*[tfjs-backend-cpu/src/kernels/Fill.ts, line 29 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4096#-MK1qTZ98LUT9dGOpCva:-MK1qTZA5EXd5KIvi8ok:bytt7v9) ([raw file](https://github.com/tensorflow/tfjs/blob/f69888f2e34469914e998f8c2c31723b852e8709/tfjs-backend-cpu/src/kernels/Fill.ts#L29)):*
> ```XML
>   const $dtype = dtype || util.inferDtype(value);
>   const values =
>       util.getArrayFromDType($dtype, util.sizeFromShape(shape)) as TypedArray;
> ```

I think if we get rid of this cast we might be able to get rid the cast below. The return type of the this function sometimes returns a string[]

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4096)* status: 0 of 1 approvals obtained (waiting on @annxingyuan and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 87 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4096#-MK1qfnpBz11Z3HhpHhs:-MK6ZdJ_34cC1OAYfs_5:b6ue9h0) ([raw file](https://github.com/tensorflow/tfjs/blob/f69888f2e34469914e998f8c2c31723b852e8709/tfjs-backend-cpu/src/backend_cpu.ts#L87)):*

the type for values here should probably be something like `DataTypeMap<D>` where dtype is also of type `D` and `D` extends `DataType` (D would be a type parameter for the function). (related to comments below on getting rid of the casting).
</blockquote></details>

Added string[] to allowed values type. DataTypeMap<D> does have string[], but does not have Uint8[], some kernels can pass in Uint8[].
___
*[tfjs-backend-cpu/src/kernels/Fill.ts, line 29 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4096#-MK1qTZ98LUT9dGOpCva:-MK6ZaeX4xX_QcnDhGLa:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/f69888f2e34469914e998f8c2c31723b852e8709/tfjs-backend-cpu/src/kernels/Fill.ts#L29)):*

I think if we get rid of this cast we might be able to get rid the cast below. The return type of the this function sometimes returns a string[]
</blockquote></details>

Done.

tafsiri (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4096)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 87 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4096#-MK1qfnpBz11Z3HhpHhs:-MK7k50Q-APllkizXLGT:b-16lsr8) ([raw file](https://github.com/tensorflow/tfjs/blob/f69888f2e34469914e998f8c2c31723b852e8709/tfjs-backend-cpu/src/backend_cpu.ts#L87)):*

Added string[] to allowed values type. DataTypeMap<D> does have string[], but does not have Uint8[], some kernels can pass in Uint8[].
</blockquote></details>

DataTypeMap does have Uint8Array (see bool). Are those kernels passing non-boolean UintArrays? Do we need to add a new dtype?

I'm okay with merging this, but I think DataTypeMap/DataValues better expresses the semantics of what is being taken in here (backendValues expressing the correct semantics for write). I could be wrong on this so would be happy to look at some of the kernels involved.  (its also not that big of a deal).
___
*[tfjs-backend-cpu/src/utils/fill_utils.ts, line 20 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4096#-MK7lqHaBCTylkESL0Co:-MK7kJ6JCFEAWh1t1h9k:b-g4pgvz) ([raw file](https://github.com/tensorflow/tfjs/blob/5761a8a082ff3334fb871b8236754e9082a11e7f/tfjs-backend-cpu/src/utils/fill_utils.ts#L20)):*
> ```XML
> import {DataType, DataValues, TypedArray} from '@tensorflow/tfjs-core';
> 
> export function fillValues(
> ```

Will this be used by things other than fill (maybe zerosLike and onesLike)? If not I wouldn't make a separate file for this function (and since zerosLike and onesLike don't take strings they can just call [].fill directly). 

If it would be used by those kernels, I wonder if its just a general backend util that should be in the backend or backend_utils?

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4096)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 87 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4096#-MK1qfnpBz11Z3HhpHhs:-MK7uJnJ9oDT5bcLTQbx:b1fp9vu) ([raw file](https://github.com/tensorflow/tfjs/blob/f69888f2e34469914e998f8c2c31723b852e8709/tfjs-backend-cpu/src/backend_cpu.ts#L87)):*

DataTypeMap does have Uint8Array (see bool). Are those kernels passing non-boolean UintArrays? Do we need to add a new dtype?

I'm okay with merging this, but I think DataTypeMap/DataValues better expresses the semantics of what is being taken in here (backendValues expressing the correct semantics for write). I could be wrong on this so would be happy to look at some of the kernels involved.  (its also not that big of a deal).
</blockquote></details>

I mean Uint8Array[]. Kernels that dealt with string can do the decoding, for example, using BufferSync, the result tensorBuffer will return Uint8Array[]. Take a look at Unique or Slice. I think this typing is pretty expressive, it explicitly spell out the edge case that needs special treatment before writing to backend. 
___
*[tfjs-backend-cpu/src/utils/fill_utils.ts, line 20 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4096#-MK7lqHaBCTylkESL0Co:-MK7xA8749ys6ACEOEmJ:b-arn204) ([raw file](https://github.com/tensorflow/tfjs/blob/5761a8a082ff3334fb871b8236754e9082a11e7f/tfjs-backend-cpu/src/utils/fill_utils.ts#L20)):*

Will this be used by things other than fill (maybe zerosLike and onesLike)? If not I wouldn't make a separate file for this function (and since zerosLike and onesLike don't take strings they can just call [].fill directly). 

If it would be used by those kernels, I wonder if its just a general backend util that should be in the backend or backend_utils?
</blockquote></details>

Moved into Fill.ts

Issue #3494, opened 2020-06-23
Problem with Inferencing with Custom Object Detection Model


jacklynlxq:
#### TensorFlow.js version
2.0.0

#### Browser version
Chrome Version 83.0.4103.61 (Official Build) (64-bit)

#### Describe the problem or feature request
I tried to train my own custom object detection model using Tensorflow's Object Detection API (using Tensorflow 1.14) and converted the model in Saved Model format into tfjs model. I followed this [tutorial](https://tensorflow-object-detection-api-tutorial.readthedocs.io/en/latest/training.html#) and used SSD Mobilenet v2. 

I was inferencing using executeAsync() for an object detection task. The function works well when I have a detection continuously in my live video. However, if I have no detection in any particular frame, there will be an error while it executes the executeAsync() function.

> TypeError: Cannot read property '0' of undefined

And here's the memory leak that happened when there's no detection. I am thinking the issue lies in the dangling tensors that are created in executeAsync() since if there's a detection I would dispose the tensors, but if there's no detection, I couldn't get a reference of the dangling tensors since the ```test_result``` variable hasn't been created yet due to the aforementioned error.

> {unreliable: false, numBytesInGPU: 422861990, numTensors: 1111, numDataBuffers: 673, numBytes: 320663096}
> {unreliable: false, numBytesInGPU: 490830799, numTensors: 1263, numDataBuffers: 752, numBytes: 371606568}
> {unreliable: false, numBytesInGPU: 558799608, numTensors: 1415, numDataBuffers: 831, numBytes: 422550040}

I tried converting from a pretrained inference graph (in Tensorflow Saved Model format) to a tfjs model and it works fine - if there's no detection executeAsync() would just return an empty array. It is only the problem with custom trained models.

#### Code to reproduce the bug / link to feature request
```
async detect(video){
    console.log("detecting...");

    var example = tf.browser.fromPixels(video);
    var tf4d = example.expandDims(0);

    try{
        // error if there's no detection
        const test_result = await this.model.executeAsync({ image_tensor: tf4d }, ['detection_boxes', 'num_detections', 'detection_classes', 'detection_scores']) as tf.Tensor[];
        
        this.no_detection = false;
        var detection_boxes = test_result[0].dataSync();
        var num_detections = test_result[1].dataSync();
        var detection_classes = test_result[2].dataSync();
        var detection_scores = test_result[3].dataSync();

        // dispose tensors to avoid memory leak
[...]

PR #4122, opened 2020-10-23
replace instanceof Promise with check of then function on the object

pyu10055:
In Angular Promise objects are monkey patched for zone.js
The type check `object instanceof Promise` will return false.
This PR created a util method that check the type of `then` field of the object instead.

fix https://github.com/tensorflow/tfjs/issues/3494

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4122)


lina128 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4122)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[tfjs-core/src/util.ts, line 800 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4122#-MKLeZb97GESoeLtbgaa:-MKLeZb97GESoeLtbgab:b-ts7jjd) ([raw file](https://github.com/tensorflow/tfjs/blob/50741e86d4c77f8acd77f9dce02af5f44228f53f/tfjs-core/src/util.ts#L800)):*
> ```XML
> 
> /**
>  * This method to check of the object can be interpreted as `Promise` instance.
> ```

This method asserts whether an object is a `Promise` instance.
___
*[tfjs-core/src/util.ts, line 802 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4122#-MKLhleu3iSsNRcR4UGI:-MKLhlev4g5_guTQEJtZ:b-89ssh9) ([raw file](https://github.com/tensorflow/tfjs/blob/50741e86d4c77f8acd77f9dce02af5f44228f53f/tfjs-core/src/util.ts#L802)):*
> ```
>  * Since `instanceof Promise` only works for es6 `Promise`, some framework like
>  * Angular has custom promise where `instanceof Promise` return false.
> ```
Should we move this part inside the function? Because it is implementation detail.
___
*[tfjs-core/src/util.ts, line 806 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4122#-MKLhsqpBGpCygXQKVR0:-MKLhsqq83vmaqflh0Rr:buuum0r) ([raw file](https://github.com/tensorflow/tfjs/blob/50741e86d4c77f8acd77f9dce02af5f44228f53f/tfjs-core/src/util.ts#L806)):*
> ```XML
>  */
> export function isPromise(object: any) {
>   return object && object.then && typeof object.then === 'function';
> ```

Add the annotation here, something like:
// We chose to not use 'obj instanceOf Promise' for two reasons: 1. It only reliably works for es6 Promise, not other Promise implementations. 2. It doesn't work with framework that uses zone.js. zone.js monkey patch the async calls, so it is possible the obj (patched) is comparing to a pre-patched Promise.

pyu10055 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4122)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

___
*[tfjs-core/src/util.ts, line 800 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4122#-MKLeZb97GESoeLtbgaa:-MKM84z0AFVrisT-Lp2m:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/50741e86d4c77f8acd77f9dce02af5f44228f53f/tfjs-core/src/util.ts#L800)):*

This method asserts whether an object is a `Promise` instance.
</blockquote></details>

Done.
___
*[tfjs-core/src/util.ts, line 802 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4122#-MKLhleu3iSsNRcR4UGI:-MKM85GhD4zjQIXslyJ7:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/50741e86d4c77f8acd77f9dce02af5f44228f53f/tfjs-core/src/util.ts#L802)):*

> ```
>  * Since `instanceof Promise` only works for es6 `Promise`, some framework like
>  * Angular has custom promise where `instanceof Promise` return false.
> ```
Should we move this part inside the function? Because it is implementation detail.
</blockquote></details>

Done.
___
*[tfjs-core/src/util.ts, line 806 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4122#-MKLhsqpBGpCygXQKVR0:-MKM85n8Ex6JLwzHpDVW:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/50741e86d4c77f8acd77f9dce02af5f44228f53f/tfjs-core/src/util.ts#L806)):*

Add the annotation here, something like:
// We chose to not use 'obj instanceOf Promise' for two reasons: 1. It only reliably works for es6 Promise, not other Promise implementations. 2. It doesn't work with framework that uses zone.js. zone.js monkey patch the async calls, so it is possible the obj (patched) is comparing to a pre-patched Promise.
</blockquote></details>

Done.

Issue #4130, opened 2020-10-26
tf.grad() no longer works on Sequential models after 2.4.0


kennysong:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): Yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): MacOS 10.15.6
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device: No
- TensorFlow.js installed from (npm or script link): https://cdn.jsdelivr.net/npm/@tensorflow/tfjs@2.4.0/dist/tf.min.js
- TensorFlow.js version (use command below): 2.4.0
- Browser version: Chrome 86.0.4240.80
- Tensorflow.js Converter Version: None

**Describe the current behavior**

`tf.grad()` works correctly on a Sequential model in 2.4.0.

**Describe the expected behavior**

`tf.grad()` fails on the same Sequential model in 2.5.0 and 2.6.0.

(In 2.5.0, I used `tf.env().set('WEBGL_CPU_FORWARD', false);` to [get the code running](https://github.com/tensorflow/tfjs/issues/4029#issuecomment-704832446).)

**Standalone code to reproduce the issue**

This simple example works on 2.4.0 but not on 2.5.0 and 2.6.0.

```
let model = tf.sequential();
model.add(tf.layers.dense({inputShape: [1], units: 1, activation: 'linear'}));

let f = x => model.predict(x);
let g = tf.grad(f);

let t = tf.tensor([1]);
f(t).print();
g(t).print();
```

It fails with this stack trace:

```
Uncaught TypeError: Cannot read property 'shape' of undefined
    at gradFunc (Concat_grad.js:29)
    at Object.s.gradient (engine.js:931)
    at a (tape.js:158)
    at tape.js:136
    at engine.js:1038
    at engine.js:433
    at e.t.scopedRun (engine.js:444)
    at e.t.tidy (engine.js:431)
    at e.t.gradients (engine.js:1033)
    at gradients.js:69
```

lina128:
Fixed by https://github.com/tensorflow/tfjs/pull/4064

tafsiri:
@kennysong note that this is fixed in 2.7.0.

PR #4136, opened 2020-10-26
Fix gradient for min and max

### LINKED ISSUE #4135: grad fails for max
- TensorFlow.js version (use command below): 2.0+

The following (new) test fails

```js
const x = tf.ones([2, 1, 250]);
const axis = 1;
expect(() => {
	const gradients = tf.grad(v => tf.max(v, axis))(x);
    console.log('grad grad', gradients.shape);
}).not.toThrow();
```

Seem to be a regression starting at 2.0.

Error is 

```
Chrome 86.0.4240 (Mac OS X 10.15.6) max cpu {} max gradient: 3D, axis=1 keepDims=true FAILED
	Expected function not to throw, but it threw Error: Error in gradient for op Max. The gradient of input 'x' has shape '1,250,2', which does not match the shape of the input '2,1,250'.
```

tafsiri:
Closes #4135 
Adds a regression test to guard against #4130 (which seems to be fixed at HEAD at the time of that issue being reported).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4136)

### DIFF
--- a/e2e/integration_tests/grad_layers_test.ts
+++ b/e2e/integration_tests/grad_layers_test.ts
@@ -1,0 +1,52 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import '@tensorflow/tfjs-backend-cpu';
+import '@tensorflow/tfjs-backend-webgl';
+
+import * as tfc from '@tensorflow/tfjs-core';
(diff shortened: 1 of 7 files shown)


lina128 (approved):
Thank you Yannick! Can you provide some context here, why we don't need transpose any more?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4136#-:-MKbBc0J6Y7g7dKj6SLE:b-ah9m2e)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan and @lina128)

tafsiri:
Good question. There were a few things I noticed. 

- The grad function would itself try and transpose the output if permutedAxes was not null. However the gradForMinAndMax helper function would also do this as part of the helper function, effectively nullifying each other (though the outer transpose was called without permutedAxes which just seemed odd).
- The min and max grad functions seemed to be written against the eariler definition of the kernel in the forward func. For example looking at how the code [was at 1.7.4](https://github.com/tensorflow/tfjs/blob/tfjs_1.7.4/tfjs-core/src/ops/reduction_ops.ts#L384), you can see that that op would sometimes to an internal transpose before passing the input to the backend implementation, however the original input is what gets passed to the gradient. So I believe the gradient was trying to match the shape of what gets passed to backend.max internally. However in new modular kernels there is no op level transformation and thus no transpose to account for (any internal transposition done within the forward kernel are completely hidden). So the extra 'transposition handling' code becomes unnecessary.

lina128:
Ah, makes sense!

PR #4158, opened 2020-10-29
modularize more node kernels

tafsiri:
Modularize all, any, concat, expm1, mod, pow, reverse,sub, topK, atan2, batchToSpaceND, eluGrad, gatherNd, padV2, scatterNd, spaceToBatchND, tile, transpose.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4158)

### DIFF
--- a/tfjs-core/src/ops/broadcast_to.ts
+++ b/tfjs-core/src/ops/broadcast_to.ts
@@ -27,6 +27,7 @@ import {Rank, ShapeMap, TensorLike} from '../types';
 import {clone} from './clone';
 import {op} from './operation';
 import {reshape} from './reshape';
+import {tile} from './tile';
 
 /**
  * Broadcast an array to a compatible shape NumPy-style.
@@ -80,7 +81,7 @@ function broadcastTo_<R extends Rank>(
     return clone(input) as Tensor<R>;
   }
 
-  const forward = (backend: KernelBackend) => backend.tile(input, reps);
+  const forward = (backend: KernelBackend) => tile(input, reps);
 
   const inputs: BroadcastToInputs = {x: input};
   const attrs: BroadCastToAttrs = {shape, inputShape};
--- a/tfjs-core/src/ops/concat.ts
+++ b/tfjs-core/src/ops/concat.ts
@@ -95,7 +95,7 @@ function concat_<T extends Tensor>(tensors: Array<T|TensorLike>, axis = 0): T {
     // Keep only non-empty tensors (ignore tensors with 0 in their shape).
     $tensors = $tensors.filter(t => t.size > 0);
     if ($tensors.length === 1) {
-      return $tensors[0];
+      return clone($tensors[0]);
     }
 
     const shapes = $tensors.map(t => t.shape);
--- a/tfjs-core/src/ops/concat.ts
+++ b/tfjs-core/src/ops/concat.ts
@@ -23,6 +23,7 @@ import {convertToTensorArray} from '../tensor_util_env';
 import {TensorLike} from '../types';
 import {assert, parseAxisParam, sizeFromShape} from '../util';
 
+import {clone} from './clone';
 import {assertParamsConsistent, computeOutShape} from './concat_util';
 import {op} from './operation';
 import {tensor} from './tensor';
@@ -79,6 +80,10 @@ function concat_<T extends Tensor>(tensors: Array<T|TensorLike>, axis = 0): T {
           with dtype ${tensor.dtype}. `);
       }
     });
+  }
+
+  if ($tensors.length === 1) {
+    return clone($tensors[0]);
   }
 
   const forward: ForwardFunc<Tensor> = (backend, save) => {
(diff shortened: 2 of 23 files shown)


lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4158)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-core/src/ops/concat.ts, line 85 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4158#-MKq6PIc5Egc7kXf1riv:-MKq6PIc5Egc7kXf1riw:boaivy5) ([raw file](https://github.com/tensorflow/tfjs/blob/75bbc4eaab217611b01fbfa94dcf05210aadd827/tfjs-core/src/ops/concat.ts#L85)):*
> ```XML
>   }
> 
>   if ($tensors.length === 1) {
> ```

This transformation should be handled by kernel, right? Also see my comment below.
___
*[tfjs-core/src/ops/concat.ts, line 98 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4158#-MKq6K7aEumw0SShsp2y:-MKq6K7b885QqW9lMh_F:b-x2csui) ([raw file](https://github.com/tensorflow/tfjs/blob/75bbc4eaab217611b01fbfa94dcf05210aadd827/tfjs-core/src/ops/concat.ts#L98)):*
> ```XML
>     $tensors = $tensors.filter(t => t.size > 0);
>     if ($tensors.length === 1) {
>       return $tensors[0];
> ```

Should this be clone($tensors[0]);

tafsiri (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4158)* status: 0 of 1 approvals obtained (waiting on @jinjingforever and @lina128)

___
*[tfjs-core/src/ops/concat.ts, line 85 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4158#-MKq6PIc5Egc7kXf1riv:-MKv5PonAkVmXqUa3W1G:b-bdkyzw) ([raw file](https://github.com/tensorflow/tfjs/blob/75bbc4eaab217611b01fbfa94dcf05210aadd827/tfjs-core/src/ops/concat.ts#L85)):*

This transformation should be handled by kernel, right? Also see my comment below.
</blockquote></details>

So the TF ConcatV2 kernel does not support passing a single length array to the kernel. If you do it throws an error. So that suggested to me that it's op level sugar and kernels do not need to handle it (though it is fine if they do). 

Here is the error

```
Error: Invalid TF_Status: 3
    Message: Value for attr 'N' of 1 must be at least minimum 2
    	; NodeDef: {{node ConcatV2}}; Op<name=ConcatV2; signature=values:N*T, axis:Tidx -> output:T; attr=N:int,min=2; attr=T:type; attr=Tidx:type,default=DT_INT32,allowed=[DT_INT32, DT_INT64]>
```

This has been one nice side effect of modularizing the node kernels, it helps validate our assumptions about the op/kernel boundary.
___
*[tfjs-core/src/ops/concat.ts, line 98 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4158#-MKq6K7aEumw0SShsp2y:-MKv6sdMCWfTaFAI5Zbu:bryzb1g) ([raw file](https://github.com/tensorflow/tfjs/blob/75bbc4eaab217611b01fbfa94dcf05210aadd827/tfjs-core/src/ops/concat.ts#L98)):*

Should this be clone($tensors[0]);
</blockquote></details>

Yes that would probably be good! Done

lina128 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4158)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-core/src/ops/concat.ts, line 85 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4158#-MKq6PIc5Egc7kXf1riv:-MKvHSs83m8fxacS4IXg:bkpwmk3) ([raw file](https://github.com/tensorflow/tfjs/blob/75bbc4eaab217611b01fbfa94dcf05210aadd827/tfjs-core/src/ops/concat.ts#L85)):*

So the TF ConcatV2 kernel does not support passing a single length array to the kernel. If you do it throws an error. So that suggested to me that it's op level sugar and kernels do not need to handle it (though it is fine if they do). 

Here is the error

```
Error: Invalid TF_Status: 3
    Message: Value for attr 'N' of 1 must be at least minimum 2
    	; NodeDef: {{node ConcatV2}}; Op<name=ConcatV2; signature=values:N*T, axis:Tidx -> output:T; attr=N:int,min=2; attr=T:type; attr=Tidx:type,default=DT_INT32,allowed=[DT_INT32, DT_INT64]>
```

This has been one nice side effect of modularizing the node kernels, it helps validate our assumptions about the op/kernel boundary.
</blockquote></details>

Ah, nice. Good to know :)

PR #4215, opened 2020-11-11
Modularize stridedSlice.

lina128:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4215)


tafsiri (approved):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4215)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 465 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4215#-MLs6h6U4KJVW0EgSvI1:-MLs6h6VDLnGXb8Q5NeJ:byr02jz) ([raw file](https://github.com/tensorflow/tfjs/blob/ab4c1430c3556d7e898235b5e5fb9374a4d7f258/tfjs-backend-webgl/src/backend_webgl.ts#L465)):*
> ```XML
>   }
> 
>   bufferSync<R extends Rank>(t: TensorInfo): TensorBuffer<R> {
> ```

Discussion/Food for thought: Now having seen this twice I'm wondering if the right level of abstraction here is having a dataSync on the backend. So that bufferSync can be called on the return value from that (or buffer can just be called directly in the kernel).

That would effectively mirror the structure used on the tensor class which effectively has the same methods with slight differences in code and possibly make it easier to unify that in future. 

This is non blocking (I wouldn't expect anything to change wrt this in this PR) I'm just noticing how similar the code is and thinking about how it might evolve. The other alternative would be to allow calling dataSync, bufferSync on tensorinfo's, i'm imagining that was a constraint that forced the addition of these utility functions?

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4215)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, and @tafsiri)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 465 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4215#-MLs6h6U4KJVW0EgSvI1:-MLsPyg2BXifB-hlPT_N:bsdaekg) ([raw file](https://github.com/tensorflow/tfjs/blob/ab4c1430c3556d7e898235b5e5fb9374a4d7f258/tfjs-backend-webgl/src/backend_webgl.ts#L465)):*

Discussion/Food for thought: Now having seen this twice I'm wondering if the right level of abstraction here is having a dataSync on the backend. So that bufferSync can be called on the return value from that (or buffer can just be called directly in the kernel).

That would effectively mirror the structure used on the tensor class which effectively has the same methods with slight differences in code and possibly make it easier to unify that in future. 

This is non blocking (I wouldn't expect anything to change wrt this in this PR) I'm just noticing how similar the code is and thinking about how it might evolve. The other alternative would be to allow calling dataSync, bufferSync on tensorinfo's, i'm imagining that was a constraint that forced the addition of these utility functions?
</blockquote></details>

Yes, if we move the readSync outside, the rest can be in tfjs-core and shared by any backends. I can make the change in a separate PR.

PR #4228, opened 2020-11-13
webgpu: Add matmul vec4 version

qjia7:
PERF

See 20% improvement for mobilenet model on Intel windows device.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4228)

### DIFF
--- a/tfjs-backend-webgpu/karma.conf.js
+++ b/tfjs-backend-webgpu/karma.conf.js
@@ -69,7 +69,7 @@ module.exports = function(config) {
     customLaunchers: {
       chrome_webgpu: {
         base: 'Chrome',
-        flags: ['--enable-unsafe-webgpu', '--disable-dawn-robustness'],
+        flags: ['--enable-unsafe-webgpu'],
       }
     },
     client: {jasmine: {random: false}, args: args}
--- a/tfjs-backend-webgpu/karma.conf.js
+++ b/tfjs-backend-webgpu/karma.conf.js
@@ -21,14 +21,12 @@ const karmaTypescriptConfig = {
   coverageOptions: {instrumentation: false},
   reports: {},
   bundlerOptions: {
-    transforms: [
-      require('karma-typescript-es6-transform')({
-        presets: [
-          // ensure we get es5 by adding IE 11 as a target
-          ['@babel/env', {'targets': {'ie': '11'}, 'loose': true}]
-        ]
-      })
-    ],
+    transforms: [require('karma-typescript-es6-transform')({
+      presets: [
+        // ensure we get es5 by adding IE 11 as a target
+        ['@babel/env', {'targets': {'ie': '11'}, 'loose': true}]
+      ]
+    })],
     // worker_node_test in tfjs-core contains a conditional require statement
     // that confuses the bundler of karma-typescript.
     ignore: ['./worker_node_test']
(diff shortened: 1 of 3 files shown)


qjia7:
With this change, we see big improvement both for resnet50 and mobilenet models on Intel devices. And webgpu is firstly better than webgl for mobilenet benchmark on Intel device.

tfjs-backend-webgpu/src/kernels/matmul_packed_vec4_webgpu.ts
```diff
@@ -0,0 +1,188 @@
...
+    this.dispatchLayout = {x: [2], y: [1], z: [0]};
+    const vecSize = 4;
+    this.dispatch = computeDispatch(
+        this.dispatchLayout, this.outputShape, this.workGroupSize,
+        [vecSize, workPerThread, 1]);
+
+    this.workPerThread = workPerThread;
+    const tileAOuter = this.workGroupSize[1] * workPerThread;
+    const tileBOuter = this.workGroupSize[0] * vecSize;
+    const tileInner = tileBOuter;  // Make sure tileInner is divided by 4.
```
kainino0x, line 136:
```suggestion
    const tileInner = tileBOuter;  // Make sure tileInner is divisible by 4.
```

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -1141,14 +1142,23 @@ export class WebGPUBackend extends KernelBackend {
...
     // is faster.
     if (env().get('WEBGPU_MATMUL_WORK_PER_THREAD') === 0) {
       program = new MatMulProgram(
           a.shape, output.shape as [number, number, number], transposeA,
           transposeB);
+    } else if (
+        a.shape[2] % 4 === 0 && b.shape[2] % 4 === 0 && !transposeA &&
+        !transposeB) {
+      // TODO: Currently we need to make sure that a.shape[2] and b.shape[2] are
+      // divided by 4 since we use vec4 to get data. In future, we can remove
```
kainino0x, line 1157:
```suggestion
      // divisible by 4 since we use vec4 to get data. In future, we can remove
```

tfjs-backend-webgpu/karma.conf.js
```diff
@@ -71,7 +69,7 @@ module.exports = function(config) {
     customLaunchers: {
       chrome_webgpu: {
         base: 'Chrome',
-        flags: ['--enable-unsafe-webgpu'],
+        flags: ['--enable-unsafe-webgpu', '--disable-dawn-robustness'],
```
kainino0x, line 72: I assume this change wasn't included with the 20% number?

kainino0x (approved):
Only skimmed but overall LGTM. Interesting that vec4 gives a big improvement. Do you know why this is? Does the driver implement a vec4 load differently from four float loads, or could we do four consecutive float loads and get similar results (due to memory access patterns)? Does it generalize to larger chunks (i.e. would there be a benefit from going all the way up to mat4 to load 16 values at a time)?

tfjs-backend-webgpu/karma.conf.js
qjia7: Yes, you are right. Split it into another patch #4238 in case causing confusion. Thanks.

qjia7:
> Only skimmed but overall LGTM. Interesting that vec4 gives a big improvement. Do you know why this is? Does the driver implement a vec4 load differently from four float loads, or could we do four consecutive float loads and get similar results (due to memory access patterns)? Does it generalize to larger chunks (i.e. would there be a benefit from going all the way up to mat4 to load 16 values at a time)?

I think 20% improvement is mainly for Intel GPU. For other GPUs, maybe it only has slightly improvement. I tried to do four consecutive float loads and found it also could improve perf greatly. So I think it hits an unoptimized path for the original matmul algorithm on Intel device. I will file a bug to our driver team simultaneously.

annxingyuan (approved):

PR #4230, opened 2020-11-13
[webgl] Modularize Sign and Floor

mattsoulanille:
Modularize the Sign and Floor kernels of the WebGL backend.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4230)


mattsoulanille:
I'm not quite sure how to implement the packed version of the Floor kernel. I don't see any other modularized kernels checking for `WEBGL_PACK_UNARY_OPERATIONS`. Are we no longer using packed operations, or is this just the first kernel to be modularized that supports them?

Also, I'm not sure if my Floor CPU implementation is correct since I don't think we test CPU in the unit tests. Are they tested in nightly?

PR #4261, opened 2020-11-18
[WebGL] modularize binary kernels

tafsiri:
Modularizes mod, equal, greater, greaterEqual, less, lessEqual, logicalAnd, logicalOr, pow, prelu, floorDiv, eluDer/eluGrad

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4261)


mattsoulanille (approved):
Looks great! I noticed the times you weren't able to use binaryKernelFunc were due to the input variables not being named {a, b}. If there are lots more kernels that have differently named arguments, we might want to refactor binaryKernelFunc or factor some of its logic out into other helper methods, but I think it's fine as-is for now. 

LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4261#-:-MMS8oUD6jbarty9GhYB:b-ihmx62)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan and @mattsoulanille)

tafsiri:
Yep I checked and it looks like its just the two that are included in this PR. Only two more are left to do, which I'll do after Na modularizes their corresponding cpu kernels.

PR #4265, opened 2020-11-19
Modularize 8 more kernels.

lina128:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4265)


tafsiri (changes requested):
Overall looks good, one change requested to revert back to existing behaviour.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4265#-:-MMTaeL7Dddh9ns7zBm5:b5nsgls)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, @mattsoulanille, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/Multinomial.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4265#-MMTZmsv9-fOwoiIcZin:-MMTZmsw5Io10XXtt2bU:b-h63ju5) ([raw file](https://github.com/tensorflow/tfjs/blob/e34ab8867e9d17725d5aeed92317ae0703ede312/tfjs-backend-cpu/src/kernels/Multinomial.ts#L36)):*
> ```XML
>   assertNotComplex(logits, 'multinomial');
> 
>   const $seed = seed || Math.random();
> ```

see comment below about seed at the op level. Might be worth leaving this line out until we have that sorted and tests in place.
___
*[tfjs-core/src/ops/multinomial.ts, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4265#-MMTY0Dj7SVAQjtuY8gl:-MMTY0Dk8yQKDlIDtrKk:b2tegaf) ([raw file](https://github.com/tensorflow/tfjs/blob/e34ab8867e9d17725d5aeed92317ae0703ede312/tfjs-core/src/ops/multinomial.ts#L69)):*
> ```XML
> 
>   const forward: ForwardFunc<Tensor> = (backend) => {
>     seed = seed || Math.random();
> ```

Let's not move this to the forward func in this PR (though I do understand why you made this change)

A few reasons
- It turns out we don't have test for when no seed is passed. I looked at the node implementation and a NaN would be produced in undefined is passed through (webgl backend would also break, though tests wouldn't catch it). We should add some tests that don't necessarily care what the value is but that there is no error.
- I think the way we set this (since it was originally written) actually has a bug. A user cannot manually set the seed to 0.
- The correct default value if we are trying to match tf) for this may be 0 (at least according to my quick skim of ops.pbtxt), this may need further investigation as that may be a placeholder.

Fixing 3 and 4 might technically be breaking/change behavior in certain edge cases (even though it would be a bug fix). Probably worth separating this into a diff pr that we can fully investigate.

mattsoulanille (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4265)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, @mattsoulanille, and @tafsiri)

___
*[tfjs-core/src/ops/multinomial.ts, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4265#-MMTY0Dj7SVAQjtuY8gl:-MMVttUq5HH00aTJ-3DB:b-sg2dlh) ([raw file](https://github.com/tensorflow/tfjs/blob/e34ab8867e9d17725d5aeed92317ae0703ede312/tfjs-core/src/ops/multinomial.ts#L69)):*

Let's not move this to the forward func in this PR (though I do understand why you made this change)

A few reasons
- It turns out we don't have test for when no seed is passed. I looked at the node implementation and a NaN would be produced in undefined is passed through (webgl backend would also break, though tests wouldn't catch it). We should add some tests that don't necessarily care what the value is but that there is no error.
- I think the way we set this (since it was originally written) actually has a bug. A user cannot manually set the seed to 0.
- The correct default value if we are trying to match tf) for this may be 0 (at least according to my quick skim of ops.pbtxt), this may need further investigation as that may be a placeholder.

Fixing 3 and 4 might technically be breaking/change behavior in certain edge cases (even though it would be a bug fix). Probably worth separating this into a diff pr that we can fully investigate.
</blockquote></details>

FYI for point 2 you can use nullish coalescing: `seed = seed ?? Math.random();`

mattsoulanille (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4265)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @lina128, and @tafsiri)

___
*[tfjs-core/src/ops/multinomial.ts, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4265#-MMTY0Dj7SVAQjtuY8gl:-MMVuyjs0Q-XAY06keng:b-atvijh) ([raw file](https://github.com/tensorflow/tfjs/blob/e34ab8867e9d17725d5aeed92317ae0703ede312/tfjs-core/src/ops/multinomial.ts#L69)):*

FYI for point 2 you can use nullish coalescing: `seed = seed ?? Math.random();`
</blockquote></details>

Didn't mean to approve on the last comment. Oops!

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4265)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, and @tafsiri)

___
*[tfjs-backend-cpu/src/kernels/Multinomial.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4265#-MMTZmsv9-fOwoiIcZin:-MMWQiLy9Xocx02o7Sz7:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/e34ab8867e9d17725d5aeed92317ae0703ede312/tfjs-backend-cpu/src/kernels/Multinomial.ts#L36)):*

see comment below about seed at the op level. Might be worth leaving this line out until we have that sorted and tests in place.
</blockquote></details>

Done.
___
*[tfjs-core/src/ops/multinomial.ts, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4265#-MMTY0Dj7SVAQjtuY8gl:-MMWUvkd1yzD3hCmuqBi:b-euv89a) ([raw file](https://github.com/tensorflow/tfjs/blob/e34ab8867e9d17725d5aeed92317ae0703ede312/tfjs-core/src/ops/multinomial.ts#L69)):*

Didn't mean to approve on the last comment. Oops!
</blockquote></details>

@tafsiri Got it, reverted to previous code, added TODO to investigate as you stated. @mattsoulanille nullish coalescing and optional chaining are ES2020 specification, I think we need to change tsconfig target to es2020 to allow them to compile properly, which is beyond the scope of this PR.

tafsiri (approved):
LGTM. Thanks

> Reviewed 3 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4265#-:-MMWs7sE8cl3w7Oxg4Fq:b-sglfl8)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @mattsoulanille, and @tafsiri)

mattsoulanille (approved):
> Reviewed 11 of 15 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4265)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @mattsoulanille, and @tafsiri)

PR #4268, opened 2020-11-19
[webgl] modularize kernels

tafsiri:
Modularize pad, batchToSpaceND, spacetoBatchND, LRN, LRNGrad.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4268)


mattsoulanille (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4268)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @lina128, @mattsoulanille, and @tafsiri)

___
*[tfjs-backend-webgl/src/kernels/BatchToSpaceND.ts, line 79 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4268#-MMW0PvC57UyeNSPqGEr:-MMW0PvC57UyeNSPqGEs:b9pfsel) ([raw file](https://github.com/tensorflow/tfjs/blob/b97115fd7a4bf9c5438155cfc5a8c05a7e92ffbe/tfjs-backend-webgl/src/kernels/BatchToSpaceND.ts#L79)):*
> ```XML
>   kernelName: BatchToSpaceND,
>   backendName: 'webgl',
>   kernelFunc: batchToSpaceND as {} as KernelFunc
> ```

These unsafe casts to KernelFunc might not be necessary in all cases if we can declare the function as a KernelFunc in its definition: `export const batchToSpaceND: KernelFunc = (...`. I haven't tried it with this kernel, but it works for the `Real.ts` kernel, so I'm guessing it probably generalizes.

LGTM modulo the above nit.

tafsiri (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4268)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan and @lina128)

___
*[tfjs-backend-webgl/src/kernels/BatchToSpaceND.ts, line 79 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4268#-MMW0PvC57UyeNSPqGEr:-MMW43QL086Rl5icDrat:b-3btx04) ([raw file](https://github.com/tensorflow/tfjs/blob/b97115fd7a4bf9c5438155cfc5a8c05a7e92ffbe/tfjs-backend-webgl/src/kernels/BatchToSpaceND.ts#L79)):*

These unsafe casts to KernelFunc might not be necessary in all cases if we can declare the function as a KernelFunc in its definition: `export const batchToSpaceND: KernelFunc = (...`. I haven't tried it with this kernel, but it works for the `Real.ts` kernel, so I'm guessing it probably generalizes.

LGTM modulo the above nit.
</blockquote></details>

I think we've found in the past that it produces a different set of casts in the kernel func definition itself so went with this convention. I'm not sure what you mean by it works in the Real.ts kernel, i'm looking at that file and not seeing what you are describing.

However one thing your comment did make me notice is that I shouldn't use a const variable for this function. We should prefer named functions where possible and in this case it is quite possible so I'll make that change.

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4268)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan and @lina128)

___
*[tfjs-backend-webgl/src/kernels/BatchToSpaceND.ts, line 79 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4268#-MMW0PvC57UyeNSPqGEr:-MMWAey11zB_fs4FGK89:b-yvrgpa) ([raw file](https://github.com/tensorflow/tfjs/blob/b97115fd7a4bf9c5438155cfc5a8c05a7e92ffbe/tfjs-backend-webgl/src/kernels/BatchToSpaceND.ts#L79)):*

I think we've found in the past that it produces a different set of casts in the kernel func definition itself so went with this convention. I'm not sure what you mean by it works in the Real.ts kernel, i'm looking at that file and not seeing what you are describing.

However one thing your comment did make me notice is that I shouldn't use a const variable for this function. We should prefer named functions where possible and in this case it is quite possible so I'll make that change. 
</blockquote></details>

By "it works in the `Real.ts` kernel," I meant that, as a local test, I refactored that kernel to conform with my comment. Now that I reread my comment, I realize it reads as if I'm saying `Real.ts` already conforms to my recommended changes. My bad! 

Good point about casting directly to KernelFunc. I forgot we may need to use those functions outside of the KernelConfig, and if they're already cast, we don't know their type. 

We can probably remove the casts in the KernelConfig definition, though, since batchToSpaceND (and other kernel implementations) should be a subtype of KernelFunc (I think).

lina128 (approved):
Thank you Yannick. LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4268#-:-MMWaghVEnZuAKUiFlFH:blg4mxf)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @annxingyuan and @lina128)

PR #4272, opened 2020-11-19
Run ts-node tests with --transpile-only to increase speed

mattsoulanille:
This change nearly halves the CI run time and also significantly reduces local test time. It's safe to run the tests in transpile-only mode since the source that is being tested has already been typechecked when it was compiled. **The change may make it harder to find type errors in the test code, which is definitely a trade off to consider before merging.**

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272)


lina128 (approved):
Cutting by half is a great improvement, dev velocity is definitely worth the tradeoff imho. Thank you Matt! Wait for Yannick and Ping's comments too.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272#-:-MMWZ8SS3aHNkQXEITnL:b-m65wln)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @pyu10055, and @tafsiri)

pyu10055 (changes requested):
Thank you!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272#-:-MMWbC-S6P314on_MwD9:b-9755fn)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @mattsoulanille, @pyu10055, and @tafsiri)

___
*[tfjs-data/package.json, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWaplu83BXbcYHT4o3:-MMWaplu83BXbcYHT4o4:bcb3yu7) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs-data/package.json#L61)):*
> ```JSON
>     "publish-local": "rimraf dist/ && yarn build-npm && yalc push",
>     "publish-npm": "npm publish",
>     "test": "yarn && yarn build-deps && ts-node --transpile-only --project tsconfig.test.json src/test_node.ts",
> ```

here since it is building the deps, it would be good to add `yarn build` to ensure the package is also built. This should apply to other packages as well.

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @mattsoulanille and @tafsiri)

___
*[tfjs-data/package.json, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWaplu83BXbcYHT4o3:-MMWcg2m2_UzbmeX5pvJ:b-ukdx75) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs-data/package.json#L61)):*

here since it is building the deps, it would be good to add `yarn build` to ensure the package is also built. This should apply to other packages as well. 
</blockquote></details>

The point is we want to be consistent across packages on what are the steps for running the test locally? yarn build && yarn test or something else.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055 and @tafsiri)

___
*[tfjs-data/package.json, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWaplu83BXbcYHT4o3:-MMWlmmLBdag7DNhEcCg:b-lyh9h7) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs-data/package.json#L61)):*

The point is we want to be consistent across packages on what are the steps for running the test locally? yarn build && yarn test or something else.
</blockquote></details>

We leave the build step out to make local testing as light as possible. But if leaving build step out will make it harder to find type error in the package to test, then we should add building step.

mattsoulanille:
Good point! I've added `yarn build` to all the tests that run with ts-node.

tafsiri (changes requested):
> Reviewed 5 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @mattsoulanille, @pyu10055, and @tafsiri)

___
*[package.json, line 29 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWxr5z8HEFaRoOOruy:-MMWxr5z8HEFaRoOOruz:b-qczmx6) ([raw file](https://github.com/tensorflow/tfjs/blob/e6a316567d9d75cf0ba54d6a6dd5e49f147bc407/package.json#L29)):*
> ```JSON
>     "publish-npm": "ts-node -s ./scripts/publish-npm.ts",
>     "release-notes": "ts-node -s ./scripts/release_notes/release_notes.ts",
>     "test-release-notes": "ts-node --transpile-only -s ./scripts/release_notes/run_tests.ts",
> ```

same comment as elsewhere about this being the only place this might get typed checked as it is not part of src.
___
*[tfjs/integration_tests/validate_converter.sh, line 146 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWx8N_43ZurGlZUupW:-MMWx8Na6Q-V2Xhna32F:b-cvgwzb) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs/integration_tests/validate_converter.sh#L146)):*
> ```Shell
> 
>   echo "Starting validation karma tests in Node.js..."
>   yarn ts-node --transpile-only run_node_tests.ts \
> ```

is run_node_tests.ts built by tsc/yarn build. If not then i think this is the only opportunity for type checking these files. 
___
*[tfjs-data/package.json, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWaplu83BXbcYHT4o3:-MMWueVA7q_ShI9HoBrN:b-3ojb1t) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs-data/package.json#L61)):*

We leave the build step out to make local testing as light as possible. But if leaving build step out will make it harder to find type error in the package to test, then we should add building step.
</blockquote></details>

If we add the build step here would we not lose the speed benefits of transpile-only?  (i.e. is `tsc + ts-node --transpile-only ...` equivalent to `ts-node ...`) . Are these local builds faster than they were before now that you have added yarn build? Could you add some timings to the PR description?

One approach we could do if we only use transpile-only in test-ci, since we know the package would have already been built, and locally leave things as is (somewhat dependent on the answer to the question above?
___
*[tfjs-node/integration/typescript/package.json, line 8 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWwecG3zXX3ut8Rasz:-MMWwecG3zXX3ut8Rat-:bgq7sr0) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs-node/integration/typescript/package.json#L8)):*
> ```JSON
>   "scripts": {
>     "prep": "yalc link @tensorflow/tfjs-node && cd node_modules/@tensorflow/tfjs-node && yarn clean-deps && yarn",
>     "test": "ts-node --transpile-only -P tsconfig.test.json src/test.ts && tsc && node dist/test.js"
> ```

Is this whole folder obsolete? It doesn't seem to have srcs anymore. I can't see how yarn test would work? (not exactly related to this PR)

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @tafsiri)

___
*[tfjs-data/package.json, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWaplu83BXbcYHT4o3:-MMWyveoBadETXZEHcKF:bcuytpv) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs-data/package.json#L61)):*

If we add the build step here would we not lose the speed benefits of transpile-only?  (i.e. is `tsc + ts-node --transpile-only ...` equivalent to `ts-node ...`) . Are these local builds faster than they were before now that you have added yarn build? Could you add some timings to the PR description?

One approach we could do if we only use transpile-only in test-ci, since we know the package would have already been built, and locally leave things as is (somewhat dependent on the answer to the question above?
</blockquote></details>

agree, we should definitely add --transpile-only to ci targets. for local test, the 'test' target is currently already including everything, it is not consistent to omit build step.
If we want to make test lightweight, we should another target, 'test-dev' does nothing but ts-node --transplie-only

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @mattsoulanille, @pyu10055, and @tafsiri)

___
*[package.json, line 29 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWxr5z8HEFaRoOOruy:-MMX1-Lu-FU75IWq2DR_:b-fbntwd) ([raw file](https://github.com/tensorflow/tfjs/blob/e6a316567d9d75cf0ba54d6a6dd5e49f147bc407/package.json#L29)):*

same comment as elsewhere about this being the only place this might get typed checked as it is not part of src.
</blockquote></details>

Nice catch. On another hand, do we care whether these tests are typechecked or not?
___
*[tfjs-data/package.json, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWaplu83BXbcYHT4o3:-MMX1j2r40m9hrnT9PYO:bjd4ejc) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs-data/package.json#L61)):*

agree, we should definitely add --transpile-only to ci targets. for local test, the 'test' target is currently already including everything, it is not consistent to omit build step.
If we want to make test lightweight, we should another target, 'test-dev' does nothing but ts-node --transplie-only 
</blockquote></details>

Ah, great point!!! Agree with both Yannick and Ping, use transpile-only for ci and test-dev. Please add test-dev to all packages for consistency.

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055 and @tafsiri)

___
*[package.json, line 29 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWxr5z8HEFaRoOOruy:-MMX5BDP0Zdq-YFLztiN:b-7356jj) ([raw file](https://github.com/tensorflow/tfjs/blob/e6a316567d9d75cf0ba54d6a6dd5e49f147bc407/package.json#L29)):*

Nice catch. On another hand, do we care whether these tests are typechecked or not?
</blockquote></details>

It takes 1.5 seconds with typechecking, so I've added it back in. I should mention that there are a few other cases where the test launchers won't get typechecked if we go the transpile-only route.  Anything with a `run_tests.ts` file at the root won't have that file typechecked:

./scripts/release_notes/run_tests.ts
./tfjs/run_tools_tests.ts
./tfjs/integration_tests/run_node_tests.ts
./tfjs-converter/run_tests.ts
./tfjs-backend-cpu/run_tests.ts


It might be better to compile `run_tests.ts` against the outputs of `yarn build` with tsc and run it with node. That way, when it gets typechecked, we don't re-typecheck the entire project. I can look into this.
___
*[tfjs/integration_tests/validate_converter.sh, line 146 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWx8N_43ZurGlZUupW:-MMX6KgcANXg0BjrylJR:blix6gz) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs/integration_tests/validate_converter.sh#L146)):*

is run_node_tests.ts built by tsc/yarn build. If not then i think this is the only opportunity for type checking these files. 
</blockquote></details>

I don't think it is built by `yarn build`.See my other comment about compiling tests with tsc as a possible solution.
___
*[tfjs-data/package.json, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWaplu83BXbcYHT4o3:-MMX7YamByO_4ZXiO4Vq:b-rq2xo1) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs-data/package.json#L61)):*

Ah, great point!!! Agree with both Yannick and Ping, use transpile-only for ci and test-dev. Please add test-dev to all packages for consistency.
</blockquote></details>

@tafsiri Running `yarn build` with transpile-only after building the package in a previous step is actually a bit faster due to incremental typescript compilation, which ts-node can't take advantage of. I get 0m52.220s when running `yarn build` and testing with `transpile-only`, and I get 1m7.020s without `yarn build` and without `transpile-only`. 

It would definitely still be better not to run `yarn build` again in CI like you mentioned, since `yarn build` also runs rollup, which I haven't figured out how to make incremental yet (although Bazel could help with that). I'll add a `test-ci` rule to each package as mentioned above.
___
*[tfjs-node/integration/typescript/package.json, line 8 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWwecG3zXX3ut8Rasz:-MMXEadH3wO5zz4S0uIS:b-573a89) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs-node/integration/typescript/package.json#L8)):*

Is this whole folder obsolete? It doesn't seem to have srcs anymore. I can't see how yarn test would work? (not exactly related to this PR)
</blockquote></details>

I'm pretty sure it is. I was doing a find-replace to find these instances, and I didn't check this one's files. I'll remove it.

pyu10055 (approved):
> Reviewed 13 of 16 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @pyu10055 and @tafsiri)

lina128 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @pyu10055 and @tafsiri)

___
*[package.json, line 29 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWxr5z8HEFaRoOOruy:-MMXPqyOB07AuywFlo8B:bx3q2yu) ([raw file](https://github.com/tensorflow/tfjs/blob/e6a316567d9d75cf0ba54d6a6dd5e49f147bc407/package.json#L29)):*

It takes 1.5 seconds with typechecking, so I've added it back in. I should mention that there are a few other cases where the test launchers won't get typechecked if we go the transpile-only route.  Anything with a `run_tests.ts` file at the root won't have that file typechecked:

./scripts/release_notes/run_tests.ts
./tfjs/run_tools_tests.ts
./tfjs/integration_tests/run_node_tests.ts
./tfjs-converter/run_tests.ts
./tfjs-backend-cpu/run_tests.ts


It might be better to compile `run_tests.ts` against the outputs of `yarn build` with tsc and run it with node. That way, when it gets typechecked, we don't re-typecheck the entire project. I can look into this.
</blockquote></details>

This is a great point. Should we just move run_tests to src/

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @pyu10055 and @tafsiri)

___
*[package.json, line 29 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWxr5z8HEFaRoOOruy:-MMXYAaW6Uzv7kP0xiIt:b-up0rcl) ([raw file](https://github.com/tensorflow/tfjs/blob/e6a316567d9d75cf0ba54d6a6dd5e49f147bc407/package.json#L29)):*

This is a great point. Should we just move run_tests to src/
</blockquote></details>

I've been working on that, but I'm running into problems with running the build outputs with node. Everything gets compiled into .js files that use import syntax, which node doesn't really support, instead of require syntax.

mattsoulanille (commented):
This is a bit stale, but I think it should still be merged. I've solved the typechecking issue by first running tsc to check types and then running ts-node in transpile-only mode. There were a few cases where there wasn't an easy way to compile with tsc (some .ts files were in the /scripts directory and I didn't think it would be good to move them to /src), so I'm running those with typechecking enabled in ts-node. The most expensive .ts files were the test files, and they were all moved to /src of their respective packages.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272#-:-MNig8VE9i2jUfQ6Dj0x:b-9a49ls)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @pyu10055 and @tafsiri)

___
*[package.json, line 29 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWxr5z8HEFaRoOOruy:-MNe_4WP6lM09frqF8j0:b-wjrb3) ([raw file](https://github.com/tensorflow/tfjs/blob/e6a316567d9d75cf0ba54d6a6dd5e49f147bc407/package.json#L29)):*

I've been working on that, but I'm running into problems with running the build outputs with node. Everything gets compiled into .js files that use import syntax, which node doesn't really support, instead of require syntax.
</blockquote></details>

I was initially trying to run the compiled outputs directly with node, which didn't work since node doesn't support esm yet. I was also trying to create a custom bundle for tests using webpack, but it's much easier to just run ts-node on the typescript files. I've moved the test files to /src and made sure to only run ts-node in transpile mode if the sources it's running against have recently been compiled (each test that runs ts-node in transpile mode now runs tsc beforehand).
___
*[tfjs-data/package.json, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWaplu83BXbcYHT4o3:-MNig0949zdyEBbrIlpQ:b-livubv) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs-data/package.json#L61)):*

@tafsiri Running `yarn build` with transpile-only after building the package in a previous step is actually a bit faster due to incremental typescript compilation, which ts-node can't take advantage of. I get 0m52.220s when running `yarn build` and testing with `transpile-only`, and I get 1m7.020s without `yarn build` and without `transpile-only`. 

It would definitely still be better not to run `yarn build` again in CI like you mentioned, since `yarn build` also runs rollup, which I haven't figured out how to make incremental yet (although Bazel could help with that). I'll add a `test-ci` rule to each package as mentioned above.
</blockquote></details>

I've added test-dev to all packages.

tafsiri (approved):
I think i found one are where the test scripts are never typechecked, PTAL. 

Generally I am of the opinion if we have a ts file it should be run through the typechecker at some point in time.

> Reviewed 2 of 16 files at r1, 21 of 21 files at r4.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272#-:-MO2We8q6sz2DgL4RF4A:bhp88n4)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @mattsoulanille, @pyu10055, and @tafsiri)

___
*[tfjs/integration_tests/validate_converter.sh, line 146 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWx8N_43ZurGlZUupW:-MO2VCTw1ncbwkLZxK_L:bsyijfe) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs/integration_tests/validate_converter.sh#L146)):*

I don't think it is built by `yarn build`.See my other comment about compiling tests with tsc as a possible solution.
</blockquote></details>

We should then add a `yarn build` step somewhere in the test scripts for these intrgration tests, else we never typecheck these files. (Or I would not use transpile-only here).
___
*[tfjs-core/package.json, line 82 at r4](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MO2UAl41NutIORK5GmK:-MO2UAl41NutIORK5GmL:b9oprkx) ([raw file](https://github.com/tensorflow/tfjs/blob/67e3d830c87309529486e32951fb202921b79b0b/tfjs-core/package.json#L82)):*
> ```JSON
>     "test-async-backends": "rimraf dist/ && yarn build && ts-node --transpile-only -P tsconfig.test.json dist/test_async_backends.js",
>     "test-async-backends-ci": "ts-node --transpile-only -P tsconfig.test.json dist/test_async_backends.js",
>     "test-snippets": "yarn build && yarn build-cpu-backend && ts-node --transpile-only -P tsconfig.test.json ./scripts/test_snippets/test_snippets.ts"
> ```

this `test-snippets` is different from the others (they just use ts-node).

mattsoulanille (commented):
Thanks for catching those cases! I completely agree that all .ts files should be typechecked at some point.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4272#-:-MO2_Mqd1JXNcSypN-JT:b-o6y9ic)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @pyu10055 and @tafsiri)

___
*[tfjs/integration_tests/validate_converter.sh, line 146 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MMWx8N_43ZurGlZUupW:-MO2_BXd9WkzjR88h6Za:b-ypgleh) ([raw file](https://github.com/tensorflow/tfjs/blob/8f79cfdc75e69205802d7ffc26957a903cfb811f/tfjs/integration_tests/validate_converter.sh#L146)):*

We should then add a `yarn build` step somewhere in the test scripts for these intrgration tests, else we never typecheck these files. (Or I would not use transpile-only here).
</blockquote></details>

Nice catch! I missed this one when removing `--transpile-only` from targets that don't get built with tsc.
___
*[tfjs-core/package.json, line 82 at r4](https://reviewable.io/reviews/tensorflow/tfjs/4272#-MO2UAl41NutIORK5GmK:-MO2ZyYT-oKeanoQ2-QA:bk3g6ly) ([raw file](https://github.com/tensorflow/tfjs/blob/67e3d830c87309529486e32951fb202921b79b0b/tfjs-core/package.json#L82)):*

this `test-snippets` is different from the others (they just use ts-node). 
</blockquote></details>

Good catch! Switched it back to not using --transpile-only.

PR #4275, opened 2020-11-19
Bug fixes for local benchmark tool

jinjingforever:
- Correctly show WASM_HAS_MULTITHREAD_SUPPORT flag value, and make it "tunable". Also sort all flags to make them easier to read.
  Before: https://screenshot.googleplex.com/5iyTQriVUfeGjAg
  After: https://screenshot.googleplex.com/A57sSAuoBvxogJs

- Currently, when benchmarking a built-in model, it will first ask users to replace "-1" in the input tensor (https://screenshot.googleplex.com/5eTX73hkUBW7YNT). Then if user clicks "run benchmark" button *again*, it will start the benchmark. This is kind of broken, because users cannot actually edit the input tensor field for built-in models. This PR fixes this logic by only prompting about replacing "-1" in input tensor when benchmarking a **custom model**. It will also make benchmark to run after user clicks the button the first time.

- The following steps are broken in the current tool.
    1. After page is loaded, start a benchmark with WASM backend with simd checked.
    1. After benchmark is done, uncheck the simd option, and run benchmark again.
    1. An error message is shown: https://screenshot.googleplex.com/ArQe6p97n3aAGyq
      (This also happens in webgl backend when users toggle flags in different runs) <br></br>

  The root cause is: the model loading logic only happens when users change the model, NOT when users change flags. In the example above, when user unchecks the simd option, a new WASM backend will be registered, but the tensors allocated during tf.loadGraphModel (when model was loaded) will still [point to](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/engine.ts#L1136) the *old* WASM backend which is now disposed (and the wasm field is [set to null](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/backend_wasm.ts#L130)).
  
  This PR fixes this by re-loading the model when the flags are changed.
  After the fix: https://screencast.googleplex.com/cast/NTc1NjUzOTkyNTE2ODEyOHw0Njc1ZmQyMi0yMg 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4275)


pyu10055 (changes requested):
Thanks for fixing the bugs!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4275#-:-MMXJz-uDbNYQju-joj-:b-vxnkvr)* status: 0 of 1 approvals obtained (waiting on @jinjingforever)

___
*[e2e/benchmarks/local-benchmark/index.html, line 500 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4275#-MMXJTzC8Tg4BxWiYPih:-MMXJTzC8Tg4BxWiYPii:b-gznkot) ([raw file](https://github.com/tensorflow/tfjs/blob/eac9213b2fc392658852cca0a3df0898e5a4fe33/e2e/benchmarks/local-benchmark/index.html#L500)):*
> ```HTML
> 
>     async function loadModel() {
>       if (state.isModelChanged || !state.isModelLoaded || state.isFlagChanged) {
> ```

mostly of time flag changes does not require a model reload, this flag cause a backend change, maybe track this with a new state flag (i.e. isBackendChanged)?

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4275)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

___
*[e2e/benchmarks/local-benchmark/index.html, line 500 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4275#-MMXJTzC8Tg4BxWiYPih:-MMXfH6F-iWE8eCESp47:bfvgz15) ([raw file](https://github.com/tensorflow/tfjs/blob/eac9213b2fc392658852cca0a3df0898e5a4fe33/e2e/benchmarks/local-benchmark/index.html#L500)):*

mostly of time flag changes does not require a model reload, this flag cause a backend change, maybe track this with a new state flag (i.e. isBackendChanged)? 
</blockquote></details>

Thanks Ping!

Actually I just checked. Currently, we have 7 tunable flags. 2 for wasm (https://screenshot.googleplex.com/5VZpgShZBUnmM8o) and 5 for webgl (https://screenshot.googleplex.com/AMTwQgtRrWvxZwX). Any changes among these flags (**without changing backend**) will require a model reload, or errors will be thrown. For example, here is what happens if I run a benchmark using webgl, uncheck "cpu-forwarding", and run benchmark again: https://screenshot.googleplex.com/RdozZ3syDewJrD9. So the "stat.isFlagChanged" check I added should fix these cases.

On the other hand, if user changes backend (e.g. from wasm to webgl), we don't actually need to reload model. It will just work.

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4275)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever)

___
*[e2e/benchmarks/local-benchmark/index.html, line 500 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4275#-MMXJTzC8Tg4BxWiYPih:-MMadvHY8t9msavxoPZe:b-7a0120) ([raw file](https://github.com/tensorflow/tfjs/blob/eac9213b2fc392658852cca0a3df0898e5a4fe33/e2e/benchmarks/local-benchmark/index.html#L500)):*

Thanks Ping!

Actually I just checked. Currently, we have 7 tunable flags. 2 for wasm (https://screenshot.googleplex.com/5VZpgShZBUnmM8o) and 5 for webgl (https://screenshot.googleplex.com/AMTwQgtRrWvxZwX). Any changes among these flags (**without changing backend**) will require a model reload, or errors will be thrown. For example, here is what happens if I run a benchmark using webgl, uncheck "cpu-forwarding", and run benchmark again: https://screenshot.googleplex.com/RdozZ3syDewJrD9. So the "stat.isFlagChanged" check I added should fix these cases.

On the other hand, if user changes backend (e.g. from wasm to webgl), we don't actually need to reload model. It will just work.


</blockquote></details>

cool, thanks for checking.

pyu10055 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4275)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever)

PR #4277, opened 2020-11-19
Modularize expandDims kernel.

lina128:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4277)

### DIFF
--- a/tfjs-core/src/ops/expand_dims_test.ts
+++ b/tfjs-core/src/ops/expand_dims_test.ts
@@ -34,7 +34,7 @@ describeWithFlags('expandDims', ALL_ENVS, () => {
   it('1d, axis=-3', () => {
     expect(() => {
       tf.tensor1d([1, 2, 3]).expandDims(-3);
-    }).toThrowError('Axis must be in the interval [-2, 1]');
+    }).toThrowError();
   });
 
   it('1d, axis=-2', async () => {
@@ -64,7 +64,7 @@ describeWithFlags('expandDims', ALL_ENVS, () => {
   it('2d, axis=-4', () => {
     expect(() => {
       tf.tensor2d([[1, 2], [3, 4], [5, 6]]).expandDims(-4 /* axis */);
-    }).toThrowError('Axis must be in the interval [-3, 2]');
+    }).toThrowError();
   });
 
   it('2d, axis=-3', async () => {
--- a/tfjs-layers/src/models_test.ts
+++ b/tfjs-layers/src/models_test.ts
@@ -842,53 +842,51 @@ describeMathCPU('loadLayersModel from URL', () => {
   });
 
   it('loadLayersModel: no memory leak with nested LayersModel layer',
-    async() => {
-      const modelTopology =
-          JSON.parse(JSON.stringify(fakeSequentialModelWithNestedContainer))
-              .modelTopology;
+     async () => {
+       const modelTopology =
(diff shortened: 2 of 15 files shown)


tfjs-layers/src/models_test.ts
```diff
@@ -842,53 +842,51 @@ describeMathCPU('loadLayersModel from URL', () => {
   });
 
   it('loadLayersModel: no memory leak with nested LayersModel layer',
-    async() => {
```
tafsiri, line 845: Is this related to this change? Curious why this test had to change for this PR?

tafsiri (commented):
Just a couple of high level questions before doing a full review, two are inline, but i'm also just curious if there was any particular reason to implement ExpandDims as a full kernel. Nothing wrong with that though and this overall looks good, just curious as to whether there was something else motivating this.

tfjs-core/src/ops/expand_dims_test.ts
```diff
@@ -34,7 +34,7 @@ describeWithFlags('expandDims', ALL_ENVS, () => {
   it('1d, axis=-3', () => {
     expect(() => {
       tf.tensor1d([1, 2, 3]).expandDims(-3);
-    }).toThrowError('Axis must be in the interval [-2, 1]');
+    }).toThrowError();
```
tafsiri, line 37: Looks like GH didn't save my comment here. Was also curious about this change to the test? I'm guessing its because node throws a different error but just want to double check. Or maybe node does not thrown any error?

lina128 (commented):
Hi Yannick, when I tried to modularize unsortedSegmentSum, it uses tf.expandDims and tf.stack. Both are op-level implementation. I checked these two in TF, both are implemented as kernels. So I think we should implement them as kernels too.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4277#-:-MMaezcUCu2oXhQW9fE-:b-8jdhvh)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, @mattsoulanille, and @tafsiri)

___
*[tfjs-core/src/ops/expand_dims_test.ts, line 37 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4277#-MMXh6VAScJPD81K7o99-r1-37:-MMag1L05hWi6QrVuydE:b785ajt) ([raw file](https://github.com/tensorflow/tfjs/blob/331766f3713b560c5968947d4795ea35c42d25a6/tfjs-core/src/ops/expand_dims_test.ts#L37)):*

Looks like GH didn't save my comment here. Was also curious about this change to the test? I'm guessing its because node throws a different error but just want to double check. Or maybe node does not thrown any error?
</blockquote></details>

Node throws a similar error, but in different language. This is actually another benefit I see moving the implementation to kernel, which allowed us to validate the result with TF through Node :)

tfjs-layers/src/models_test.ts
lina128: When I ran layers test in local, it gives me the result of 120.99999999, so I change it to closeTo.

tafsiri (approved):
Thanks Na, great stuff. I asked about the layers test because the diff appeared to be pretty big, probably reformatting.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4277#-:-MMbXKAO6SiB1cw4d_lM:bjuy6o0)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @jinjingforever, and @mattsoulanille)

PR #4285, opened 2020-11-20
Quantization fix

patlevin:
## Motivation

The Python converter's de-quantisation function doesn't match the tfjs converter and causes data loss:

```python
import tensorflowjs as tfjs
import numpy as np

d = np.array([-0.2345, 0.0468, 0.3456, 0.], dtype=np.float32)
q, metadata = tfjs.quantization.quantize_weights(d, np.uint16)
de_q = tfjs.quantization.dequantize_weights(q, metadata)
np.testing.assert_array_almost_equal(de_q, d)                           # this will fail
```

The tfjs de-quantisation works properly and the problem only exists with the python converter.

## Explanation

The de-quantisation rounds the weights to integers before converting them to the target `dtype` (e.g. `float32`), see [quantization.py line 160](https://github.com/tensorflow/tfjs/blob/17d79c13b7784409623b75f7bdcfb499a8e6ffd0/tfjs-converter/python/tensorflowjs/quantization.py#L160).

This is not only inconsistent with the behaviour of the TS version (see [tfjs-core io_utils.ts line 155](https://github.com/tensorflow/tfjs/blob/479b02311cc9930619e8e068dce211f8315f5372/tfjs-core/src/io/io_utils.ts#L158)), but also causes data loss; this primarily affects normalised FP weights (e.g. values in the range [0,1] or [-1,1] resp.).

## Proposed Fix

The changes test whether the de-quantisation should result in `float32` values and doesn't round values after scaling if that is the case. The behaviour of integer case (e.g. de-quantisation to `int32`) remains unchanged.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4285)


tfjs-converter/python/tensorflowjs/quantization_test.py
```diff
@@ -118,6 +121,13 @@ def testAffineQuantizePositiveFloats(self):
         1, 3, np.float32, np.uint16,
         expected_metadata={'scale': 2/65536})
 
+  def testAffineQuantizeNormalizedFloats(self):
```
Gentlemen-s, line 124:
```suggestion
  def testAffineQuantizeNormalizedFloats(self):
```

Gentlemen-s (approved):

pyu10055 (changes requested):
Thank you for the contribution, I have a small comment.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4285#-:-MNOyrEY6tl_SDEL_doa:brnx19e)* status: 0 of 1 approvals obtained (waiting on @patlevin and @pyu10055)

___
*[tfjs-converter/python/tensorflowjs/quantization.py, line 160 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4285#-MNOyy-u-vfzFGIIckXD:-MNOyy-u-vfzFGIIckXE:bayp4jq) ([raw file](https://github.com/tensorflow/tfjs/blob/a4b6c08c656f7fcb00e0cd55b018fab491e25c5d/tfjs-converter/python/tensorflowjs/quantization.py#L160)):*
> ```Python
>     scale = metadata['scale']
>     min_val = metadata['min']
>     if original_dtype == np.int32:
> ```

I think even for int32 case, np.round is not needed, the astype() should have taken care of the rounding. can you leave it as
`return (data * scale + min_val).astype(original_dtype)`

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4285)* status: 0 of 1 approvals obtained (waiting on @patlevin)

___
*[tfjs-converter/python/tensorflowjs/quantization_test.py, line 126 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4285#-MNP6xNa6GTLDaNBX3KL:-MNP6xNa6GTLDaNBX3KM:bylm3ss) ([raw file](https://github.com/tensorflow/tfjs/blob/1606c592b2eae0ea353a1a6e3712583d71be7919/tfjs-converter/python/tensorflowjs/quantization_test.py#L126)):*
> ```Python
>   def testAffineQuantizeNormalizedFloats(self):
>     data = np.array(
>       [-0.29098126, -0.24776903, -0.27248842, 0.23848203], dtype=np.float32)
> ```

C:126, 0: Wrong hanging indentation (add 2 spaces).
      [-0.29098126, -0.24776903, -0.27248842, 0.23848203], dtype=np.float32)
      ^ | (bad-continuation)

patlevin:
@pyu10055 

> I think even for int32 case, np.round is not needed, the astype() should have taken care of the rounding. can you leave it as
return (data * scale + min_val).astype(original_dtype)

I tried that but this leads to int32 unit tests failing because the result of `astype()` is basically equivalent to `trunc()`, which doesn't yield the expected results in the int32 case.

patlevin:
___
*[tfjs-converter/python/tensorflowjs/quantization.py, line 160 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4285#-MNOyy-u-vfzFGIIckXD:-MNQPzKfBaRLRKkL9S4Q:b-rdwvno) ([raw file](https://github.com/tensorflow/tfjs/blob/1606c592b2eae0ea353a1a6e3712583d71be7919/tfjs-converter/python/tensorflowjs/quantization.py#L160)):*

I think even for int32 case, np.round is not needed, the astype() should have taken care of the rounding. can you leave it as
`return (data * scale + min_val).astype(original_dtype)`
</blockquote></details>

This would result in failing unit tests for the int32 case since astype() acts similarly (identically?) to trunc(), which leads to differing results.

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4285)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @patlevin and @pyu10055)

___
*[tfjs-converter/python/tensorflowjs/quantization.py, line 160 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4285#-MNOyy-u-vfzFGIIckXD:-MNZIiv55204tUwQNRJ5:bbv5md7) ([raw file](https://github.com/tensorflow/tfjs/blob/1606c592b2eae0ea353a1a6e3712583d71be7919/tfjs-converter/python/tensorflowjs/quantization.py#L160)):*
<details><summary><i>Previously, patlevin (Patrick Levin) wrote…</i></summary><blockquote>

This would result in failing unit tests for the int32 case since astype() acts similarly (identically?) to trunc(), which leads to differing results.
</blockquote></details>

thank you for clarifying.

PR #4292, opened 2020-11-21
Improve nested array creation example for tensor5d

joemullenix-ks:
- Change the nested-arrays example to match the flat-array example,
as is done with the other tensor classes (tensor1d, tensor2d, etc...).

Explanation:
When viewed on https://js.tensorflow.org/api/2.7.0, there is a small but meaningful inconsistency in the tensor constructor examples.
Each tensor (1d, 2d, ..., 6d) has a nested-arrays example and a flat-array example. These two examples use the same data. That is, they show nested first (shape is inferred), then flat with an explicit shape. This is true in every case _except_ the 5d tensor, in which nested and flat data do not match. The difference can be misleading to readers new to tensors.
This appears to have been a kind of copy-paste error (an easy one to make).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4292)


tafsiri (approved):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4292#-:-MNP11L3Byeiq4kZY9Lf:b-ucfknh)* status: :shipit: complete! 1 of 1 approvals obtained

PR #4303, opened 2020-11-25
core: Disable checkComputationForErrors for webgpu backend.

qjia7:
BUG

Add a flag shouldCheckComputationForErrors to control checkComputationForErrors.
Disable checkComputationForErrors for webgpu backend sine it may result the tensor
is disposed before it is read.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4303)

### DIFF
--- a/tfjs-backend-webgpu/src/index.ts
+++ b/tfjs-backend-webgpu/src/index.ts
@@ -18,19 +18,22 @@
 import './flags_webgpu';
 import './register_all_kernels';
 
-import * as tf from '@tensorflow/tfjs-core';
+import {env, registerBackend} from '@tensorflow/tfjs-core';
 import glslangInit from '@webgpu/glslang/dist/web-devel/glslang.onefile';
 
 import {WebGPUBackend} from './backend_webgpu';
 import {fromPixelsAsync} from './ops/from_pixels_async';
 import * as webgpu from './webgpu';
 
-tf.registerBackend('webgpu', async () => {
+registerBackend('webgpu', async () => {
+  // Remove it once we figure out how to correctly read the tensor data before
+  // the tensor is disposed in profiling mode.
+  env().setFlags({'CHECK_COMPUTATION_FOR_ERRORS': false});
+
   const glslang = await glslangInit();
   const gpuDescriptor: GPURequestAdapterOptions = {
-    powerPreference: tf.env().get('WEBGPU_USE_LOW_POWER_GPU') ?
-        'low-power' :
-        'high-performance'
+    powerPreference: env().get('WEBGPU_USE_LOW_POWER_GPU') ? 'low-power' :
+                                                             'high-performance'
   };
 
   const adapter = await navigator.gpu.requestAdapter(gpuDescriptor);
--- a/tfjs-core/src/engine.ts
+++ b/tfjs-core/src/engine.ts
@@ -557,12 +557,6 @@ export class Engine implements TensorTracker, DataMover {
     }
   }
 
-  private shouldCheckComputationForErrors(): boolean {
-    // Don't do checkComputationForErrors for webgpu backend since it's in an
-    // async function. It may result the tensor is disposed before it's read.
-    return this.backendName !== 'webgpu';
-  }
-
   /**
    * @deprecated Use `runKernel` for newly added kernels. Keep using this method
    *     only for kernels that are not yet fully modularized.
@@ -666,8 +660,7 @@ export class Engine implements TensorTracker, DataMover {
             outputs = kernelFunc();
           } else {
             kernelProfile = this.profiler.profileKernel(
-                kernelName, inputs, () => kernelFunc(),
-                this.shouldCheckComputationForErrors());
+                kernelName, inputs, () => kernelFunc());
             if (this.ENV.getBool('DEBUG')) {
               this.profiler.logKernelProfile(kernelProfile);
             }
--- a/tfjs-core/src/engine.ts
+++ b/tfjs-core/src/engine.ts
@@ -557,6 +557,12 @@ export class Engine implements TensorTracker, DataMover {
     }
   }
 
+  private shouldCheckComputationForErrors(): boolean {
+    // Don't do checkComputationForErrors for webgpu backend since it's in an
+    // async function. It may result the tensor is disposed before it's read.
+    return this.backendName !== 'webgpu';
+  }
+
   /**
    * @deprecated Use `runKernel` for newly added kernels. Keep using this method
    *     only for kernels that are not yet fully modularized.
@@ -660,7 +666,8 @@ export class Engine implements TensorTracker, DataMover {
             outputs = kernelFunc();
           } else {
             kernelProfile = this.profiler.profileKernel(
-                kernelName, inputs, () => kernelFunc());
+                kernelName, inputs, () => kernelFunc(),
+                this.shouldCheckComputationForErrors());
             if (this.ENV.getBool('DEBUG')) {
               this.profiler.logKernelProfile(kernelProfile);
             }
--- a/tfjs-core/src/flags.ts
+++ b/tfjs-core/src/flags.ts
@@ -14,9 +14,8 @@
  * limitations under the License.
  * =============================================================================
  */
-import './engine';
-
 import * as device_util from './device_util';
+import {ENGINE} from './engine';
 import {env} from './environment';
 
 const ENV = env();
@@ -70,3 +69,7 @@ ENV.registerFlag('DEPRECATION_WARNINGS_ENABLED', () => true);
 
 /** True if running unit tests. */
 ENV.registerFlag('IS_TEST', () => false);
+
+/**Whether to check computation result for errors. */
+ENV.registerFlag(
+    'CHECK_COMPUTATION_FOR_ERRORS', () => ENGINE.backendName !== 'webgpu');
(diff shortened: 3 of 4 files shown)


qjia7:
@annxingyuan @tafsiri Please take a look. Thanks.

tafsiri:
I agree there is a bug/issue here, i'm not sure this is the right fix (though we might want to put the checkComputationForErrors behind a flag so it can be disabled), as there would be a similar issue in the `logKernelProfile` function reight below this one which also is not async. In webgls `disposeData` i believe we do not dispose the data if there are pending reads on a tensor, i don't see similar logic in `disposeData`. Maybe similar logic is needed there to resolve this?

@qjia7 just as a heads up, because of the holiday we won't really be able to dig into until next week. 

With regard to performance i think we expect a perf hit when profiling is on, but that it would generally be fairly consistent across kernels as long as that is true, while it is slower than true execution time it still allows for identifying bottlenecks and measuring improvements. Are you finding that is not the case?

qjia7:
@tafsiri Thanks for your comment. 
>In webgls disposeData i believe we do not dispose the data if there are pending reads on a tensor, i don't see similar logic in disposeData. Maybe similar logic is needed there to resolve this?

Yes, it can resolve this issue. Thanks. Since there are still some profile cases fail, I need more time to figure out the reason. Will ping you again when it's ready for review.

>With regard to performance i think we expect a perf hit when profiling is on, but that it would generally be fairly consistent across kernels as long as that is true, while it is slower than true execution time it still allows for identifying bottlenecks and measuring improvements. Are you finding that is not the case?

Agree with your opinion. I just want to express that we should try to reduce unnecessary overhead when profiling. But if `checkComputationForErrors` is wanted for each kernel function. We should leave it there.

qjia7:
> > In webgls disposeData i believe we do not dispose the data if there are pending reads on a tensor, i don't see similar logic in disposeData. Maybe similar logic is needed there to resolve this?
> 
> Yes, it can resolve this issue. Thanks. Since there are still some profile cases fail, I need more time to figure out the reason. Will ping you again when it's ready for review.
> 

Sorry for taking back my previous rely.  I find it still can't resolve the issue by caching pending reads in the backend. See my draft PR #4314 . I think it uses the same logic with webgl. However, the [profile test case](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/engine_test.ts#L426) still explains that 'Buffer is destroyed before the mapping is resolved'. My guess is that if we put the cache logic in the backend, it's already in two different threads. You can't make sure that caching pending read is executed before disposeTensor (there are always race condition). Maybe we should cache the pending read tensor in frontend before divarication.

qjia7:
@tafsiri Could you help guild me how to further resolve this issue?

tafsiri:
One idea that comes to mind is to clone the tensors in the outputs array before entering that block.

Something like 

```ts
let outputsToCheck = []
for (let i = 0; i < outputs.length; i++) {
	outputsToCheck.push(outputs[i].clone());
}

for (let i = 0; i < outputsToCheck.length; i++) {
	const output = outputsToCheck[i];
	// Dangling promise here because we don't want to propagate up
	// asynchronicity.
	output.data().then(tensorVals => {
		checkComputationForErrors(tensorVals, output.dtype, kernelName);
		output.dispose();
	});
}
```

That should allow safe disposal of the actual output while allowing deferred checking of the output results.

@qjia7 could you give that a try and see if it solves the issue in the webgpu backend?

@annxingyuan what do you think of this approach?

qjia7:
> One idea that comes to mind is to clone the tensors in the outputs array before entering that block.
> 
> Something like
> 
> ```ts
> let outputsToCheck = []
> for (let i = 0; i < outputs.length; i++) {
> 	outputsToCheck.push(outputs[i].clone());
> }
> 
> for (let i = 0; i < outputsToCheck.length; i++) {
> 	const output = outputsToCheck[i];
> 	// Dangling promise here because we don't want to propagate up
> 	// asynchronicity.
> 	output.data().then(tensorVals => {
> 		checkComputationForErrors(tensorVals, output.dtype, kernelName);
> 		output.dispose();
> 	});
> }
> ```
> 
> That should allow safe disposal of the actual output while allowing deferred checking of the output results.
> 
> @qjia7 could you give that a try and see if it solves the issue in the webgpu backend?
> 
> @annxingyuan what do you think of this approach?

In fact, it will cause below errors both for webgl and webgpu:
```
FAILED profile webgl1 {"WEBGL_VERSION":1,"WEBGL_CPU_FORWARD":false,"WEBGL_SIZE_UPLOAD_UNIFORM":0} squaring
tensor.ts:272 Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at Array.slice (<anonymous>)
    at new Tensor (tensor.ts:272)
    at Engine.makeTensorFromDataId (engine.ts:780)
    at engine.ts:604
    at Array.map (<anonymous>)
    at kernelFunc (engine.ts:596)
    at engine.ts:659
    at holdResultWrapperFn (profiler.ts:43)
    at MathBackendWebGL.<anonymous> (backend_webgl.ts:526)
    at step (backend_webgl.ts:16)
    at Object.next (backend_webgl.ts:16)
    at backend_webgl.ts:16
    at new Promise (<anonymous>)
    at __awaiter (backend_webgl.ts:16)
    at MathBackendWebGL.time (backend_webgl.ts:513)
    at Profiler.profileKernel (profiler.ts:45)
    at engine.ts:658
    at Engine.scopedRun (engine.ts:453)
    at Engine.runKernelFunc (engine.ts:653)
    at clone_ (clone.ts:50)
    at Object.clone__op [as clone] (operation.ts:51)
    at Tensor.clone (tensor.ts:421)
    at Profiler.profileKernel (profiler.ts:49)
    at engine.ts:658
    at Engine.scopedRun (engine.ts:453)
    at Engine.runKernelFunc (engine.ts:653)
    at clone_ (clone.ts:50)
    at Object.clone__op [as clone] (operation.ts:51)
    at Tensor.clone (tensor.ts:421)
    at Profiler.profileKernel (profiler.ts:49)
    at engine.ts:658
    at Engine.scopedRun (engine.ts:453)
    at Engine.runKernelFunc (engine.ts:653)
    at clone_ (clone.ts:50)
    at Object.clone__op [as clone] (operation.ts:51)
    at Tensor.clone (tensor.ts:421)
    at Profiler.profileKernel (profiler.ts:49)
    at engine.ts:658
    at Engine.scopedRun (engine.ts:453)
    at Engine.runKernelFunc (engine.ts:653)
... ...

tafsiri:
Thanks for trying that. I'll sync with @annxingyuan over video chat and get back to you. The other idea I had to unblock this would be to put that section behind a flag that the webgpu backend could set to false.

annxingyuan:
Hi @qjia7 - apologies for the delay in getting back to you on this - I just spoke with @tafsiri, and we were thinking of this approach:

1. Before checking the output of the computation for errors, check to make sure the tensor is not disposed.
2. Add an argument to `profileKernel` - shouldCheckComputationForErrors - that defaults to `true`, but that is set to `false` for WebGPU.

Hopefully this would address the issue!

qjia7:
> 1. Before checking the output of the computation for errors, check to make sure the tensor is not disposed.
> 2. Add an argument to `profileKernel` - shouldCheckComputationForErrors - that defaults to `true`, but that is set to `false` for WebGPU.
> 

Done for step 2. 
For 1, I think it's not necessary. Once checkComputationForErrors is called, it means you already get the tensor data. All the check should happen on the backend.

@tafsiri @annxingyuan Please take another look. Thanks.

annxingyuan:
Thanks @qjia7 - however it would be ideal if core did not contain this logic. So perhaps the easiest thing would be to add an environment flag controlling whether to check computation for errors. Sorry for the back and forth :)

qjia7:
> Thanks @qjia7 - however it would be ideal if core did not contain this logic. So perhaps the easiest thing would be to add an environment flag controlling whether to check computation for errors. Sorry for the back and forth :)

Done. Please take another look. Thanks.

qjia7:
@tafsiri @annxingyuan Could you please take another look? It blocks us to enable the profiling mode in webgpu backend. Thanks.

tafsiri (commented):
Left a comment but I'll leave the final review with @annxingyuan as I think she has more context on this.

> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4303#-:-MO9mVoN8TuhaVw7C9vj:b-f5me1h)* status: 0 of 1 approvals obtained (waiting on @qjia7)

___
*[tfjs-core/src/flags.ts, line 75 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4303#-MO9m-xM3CAt4DXyI6aA:-MO9m-xM3CAt4DXyI6aB:b-729vhc) ([raw file](https://github.com/tensorflow/tfjs/blob/068ac4ca64dce162d16f14aeb9ea65986355b606/tfjs-core/src/flags.ts#L75)):*
> ```XML
> /** Whether to check computation result for errors. */
> ENV.registerFlag(
>     'CHECK_COMPUTATION_FOR_ERRORS', () => ENGINE.backendName !== 'webgpu');
> ```

@annxingyuan did you want this to check for webgpu here?

I would have thought this can default to `true` but in the the WebGPU backend it explicitly does

`env().setFlags({'CHECK_COMPUTATION_FOR_ERRORS': false});`

that way core doesn't explicitly know about webgpu and it can be enabled in future without changing core.

qjia7:
> I would have thought this can default to `true` but in the the WebGPU backend it explicitly does
> 
> `env().setFlags({'CHECK_COMPUTATION_FOR_ERRORS': false});`
> 
> that way core doesn't explicitly know about webgpu and it can be enabled in future without changing core.

Agree with you. Done. Thanks.

annxingyuan (approved):
LGTM - thanks @qjia7 !

PR #4306, opened 2020-11-25
[webgl] Modularize the following kernels:

mattsoulanille:
* logicalNot
* round
* isNaN
* isInf
* isFinite
* log1p
* reciprocal
* selu
* step

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4306)


mattsoulanille:
I refactored `unaryKernelFunc` in `kernel_func_utils.ts` to allow for attributes to be passed to the unary op (this better matches how we do things in the cpu version of `unaryKernelFunc`). `opSnippet` and  `packedOpSnippet`  can be passed as strings like they have been, but they can also be passed as functions that take attributes as their argument and generate the op snippet string. This is to remain compatible with all the other unary kernels while also allowing new ones to use attributes, but it might not be the best interface.

lina128 (commented):
The approach looks good to me. Though a counter argument is, if using attr in unary kernels is rare (special case), might as well not use the unaryKernelFunc and just write in regular kernelFunc style. I'll leave it to @tafsiri and @annxingyuan to decide.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4306#-:-MN0merm-0r3kxcQRcxS:byxn6rp)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, @mattsoulanille, and @tafsiri)

___
*[tfjs-backend-webgl/src/kernels/Step.ts, line 22 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4306#-MN0mQSP75Z6gmRWorFZ:-MN0mQSP75Z6gmRWorF_:buhm5h9) ([raw file](https://github.com/tensorflow/tfjs/blob/a5c7946370a38d8d4719e98d05f5229a369cf461/tfjs-backend-webgl/src/kernels/Step.ts#L22)):*
> ```XML
> import {CHECK_NAN_SNIPPET} from '../unaryop_gpu';
> 
> export function STEP(attrs: NamedAttrMap) {
> ```

Can we specify attrs as StepAttrs?

tafsiri (changes requested):
I agree the approach is fine, but from what I can tell step is the only thing that would use it in our current set of kernels. That said the change seems reasonably small and allows for other kernel to use UnaryOpProgram while having a little bit of customization. So I'm okay with keeping it modulo the main comment below: 

I left a comment alongside Na's about using StepAttrs rather then NamedAttrMap in the STEP function. If that is not doable (and I suspect it may be tricky) then I would just write step as a regular kernel and undo the refactoring of unaryKernelFunc. I also wouldn't spend a lot of time on this (trying to fix the types), not using the unaryKernelFunc helper is totally fine.

> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4306#-:-MNOmqh6-JZzHceJTBbZ:br6qr18)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, @mattsoulanille, and @tafsiri)

___
*[tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4306#-MNOobcR3F7SfFvYqnTx:-MNOlmnS4ACCgNjtBiYT:bhnkcpr) ([raw file](https://github.com/tensorflow/tfjs/blob/31b871bd408acd092333d8457595b03b092ed8db/tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts#L46)):*
> ```XML
> `;
> 
> type OpGenerator = (attrs?: NamedAttrMap) => string;
> ```

Rename to OpSnippetGenerator
___
*[tfjs-backend-webgl/src/kernels/Step.ts, line 22 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4306#-MN0mQSP75Z6gmRWorFZ:-MNOohPZ8YP8Nx5a_CUc:blrptiv) ([raw file](https://github.com/tensorflow/tfjs/blob/31b871bd408acd092333d8457595b03b092ed8db/tfjs-backend-webgl/src/kernels/Step.ts#L22)):*

Can we specify attrs as StepAttrs?
</blockquote></details>

+1 (and see comment below). However if the typing becomes complicated because of how OpGenerator is defined, then I would just undo the refactor and write this in the typical kernel style and not use use the unaryKernelFunc helper.
___
*[tfjs-backend-webgl/src/kernels/Step.ts, line 24 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4306#-MNOn1vyDwGpBGtVljU8:-MNOn1vyDwGpBGtVljU9:b-3gsewy) ([raw file](https://github.com/tensorflow/tfjs/blob/31b871bd408acd092333d8457595b03b092ed8db/tfjs-backend-webgl/src/kernels/Step.ts#L24)):*
> ```XML
> export function STEP(attrs: NamedAttrMap) {
>   let alpha = 0.0;
>   if (attrs.alpha !== undefined) {
> ```

This check (and even the default assignment) should not be necessary alpha is not an optional attribute. Implementing Na's suggestion to use StepAttrs will also make that type check.

mattsoulanille (commented):
Like you both said, It looks like this pattern is only used by the Step kernel. I thought there were a few others (reverse, neg, gather), but they are actually a bit different (they use a class instead of a function for the op text). I'm reverting the changes and special-casing Step.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4306#-:-MNOv8_c2Fqce3hwG3Q-:brk3lc1)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4306#-MNOobcR3F7SfFvYqnTx:-MNOvRfd00404ddyezC-:bp3aow6) ([raw file](https://github.com/tensorflow/tfjs/blob/31b871bd408acd092333d8457595b03b092ed8db/tfjs-backend-webgl/src/kernel_utils/kernel_funcs_utils.ts#L46)):*

Rename to OpSnippetGenerator
</blockquote></details>

Reverted.
___
*[tfjs-backend-webgl/src/kernels/Step.ts, line 22 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4306#-MN0mQSP75Z6gmRWorFZ:-MNOvB7F7u44TG1S0i-Y:b-5vvaxn) ([raw file](https://github.com/tensorflow/tfjs/blob/31b871bd408acd092333d8457595b03b092ed8db/tfjs-backend-webgl/src/kernels/Step.ts#L22)):*

+1 (and see comment below). However if the typing becomes complicated because of how OpGenerator is defined, then I would just undo the refactor and write this in the typical kernel style and not use use the unaryKernelFunc helper.
</blockquote></details>

This makes the typing not match. I'm reverting and special-casing Step instead.
___
*[tfjs-backend-webgl/src/kernels/Step.ts, line 24 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4306#-MNOn1vyDwGpBGtVljU8:-MNOvOOwC3gNDrnMz5O6:bp3aow6) ([raw file](https://github.com/tensorflow/tfjs/blob/31b871bd408acd092333d8457595b03b092ed8db/tfjs-backend-webgl/src/kernels/Step.ts#L24)):*

This check (and even the default assignment) should not be necessary alpha is not an optional attribute. Implementing Na's suggestion to use StepAttrs will also make that type check.
</blockquote></details>

Reverted.

tafsiri (approved):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4306#-:-MNPeDRMB5SP_qu2OPQ5:b-ucfknh)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

PR #4374, opened 2020-12-08
Use CopyImageBitmapToTexture for webgpu backend

shaoboyan:
ImageBitmap is also a possible input from HTML. Adding ImageBitmap as
fromPixels valid input type.

CopyImageBitmapToTexture is usual API for uploading in WebGPU.
Integrate this API into fromPixels in tfjs webgpu backend.

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4374)

PR #4211, opened 2020-11-10
[core] Remove fromPixelsAsync

annxingyuan:
This fixes https://github.com/tensorflow/tfjs/issues/4210

Reference PR: https://github.com/tensorflow/tfjs/pull/4139

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4211)

PR #4375, opened 2020-12-08
Add fromPixelAsync in tfjs-core

shaoboyan:
FromPixelAsync the first async kernel. This CL try to add a path to
run async kernel in tfjs.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)


shaoboyan:
@qjia7  PTAL

qjia7:
The original one was ever reverted by https://github.com/tensorflow/tfjs/pull/4211. So I prefer @annxingyuan @pyu10055 @tafsiri's review comments. Thanks.

tafsiri:
A couple of high level questions:

1. Could you explain again why the fromPixels _kernel_ cannot be synchronous in webgpu? (i.e. having an async op that calls a synchronous kernel— see where_async.ts for an example).
    - From a brief look at fromPixelsAsync in webgpu, the call to `createImageBitmap` is the only async part. If that were pulled into the op itself could the kernel be synchronous?
    - However looking at the docs for [createImageBitmap]( https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap) it does not seem to be supported in Safari? Is that the reason that part cannot be in the op and thus the kernel is synchronous?
2. Why add this to core and the public api now if webgpu is the only backend that has an implementation (as opposed to the current approach where it is an op exported by webgpu)? Will other backends need/be able to implement this in a way that is substantially different from fromPixels?

shaoboyan:
> A couple of high level questions:
> 
> 1. Could you explain again why the fromPixels _kernel_ cannot be synchronous in webgpu? (i.e. having an async op that calls a synchronous kernel— see where_async.ts for an example).
>    
>    * From a brief look at fromPixelsAsync in webgpu, the call to `createImageBitmap` is the only async part. If that were pulled into the op itself could the kernel be synchronous?

I'm sorry but I'm not quite understand "If that were pulled into the op itself could the kernel be synchronous". Could you pls provide some code clips as example ? Thanks. 
I've tried some `Promise.then` approach, which doesn't work. For `fromPixels` is an op with return value. Using `Promise.then` to implement this op needs to place `return state` into `.then` scope. This will trigger compile error which yield `fromPixels`  need to return tensor.

>    * However looking at the docs for [createImageBitmap](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap) it does not seem to be supported in Safari? Is that the reason that part cannot be in the op and thus the kernel is synchronous?

Yes, Safari doesn't support ImageBitmap. But for WebGPU, [copyImageBitmapToTexture](https://gpuweb.github.io/gpuweb/#dom-gpuqueue-copyimagebitmaptotexture) is the only fast uploading path. And I think with Apple in WebGPU working group, they'll make imageBitmap work in safari (Maybe).

> 2. Why add this to core and the public api now if webgpu is the only backend that has an implementation (as opposed to the current approach where it is an op exported by webgpu)? Will other backends need/be able to implement this in a way that is substantially different from fromPixels?

The reason why this API needs to be in core is that we need to run tfjs-models (e.g. resnet50). The samples in tfjs-models only interact with tfjs-core APIs. And I don't think other backends needs to be implement this async uploading API.

@qjia7 may have more inputs.

tafsiri:
> I'm sorry but I'm not quite understand "If that were pulled into the op itself could the kernel be synchronous". Could you pls provide some code clips as example ? Thanks.

Here is a link to the [whereAsync op code](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/where_async.ts#L42), the op itself is asynchronous but the implementation of the functionality (in whereImpl—which is similar executing a kernel) is a synchronous function. I was wondering if a similar approach could be taken her.

> The reason why this API needs to be in core is that we need to run tfjs-models (e.g. resnet50). The samples in tfjs-models only interact with tfjs-core APIs. And I don't think other backends needs to be implement this async uploading API.

I don't understand this, if this new method is added to core, how will those model samples (could you link to an examples) call it if they do not do so today? Do you plan to submit changes to those models? If so then what will the behaviour be on non webgpu backends?

shaoboyan:
> > I'm sorry but I'm not quite understand "If that were pulled into the op itself could the kernel be synchronous". Could you pls provide some code clips as example ? Thanks.
> 
> Here is a link to the [whereAsync op code](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/where_async.ts#L42), the op itself is asynchronous but the implementation of the functionality (in whereImpl—which is similar executing a kernel) is a synchronous function. I was wondering if a similar approach could be taken her.
> 

Thanks for the link. And I made an update to this CL(a prototyping). Pls double check whether this is the form more proper :).

> > The reason why this API needs to be in core is that we need to run tfjs-models (e.g. resnet50). The samples in tfjs-models only interact with tfjs-core APIs. And I don't think other backends needs to be implement this async uploading API.
> 
> I don't understand this, if this new method is added to core, how will those model samples (could you link to an examples) call it if they do not do so today? Do you plan to submit changes to those models? If so then what will the behaviour be on non webgpu backends?

Yes, you're right. The model needs to call the API too. And I think other backends will support this async fromPixels too by calling sync kernel in an async op. And the sample I mentioned is [here](https://github.com/tensorflow/tfjs-models/blob/master/posenet/demos/camera.js#L399) which is a sample of posenet.

tfjs-core/src/ops/browser.ts
```diff
@@ -154,6 +154,138 @@ function fromPixels_(
...
+      (pixels as HTMLVideoElement).readyState <
+          HAVE_CURRENT_DATA_READY_STATE) {
+    throw new Error(
+        'The video element has not loaded data yet. Please wait for ' +
+        '`loadeddata` event on the <video> element.');
+  }
+}
+// If the current backend has 'FromPixels' registered, it has a more
+// efficient way of handling pixel uploads, so we call that.
+const kernel = getKernel(FromPixels, ENGINE.backendName);
```
tafsiri, line 233: Given this structure can this method just call fromPixels_ directly rather using ENGINE.runKernel/getKernel? I think you would get a lot of code-reuse that way and reduce duplication.

It would basically test if createImageBitmap is available (since its not supported in all browsers) do the `await imageBitmap = await createImageBitmap(pixels);` you have below and then forward to fromPixels. If createImageBitmap is not available it can return await fromPixels(...) that way it always has a fallback across browsers.

tafsiri (dismissed):
Thanks for trying this! If this approach works for you I have a few high level comments to reduce duplication in the code and improve behaviour in browsers that do not support createImageBitmap.

My other high level comment is that this op needs tests (should be similar/same as fromPixels).

tfjs-core/src/ops/browser.ts
qjia7: @tafsiri fyi, #4374 is another patch to support `ImageBitmap` as a valid input of `fromPixels`

shaoboyan: @tafsiri Thanks for your reviewing and valuable suggestions! Totally agree with the direction to add this async op. And as @qjia7  said ,#4374 will be the base for this patch. So please take a look in that patch if you're available :)

shaoboyan:
@tafsiri (No need to reply in hurry since it's in holiday) FYI that I update the prototyping and test cases to cover fromPixelsAsync. After #4374 I'll bake it mature.

shaoboyan:
@qjia7  Please take a look at this!

qjia7 (approved):
LGTM.
@kainino0x @lina128 @tafsiri take a look, thanks.

lina128 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @kainino0x, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/browser.ts, line 187 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjCWSi8vHb4z_P_JKP:-MSjCWSi8vHb4z_P_JKQ:b584aeg) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L187)):*
> ```XML
> export async function fromPixelsAsync(
>   pixels: PixelData|ImageData|HTMLImageElement|HTMLCanvasElement|
>   HTMLVideoElement,
> ```

Should ImageBitmap in this list too?
___
*[tfjs-core/src/ops/browser.ts, line 199 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjDGxR9hO-ngdpBhKK:-MSjDGxR9hO-ngdpBhKL:brwheqz) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L199)):*
> ```XML
> 
>   // Check whether browser support ImageBitmap or the input is PixelData.
>   if (typeof ImageBitmap === 'undefined' ||
> ```

if(ImageBitmap == null
___
*[tfjs-core/src/ops/browser.ts, line 204 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjEcLp8ALArWwfmPpH:-MSjEcLp8ALArWwfmPpI:b-47tpun) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L204)):*
> ```XML
>     inputs = pixels;
>   } else {
>     // Force the imageBitmap creation to not do any premultiply alpha
> ```

Sugg: // Test if createImageBitmap is available, since it is not available in all browsers, if yes, create the imageBitmap and not do any premultiply alpha ops, if no, fallback to PixelData.
___
*[tfjs-core/src/ops/browser.ts, line 208 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjFAtY2qDNSM8dTXPj:-MSjFAtY2qDNSM8dTXPk:b-dtakc6) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L208)):*
> ```XML
>     const imageBitmap = 
>         // tslint:disable-next-line: no-any
>         await (createImageBitmap as any)(pixels as ImageBitmapSource,
> ```

What's the behavior when createImageBitmap is not available? Do you need to catch?
___
*[tfjs-core/src/ops/browser.ts, line 215 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjGh-Z-wazhrVeSmT5:-MSjGh-Z-wazhrVeSmT6:bv8fxy0) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L215)):*
> ```XML
>     // e.g. new Image(10, 10) but with 1 x 1 content. Avoid using
>     // imageBitmap as input in these cases. 
>     if (imageBitmap.width !== pixels.width ||
> ```

Do you need to check imageBitmap != null before accessing properties?
___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 18 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjHFsD2sURwvkTkIdg:-MSjHFsD2sURwvkTkIdh:bfqu4gs) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/from_pixels_async_test.ts#L18)):*
> ```XML
>  */
> 
> import * as tf from '../index';
> ```

Why this change? Also please add this test to tfjs-core/src/tests.ts
___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 206 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjIITE6h8P6rjyQoaz:-MSjIITE6h8P6rjyQob-:b-dmeep5) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/from_pixels_async_test.ts#L206)):*
> ```XML
>     img.src = 'data:image/gif;base64' +
>         ',R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw==';
>     
> ```

Is this .... in the code?

shaoboyan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/browser.ts, line 187 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjCWSi8vHb4z_P_JKP:-MSjVKhw6JTDmi60VgzL:b-ln775k) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L187)):*

Should ImageBitmap in this list too?
</blockquote></details>

Actually this API tend to limit the imageBitmap uses to WebGPU paths. For it will benefit WebGPU a lot. But for other paths, using other input format will be better.
___
*[tfjs-core/src/ops/browser.ts, line 204 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjEcLp8ALArWwfmPpH:-MSjVl8dD7FO7vyF49IR:b-6usgbq) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L204)):*

Sugg: // Test if createImageBitmap is available, since it is not available in all browsers, if yes, create the imageBitmap and not do any premultiply alpha ops, if no, fallback to PixelData.
</blockquote></details>

You're right. And another problem here is  that the ts check doesn't support createImageBitmap with options as parameter. https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap
Do you have any suggestions?
___
*[tfjs-core/src/ops/browser.ts, line 208 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjFAtY2qDNSM8dTXPj:-MSjWT5rD0qen3daKRC7:b-3k4and) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L208)):*

What's the behavior when createImageBitmap is not available? Do you need to catch?
</blockquote></details>

You're right. And I think it should fallback to pixelData path too.a
___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 18 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjHFsD2sURwvkTkIdg:-MSjZEP4-TeJuuVAU-6D:blz6qja) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/from_pixels_async_test.ts#L18)):*

Why this change? Also please add this test to tfjs-core/src/tests.ts
</blockquote></details>

I think this is due to I move the test case from WebGPU backend to core.
___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 206 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjIITE6h8P6rjyQoaz:-MSjZPsE7hlRGWTee0VF:bogg9z1) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/from_pixels_async_test.ts#L206)):*

Is this .... in the code?
</blockquote></details>

Weired. Actually no '....' shows in my editor. But there is some extra spaces. I'll delete them.a

shaoboyan:
@lina128  Please take another look ! I think the latest bots fail is not related to the CL. But I've fought with the bots on checking whether the environement support "window", "ImageBitmap" and "createImageBitmap". Maybe you have some suggestions.

shaoboyan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/browser.ts, line 199 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjDGxR9hO-ngdpBhKK:-MSlmKtA96MRwJQsP0Oo:b8qbx0t) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L199)):*

if(ImageBitmap == null
</blockquote></details>

I'm sorry but this check seems not work. So I change it to window.hasOwnProperty.
___
*[tfjs-core/src/ops/browser.ts, line 204 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjEcLp8ALArWwfmPpH:-MSlmStR2wU6sw10kYrE:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L204)):*
<details><summary><i>Previously, shaoboyan (Yan, Shaobo) wrote…</i></summary><blockquote>

You're right. And another problem here is  that the ts check doesn't support createImageBitmap with options as parameter. https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap
Do you have any suggestions?
</blockquote></details>

Done.
___
*[tfjs-core/src/ops/browser.ts, line 208 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjFAtY2qDNSM8dTXPj:-MSlmSFr37JP-n3Bt3jw:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L208)):*
<details><summary><i>Previously, shaoboyan (Yan, Shaobo) wrote…</i></summary><blockquote>

You're right. And I think it should fallback to pixelData path too.a
</blockquote></details>

Done.
___
*[tfjs-core/src/ops/browser.ts, line 215 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjGh-Z-wazhrVeSmT5:-MSlmUQH-47tiM8DNzEB:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L215)):*

Do you need to check imageBitmap != null before accessing properties?
</blockquote></details>

Done.
___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 206 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjIITE6h8P6rjyQoaz:-MSlmW2K902loSbfo6MY:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/from_pixels_async_test.ts#L206)):*
<details><summary><i>Previously, shaoboyan (Yan, Shaobo) wrote…</i></summary><blockquote>

Weired. Actually no '....' shows in my editor. But there is some extra spaces. I'll delete them.a
</blockquote></details>

Done.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/browser.ts, line 187 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjCWSi8vHb4z_P_JKP:-MSnDwOy4ZiK2hUI_duL:b5cgqxt) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L187)):*
<details><summary><i>Previously, shaoboyan (Yan, Shaobo) wrote…</i></summary><blockquote>

Actually this API tend to limit the imageBitmap uses to WebGPU paths. For it will benefit WebGPU a lot. But for other paths, using other input format will be better.
</blockquote></details>

So if pixels is ImageBitmap, which api should user use? Can you also update the doccomment above about when user should use this api?
___
*[tfjs-core/src/ops/browser.ts, line 203 at r4](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSnFiRk3ebTheEqo_Q5:-MSnFiRk3ebTheEqo_Q6:bzeo7il) ([raw file](https://github.com/tensorflow/tfjs/blob/d1d42da7273d0d0dde3043a46c4fef7da4c53f0a/tfjs-core/src/ops/browser.ts#L203)):*
> ```XML
>       window.hasOwnProperty('createImageBitmap') &&
>       kernel != null &&
>       !((pixels as PixelData).data instanceof Uint8Array)) {
> ```

Take a look at typescript Type Guard, is it possible to use that? It's a much better code style: https://www.typescriptlang.org/docs/handbook/advanced-types.html#typeof-type-guards

So something like:
```
function isPixelData(image:  PixelData|ImageData|HTMLImageElement|HTMLCanvasElement|
  HTMLVideoElement | ImageBitmap): image is PixelData {
  return (image != null) && (image.data instanceof Uint8Array);
}

if(window.ImageBitmap != null && window.createImageBitmap != null && kernel != null && isPixelData(pixels)) {
  // Create bitmap.
} else {
  // fall back.
}

```

shaoboyan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/browser.ts, line 187 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjCWSi8vHb4z_P_JKP:-MSuCaph9-M4tKwIljBV:b87dgfc) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L187)):*

So if pixels is ImageBitmap, which api should user use? Can you also update the doccomment above about when user should use this api?
</blockquote></details>

Update the doc to describe the recommend cases to use this new async API. Please take a look, thanks!
___
*[tfjs-core/src/ops/browser.ts, line 203 at r4](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSnFiRk3ebTheEqo_Q5:-MSuBiRZ8dePieu1OOad:btmmt1a) ([raw file](https://github.com/tensorflow/tfjs/blob/d1d42da7273d0d0dde3043a46c4fef7da4c53f0a/tfjs-core/src/ops/browser.ts#L203)):*

Take a look at typescript Type Guard, is it possible to use that? It's a much better code style: https://www.typescriptlang.org/docs/handbook/advanced-types.html#typeof-type-guards

So something like:
```
function isPixelData(image:  PixelData|ImageData|HTMLImageElement|HTMLCanvasElement|
  HTMLVideoElement | ImageBitmap): image is PixelData {
  return (image != null) && (image.data instanceof Uint8Array);
}

if(window.ImageBitmap != null && window.createImageBitmap != null && kernel != null && isPixelData(pixels)) {
  // Create bitmap.
} else {
  // fall back.
}

```
</blockquote></details>

That will be a great form but unfortunately I've tried and got some errors. My trying is 
1. 
```
function isPixelData(image:  PixelData|ImageData|HTMLImageElement|HTMLCanvasElement|
  HTMLVideoElement | ImageBitmap): image is PixelData {
  return (image != null) && (image.data instanceof Uint8Array);
}
```
And this will report error on 'image.data', which yield that HTMLImageElement etc don't have member data.

2.
```
function isPixelData(image:  any): image is PixelData {
  return (image != null) && 
             (typeof image === "PixelData")
             (image.data instanceof Uint8Array);
}
```
this will report error on `typeof image === "PixelData"`, which yield that Number | Bool | Undefined etc doesn't cover type "PixelData".

Do you have any suggestions?

shaoboyan:
@qjia7 @lina128  Please take another look since adding back ImageBitmap as valid input for FromPixelsAsync

qjia7:
> @qjia7 @lina128 Please take another look since adding back ImageBitmap as valid input for FromPixelsAsync

LGTM, thanks. Would like @lina128 to have the final review.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/browser.ts, line 203 at r4](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSnFiRk3ebTheEqo_Q5:-MT9-1cM5swEJsLSoHg9:b-nfb8y6) ([raw file](https://github.com/tensorflow/tfjs/blob/d1d42da7273d0d0dde3043a46c4fef7da4c53f0a/tfjs-core/src/ops/browser.ts#L203)):*
<details><summary><i>Previously, shaoboyan (Yan, Shaobo) wrote…</i></summary><blockquote>

That will be a great form but unfortunately I've tried and got some errors. My trying is 
1. 
```
function isPixelData(image:  PixelData|ImageData|HTMLImageElement|HTMLCanvasElement|
  HTMLVideoElement | ImageBitmap): image is PixelData {
  return (image != null) && (image.data instanceof Uint8Array);
}
```
And this will report error on 'image.data', which yield that HTMLImageElement etc don't have member data.

2.
```
function isPixelData(image:  any): image is PixelData {
  return (image != null) && 
             (typeof image === "PixelData")
             (image.data instanceof Uint8Array);
}
```
this will report error on `typeof image === "PixelData"`, which yield that Number | Bool | Undefined etc doesn't cover type "PixelData".

Do you have any suggestions?
</blockquote></details>

You can just use what you used `(pixels as PixelData).data instanceof Uint8Array)` in that type guard function, the benefit is, then pixels will be casted to PixelData, so that later you can just use it as PixelData without casting. Also this check condition has too many conditions, it's hard to understand what it does from a glance. Can you think of ways to refactor these condition checking into a readable way? By wrapping them into functions with semantic meaningful names.

shaoboyan:
@lina128  Thanks for the comments. I did some wraps for the conditions . PTAL again, thanks!

shaoboyan:
BTW, the bots fail seems not related to this CL, but I cannot triage the bots again

shaoboyan:
@lina128 now the bots are all green, please take a look if you have time, thanks!

lina128 (approved):
The refactor looks great, it's much clearer and more readable. Also overall the work is great addition to TF.js, thank you!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375#-:-MTqXux21oacw-UDR9jK:bjgch44)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/browser.ts, line 215 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjGh-Z-wazhrVeSmT5:-MTqXihXD3f6Fvj5_F9T:b-v8d50b) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L215)):*
<details><summary><i>Previously, shaoboyan (Yan, Shaobo) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

You don't need this check anymore right? Because you checked isNonEmptyPixels in the canWrapPixelsToImageBitmap(pixels).

shaoboyan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/browser.ts, line 215 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjGh-Z-wazhrVeSmT5:-MTs4EMn8wP4OtNoE7Dc:b-n58o0j) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L215)):*

You don't need this check anymore right? Because you checked isNonEmptyPixels in the canWrapPixelsToImageBitmap(pixels).
</blockquote></details>

I hope but no, we still need this check :(. For the reason that createImageBitmap may generate imageBitmap that clip to the image content size and createImageBitmaap may fail too. This will cause the imageBitmap size and pixel size are not the same. One of the test case catch this problem.

shaoboyan:
@lina128 @tafsiri After some discussion with @qjia7 , we adopt the approach that use the flag to control whether fromPixelsAsync try to wrap pixels into imageBitmap. The reason is that:
- Use ImageBitmap as input will benefit WebGPU backend a lot, but for other backends, it is not needed.
- Wrap pixels with transparent content to imageBitmap and draw to canvas will have value difference, only WebGL and WebGPU 
   can handle the  transparent content.
PTAL this again, thanks!

lina128 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-backend-webgpu/src/index.ts, line 33 at r10](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MUA6z0-8JUyGbkzwQpv:-MUA6z0-8JUyGbkzwQpw:b-wkk7zx) ([raw file](https://github.com/tensorflow/tfjs/blob/6c5d68f2e0e118b10b1fd7103561a3f30447ff47/tfjs-backend-webgpu/src/index.ts#L33)):*
> ```XML
> 
>   // WebGPU backend prefers ImageBitmap as input to use fast uploading path.
>   env().set('WRAP_TO_IMAGEBITMAP', true);
> ```

This flag is global, once webgpu backend is registered, the flag will be true. Imagine a scenario, where we switch to another backend, then the flag will be true for that backend as well. Maybe just leave it to user to set? Because they own the logic of when to setBackend too.
___
*[tfjs-core/src/ops/browser.ts, line 215 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MSjGh-Z-wazhrVeSmT5:-MUA6NDL45vhTGOM2wzs:b5eewv9) ([raw file](https://github.com/tensorflow/tfjs/blob/e0690976544d08c6fd24e7a13b6758f847b0ff81/tfjs-core/src/ops/browser.ts#L215)):*
<details><summary><i>Previously, shaoboyan (Yan, Shaobo) wrote…</i></summary><blockquote>

I hope but no, we still need this check :(. For the reason that createImageBitmap may generate imageBitmap that clip to the image content size and createImageBitmaap may fail too. This will cause the imageBitmap size and pixel size are not the same. One of the test case catch this problem.
</blockquote></details>

Ah, I missed that. Thanks for the explanation, makes perfect sense.

shaoboyan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-backend-webgpu/src/index.ts, line 33 at r10](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MUA6z0-8JUyGbkzwQpv:-MUCAd2s5W_NvwN6HjB3:bybt7lr) ([raw file](https://github.com/tensorflow/tfjs/blob/6c5d68f2e0e118b10b1fd7103561a3f30447ff47/tfjs-backend-webgpu/src/index.ts#L33)):*

This flag is global, once webgpu backend is registered, the flag will be true. Imagine a scenario, where we switch to another backend, then the flag will be true for that backend as well. Maybe just leave it to user to set? Because they own the logic of when to setBackend too.
</blockquote></details>

Ah, that's the case. Thanks for point it out! Remove this setting and left to customer to set it.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 62 at r11](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MUEzH-pEQhwBkBDLhpV:-MUEzH-pEQhwBkBDLhpW:blnzbn2) ([raw file](https://github.com/tensorflow/tfjs/blob/bf5fc78503b4270a069333f02c5c6b998991b468/tfjs-core/src/ops/from_pixels_async_test.ts#L62)):*
> ```XML
> });
> 
> describeWithFlags('fromPixelsAsync', BROWSER_ENVS, () => {
> ```

You also need to test WRAP_TO_IMAGEBITMAP=true case.

lina128 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 284 at r12](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MUHdcLK3FrcauamSS0S:-MUHdcLK3FrcauamSS0T:b-hhe5b3) ([raw file](https://github.com/tensorflow/tfjs/blob/56d8ac4b5df9ed98fc9d3a8b6c3de3b348ee9d1f/tfjs-core/src/ops/from_pixels_async_test.ts#L284)):*
> ```XML
>       tf.env().set('WRAP_TO_IMAGEBITMAP', true);
>   });
> ```

Also add afterAll, set back to false.

shaoboyan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 62 at r11](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MUEzH-pEQhwBkBDLhpV:-MUHn7OSB6ne8UhURR3p:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/bf5fc78503b4270a069333f02c5c6b998991b468/tfjs-core/src/ops/from_pixels_async_test.ts#L62)):*

You also need to test WRAP_TO_IMAGEBITMAP=true case.
</blockquote></details>

Done.
___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 284 at r12](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MUHdcLK3FrcauamSS0S:-MUHmVGh7ThVPL0K0Wk3:b-glzzm0) ([raw file](https://github.com/tensorflow/tfjs/blob/56d8ac4b5df9ed98fc9d3a8b6c3de3b348ee9d1f/tfjs-core/src/ops/from_pixels_async_test.ts#L284)):*

Also add afterAll, set back to false.
</blockquote></details>

Thanks for the suggestions and patience :) Done.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 369 at r13](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MUKOR4zAKM9qfSJvXun:-MUKOR4zAKM9qfSJvXuo:b-vqz9pg) ([raw file](https://github.com/tensorflow/tfjs/blob/3d2c3c3466605b80af5aa47d690c65a1c535edaa/tfjs-core/src/ops/from_pixels_async_test.ts#L369)):*
> ```XML
>   });
> 
>   afterAll(() => {
> ```

Please place the afterAll after beforeAll

shaoboyan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4375)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kainino0x, @lina128, @shaoboyan, and @tafsiri)

___
*[tfjs-core/src/ops/from_pixels_async_test.ts, line 369 at r13](https://reviewable.io/reviews/tensorflow/tfjs/4375#-MUKOR4zAKM9qfSJvXun:-MULrWQC4NgOTVqdww6T:b-uo70wc) ([raw file](https://github.com/tensorflow/tfjs/blob/3d2c3c3466605b80af5aa47d690c65a1c535edaa/tfjs-core/src/ops/from_pixels_async_test.ts#L369)):*

Please place the afterAll after beforeAll
</blockquote></details>

Thanks for the comments, Done.

PR #4422, opened 2020-12-17
remove auto registration of gradients in core [3.x_dev]

tafsiri:
Also add some tests to make sure gradients and chain api are present when using the union script tag bundle.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4422)

### DIFF
--- a/e2e/integration_tests/setup_test.ts
+++ b/e2e/integration_tests/setup_test.ts
@@ -20,6 +20,8 @@ import '@tensorflow/tfjs-core';
 // tslint:disable-next-line: no-imports-from-dist
 import '@tensorflow/tfjs-core/dist/public/chained_ops/register_all_chained_ops';
 // tslint:disable-next-line: no-imports-from-dist
+import '@tensorflow/tfjs-core/dist/register_all_gradients';
+// tslint:disable-next-line: no-imports-from-dist
 import {parseTestEnvFromKarmaFlags, registerTestEnv, setTestEnvs, TEST_ENVS} from '@tensorflow/tfjs-core/dist/jasmine_util';
 
 import {TAGS} from './constants';
--- a/e2e/integration_tests/setup_test.ts
+++ b/e2e/integration_tests/setup_test.ts
@@ -20,8 +20,6 @@ import '@tensorflow/tfjs-core';
 // tslint:disable-next-line: no-imports-from-dist
 import '@tensorflow/tfjs-core/dist/public/chained_ops/register_all_chained_ops';
 // tslint:disable-next-line: no-imports-from-dist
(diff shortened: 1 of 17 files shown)


lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4422)* status: 0 of 1 approvals obtained (waiting on @annxingyuan, @lina128, and @tafsiri)

___
*[e2e/scripts/test-ci.sh, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4422#-MOmL40OA1ygHGJkNZ3i:-MOmL40OA1ygHGJkNZ3j:b-puylfl) ([raw file](https://github.com/tensorflow/tfjs/blob/283c010eaee821a004cf6aeaabe0357947d95389/e2e/scripts/test-ci.sh#L69)):*
> ```Shell
>   yarn run-browserstack --browsers=bs_chrome_mac --tags $TAGS
>   # Test script tag bundles
>   karma start ./script_tag_tests/karma.conf.js --browserstack --browsers=bs_chrome_mac --testBundle tf.min.js
> ```

Why do we move it back to test in every PR?

tafsiri (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4422)* status: 0 of 1 approvals obtained (waiting on @annxingyuan and @lina128)

___
*[e2e/scripts/test-ci.sh, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4422#-MOmL40OA1ygHGJkNZ3i:-MOmg-BA5cwUM-alCR52:blw1dbn) ([raw file](https://github.com/tensorflow/tfjs/blob/283c010eaee821a004cf6aeaabe0357947d95389/e2e/scripts/test-ci.sh#L69)):*

Why do we move it back to test in every PR?
</blockquote></details>

Because nightly is flaky and these changes directly affect the contents of the builds I think it is good to have this in every PR in this branch. It will also help test correctness of merging 3.x_dev into master before waiting for a nightly run.

We can change it back to nightly once this branch has merging into master, but I would still advocate for keeping them in every PR even then, they do not take long to run as they primarily rely on the bundle being built which is done anyway in previous steps and they are a good sanity check that our precompiled package is working. 

I've added a todo to revisit/remove this once we have merged. Wdyt?

lina128 (approved):

tafsiri:
To add a bit more context as well, i wanted to add these specific tests because the way our test infra works would mask a bug in the final build if the imports aren't in the right place (the test infra does imports of its own), so testing the final build is also useful as it isolates some of the action of what we do in some of our test infra (or at least that is the goal).

lina128:
Got it, thanks for the detailed explanation Yannick!

Issue #4429, opened 2020-12-18
tfjs 2.8.1 change in behavior in tf.reshape causes errors


vladmandic:
upon upgrade from 2.7.0 to 2.8.1, my existing app fails with:
```
Uncaught (in promise) Error: The implicit shape can't be a fractional number. Got 266 / 3
    at Object.inferFromImplicitShape (util_base.ts:328)
    at Object.reshape76 [as kernelFunc] (Reshape.ts:35)
    at kernelFunc3 (engine.ts:590)
    at engine.ts:660
    at Engine.scopedRun (engine.ts:453)
    at Engine.runKernelFunc (engine.ts:657)
    at Engine.runKernel (engine.ts:522)
    at reshape_ (reshape.ts:60)
    at reshape__op (operation.ts:51)
```

this may be desired behavior, but it breaks several existing models published on `tfhub`!

e.g, i'm using `blazeface` where `reshape` is failing on results of model inference, so not like that can be changed easily:
```js
const [contours, confidence, contourCoords] = this.meshDetector.predict(face);
const coordsReshaped = tf.reshape(contourCoords, [-1, 3]);
let rawCoords = coordsReshaped.arraySync();
```

lina128:
Which backend has this behavior?

vladmandic:
webgl backend

scottzockoll:
@lina128 I am also receiving this error after upgrading to 2.8.1 with webgl.

p-dimi:
Same thing here!

designerzen:
Same here! When trying to fetch face landmarks :

` const predictions = await model.estimateFaces({ input: video }) `

Browser returns 

`util_base.ts:328 Uncaught (in promise) Error: The implicit shape can't be a fractional number. Got 266 / 3`

rthadur:
cc @pyu10055 @annxingyuan @tafsiri

dewball345:
Yup, I'm getting the same error as well! I still don't understand why demos like these https://mediapipe-facemesh-3js-tf2.glitch.me/ work. I tried changing the version of my facemesh model but that doesn't do anything

vladmandic:
@dewball345 because it's hard coded to load tfjs 2.4.0. anything up to 2.7.0 works fine, 2.8.0 and 2.8.1 don't.

dewball345:
yeah i realized that now. A temporary fix would be to just use the older version of tensorflow?

vladmandic:
[...]

PR #4436, opened 2020-12-21
Fix signature field

lina128:
Tested with Face-landmarks-detection model, confirmed it fixes https://github.com/tensorflow/tfjs/issues/4429

Instead of checking version, we should just check whether signature is in userDefinedMetadata, otherwise it is in top level. This is because the version checking method is vulnerable, it relies on hard coded marker in the convertedBy field.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4436)

### DIFF
--- a/tfjs-converter/src/executor/graph_model.ts
+++ b/tfjs-converter/src/executor/graph_model.ts
@@ -43,6 +43,7 @@ export class GraphModel implements InferenceModel {
   private artifacts: io.ModelArtifacts;
   private initializer: GraphExecutor;
   private resourceManager: ResourceManager;
+  private signature: tensorflow.ISignatureDef;
 
   // Returns the version information for the tensorflow model GraphDef.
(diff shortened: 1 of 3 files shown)


annxingyuan (approved):
Thanks Na! Out of curiosity how exactly does this eventually translate to the reshape error?

annxingyuan (approved):
Thanks Na! Out of curiosity how exactly does this eventually translate to the reshape error?

lina128 (commented):
It couldn't read the signature field, so I guess the shape info in the signature is lost.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4436#-:-MPBMSgU1t00n03MWRYz:bk7yq0z)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

PR #4445, opened 2020-12-25
Enable float16 feature and add float16 type

xhcao:
DEV


xhcao:

tfjs-core/src/globals.ts
```diff
@@ -52,6 +52,28 @@ export function enableDebugMode(): void {
...
+ *
+ * Currently, the float16 mode is only available on webgpu backend.
+ *
+ * Returns true if underlying driver supports float16 feature and enable the
+ * feature successfully.
+ *
+ * @doc {heading: 'Environment'}
+ */
+export function tryToEnableFloat16Mode(): boolean {
+  if (ENGINE.backendName === 'webgpu' &&
```
qjia7, line 67: Remove `ENGINE.backendName === 'webgpu' &&`.

qjia7 (commented):
+@pyu10055 @tafsiri @lina128, please take a look if this is in the right way. Thanks.

xhcao:
I encountered the issue about how to introduce this feature to developers, in this patch,
1. Introduced a "DRIVER_SUPPORT_FLOAT16" flag;
2. Introduced a "enableFloat16Mode()" API.
If developers want to use the float16 feature, they must be use "enableFloat16Mode()" to explicitly enable this feature. If developers does not enable this feature but use this feature, tfjs-core could throw errors.

There is another way to introduce float16 feature. We could only introduce a "DRIVER_SUPPORT_FLOAT16" flag, and developers are responsible for their usages. If developers use the float16 but driver does not support float16, they would get error and undefined result.

Which way is more reasonable？

lina128 (commented):
Can you clarify what's the difference between (1) Only introduce a "DRIVER_SUPPORT_FLOAT16" flag and allow developers to set it. and (2) enableFloat16Mode().

What indicator can we use to detect driver float16 support and is it reliable?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4445#-:-MRW9AdJ9syZYX5ugR-n:b-f8xlug)* status: 0 of 1 approvals obtained (waiting on @pyu10055 and @xhcao)

lina128 (commented):
Hi @xhcao , so I further looked at the code, now I understand your comments better. First all, if this is webgpu specific flag, can we add it under flags_webgpu.ts? If there's not one, please create the file. Second, if float16 is supported, do we want to force users to use it or not? Is it a universally better option? If so, you don't need a FLOAT16 flag, if not then adding a user tunable flag makes sense, the name is too general, consider something like 'WEBGPU_ENABLE_FLOAT16'.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4445#-:-MRWPFah4eOjKmhAT4_1:b-n57t45)* status: 0 of 1 approvals obtained (waiting on @pyu10055 and @xhcao)

xhcao:
Hi, @lina128 ,
1. Move the flag to flags_webgpu.ts;
2. It is not a universally better option. Float16 is a good option that users pay more attention to computation speed. But if users want high accuracy, I think it is better using float32.

I also removed the "enableFloat16Mode" api here, I think that users does not need to explicitly enable float16 feature. Whether or not to use float16 feature depends on model's data type or input data type.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4445)* status: 0 of 1 approvals obtained (waiting on @pyu10055 and @xhcao)

___
*[tfjs-core/src/types.ts, line 42 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4445#-MS0lyNuB-IEZNfkuwHC:-MS0lyNv2UEJq6L6fYQ_:bq8838c) ([raw file](https://github.com/tensorflow/tfjs/blob/3f0538fa69c97c923bd69214bb06b7cc2057052a/tfjs-core/src/types.ts#L42)):*
> ```XML
> export interface DataTypeMap {
>   float32: Float32Array;
>   float16: Uint16Array;
> ```

Can you add an annotation on how float16 is represented by Uint16Array?

xhcao:
> Can you add an annotation on how float16 is represented by Uint16Array?

Sorry, I do not fully understand your question. Uint16Array is only the storage for float16, because the element size of Uint16Array and float16 are the same(16 bits), and float16 is not the standard data type on CPU side, so we save float16 into an Uint16Array object.
float16's format is as https://en.wikipedia.org/wiki/Half-precision_floating-point_format, and we could convert a float to float16 according to http://www.fox-toolkit.org/ftp/fasthalffloatconversion.pdf algorithm. 
It is the same style as C++ code, we will use uint16_t to save float16 data on CPU side.

lina128 (approved):
Thank you for the detailed explanation! This is exactly what I was asking for, should be documented somewhere in the code. I'll leave it to you to decide whether this is suitable to add in types.ts or elsewhere e.g. conversion method. But this is not a blocker for this PR. LGTM.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4445#-:-MS5xsJq3HB0molduP7e:bixcfpy)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055 and @xhcao)

xhcao:
@tafsiri @pyu10055 Please take a look, thank you.

pyu10055:
> hi @lina128 I was wondering if there was any resolution to this issue. Support for float16 (and int16) dtype would be very valuable as a feature for medical imaging use cases.

Hi @Ouwen We would love to understand your use cases in more details, since javascript does not support float16 or int16 natively, if you are loading a certain file/data, this would not necessarily help, this is mainly for improve accelerator (GPU/WASM) utilization.

Ouwen:
hi @pyu10055, thanks for the quick response! While half float is not natively supported by CPU javascript it is supported in WebGL. Additionally, uint16 and int16 datatypes are supported by CPU javascript via TypedArrays.

Our use case is in the medical imaging field where images are commonly stored natively in uint16 and int16. This would allow for lossless input without inflating images into Float32 form. This can be significant for medical imaging volumes such as CT scans, MRI scans etc which can be 512x512x512 cubes.

Issue #4378, opened 2020-12-09
GPU memory leak of tf.signal.stft


huningxin:
<em>Please make sure that this is a bug. As per our
[GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md),
we only address code/doc bugs, performance issues, feature requests and
build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): Yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): macOS 10.15.7
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device: N/A
- TensorFlow.js installed from (npm or script link): <script src="https://cdn.jsdelivr.net/npm/@tensorflow/tfjs@2.7.0"> </script>
- TensorFlow.js version (use command below): 2.7.0
- Browser version: Chrome 87.0.4280.88
- Tensorflow.js Converter Version: N/A

**Describe the current behavior**
There is GPU memory leak after use tf.signal.stft. If use it continuously, it leads to crash the Chrome GPU process.  

**Describe the expected behavior**
There is no GPU memory leak after use tf.signal.stft.

**Standalone code to reproduce the issue**
Provide a reproducible test case that is the bare minimum necessary to generate
the problem. If possible, please share a link to Colab/CodePen/any notebook.

```js
function hanningWindow(N) {
  return tf.tidy(() => {
    const window = new Float32Array(N);
    for (let i = 0; i < N - 1; ++i) {
      window[i] = 0.5*(1 - Math.cos(6.283185307179586*i/(N-1)));
    }
    return tf.sqrt(window);
  });
}

console.log(tf.engine().memory());
// e.g. {unreliable: false, numBytesInGPU: 0, numBytesInGPUAllocated: 0, numBytesInGPUFree: 0, numTensors: 0, …}

const N = 100; // Reproduce the GPU process crash by setting large number, e.g. 10000 
for (let i = 0; i < N; ++i) {
  const win = 320;
  const fft = 320;
  const hop = 160;
  const input = tf.zeros([1760]);
  const output = tf.signal.stft(input, win, hop, fft, hanningWindow);
  input.dispose();
  output.dispose();
}

console.log(tf.engine().memory());
[...]

PR #4448, opened 2020-12-28
[webgl] Reference for WebGL stft memLeak fix.

annxingyuan:
Fix issue https://github.com/tensorflow/tfjs/issues/4378

There are two issues related to STFT failure
1. Complex components cannot be release if there are multiple references on the components and those reference are disposed before the complex tensor is disposed.
2. STFT uses RFFT which uses a shallow_slice, there is a bug in shallow_slice that copied refCount and complexParentRefCount from the original tensor to the output tensor, which cause the tensors to have extra refCount and complexParentRefCount.

Problem 1 is solved by sync engine ref count to backend, reset the output tensor ref counts in the shallow_slice fixed problem 2. 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4448)


lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4448)* status: 0 of 1 approvals obtained (waiting on @annxingyuan)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 557 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4448#-MR2PPIr1nFEOPeDzaaH:-MR2PPIr1nFEOPeDzaaI:b-2h2unw) ([raw file](https://github.com/tensorflow/tfjs/blob/7fbd5547b52f4afe6c5f545cdcee481694c789e9/tfjs-backend-webgl/src/backend_webgl.ts#L557)):*
> ```XML
>     if (this.texData.get(dataId).complexParentRefCount > 0) {
>       this.texData.get(dataId).refCount--;
>       this.texData.get(dataId).complexParentRefCount--;
> ```

If this tensor is an underlying part of complex, e.g. real or imag, when this tensor is disposed, we don't have information whether the parent tensor is disposed, so why do we reduce the parent tensor's reference? For example, I create a real tensor a, then I create a complex tensor from a, then I dispose tensor a, it shouldn't affect the complex tensor. If I read the value from the complex tensor, I should be able to get the value of a. But after this change, above scenario will fail, it will not be able to read a's value because it is disposed.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 569 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4448#-MR2M_xTCeTTFTGrCg2-:-MR2M_xU3IbNXd8sfwx_:b2zr9w8) ([raw file](https://github.com/tensorflow/tfjs/blob/7fbd5547b52f4afe6c5f545cdcee481694c789e9/tfjs-backend-webgl/src/backend_webgl.ts#L569)):*
> ```XML
>       const realDataInfo = this.texData.get(complexTensorInfos.real.dataId);
>       realDataInfo.complexParentRefCount--;
> ```

The block of change here has the same effect as before, right?

lina128 (approved):
Thank you Ping!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4448#-:-MR6eba72Cpf6i77eGbl:b-t1trzh)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan)

___
*[tfjs-backend-webgl/src/kernels/STFT_test.ts, line 22 at r4](https://reviewable.io/reviews/tensorflow/tfjs/4448#-MR6eDEe7zp9uac45amT:-MR6eDEe7zp9uac45amU:b-l8l8cc) ([raw file](https://github.com/tensorflow/tfjs/blob/856a05cd5fdb1ce1ade79909dad6915c86417fcf/tfjs-backend-webgl/src/kernels/STFT_test.ts#L22)):*
> ```XML
> import {ALL_ENVS, describeWithFlags} from '@tensorflow/tfjs-core/dist/jasmine_util';
> 
> describeWithFlags('stft memory test', ALL_ENVS, () => {
> ```

Please add the same test to cpu too.

Issue #4480, opened 2021-01-04
Error running LinSpace


epicure:
When I run **[tf.linspace](https://js.tensorflow.org/api/2.8.2/#linspace)** in the 2.8.2  API documentation.
I got this "An error occured on line: 1 Error running LinSpace: Neither modular kernel nor forward func passed".

![image](https://user-images.githubusercontent.com/422362/103514521-23658b00-4eb0-11eb-9284-8197ab47d166.png)

The 2.7.0 vsersion of tf.linspace works but 2.8.0, 2.8.1, 2.8.2  are not working for now.

**System information**

- TensorFlow.js version: 2.8.0, 2.8.1, 2.8.2

- OS Platform and Distribution (MacOS 10.15, Windows 10):

tafsiri:
Re-opening as this will need a new release to publish the fix.

For context this will only happen if linspace is the first op called. As a workaround you call 
`tf.ready()` or `await tf.ready()` before calling any ops.

rafallukasik123:
@tafsiri  your workground doesn't work for me. 
When i call tf.ready() before calling any ops i have error in one of kernel func. 
More specifically in fromPixels.js and the error content is "backend.makeTensorInfo is not a function"

tafsiri:
@rafallukasik123 could you post a reprodcution (maybe in codepen form) so that we can take a look.

rafallukasik123:
Unfortunately I can't share all the code. Here is a small snipe of code https://codepen.io/rafallukasik123/pen/LYRBpVj 
When i remove tf.ready() from the beginning i have error similar to you "Error running Round: Neither modular kernel nor forward func passed".
If my code contain tf.ready() i have mentioned error "backend.makeTensorInfo is not a function" in @tensorflow/tfjs-backend-webgl/dist/kernels/FromPixels.js

rafallukasik123:
in the latest version 2.8.3 I have error `Error running Reciprocal: Neither modular kernel nor forward func passed`. 
should I create a new issue?

tafsiri:
@rafallukasik123 Yes I think a new issue would be best. In terms of reproduction, you don't have to post your whole code, ideally it would be the smallest snippet of code that shows the error. It can even just load the library and directly call tf.reciprocal or tf.browser.fromPixels. In the codepen you posted I get a syntax error so the code doesn't run.

tafsiri:
[...]

PR #4485, opened 2021-01-05
initialize/set backend in runKernelFunc

tafsiri:
Fixes https://github.com/tensorflow/tfjs/issues/4480

Initialize backend in runKernelFunc if its not already initialized/set.

It manifests in the bug above because linspace no longer calls other ops which would initialize the backend (usually by creating some tensor). It also affects op like range which call runKernel before doing any tensor creation or anything else. The bug will _only appear if linspace or range is the first op_ called, and no any other action that would cause the lazy setting of a backend to have completed (like tensor creation).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4485)

### DIFF
--- a/tfjs-core/src/engine.ts
+++ b/tfjs-core/src/engine.ts
@@ -585,7 +585,8 @@ export class Engine implements TensorTracker, DataMover {
     if (this.backendName == null) {
       // backend has not been initialized yet. Fetch it to trigger
       // initialization.
-      this.backend;
+      // tslint:disable-next-line: no-unused-expression
+      this.backend;  // this getter has side effects
     }
     const kernel = getKernel(kernelName, this.backendName);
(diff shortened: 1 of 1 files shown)


annxingyuan (approved):
Interesting! Thanks Yannick. Could you add a bit more context to the comment - e.g. mention that certain ops don't cause backend initialization anymore?

lina128 (approved):

tafsiri:
@annxingyuan Thanks. I've updated the comment to mention lazy backend initialization and how that interacts with ops/kernels broadly, I didn't want the comment to be too tied to the ops mentioned in this PR description to keep it as relevant as possible 6 months+ from now, so I tried to find a general way to express it.

PR #4486, opened 2021-01-05
[wasm] Implement the Round kernel

### LINKED ISSUE #4434: [wasm] Add Round kernel. Unblocks efficientdet models.
**System information**
- TensorFlow.js version (you are using): 2.7.0

There are plans to add an implementation for efficientdet models in the wasm backend ? 
It is related to the issue https://github.com/tensorflow/tfjs/issues/4408.
When i use webgl backend i  have a problem with precision (my device doesn't support float32) 
I took the advice from this issue. I tried to use the wasm backend but i have exception 
"Error: 'round' not yet implemented or not found in the registry. This kernel may not be supported by the tfjs backend you have chosen"

mattsoulanille:
Implement the Round kernel using XNNPACK's banker's rounding. Banker's rounding rounds to the nearest integer, but breaks ties in the direction of the nearest even integer. It passes our tests, but might not be desirable. XNNPACK doesn't seem to expose a 'normal' rounding operator that always rounds up in the case of a tie.

Closes #4434 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4486)

### DIFF
--- a/tfjs-backend-wasm/src/cc/kernels/Round.cc
+++ b/tfjs-backend-wasm/src/cc/kernels/Round.cc
@@ -1,0 +1,38 @@
+/* Copyright 2020 Google LLC. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


mattsoulanille:
I should have checked CPU and WebGL first. They also use banker's rounding, so this PR's  implementation should be fine.

annxingyuan (approved):
LGTM - thanks Matt!

PR #4492, opened 2021-01-06
webgpu: Fix the incorrect use of setFlags

qjia7:
BUG

setFlags will reset all flags with the given list. However, in our case,
we only need to update flag CHECK_COMPUTATION_FOR_ERRORS's value and not
override other flags values. So we should use set instead of setFlags function.

However, this flag CHECK_COMPUTATION_FOR_ERRORS is added in tfjs-core 2.8.0.
The current depentent tfjs-core is 2.7.0, We have to disable it until the
tfjs-core gets upgraded.

This PR also suppresses some failure cases in webgpu.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4492)

### DIFF
--- a/tfjs-backend-webgpu/src/index.ts
+++ b/tfjs-backend-webgpu/src/index.ts
@@ -28,7 +28,9 @@ import * as webgpu from './webgpu';
 registerBackend('webgpu', async () => {
   // Remove it once we figure out how to correctly read the tensor data before
   // the tensor is disposed in profiling mode.
-  env().setFlags({'CHECK_COMPUTATION_FOR_ERRORS': false});
+  // Comment it out temporarily since this flag is added in tfjs-core 2.8.0.
+  // Otherwise, it will report 'it has not been registered' error.
+  // env().set('CHECK_COMPUTATION_FOR_ERRORS', false);
 
   const glslang = await glslangInit();
   const gpuDescriptor: GPURequestAdapterOptions = {
@@ -39,12 +41,10 @@ registerBackend('webgpu', async () => {
   const adapter = await navigator.gpu.requestAdapter(gpuDescriptor);
   let deviceDescriptor: GPUDeviceDescriptor = {};
   const supportTimeQuery =
-    adapter.extensions.includes('timestamp-query' as GPUExtensionName);
+      adapter.extensions.includes('timestamp-query' as GPUExtensionName);
 
   if (supportTimeQuery) {
-    deviceDescriptor = {
-      extensions: ['timestamp-query' as const]
-    };
+    deviceDescriptor = {extensions: ['timestamp-query' as const ]};
   }
   const device: GPUDevice = await adapter.requestDevice(deviceDescriptor);
   return new WebGPUBackend(device, glslang, supportTimeQuery);
--- a/tfjs-backend-webgpu/src/setup_test.ts
+++ b/tfjs-backend-webgpu/src/setup_test.ts
@@ -252,6 +252,7 @@ const TEST_FILTERS: TestFilter[] = [
       'has zero in its shape',           // Test times out.
       'valueAndGradients',               // backend.sum() not yet implemented.
       'upcasts when dtypes dont match',  // Missing cast().
+      'broadcast',  // matmul broadcasting not yet implemented.
     ]
   },
   {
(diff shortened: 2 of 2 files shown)


qjia7:
@pyu10055 @lina128 Please take a look, thanks.
Without this fixing, lots of webgpu tests fail due to `'WEBGPU_CPU_FORWARD': false` not working. All failed cases are going to cpu path. This is another issue. Our modularization work will fix it.

pyu10055 (changes requested):
thanks

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4492#-:-MQNjx278gbtAXt4ZF1J:b-k73z76)* status: 0 of 1 approvals obtained (waiting on @qjia7)

___
*[tfjs-backend-webgpu/src/index.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4492#-MQNjo5o67RidRaOeRyG:-MQNjo5o67RidRaOeRyH:b-i20wwx) ([raw file](https://github.com/tensorflow/tfjs/blob/ee4cbf046d7406abfedd97fa2242afdd4c517bce/tfjs-backend-webgpu/src/index.ts#L33)):*
> ```XML
>   // Comment it out temporarily since this flag is added in tfjs-core 2.8.0.
>   // Otherwise, it will report 'it has not been registered' error.
>   // env().set('CHECK_COMPUTATION_FOR_ERRORS', false);
> ```

this is still commented out, do you need to disable this flag in order for test to pass?

qjia7:
> thanks
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4492#-:-MQNjx278gbtAXt4ZF1J:b-k73z76)_ status: 0 of 1 approvals obtained (waiting on @qjia7)
> 
> _[tfjs-backend-webgpu/src/index.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4492#-MQNjo5o67RidRaOeRyG:-MQNjo5o67RidRaOeRyH:b-i20wwx) ([raw file](https://github.com/tensorflow/tfjs/blob/ee4cbf046d7406abfedd97fa2242afdd4c517bce/tfjs-backend-webgpu/src/index.ts#L33)):_
> 
> > ```
> >   // Comment it out temporarily since this flag is added in tfjs-core 2.8.0.
> >   // Otherwise, it will report 'it has not been registered' error.
> >   // env().set('CHECK_COMPUTATION_FOR_ERRORS', false);
> > ```
> 
> this is still commented out, do you need to disable this flag in order for test to pass?

It's on purpose. Otherwise, it will complain that `Error: Cannot set flag CHECK_COMPUTATION_FOR_ERRORS as it has not been registered.`. Commenting out this line can also make test pass. This flag is only for profiling mode. But currently we haven't enable the profiling tests. SO it's ok to just comment it out. I will remove `//` before `env().set` after upgrading to tfjs-2.8.0 for webgpu backend so that we can enable the profiling mode.

qjia7:
Suppress matmul broadcasting cases. After applying this PR, there are still 3 cases failed when running `yarn test --no-single-run`.  Our following patches will fix them. `matmul braodcasting` and `mirrorPad` won't be fixed in short term. So just suppress them now.

Please take another look. Thanks.

pyu10055 (commented):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4492)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4492)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[tfjs-backend-webgpu/src/index.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4492#-MQNjo5o67RidRaOeRyG:-MQU39eq9TzgA_KyCxgC:bd5l6mr) ([raw file](https://github.com/tensorflow/tfjs/blob/ee4cbf046d7406abfedd97fa2242afdd4c517bce/tfjs-backend-webgpu/src/index.ts#L33)):*

It's on purpose. Otherwise, it will complain that `Error: Cannot set flag CHECK_COMPUTATION_FOR_ERRORS as it has not been registered.`. Commenting out this line can also make test pass. This flag is only for profiling mode. But currently we haven't enable the profiling tests. SO it's ok to just comment it out. I will remove `//` before `env().set` after upgrading to tfjs-2.8.0 for webgpu backend so that we can enable the profiling mode. 
</blockquote></details>

thanks!

qjia7:
@pyu10055 @lina128, Have rebased it. Please help merge it, thanks.

PR #4496, opened 2021-01-06
enable createImageBitmap test only for chrome

pyu10055:
This will fix the nightly build failure that is caused by https://github.com/tensorflow/tfjs/pull/4374

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4496)

### DIFF
--- a/tfjs-core/src/ops/from_pixels_test.ts
+++ b/tfjs-core/src/ops/from_pixels_test.ts
@@ -283,25 +283,28 @@ describeWithFlags('fromPixels', BROWSER_ENVS, () => {
     expectArraysClose(pixelsData, actualInt32, 10);
   });
 
-  it('fromPixels for ImageBitmap', async () => {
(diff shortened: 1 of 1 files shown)


mattsoulanille:
LGTM. Thanks!

shaoboyan:
Oops, Really sorry. ImageBitmap is not supported by all browsers now (https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap). But I think all browsers will commit to support it for WebGPU (@kainino0x  for double check) ?
And I'm not sure whether we could have try bots which covers more OS.

kainino0x:
> ImageBitmap is not supported by all browsers now (https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap). But I think all browsers will commit to support it for WebGPU (@kainino0x for double check) ?

Yup, that is the plan.

PR #4513, opened 2021-01-09
Add safeguard to initializer field.

lina128:
TFX has some logic in exporter to remove some preprocessing nodes that are not included in the final graph, but collection_def still has table_initializer with an empty node list. Ideally, this should be fixed in the exporter code. Adding safeguard code on our side to only set modelInitializer field if node list is not empty.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4513)


pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4513)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

PR #4518, opened 2021-01-11
Avoid building tfjs-core twice in CI

mattsoulanille:
tfjs-core's 'build-deps-always' cloudbuild rule runs 'yarn build-deps-ci', which, due to a circular dependency with tfjs-backend-cpu, first builds tfjs-core before building its dependencies. The 'build' cloudbuild rule then builds tfjs-core again right after tfjs-backend-cpu is built. Since tfjs-core imports tfjs-backend-cpu for testing only, and uses `require()` notation to do so, this rebuild is unnecessary. This commit removes the extra rebuild of tfjs-core.

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4518)

PR #4523, opened 2021-01-12
Always test generate-cloudbuild against the golden files

mattsoulanille:
PR #4518 modified cloudbuild files without updating the golden files used for testing the cloudbuild file generator, but it passed CI testing. This PR fixes that by ~~making the scripts package depend on all other packages, which means the cloudbuild generator will always be tested if a cloudbuild file could have been changed~~ special-casing the running of cloudbuild generator tests on every CI run. It also updates the golden files, fixing #4518's issue.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4523)


mattsoulanille:
Actually, this is unfortunately going to have to be special-cased instead of handled via the dependency tree. Making `scripts` depend on every other package means whenever we build a package, we also build and test `scripts`, which then requires every other package to be built. If we had dependencies on a file-level instead of package-level, this would be possible.

lina128 (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4523#-:-MQrmO_e99UAAMXYzXjN:bb6ywmx)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

Issue #4539, opened 2021-01-15
Regresssion: WASM fails to load due to _scriptDir not defined


vladmandic:
This is a regression in `tfjs` 2.8.4 compared to 2.8.3  

Setting backend to `wasm` on the main thread results in error:
```
worker.js onmessage() captured an uncaught exception: ReferenceError: _scriptDir is not defined
  threadPrintErr @ e67a52fd-9fdd-4d80-87b0-5a39d7cd567d:1
  onmessage @ e67a52fd-9fdd-4d80-87b0-5a39d7cd567d:1

ReferenceError: _scriptDir is not defined
  WasmBackendModuleThreadedSimd (a0196e5e-0cae-4d78-90e3-7f0714096dfc:1)
  onmessage (e67a52fd-9fdd-4d80-87b0-5a39d7cd567d:1)
```

No issues running TFJS 2.8.3 or older.

This is likely related to pull request #4528

jinjingforever:
Hi Vladimir, 

Thanks for reporting! This is a little bit strange because we did test it in 2.8.4. See this codepin: https://codepen.io/jing-jin/pen/BaLqJoE. If you click "Settings", you can see that the program is loading tfjs core and wasm backend @ 2.8.4.

Could you please provide more context/sample code about this issue? Thanks!

vladmandic:
Ok, this is related to both #4392 and #4528.  

Originally `tfjs-backend-wasm` was not fully minification safe when running through a bundler - minify whitespace and minify syntax was ok, but minify identifiers was not. Then #4528 fixed it and everything was fine for a while.

Now #4392 broke it once more and it's even worse since not even minify syntax works anymore.  
So yes, it's a regression.  

If you want to see a simple project, it's at <https://github.com/vladmandic/stocks> and bundler runs via 'server/build.js'

vladmandic:
closing as fixed in tfjs 2.8.5.  
thank you!

PR #4562, opened 2021-01-15
Fix WASM JS script not working after minification

jinjingforever:
In my previous PR, I replaced ```if(_scriptDir)``` with ```if(typeof _scriptDir !== "undefined" && _scriptDir)``` in the emscripten-generated JS script. I did it in backend_wasm.ts where the script content is imported. But it doesn't really work when end users minify the code through a bundler, because "if(_scriptDir)" cannot be found anymore in the minified code.

To make this work, the code replacement has to happen right after the code is generated, which is in build-wasm.sh. I added a node.js script to do the code replacement there.

Another thing I had to do is that by default, terser (the plugin we use to minify code in rollup) will transform ```typeof _scriptDir == "undefined"``` into ```_scriptDir === void 0```, which will break our code (it will throw "_scriptDir is not defined" error). To prevent this, I need to turn off that option in terser in WASM's rollup.config.js. I also updated the README.md file to let users know they need to turn this option off if they want to minified the code by themselves. 

I tested this in various situations:
- Use full script through script tag
- Use our own minified script through script tag
- Use bundler with minification (esbuild, the one used in https://github.com/tensorflow/tfjs/issues/4539)

Because I based this fix on top of the version after Yannick's merge, so please ignore rollup.config.helpers.js (which was added in Yannick's PR). For the tfjs-backend-wasm/rollup.config.js file, please only look at line 124-129. The rest of the changes in that file came from Yannick's PR that makes rollup.config.helper.js work.

Thanks!

Fixed https://github.com/tensorflow/tfjs/issues/4539

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4562)

### DIFF
--- a/rollup.config.helpers.js
+++ b/rollup.config.helpers.js
@@ -1,0 +1,114 @@
+/**
+ * @license
+ * Copyright 2018 Google LLC. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


lina128 (approved):
Thank you Jing!

PR #4614, opened 2021-01-28
Mark imports as ignored in the browser

mattsoulanille:
This change marks several node-specific imports as ignored in the browser. It's necessary for tfjs-react-native to build on 3.0.x. Otherwise, Metro tries to include them in a react-native build, which breaks the build as they're not available outside of node. 

Since tfjs-react-native is not a part of the monorepo and pulls `@tensorflow/...` dependencies from npm (instead of using `link:...`), this change has to be _released_ before tfjs-react-native can be updated to 3.0.x and pass its CI tests.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4614)

### DIFF
--- a/tfjs-core/package.json
+++ b/tfjs-core/package.json
@@ -90,7 +90,8 @@
   "browser": {
     "node-fetch": false,
     "util": false,
-    "crypto": false
+    "crypto": false,
+    "worker_threads": false
   },
   "sideEffects": [
     "./dist/index.js",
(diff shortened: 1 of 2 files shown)


tafsiri (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4614)* status: :shipit: complete! 1 of 1 approvals obtained

PR #4616, opened 2021-01-29
webgpu: Support channel = 3 with conv2d vec4

qjia7:
PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4616)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
@@ -96,6 +96,9 @@ export class Conv2DMMVec4Program implements WebGPUProgram {
         if (divBy4Remainder == 0) {
           temp = curData;
         } else {
+          // TODO: This could end up being a redundant load with another one in
+          // the same shader invocation. Perhaps there's an opportunity for
+          // optimization
           vec4 nextData = x[divBy4Index + 1];
           if (divBy4Remainder == 1) {
             temp = vec4(curData.yzw, nextData.x);
--- a/tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
@@ -109,7 +109,9 @@ export class Conv2DMMVec4Program implements WebGPUProgram {
 
     const remainder = this.convInfo.inChannels % 4;
     const remainderSnippet = remainder === 0 ?
-        `resData = coordsInBounds(coord, ${
+        `// The bounds checking is always needed since we use it to pad zero for
+        // the 'same' padding type.
+        resData = coordsInBounds(coord, ${
             getShapeCoords(this.convInfo.inShape)}) ? x[getFlatIndex(coord, ${
             getShapeCoords(
                 this.convInfo.inShape)}) / 4] : vec4(0.0, 0.0, 0.0, 0.0);` :
--- a/tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
@@ -93,11 +93,57 @@ export class Conv2DMMVec4Program implements WebGPUProgram {
             getShapeCoords(this.convInfo.inShape)}) ? x[getFlatIndex(coord, ${
             getShapeCoords(
                 this.convInfo.inShape)}) / 4] : vec4(0.0, 0.0, 0.0, 0.0)`;
+    const sampleAWithRemainder = `int flatIndex = getFlatIndex(coord, ${
+        getShapeCoords(this.convInfo.inShape)});
+        int divBy4Remainder = flatIndex % 4;
+        int divBy4Index = flatIndex / 4;
+        vec4 curData = x[divBy4Index];
+        if (divBy4Remainder == 0)
+        {
+          temp = curData;
+        } else {
+          vec4 nextData = x[divBy4Index + 1];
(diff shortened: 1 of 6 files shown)


qjia7:
@lina128 @kainino0x Please take a look, thanks.
See over 40% speedup with `x[1, 231, 231, 3], filter[7, 7, 3, 64]` on UHD 630.

lina128 (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4616#-:-MSRmbmx4I_A61jse8vs:bb6ywmx)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @kainino0x)

tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
```diff
@@ -93,11 +93,57 @@ export class Conv2DMMVec4Program implements WebGPUProgram {
             getShapeCoords(this.convInfo.inShape)}) ? x[getFlatIndex(coord, ${
             getShapeCoords(
                 this.convInfo.inShape)}) / 4] : vec4(0.0, 0.0, 0.0, 0.0)`;
+    const sampleAWithRemainder = `int flatIndex = getFlatIndex(coord, ${
+        getShapeCoords(this.convInfo.inShape)});
+        int divBy4Remainder = flatIndex % 4;
+        int divBy4Index = flatIndex / 4;
+        vec4 curData = x[divBy4Index];
+        if (divBy4Remainder == 0)
+        {
```
kainino0x, line 102: μNit: Please match the style of the surrounding JS code (brace on same line)

tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
```diff
@@ -93,11 +93,57 @@ export class Conv2DMMVec4Program implements WebGPUProgram {
...
+          temp = curData;
+        } else {
+          vec4 nextData = x[divBy4Index + 1];
```
kainino0x, line 105: nonblocking: Could this end up being a redundant load with another one in the same shader invocation (for curData at another index, e.g. if work-per-thread > 1)? Perhaps there's an opportunity for optimization (though it might be tricky).

tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
```diff
@@ -113,7 +159,7 @@ export class Conv2DMMVec4Program implements WebGPUProgram {
         throw new Error('Leakyrelu is not supported.');
       } else {
         activationSnippet = `
-        vec4 activation(vec4 a, ivec4 outCoord) {
+          vec4 activation(vec4 a, ivec4 outCoord) {
```
kainino0x, line 162: nit: should be indented the same amount as the corresponding `}`

kainino0x (approved):

tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
qjia7: Done

tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
qjia7: Do you mean put nextData next to curData like below? If yes, done.
``` 
        vec4 curData = x[divBy4Index];
        vec4 nextData = x[divBy4Index + 1];

tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
qjia7: Done.

qjia7:
@kainino0x  I did some code adjustment in the last change. Please take another look, thanks.

qjia7:
Also apply the fix to conv2d non-vec4 version. The error happens when the tile size fits the [dimAOuter, dimInner], but there are paddings needed using the 'same' mode.

No new test case is added. When workgroupsize = [16, 16, 1], workPerThread=[1, 1, 1], the existed test `fused conv2d test-webgpu {"WEBGPU_CPU_FORWARD":false} prelu bias stride 2 x=[1,8,8,16] f=[3,3,16,1] s=[2,2] d=8 p=same` will catch the error.

tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
kainino0x: What I mean is, could there be two calls to mm_readA in the same thread that result in loads like:
`nextData = x[4 + 1]`
`curData = x[5]`
? Deduplicating them would be hard (because the load is deep inside mm_readA), but probably saves some bandwidth.

> Do you mean put nextData next to curData like below? If yes, done.

I don't think this should be done. It could cause an unnecessary load in some cases (though the compiler may optimize to avoid that).

kainino0x (approved):

tfjs-backend-webgpu/src/kernels/conv2d_mm_vec4_webgpu.ts
qjia7: Sorry for the misunderstanding. I thought your meaning was putting nextData and curData together would make the driver better to fetch data. Otherwise, in some case, the driver may hit unoptimized path since they are not adjacent in code even though nextData and curData are continuous in memory.

Yes, I agree that `curData` may be the same data with previous mm_readA's `nextData`. But it's hard to do the opt, which may make the code very complicated. I will leave a hint here for possible optimization.

PR #4564, opened 2021-01-16
Fix memory leak for complex tensors (#4563)

pyu10055:
BUG
* would work

* ensure real and imag tensor would not get disposed if ref count > 0

* fix typo

* fixed engine refCount mismatch with webgl backend, fixed webgl simpleSlide that accidentally copied parent tensor refCount and parentRefCount;

* add stft test to cpu

* only reduce refCount if the tensor is part of complex tensor
;

Co-authored-by: Ann Yuan <annyuan@gmail.com>

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4564)

PR #4628, opened 2021-02-01
Engine/backend refCount refactoring

pyu10055:
The PR aims to clean up the refCount mechanism in both engine and backends. This follows the [PR that fixed the complex tensor memory leak](https://github.com/tensorflow/tfjs/pull/4564), which caused by mismatch of refCount between engine and WebGL backend.
Here are main changes in this PR:
1. Removed the refCount from engine TensorInfo, engine only tracks the tensors and mapping between DataId and Backend.
2. Backend disposeData method is called on the dataId of the tensor for every disposal and backend will determine if the dataId is ready to be disposed, it will return a boolean to indicated success of the disposal. Engine will only remove the DataId mapping when it is disposed successfully in the backend.
3. When Engine calls makeTensor or makeTensorFromDataId from backend dataId, it would just track the mapping of dataId to backend, but makeVariable needs to increase backend refCount, since it is an extra reference from the variable to the dataId that exists only in the engine.
4. Engine clone tensor for gradient would need to call backend Identity op to increase ref count in the backend to prevent disposal prematurely.
5. Removed complexParentRefCount from WebGL, since complex components can be tracked sufficiently by backend refCount.
6. Added refCount tracking to node and WASM backend.
7. compileAndRun method in WebGL should return TensorInfo instead of Tensor, since it is used in the Backend ops, returning Tensor would add engine tracking to the DataId and cause unnecessary disposal by scope.
8. When DataId is moved across backends, the source backend will force disposal, and the refCount is copied to the destination backend, this will ensure no premature disposal in the destination backend.

This would affect WebGPU backend due to backend interface changes. @qjia7 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628)


mattsoulanille (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @pyu10055, and @tafsiri)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 532 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUDEotDbtxOHOzuUaO:-MSUDEou7J_YYc_H2vh2:b-kzhfg9) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-backend-webgl/src/backend_webgl.ts#L532)):*
> ```XML
> 
>   /**
>    * Dispose the memory if the dataId has 0 refCount. Return true if the memory
> ```

Nit: I'd mention that this function also decrements the refCount.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 556 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUIJII17q6aVVupCJ9:-MSUIJII17q6aVVupCJA:bhrk4ym) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-backend-webgl/src/backend_webgl.ts#L556)):*
> ```XML
>       this.pendingDisposal.add(dataId);
>       this.pendingDeletes++;
>       return false;
> ```

This data will eventually be disposed (line 385), so how do we notify things like disposeTensor (engine.ts) that this has happened? Does that matter?
___
*[tfjs-core/src/engine.ts, line 429 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUFr6k7bkBzPBwd3Y2:-MSUFr6k7bkBzPBwd3Y3:br6dmnu) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-core/src/engine.ts#L429)):*
> ```XML
>     // Delete the tensor from the old backend and move it to the new
>     // backend.
>     srcBackend.disposeData(dataId, true);
> ```

Do we need to check if the tensor was actually disposed, or does that not actually matter? `backend_webgl.ts`'s `disposeData` may return false if `this.pendingDisposal.has(dataId)` is true even if `force = true` is passed.
___
*[tfjs-core/src/engine.ts, line 903 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUHM3x9rVjVugTstwK:-MSUHM3x9rVjVugTstwL:bj10m71) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-core/src/engine.ts#L903)):*
> ```XML
> 
>     // Remove the reference to dataId if backend dispose the data successfully
>     if (info.backend.disposeData(a.dataId)) {
> ```

What happens if backend_webgl has the tensor in `pendingDisposal` and returns false here but eventually disposes it later?

pyu10055 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, and @tafsiri)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 532 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUDEotDbtxOHOzuUaO:-MSURhsm5Mp6hDpYQ0hp:bb7f2mf) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-backend-webgl/src/backend_webgl.ts#L532)):*

Nit: I'd mention that this function also decrements the refCount.
</blockquote></details>

done
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 556 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUIJII17q6aVVupCJ9:-MSURd8sBAVB3X2_QnZb:b-kd1mqu) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-backend-webgl/src/backend_webgl.ts#L556)):*

This data will eventually be disposed (line 385), so how do we notify things like disposeTensor (engine.ts) that this has happened? Does that matter?
</blockquote></details>

This is a good question, currently this is async op, it should callback engine to remove the dataId entry after done (check backend in case dataId is moved).
___
*[tfjs-core/src/engine.ts, line 429 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUFr6k7bkBzPBwd3Y2:-MSUS60WCSPqgpCAT9ZD:b-55ov42) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-core/src/engine.ts#L429)):*

Do we need to check if the tensor was actually disposed, or does that not actually matter? `backend_webgl.ts`'s `disposeData` may return false if `this.pendingDisposal.has(dataId)` is true even if `force = true` is passed.
</blockquote></details>

I think the result would not matter, since we assume the backend will disposed it as soon as it can, and track it in pending disposal queue to later disposal. We need to reduce the ref count, so it will be disposed for sure.
___
*[tfjs-core/src/engine.ts, line 903 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUHM3x9rVjVugTstwK:-MSUSjdx6FiUHGtbrH8Y:bk3c636) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-core/src/engine.ts#L903)):*

What happens if backend_webgl has the tensor in `pendingDisposal` and returns false here but eventually disposes it later?
</blockquote></details>

Good catch, we need to add engine to the pendingDisposal listener to clean up here.

mattsoulanille (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @pyu10055, and @tafsiri)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 556 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUIJII17q6aVVupCJ9:-MSUb9ps1wYZG31Uq7RN:b-77ml27) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-backend-webgl/src/backend_webgl.ts#L556)):*

This is a good question, currently this is async op, it should callback engine to remove the dataId entry after done (check backend in case dataId is moved).
</blockquote></details>

This seems like a good solution (+ the callback in read). 
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 364 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUdbD-Es2C67uKY6nC:-MSUdbD-Es2C67uKY6nD:bncgtbp) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L364)):*
> ```XML
> 
>     // local copy of the backend
>     const that = this;
> ```

nit: I don't think this is necessary. I think you can just pass `this` in `engine().removeDataId(dataId, ...)`, but I may be wrong.
___
*[tfjs-core/src/engine.ts, line 429 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUFr6k7bkBzPBwd3Y2:-MSU_bFF2N5CG55uAOt_:b-d95zgv) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-core/src/engine.ts#L429)):*

I think the result would not matter, since we assume the backend will disposed it as soon as it can, and track it in pending disposal queue to later disposal. We need to reduce the ref count, so it will be disposed for sure.
</blockquote></details>

That makes sense.

lina128 (commented):
This is really elegant solution, great work Ping!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628#-:-MSZQHqO1trVTb27-P4t:b-v94kma)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever, @mattsoulanille, @pyu10055, and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 3 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZBCw02IBwRbQUXZCC:-MSZBCw02IBwRbQUXZCD:bb6ht36) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L3)):*
> ```XML
> /**
>  * @license
>  * Copyright 2017 Google LLC. All Rights Reserved.
> ```

2021
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 98 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZAkv229DTZa0P2fg_:-MSZAkv229DTZa0P2fga:b-2qjfz3) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L98)):*
> ```XML
>   /** Return refCount of a `TensorData`. */
>   refCount(dataId: DataId): number {
>     const tensorData = this.data.get(dataId);
> ```

Do you need to first check whether dataId exists in this backend? The get api will trigger moveData if dataId doesn't exist.
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 166 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZCb-I3wCYHnvO4PL5:-MSZCb-I3wCYHnvO4PL6:b-9k00k4) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L166)):*
> ```XML
>   /**
>    * Dispose the memory if the dataId has 0 refCount. Return true if the memory
>    * is released, false otherwise.
> ```

Return true if the memory is released or memory is not managed in this backend, false if memory is not cleared.
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 170 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZB8wQ-qV9wR0_6sDz:-MSZB8wQ-qV9wR0_6sE-:b8n9fto) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L170)):*
> ```XML
>    * @oaram force Optional, remove the data regardless of refCount
>    */
>   disposeData(dataId: DataId, force?: boolean): boolean {
> ```

force: boolean = false
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 180 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZCOpZ1upQhdqIJjCr:-MSZCOpZ1upQhdqIJjCs:bqqmh2p) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L180)):*
> ```XML
> 
>       if (complexTensorInfos != null) {
>         this.disposeData(complexTensorInfos.real.dataId);
> ```

Should force be true?
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 181 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZCUF25_cZNK-LCb1g:-MSZCUF25_cZNK-LCb1h:b-816h6l) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L181)):*
> ```XML
>       if (complexTensorInfos != null) {
>         this.disposeData(complexTensorInfos.real.dataId);
>         this.disposeData(complexTensorInfos.imag.dataId);
> ```

Same.
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 189 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZGg0d9kh7bEzfkbRu:-MSZGg0d9kh7bEzfkbRv:b-ra936d) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L189)):*
> ```XML
>   }
> 
>   disposeIntermediateTensorInfo(tensorInfo: TensorInfo): void {
> ```

In this method, why not just call disposeData(tensorInfo.dataId, false /* force */)?
___
*[tfjs-backend-wasm/src/backend_wasm.ts, line 179 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZHCxu6daRHlBNfz20:-MSZHCxu6daRHlBNfz21:bgsb6ru) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-wasm/src/backend_wasm.ts#L179)):*
> ```XML
>     } else {
>       const id = this.dataIdNextNumber++;
>       dataId = {id};
> ```

Brilliant! This will make debugging memory easier!
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 39 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZHYJ861bdDfl_c7xw:-MSZHYJ861bdDfl_c7xx:b-aj3zw1) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-wasm/src/kernels/Reshape.ts#L39)):*
> ```XML
>           `shape must have the same number of elements.`);
> 
>   // Backend needs to track refCount for the dataId for reshape op
> ```

Also check any wasm kernel that calls reshape, they need to decRef.
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 40 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZPp1000bwELPdkVm9:-MSZPp1000bwELPdkVmA:btp6oqu) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-wasm/src/kernels/Reshape.ts#L40)):*
> ```XML
> 
>   // Backend needs to track refCount for the dataId for reshape op
>   args.backend.incRef(x.dataId);
> ```

e2e has a memory_leak.ts test, let's make sure wasm also pass that test. Please add wasm to test env in setup_test.ts
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 109 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZHkaY0kzzzJdWMie7:-MSZHkaY0kzzzJdWMie8:b-qiqz6q) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L109)):*
> ```XML
> 
>   private static nextDataId = 0;
>   private nextDataId(): number {
> ```

Can we have this for cpu too? This is very convenient for memory debugging.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 192 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZI0eJ4iAiuLAj1q3W:-MSZI0eJ4iAiuLAj1q3X:bmtmcom) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L192)):*
> ```XML
>   /** Return refCount of a `TensorData`. */
>   refCount(dataId: DataId): number {
>     const tensorData = this.texData.get(dataId);
> ```

Check has dataId first.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 228 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZIFed8AKLBc9-GRDO:-MSZIFed8AKLBc9-GRDP:b-212gcv) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L228)):*
> ```XML
>     const dataId = tensorInfo.dataId;
> 
>     if (this.texData.has(dataId)) {
> ```

You don't need this check, right?
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 364 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUdbD-Es2C67uKY6nC:-MSZJ2OI2MBk7supgOfj:b32v57) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L364)):*

nit: I don't think this is necessary. I think you can just pass `this` in `engine().removeDataId(dataId, ...)`, but I may be wrong.
</blockquote></details>

+1
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 536 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZJ83t6N15ARp5ITjG:-MSZJ83t6N15ARp5ITjH:bpzdzya) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L536)):*
> ```XML
> 
>   /**
>    * Decrease the RefCount on the dataId and dispose the memory if the dataId
> ```

There's more logic than what is described here, right? Maybe something like: Soft dispose data based on refCount and dataRefCount. When memory is cleared or memory no longer exists for this backend, return true.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 552 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZMJn3Beip0lKMKnOz:-MSZMJn3Beip0lKMKnP-:brf1ywf) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L552)):*
> ```XML
> 
>     // if force flag is set, change refCount to 0, this would ensure disposal
>     // when added to the pendingDisposal queue.
> ```

Sugg add at the end of above annotation: Memory may or may not be released, which also depends on dataRefCount, see `releaseGPU`.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 572 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZLips9BCzBWLoPq7R:-MSZLips9BCzBWLoPq7S:bpup4ld) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L572)):*
> ```XML
>     const {complexTensorInfos} = this.texData.get(dataId);
>       this.disposeIntermediateTensorInfo(complexTensorInfos.real);
> ```

Can we just call disposeData?
___
*[tfjs-backend-webgl/src/kernels/Complex.ts, line 29 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZM-RD8BJKHMM7_k3E:-MSZM-RE1ywa5fugWX7_:biifp12) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/kernels/Complex.ts#L29)):*
> ```XML
>  * component. The refCounts are increased by the identity call.
>  *
>  * When a complex tensor is disposed, it will reduce the refCount on the
> ```

This becomes so elegant!
___
*[tfjs-core/src/engine.ts, line 854 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZMtTB57DAyomFabvj:-MSZMtTB57DAyomFabvk:bdvnty1) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-core/src/engine.ts#L854)):*
> ```XML
>       this.state.numStringTensors++;
>     }
>     // Bytes for complex numbers are counted by their components. Bytes for
> ```

Curious, do we want bytes calculation be moved to backends too? Because different backends can use different bytes even for the same kernel.
___
*[tfjs-core/src/engine.ts, line 879 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZNQGz9rNTEA2-HUZZ:-MSZNQGz9rNTEA2-HUZ_:bkdo9kn) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-core/src/engine.ts#L879)):*
> ```XML
>   // Track the tensor by dataId and increase the refCount for the dataId in the
>   // backend.
>   incRef(a: Tensor, backend: KernelBackend): void {
> ```

Where will this be used?
___
*[tfjs-core/src/engine.ts, line 911 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZNzXp3vlRxD1y8ria:-MSZNzXp3vlRxD1y8rib:by85eu6) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-core/src/engine.ts#L911)):*
> ```XML
>     // Remove the reference to dataId if backend dispose the data successfully
>     if (info.backend.disposeData(a.dataId)) {
>       this.state.tensorInfo.delete(a.dataId);
> ```

Why not just call removeDataId?

pyu10055 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 3 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZBCw02IBwRbQUXZCC:-MSZWiT8CTfgXOnPEUIS:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L3)):*

2021
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 98 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZAkv229DTZa0P2fg_:-MSZSVFr22xCa_g5arEj:bgv60gk) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L98)):*

Do you need to first check whether dataId exists in this backend? The get api will trigger moveData if dataId doesn't exist.
</blockquote></details>

good point, added.
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 166 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZCb-I3wCYHnvO4PL5:-MSZSm7ODCBB9gCpuOYb:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L166)):*

Return true if the memory is released or memory is not managed in this backend, false if memory is not cleared.
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 170 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZB8wQ-qV9wR0_6sDz:-MSZT9cD47X6UqKwiDQt:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L170)):*

force: boolean = false
</blockquote></details>

Done.
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 180 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZCOpZ1upQhdqIJjCr:-MSZTaT33580K8vhMKoN:b9hzuxv) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L180)):*

Should force be true?
</blockquote></details>

should it use the passed in force value?
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 181 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZCUF25_cZNK-LCb1g:-MSZTeWa8O2jryU7Xsx6:bnj7hl0) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L181)):*

Same.
</blockquote></details>

same as above
___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 189 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZGg0d9kh7bEzfkbRu:-MSZU0T040RB86Ap3BEs:b6a2t1l) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L189)):*

In this method, why not just call disposeData(tensorInfo.dataId, false /* force */)?
</blockquote></details>

yes, updated.
___
*[tfjs-backend-wasm/src/backend_wasm.ts, line 179 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZHCxu6daRHlBNfz20:-MSZU4DCCTNQqVJZhxbF:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-wasm/src/backend_wasm.ts#L179)):*

Brilliant! This will make debugging memory easier!
</blockquote></details>

Done.
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 39 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZHYJ861bdDfl_c7xw:-MSZUGAC6bSpi14m2PGX:bpwh4a2) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-wasm/src/kernels/Reshape.ts#L39)):*

Also check any wasm kernel that calls reshape, they need to decRef.
</blockquote></details>

can you explain why caller needs to decRef?
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 40 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZPp1000bwELPdkVm9:-MSalTF66K0tkTzFGhi2:bcbza3a) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-wasm/src/kernels/Reshape.ts#L40)):*

e2e has a memory_leak.ts test, let's make sure wasm also pass that test. Please add wasm to test env in setup_test.ts
</blockquote></details>

There are some issue add wasm backend to integration test, I added memory test to the ops that involve reshape. I think we need to have a better memory test refactoring across all ops, seems memory test is rare.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 556 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSUIJII17q6aVVupCJ9:-MSZL1TD4ZmCycoHFryv:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/b761f2e54f4b9358c31c0ab3cd2f9aa1d00b78bb/tfjs-backend-webgl/src/backend_webgl.ts#L556)):*

This seems like a good solution (+ the callback in read). 
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 109 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZHkaY0kzzzJdWMie7:-MSZX7oG0hDAGlgEBgRD:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L109)):*

Can we have this for cpu too? This is very convenient for memory debugging.
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 192 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZI0eJ4iAiuLAj1q3W:-MSZUOilEe5SYnC12YjP:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L192)):*

Check has dataId first.
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 228 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZIFed8AKLBc9-GRDO:-MSZUYGD1be6T_YS1J1I:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L228)):*

You don't need this check, right?
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 536 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZJ83t6N15ARp5ITjG:-MSZVKoyD6Mdwva2tC7l:b5dyr55) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L536)):*
> Memory may or may not be released, which also depends on dataRefCount, see releaseGPU.
updated.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 552 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZMJn3Beip0lKMKnOz:-MSZV_VD0d5DU2456Ty8:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L552)):*

Sugg add at the end of above annotation: Memory may or may not be released, which also depends on dataRefCount, see `releaseGPU`.
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 572 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZLips9BCzBWLoPq7R:-MSZVeLr177T1XgqBgGJ:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/backend_webgl.ts#L572)):*

Can we just call disposeData?
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/kernels/Complex.ts, line 29 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZM-RD8BJKHMM7_k3E:-MSZVfH68bmLKDenI7SP:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-webgl/src/kernels/Complex.ts#L29)):*

This becomes so elegant!
</blockquote></details>

Done.
___
*[tfjs-core/src/engine.ts, line 854 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZMtTB57DAyomFabvj:-MSZVisA9RahbNBSIaGp:bog0w7p) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-core/src/engine.ts#L854)):*

Curious, do we want bytes calculation be moved to backends too? Because different backends can use different bytes even for the same kernel.
</blockquote></details>

This is a good idea, and I think it can be moved to backend as well. To keep the size of this PR, I will do it in a separate PR.
___
*[tfjs-core/src/engine.ts, line 879 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZNQGz9rNTEA2-HUZZ:-MSZW83B2nUzfxZg0-Ur:b-p38ml5) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-core/src/engine.ts#L879)):*

Where will this be used?
</blockquote></details>

For variable creation.
___
*[tfjs-core/src/engine.ts, line 911 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZNzXp3vlRxD1y8ria:-MSZWcXwA4P3vSLwdoom:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-core/src/engine.ts#L911)):*

Why not just call removeDataId?
</blockquote></details>

Done.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, @pyu10055, and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 180 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZCOpZ1upQhdqIJjCr:-MSiYmzLB6lKUnoYzcfK:b-dc9zip) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L180)):*

should it use the passed in force value?
</blockquote></details>

At this point, you do want to delete real and imag, right? We already know we are going to delete the parent tensorInfo.
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 39 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZHYJ861bdDfl_c7xw:-MSiZB_F0givS2eCzniC:b89zph0) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-wasm/src/kernels/Reshape.ts#L39)):*

can you explain why caller needs to decRef?
</blockquote></details>

Discussed offline. Intermediate reshape needs to be disposed, just like cpu and webgl backends.
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 40 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZPp1000bwELPdkVm9:-MSiZLQ10Vj_GTt7Xm85:b-pbdqz3) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-wasm/src/kernels/Reshape.ts#L40)):*

There are some issue add wasm backend to integration test, I added memory test to the ops that involve reshape. I think we need to have a better memory test refactoring across all ops, seems memory test is rare.
</blockquote></details>

Agreed.
___
*[tfjs-backend-wasm/src/kernels/Reverse.ts, line 65 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSiZUlb1KhmC7nnKDL0:-MSiZUlb1KhmC7nnKDL1:b9p8vff) ([raw file](https://github.com/tensorflow/tfjs/blob/cdea32c54c2118f98db9b2b487f2609b3f38b283/tfjs-backend-wasm/src/kernels/Reverse.ts#L65)):*
> ```XML
>   const reshaped =
>       reshape({inputs: {x: out}, attrs: {shape: x.shape}, backend});
>   if (out != null) {
> ```

Why check out != null?
___
*[tfjs-backend-wasm/src/kernels/StridedSlice.ts, line 120 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSi_6tb1YU8GxMZvUvw:-MSi_6tb1YU8GxMZvUvx:bbmeqh5) ([raw file](https://github.com/tensorflow/tfjs/blob/cdea32c54c2118f98db9b2b487f2609b3f38b283/tfjs-backend-wasm/src/kernels/StridedSlice.ts#L120)):*
> ```XML
>     const reshaped =
>         reshape({inputs: {x: xSliced}, attrs: {shape: outShape}, backend});
>     if (xSliced != null) {
> ```

Why null checking?
___
*[tfjs-core/src/engine.ts, line 879 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZNQGz9rNTEA2-HUZZ:-MSi_r3n9e-D7D6XayDO:b-cpv8yg) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-core/src/engine.ts#L879)):*

For variable creation.
</blockquote></details>

Just for my understanding, why variable refCount can't be tracked by backend? Can we call reshape where variable needs incRef?

pyu10055 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, @pyu10055, and @tafsiri)

___
*[tfjs-backend-wasm/src/kernels/Reverse.ts, line 65 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSiZUlb1KhmC7nnKDL0:-MSicm3HAWjuIeJC2v9P:b-sey6jg) ([raw file](https://github.com/tensorflow/tfjs/blob/cdea32c54c2118f98db9b2b487f2609b3f38b283/tfjs-backend-wasm/src/kernels/Reverse.ts#L65)):*
> I see other places are doing that, just assume it can be null.
___
*[tfjs-backend-wasm/src/kernels/StridedSlice.ts, line 120 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSi_6tb1YU8GxMZvUvw:-MSicaiO7rH-cCeLI2YH:b-o868vw) ([raw file](https://github.com/tensorflow/tfjs/blob/cdea32c54c2118f98db9b2b487f2609b3f38b283/tfjs-backend-wasm/src/kernels/StridedSlice.ts#L120)):*

Why null checking?
</blockquote></details>

I see other places are doing that, just assume it can be null.
___
*[tfjs-core/src/engine.ts, line 879 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZNQGz9rNTEA2-HUZZ:-MSibbPC5I7Sg73pDWZB:b-oa33t1) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-core/src/engine.ts#L879)):*

Just for my understanding, why variable refCount can't be tracked by backend? Can we call reshape where variable needs incRef?
</blockquote></details>

My understanding is that variable is engine concept and not same as tensor, it used in the back prop, hence no reshape or clone is done on the dataId. It might be able to use Identity op to do that, but I am not sure if there are any implication to the tape.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, @pyu10055, and @tafsiri)

___
*[tfjs-backend-wasm/src/kernels/Reverse.ts, line 65 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSiZUlb1KhmC7nnKDL0:-MSj3rHW0-mcHO6Vllm7:bjuvw52) ([raw file](https://github.com/tensorflow/tfjs/blob/cdea32c54c2118f98db9b2b487f2609b3f38b283/tfjs-backend-wasm/src/kernels/Reverse.ts#L65)):*

> I see other places are doing that, just assume it can be null.
</blockquote></details>

I don't think you need null check. The only places I see null check (like in CropAndResize when dispose castedData) is that they may not be instantiated. In other places, we don't need this check.
___
*[tfjs-backend-wasm/src/kernels/StridedSlice.ts, line 115 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSj47PFCwlgSejgCTq3:-MSi_1-N-MmyjvJmxfMr:b-ltqgy0) ([raw file](https://github.com/tensorflow/tfjs/blob/cdea32c54c2118f98db9b2b487f2609b3f38b283/tfjs-backend-wasm/src/kernels/StridedSlice.ts#L115)):*
> ```XML
>   if (nonStrided) {
>     const xSliced = slice({inputs: {x}, attrs: {begin, size}, backend});
>     if (xReshaped != null) {
> ```

No need for null check.
___
*[tfjs-core/src/engine.ts, line 879 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZNQGz9rNTEA2-HUZZ:-MSj4IoR1PcoyAFUJG79:b-lz13ho) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-core/src/engine.ts#L879)):*

My understanding is that variable is engine concept and not same as tensor, it used in the back prop, hence no reshape or clone is done on the dataId. It might be able to use Identity op to do that, but I am not sure if there are any implication to the tape.
</blockquote></details>

Got it. Let's keep the current mechanism for now. Can you add a TODO to investigate whether we can use Identity?

pyu10055 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, and @tafsiri)

___
*[tfjs-backend-cpu/src/backend_cpu.ts, line 180 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZCOpZ1upQhdqIJjCr:-MSj82DiCNVPoQ-jAsw5:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-cpu/src/backend_cpu.ts#L180)):*

At this point, you do want to delete real and imag, right? We already know we are going to delete the parent tensorInfo.
</blockquote></details>

Done.
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 39 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZHYJ861bdDfl_c7xw:-MSj84Ih8TnVhUJ_Pysi:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-wasm/src/kernels/Reshape.ts#L39)):*

Discussed offline. Intermediate reshape needs to be disposed, just like cpu and webgl backends.
</blockquote></details>

Done.
___
*[tfjs-backend-wasm/src/kernels/Reshape.ts, line 40 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZPp1000bwELPdkVm9:-MSj85Oy2g70mN_RNEh_:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-backend-wasm/src/kernels/Reshape.ts#L40)):*

Agreed.
</blockquote></details>

Done.
___
*[tfjs-backend-wasm/src/kernels/Reverse.ts, line 65 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSiZUlb1KhmC7nnKDL0:-MSj9rDF16Py1spGX2yu:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/cdea32c54c2118f98db9b2b487f2609b3f38b283/tfjs-backend-wasm/src/kernels/Reverse.ts#L65)):*

I don't think you need null check. The only places I see null check (like in CropAndResize when dispose castedData) is that they may not be instantiated. In other places, we don't need this check.
</blockquote></details>

Done.
___
*[tfjs-backend-wasm/src/kernels/StridedSlice.ts, line 115 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSj47PFCwlgSejgCTq3:-MSj90jV0jChMzNlFWXP:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/cdea32c54c2118f98db9b2b487f2609b3f38b283/tfjs-backend-wasm/src/kernels/StridedSlice.ts#L115)):*

No need for null check.
</blockquote></details>

Done.
___
*[tfjs-backend-wasm/src/kernels/StridedSlice.ts, line 120 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSi_6tb1YU8GxMZvUvw:-MSj91W6DHilxKCuJ7qZ:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/cdea32c54c2118f98db9b2b487f2609b3f38b283/tfjs-backend-wasm/src/kernels/StridedSlice.ts#L120)):*

I see other places are doing that, just assume it can be null.
</blockquote></details>

Done.
___
*[tfjs-core/src/engine.ts, line 879 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4628#-MSZNQGz9rNTEA2-HUZZ:-MSj93XO8X6egEghArtp:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/2123c9e75a952926d2279afc270a151819b629da/tfjs-core/src/engine.ts#L879)):*

Got it. Let's keep the current mechanism for now. Can you add a TODO to investigate whether we can use Identity?
</blockquote></details>

Done.

lina128 (approved):
Thank you Ping, great work! LGTM!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4628#-:-MSkFnAG8EynkHFB0_g_:bl6roq0)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, and @tafsiri)

PR #4642, opened 2021-02-05
webgpu: RefCount refactoring

qjia7:
DEV

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4642)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -198,6 +198,9 @@ export class WebGPUBackend extends KernelBackend {
 
   /** Increase refCount of a `TensorData`. */
   incRef(dataId: DataId): void {
+    if (!this.tensorMap.has(dataId)) {
+      throw new Error(`Tensor ${dataId} was not registered!`);
+    }
     const tensorData = this.tensorMap.get(dataId);
     tensorData.refCount++;
   }
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -19,7 +19,7 @@
 
 import './flags_webgpu';
 
-import {backend_util, DataStorage, DataType, engine, env, KernelBackend, RecursiveArray, Tensor, TensorInfo, TimingInfo, util} from '@tensorflow/tfjs-core';
+import {backend_util, DataStorage, DataType, engine, env, KernelBackend, RecursiveArray, TensorInfo, TimingInfo, util} from '@tensorflow/tfjs-core';
 import {Glslang} from '@webgpu/glslang/dist/web-devel/glslang.onefile';
 
 import {BufferManager} from './buffer_manager';
@@ -45,7 +45,8 @@ type BufferInfo = {
 type TensorBufferInfo = {
   values: backend_util.BackendValues,
   dtype: DataType,
-  bufferInfo: BufferInfo
+  bufferInfo: BufferInfo,
+  refCount: number;
 };
 
 interface DataId {}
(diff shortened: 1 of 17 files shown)


qjia7:
@gyagp @axinging @xhcao @haoyunfeix Please take a look, thanks.
CC @shaoboyan

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -170,30 +187,55 @@ export class WebGPUBackend extends KernelBackend {
...
+    if (this.tensorMap.has(dataId)) {
+      const tensorData = this.tensorMap.get(dataId);
+      tensorData.refCount--;
+    }
+  }
+
   write(values: backend_util.BackendValues, shape: number[], dtype: DataType):
       DataId {
-    const dataId = {};
+    const dataId = {id: this.nextDataId()};
```
haoyunfeix, line 215: I'm not sure about this. Because webgl and cpu backend also added `id` in `DataId`.
Seems like `id` parameter is not necessary. `dataId={}` itself could represent a unique tensorInfo in webgpu backend, and we doesn't use `id` any where.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -170,30 +187,55 @@ export class WebGPUBackend extends KernelBackend {
...
+    if (this.tensorMap.has(dataId)) {
+      const tensorData = this.tensorMap.get(dataId);
+      return tensorData.refCount;
+    }
+    return 0;
+  }
+
+  /** Increase refCount of a `TensorData`. */
+  incRef(dataId: DataId): void {
+    const tensorData = this.tensorMap.get(dataId);
```
haoyunfeix, line 201: what about add an existence check here?
`if (this.tensorMap.has(dataId)) {...`

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: I think it's convenient for debugging to know which tensor has problem. `dataId={}` is just an `Object`. I don't know the background. But it seems that it's better if `DataId` is directly a `number` not an `object`. Just keep it in this PR since we mainly resolve the memory issue.
@pyu10055 @lina128 may know more about this.

tfjs-backend-webgpu/src/backend_webgpu.ts
haoyunfeix: LGTM

qjia7:
Thanks for your review, Yunfei.

@pyu10055 @lina128 Please take a look.
This is a follow up for this PR https://github.com/tensorflow/tfjs/pull/4628.

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: Hi Yunfei, fyi. I have to recover to the original change and not check `if (this.tensorMap.has(dataId))`. `this.tensorMap.get(dataId)` will make sure that there is always a valid tensor. If there is not a tensor id, it will call `engine.moveData` to get one.
If we used `this.tensorMap.has(dataId)`, there will be no chance to call `engine.moveData`. In that case, we will get error in some operators.

lina128 (approved):
LGTM! Great work!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4642#-:-MT3tMWR-mQwyWn6bov9:b-tcm0nv)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

___
*[tfjs-backend-webgpu/src/backend_webgpu.ts, line 215 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4642#-MSlvoqCR6hJa-Ym98XD-r1-215:-MT3sXImAwwYRSpxqw0z:bt4d29) ([raw file](https://github.com/tensorflow/tfjs/blob/0c69631701709120c636c939f7e5b461e84efc36/tfjs-backend-webgpu/src/backend_webgpu.ts#L215)):*

I think it's convenient for debugging to know which tensor has problem. `dataId={}` is just an `Object`. I don't know the background. But it seems that it's better if `DataId` is directly a `number` not an `object`. Just keep it in this PR since we mainly resolve the memory issue.
@pyu10055 @lina128 may know more about this.
</blockquote></details>

+1. The reason we don't use a number but an object is because we use weakmap (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) to manage memory. But having the id information as a number in the object helps debugging.

PR #4747, opened 2021-02-25
Allows extra flexibility to the location of libtensorflow downloads.

melmerp:
Specifically, modifies the `install.js`.

Adds two more options to control `CDN_STORAGE` for modifying the base
host:

1. TFJS_NODE_CDN_STORAGE env variable to help with any naming collisions
with other packages that might be using `CDN_STORAGE` env variable.
2. TFJS_NODE_CDN_STORAGE in the `.npmrc` file

Adds two options to control the `BASE_URI` for use cases where the
url path doesn't match the default:

1. TJFS_NODE_BASE_URI env variable
2. TFJS_NODE_BASE_URI in the `.npmrc`

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4747)


tfjs-node/README.md
```diff
@@ -66,6 +66,49 @@ To use this package on Raspberry Pi, you need to rebuild the node native addon w
 $ npm rebuild @tensorflow/tfjs-node --build-from-source
 ```
 
+#### Custom binaries URI
+
+If you happen to be using a mirror for the libtensorflow binaries (default is [https://storage.googleapis.com/]), you have 3 options (in order of priority):
+
+1. Set the environment variable `CDN_STORAGE`
```
pyu10055, line 73: should this variable also support npm config file?

tfjs-node/README.md
```diff
@@ -66,6 +66,49 @@ To use this package on Raspberry Pi, you need to rebuild the node native addon w
...
+
+```sh
+CDN_STORAGE="https://yourmirrorofchoice.com" npm install <package>
+(or)
+CDN_STORAGE="https://yourmirrorofchoice.com" yarn install <package>
+```
+
+2. Set the environment variable `TFJS_NODE_CDN_STORAGE`. This has the same behavior as `CDN_STORAGE`, but introduced to prevent collisions with other npm packages that might use `CDN_STORAGE`.
```
pyu10055, line 81: This is bit duplicated, maybe better to add deprecation note on the CDN_STORAGE, since TFJS_NODE_CDN_STORAGE seems to be a better naming.

pyu10055 (commented):
Thank you for the contribution, looks good with couple minor comments.

tfjs-node/README.md
melmerp: I thought that it might be best to deprecate `CDN_STORAGE` at some major revision point and discontinue use in favor of the prefix versions: `TFJS_NODE_CDN_STORAGE` and `npm_config_TFJS_NODE_CDN_STORAGE`. As such, I didn't want to introduce the npm config file version of `CDN_STORAGE` as it would just get deprecated.

Please let me know how you'd like to proceed.

tfjs-node/README.md
melmerp: Yes, I agree. It would be great to deprecate the use of `CDN_STORAGE`. Is there an "official" way to do this other than to leave a note that `CDN_STORAGE` will be deprecated at some future date? I will do this now.

melmerp:
@googlebot I signed it!

tfjs-node/README.md
pyu10055: You can put the option 2 and 3 at the top and move the CDN_STORAGE to the end and indicate it as a deprecated options.

melmerp: I've completed this. Thanks.

pyu10055 (approved):
Thanks you!

PR #4753, opened 2021-02-26
Keep values on CPU

qjia7:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4753)


qjia7:
@lina128 @pyu10055 take a look, thanks.
Keeping values on CPU is 1) convenient for `moveData` when switching backends. For example, if we switch backend from webgl to wasm, when calling `readSync` in webgl, we don't need to read data back from gpu to cpu.
2) convenient for WebGPU since it doesn't support read data back from gpu synchronously. Otherwise, when switching from webgpu to other backends, it will report `WebGPU readSync is only available for CPU-resident tensors.` error.

Another option is totally to remove `readSync` and use `async read`. Then we need to change the current logic in tfjs-core.
Cc @haoyunfeix @axinging @xhcao

pyu10055 (commented):
I don't think it is a typical use case for user to switch from WebGL to WASM. For WebGPU this probably only helps constant tensors, that was originally on the CPU, not any output tensors, which do not have value field populated.

I think the bigger issue is how to sync data when switching from WebGPU to other backend? The current codebase has many places uses readSync, it would be a really big change to remove the API.
Can WebGPU simulate sync read with a blocking poll?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4753#-:-MUW9kPc2QDNs1_8z8mA:bm0fsxt)* status: 0 of 1 approvals obtained (waiting on @lina128)

qjia7:
> I don't think it is a typical use case for user to switch from WebGL to WASM. For WebGPU this probably only helps constant tensors, that was originally on the CPU, not any output tensors, which do not have value field populated.
> 
In current e2e benchmark framework, for the same model, we can try different backends. The error will happen when you switching from WebGPU to other backend. It's just a temporary workaround to make this situation work well for webgpu. Like you said, it can't resolve the situation that tensor doesn't have cpu values. For the latter, it may happen on hybrid mode. But it seems that there is no solution for WebGPU if we don't use the async read.

I just modified the change only for webgpu to make it work for e2e benchmark when integrating webgpu. For future, we still hope tfjs-core can use async read everywhere. 

> I think the bigger issue is how to sync data when switching from WebGPU to other backend? The current codebase has many places uses readSync, it would be a really big change to remove the API.
> Can WebGPU simulate sync read with a blocking poll?
> 
As far as I know, WebGPU only support async read. It seems that there is no way to call an async read in a sync method.

pyu10055 (approved):
I think for short term this solution to fix e2e test is fine, but we shall think about how to align WebGPU with other backends in terms of Tensor Read. Thanks!

lina128 (approved):
+1 to Ping's comment above.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4753#-:-MV0JJbwD4fwMgy0z1vT:b-o83skc)* status: :shipit: complete! 2 of 1 approvals obtained

PR #4754, opened 2021-02-26
Fix for Failed Symlink on NPM Installation of 'tfjs-node'

PhillipAWells:
Attempting to install `tfjs-node` or `tfjs-node-gpu` on the Windows platform would always result in the following error:

```
> node scripts/install.js

CPU-windows-3.2.0.zip
* Downloading libtensorflow

* Building TensorFlow Node.js bindings
symlink ./lib/napi-v7 failed:  null
```
This is due to the original code not checking the error value on the result of the symlink call. A simple conditional on whether `error`  is null or not fixes the problem.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4754)


pyu10055 (approved):
Thanks!

PR #4767, opened 2021-03-02
webgpu: add control barrier for read-write operation

xhcao:
BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4767)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
@@ -81,6 +81,7 @@ export class ReduceProgram implements WebGPUProgram {
             ${op}
           }
         }
+        barrier();
         xBestValues[gl_LocalInvocationID.x] = bestValue;
         currentSize = DIV_CEIL(currentSize, ${reductionFactor});
         ${this.reduceType === 'sum' ? 'if(currentSize > 1) bestValue=0;' : ''}


xhcao:

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
```diff
@@ -81,6 +81,7 @@ export class ReduceProgram implements WebGPUProgram {
             ${op}
           }
         }
+        barrier();
```
xhcao, line 84: I have a question about using GLSL `barrier` function or GLSL `memoryBarrier` function for shared memory. In all operators' shaders, we only use `barrier`  to sync shared memory read/write accesses. GLSL `barrier` is used for execution control and GLSL `memoryBarrier` is used for memory control, and SPIRV also distinguishes these features using `OpControlBarrier` and `OpMemoryBarrier` feature.
If we only uses  GLSL `barrier` in line 76 for Read-After-Write operation, do we have a flaky incorrect result on Vulkan backend?
HLSL does not distinguish execution control and memory control features, it merges them together, so GLSL `barrier` will be translated to HLSL `GroupMemoryBarrierWithGroupSync` function, so we could get correct result on D3D backend.

xhcao:
This patch fixes the issue that `reduce` operator sometimes gets wrong result, for example, https://github.com/qjia7/tfjs/commit/ded2de6fbe7cec511977a5519475fa0ed75c1731

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
qjia7: That's a good point. We never tried it on vulkan. According to https://stackoverflow.com/questions/39393560/glsl-memorybarriershared-usefulness and the latest spec of GLSL, it seems that barrier() by itself is enough to synchronize both control flow and memory accesses to shared variables. However, we need to check SPIRV's definition for this.  Maybe use `memoryBarrierShared(); barrier()` for the correctness on all platforms.

xhcao: Only lateset GLSL 4.60.7 spec refers that `barrier()` could synchronizes memory accesses to shared variables, `barrier()` in the GLSL 4.60 spec has not this function. And all our operators' shaders are written based on GLSL 450, if so, our shaders are not very correct. Although `gslang` tool may translate our wrong shaders to correct SPRIV binary, we still need write right shaders.
I will fix the synchronization issue of memory accesses to shared variables in another patch. This patch only fix the issue that `reduce` operator could not work correctly.

qjia7 (approved):
LGTM.
+ @kainino0x  @lina128 take a look, thanks.

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
kainino0x: I am pretty sure you can assume that a glsl `barrier()` will emit both control-flow and shared-memory barriers. This answer is enlightening:
https://stackoverflow.com/a/39395947
Basically, it was a bug in the GLSL spec that `barrier()` wasn't specified to also do a shared-memory barrier. "It's probably best to treat desktop GLSL as though it always said this."

Also note that you are using a pinned version of glslang which will always produce the same SPIR-V. If it produces the correct SPIR-V now (it should - see below), then it shouldn't break down the line. (though something to be aware of when switching to WGSL.)

glslang issue:
https://github.com/KhronosGroup/glslang/issues/502#issuecomment-474737012
> Glslang translates this to:
> ...
> OpControlBarrier with `Workgroup` execution scope and `Workgroup` memory scope, and memory semantics of ... `MemorySemanticsWorkgroupMemory` and `MemorySemanticsAcquireRelease`

i.e. a control-plus-memory barrier.

You can test this (but with the latest glslang, not the one you're using) on http://shader-playground.timjones.io/
or you can export and disassemble the SPIR-V output from glslang.js.

kainino0x (approved):
LGTM

PR #4776, opened 2021-03-04
webgpu: Batch several dispatching calls into one compute pass

qjia7:

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -77,6 +77,7 @@ export class WebGPUBackend extends KernelBackend {
   tensorMap: DataStorage<TensorBufferInfo>;
   fromPixelProgram: FromPixelsProgram;
   supportTimeQuery: boolean;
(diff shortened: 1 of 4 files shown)


qjia7:
@gyagp @xhcao @axinging @haoyunfeix Please take a look, thanks.
With this support, it's convenient for us to try the performance impact of batching different dispatching numbers.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -259,6 +261,11 @@ export class WebGPUBackend extends KernelBackend {
       // Data is on the CPU.
       return info.values;
     }
+    if (this.dispatchNumberInEncoder !== 0) {
+      this.dispatchNumberInEncoder = 0;
+      this.computePass.endPass();
+      this.submitQueue();
```
haoyunfeix, line 267: Could this submitQueue() be removed? So we could get the same behavior as old delayed mode when WEBGPU_BATCH_DISPATCHING_CALLS set to negative value.

PR #4780, opened 2021-03-04
Add a local test config for the wasm threaded-simd binary

mattsoulanille:
This PR enables locally testing the threaded simd wasm bundle with `yarn test-threaded-simd`.

It appears as though we may have a broken test on threaded-simd. Using the [spec reporter](https://www.npmjs.com/package/karma-spec-reporter), tests stop running after `pad 2d test-wasm`. It seems like the `pad 3d` tests may be causing this.

I have not yet found out how to test just simd without threaded. I haven't found a chrome flag to disable threaded wasm yet. 


jinjingforever (approved):

pyu10055 (approved):
Looks good, thanks!

PR #4813, opened 2021-03-16
webgpu: remove unnecessary barrier

xhcao:
The barrier registers are finite, and they are key factor that
determine how many work-groups can be executed simultaneously.

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/matmul_packed_vec4_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/matmul_packed_vec4_webgpu.ts
@@ -19,7 +19,15 @@ import {computeDispatch, computeWorkGroupSizeForMatMul, tilesFitEvenlyIntoShape}
 
 import {WebGPUProgram} from './webgpu_program';
 
-export function makeMatMulPackedVec4Source(workPerThread: number[]): string {
+export function makeMatMulPackedVec4Source(workPerThread: number[],
+    workGroupSize?: number[], dimInner?: number): string {
+  let numTiles = -1;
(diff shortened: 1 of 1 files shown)


xhcao:

tfjs-backend-webgpu/src/kernels/matmul_packed_vec4_webgpu.ts
```diff
@@ -19,7 +19,15 @@ import {computeDispatch, computeWorkGroupSizeForMatMul, tilesFitEvenlyIntoShape}
 
 import {WebGPUProgram} from './webgpu_program';
 
-export function makeMatMulPackedVec4Source(workPerThread: number[]): string {
+export function makeMatMulPackedVec4Source(workPerThread: number[],
+    workGroupSize?: number[], dimInner?: number): string {
+  let numTiles = -1;
+  if (workGroupSize != null && dimInner != null) {
```
qjia7, line 25: You can directly pass the `dimInner` and `tileInner`. Then calculate `numTiles`. And this part also affects the shader key. Please add a flag to shaderKey for the shader variant.

Maybe you need to do similar change for the non-vec4 matmul version.

xhcao: `numTiles` is indirectly computed by input shape, workGroupSize and ColPerThread/RowPerThread, and there is no `ColPerThread` part in original ShaderKey, so I add `ColPerThread` to ShaderKey.
Other comments are addressed. Thank you for reviewing it again.

PR #4819, opened 2021-03-17
[webgpu] Use uniform instead of constant for shape

axinging:
This reverts https://github.com/tensorflow/tfjs/pull/3647 and adds:
1. support float/int uniform buffers, merge all uniforms into one uniform buffer.
2. store shape strides and shape size in uniform buffer.
3. turn shapes to shape lengths to save shader in makeShaderKey.

These changes will save warmup time. For all e2e related model, this saves wamup time from 60% ~ 90%. And no obvious performance regression is found.

TODO: 
remove all shapes from all shader key, such as ConcatProgram,  CropAndResizeProgram, PadProgram.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4819)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -483,18 +483,120 @@ export class WebGPUBackend extends KernelBackend {
     }
   }
 
+  private makeUniformsDataView(data: DataView): GPUBindingResource {
+    const dimensionsBuffer = this.acquireBuffer(
+        data.byteLength, GPUBufferUsage.COPY_DST | GPUBufferUsage.UNIFORM);
+    this.queue.writeBuffer(dimensionsBuffer, 0, data);
+
+    return {offset: 0, size: data.byteLength, buffer: dimensionsBuffer};
+  }
+
+  private arrayToDataView(
+      arrays: Array<{type: string; data: number[]}>, length: number): DataView {
+    const BYTES_PER_ELEMENT = 4;
+    const uniformDataView =
+        new DataView(new ArrayBuffer(length * BYTES_PER_ELEMENT));
+
+    let dataViewIndex = 0;
+    arrays.forEach(array => {
+      const arrayData = array.data;
+
+      if (array.type !== 'int32' && array.type !== 'float32') {
+        throw new Error(`${array.type} not supported!`);
+      }
+
+      if (array.type === 'int32') {
+        arrayData.forEach(d => {
+          uniformDataView.setInt32(dataViewIndex * BYTES_PER_ELEMENT, d, true);
+          dataViewIndex++;
+        });
+      } else {
+        arrayData.forEach(d => {
+          uniformDataView.setFloat32(
+              dataViewIndex * BYTES_PER_ELEMENT, d, true);
+          dataViewIndex++;
+        });
+      }
+    });
+
+    return uniformDataView;
+  }
+
+  private computePadding(uniformsWithType:
+                             Array<{type: string; data: number[];}>): DataView {
+    let currentOffset = 0;
+    let padding = 0;
+    let dataViewIndex = 0;
+    const dimUniformsData: Array<{type: string; data: number[];}> = [];
+    uniformsWithType.forEach((d, i) => {
+      if (d.data.length === 0) {
+        d.data = [1];
+      }
+      // Complete std140 layout rules are documented here:
+      // tslint:disable-next-line:max-line-length
+      // https://www.khronos.org/registry/OpenGL/specs/gl/glspec45.core.pdf#page=159
+      let baseAlignment: number;
+      switch (d.data.length) {
+        case 0:
+          baseAlignment = 1;
+          break;
+        case 1:
+          baseAlignment = 1;
+          break;
+        case 2:
+          baseAlignment = 2;
+          break;
+        case 3:
+          baseAlignment = 4;
+          break;
+        case 4:
+          baseAlignment = 4;
+          break;
+        default:
+          util.assert(false, () => `Unsupported ${d.data.length}D shape`);
+      }
+
+      padding = Math.ceil(currentOffset / baseAlignment) * baseAlignment -
+          currentOffset;
+      for (let p = 0; p < padding; ++p) {
+        dimUniformsData.push({type: 'int32', data: [0]});
+        dataViewIndex++;
+      }
+      dimUniformsData.push({type: d.type, data: d.data});
+      dataViewIndex = dataViewIndex + d.data.length;
+      currentOffset += d.data.length + padding;
+    });
+
+    return this.arrayToDataView(dimUniformsData, dataViewIndex);
+  }
+
   public runWebGPUProgram(
       program: webgpu_program.WebGPUProgram, inputs: TensorInfo[],
       outputDtype: DataType,
-      programUniforms?: Uint32Array|Int32Array|Float32Array): TensorInfo {
+      programUniforms?: Array<{type: string; data: number[]}>): TensorInfo {
     const output = this.makeTensorInfo(program.outputShape, outputDtype);
 
-    let uniformDataLength;
-    let uniforms: GPUBindingResource;
-    if (program.uniforms) {
-      // TODO: handle padding of program-specific uniforms
-      uniformDataLength = programUniforms.byteLength;
-      uniforms = this.makeUniforms(programUniforms);
+    // There are four kinds of uniforms: shapes, shape strides, program
+    // size, program defined uniforms.
+    const bufferShapes = inputs.concat(output).map(d => d.shape);
+    let uniformsWithType = bufferShapes.map(d => {
+      return {type: 'int32', data: d};
+    });
+    const strides = util.computeStrides(output.shape);
+    uniformsWithType.push({type: 'int32', data: strides});
+    if (program.size != null) {
+      uniformsWithType.push({type: 'int32', data: [program.size]});
+    }
+    if (programUniforms) {
+      uniformsWithType = [...uniformsWithType, ...programUniforms];
+    }
+
+    let uniformsByteLength = 0;
+    let uniforms: GPUBindingResource = null;
+    if (uniformsWithType != null) {
+      const uniformsDataView = this.computePadding(uniformsWithType);
+      uniformsByteLength = uniformsDataView.byteLength;
+      uniforms = this.makeUniformsDataView(uniformsDataView);
     }
 
     const inputsData = inputs.map((input: TensorInfo, i: number) => {
@@ -515,13 +617,15 @@ export class WebGPUBackend extends KernelBackend {
       };
     });
     this.uploadToGPU(output.dataId);
-    const bufferShapes = inputs.concat(output).map(d => d.shape);
     const bufferTypes = inputsData.map(d => d.dtype).concat(output.dtype);
-    const key =
-        webgpu_program.makeShaderKey(program, bufferShapes, bufferTypes);
+    const broadcastDims = inputsData.map(
+        d => backend_util.getBroadcastDims(d.shape, output.shape));
+    const broadcastDimsKey = broadcastDims.map(d => d.join('_')).join(';');
+    const key = webgpu_program.makeShaderKey(
+        program, bufferShapes, bufferTypes, broadcastDimsKey);
     const {bindGroupLayout, pipeline} = this.getAndSavePipeline(key, () => {
       return webgpu_program.compileProgram(
-          this.glslang, this.device, program, inputsData, output, uniforms);
+          this.glslang, this.device, program, inputsData, output);
     });
 
     const shouldTimeProgram = this.activeTimers != null;
@@ -554,10 +658,9 @@ export class WebGPUBackend extends KernelBackend {
       this.commandQueueOwnedIds.add(input.dataId);
     });
     this.commandQueueOwnedIds.add(output.dataId);
-
-    if (program.uniforms) {
+    if (uniforms) {
       const uniformInfo = {
-        byteSize: uniformDataLength,
+        byteSize: uniformsByteLength,
         usage: GPUBufferUsage.COPY_DST | GPUBufferUsage.UNIFORM,
         buffer: (uniforms as GPUBufferBinding).buffer
       };
@@ -632,15 +735,6 @@ export class WebGPUBackend extends KernelBackend {
         GPUBufferUsage.COPY_SRC | GPUBufferUsage.QUERY_RESOLVE);
     // Return milliseconds.
     return timeElapsedNanos / 1000000;
-  }
-
-  private makeUniforms(data: Uint32Array|Int32Array|
-                       Float32Array): GPUBindingResource {
-    const dimensionsBuffer = this.acquireBuffer(
-        data.byteLength, GPUBufferUsage.COPY_DST | GPUBufferUsage.UNIFORM);
-    this.queue.writeBuffer(dimensionsBuffer, 0, data);
-
-    return {offset: 0, size: data.byteLength, buffer: dimensionsBuffer};
   }
 
   private getCPUBackend(): KernelBackend|null {
--- a/tfjs-backend-webgpu/src/kernels/AvgPool.ts
+++ b/tfjs-backend-webgpu/src/kernels/AvgPool.ts
@@ -45,15 +45,16 @@ export function avgPool(
   }
 
   const dimensions = [
-    convInfo.padInfo.top, convInfo.padInfo.left,      // Padding.
-    convInfo.strideHeight, convInfo.strideWidth,      // Stride.
-    convInfo.dilationHeight, convInfo.dilationWidth,  // Dilation.
-    convInfo.inHeight, convInfo.inWidth,              // Conv dims.
-    convInfo.effectiveFilterHeight,
-    convInfo.effectiveFilterWidth  // Filter dims.
+    {type: 'int32', data: [convInfo.padInfo.top, convInfo.padInfo.left]},
+    {type: 'int32', data: [convInfo.strideHeight, convInfo.strideWidth]},
+    {type: 'int32', data: [convInfo.dilationHeight, convInfo.dilationWidth]},
+    {type: 'int32', data: [convInfo.inHeight, convInfo.inWidth]}, {
+      type: 'int32',
+      data: [convInfo.effectiveFilterHeight, convInfo.effectiveFilterWidth]
+    }
   ];
-  const uniformData = new Int32Array(dimensions);
-  return backend.runWebGPUProgram(program, [x], x.dtype, uniformData);
+
+  return backend.runWebGPUProgram(program, [x], x.dtype, dimensions);
 }
 
 export const avgPoolConfig: KernelConfig = {
--- a/tfjs-backend-webgpu/src/kernels/FromPixelsImageBitmap.ts
+++ b/tfjs-backend-webgpu/src/kernels/FromPixelsImageBitmap.ts
@@ -30,11 +30,14 @@ export function fromPixelsImageBitmap(args: {
 
   const outShape = [imageBitmap.height, imageBitmap.width, numChannels];
   const size = util.sizeFromShape(outShape);
-  const uniformData: [number, number] = [size, numChannels];
+  const strides = util.computeStrides(outShape);
+  const uniformData = [size, numChannels, ...strides];
 
   const output = backend.makeTensorInfo(outShape, 'int32');
   if (!backend.fromPixelProgram) {
     backend.fromPixelProgram = new FromPixelsProgram(outShape);
+  } else {
+    backend.fromPixelProgram.updateOutputShape(outShape);
   }
 
   // Different outShape will affect preprocessor result,
@@ -49,11 +52,10 @@ export function fromPixelsImageBitmap(args: {
 
   const {bindGroupLayout, pipeline} = backend.getAndSavePipeline(key, () => {
     return webgpu_program.compileProgram(
-        backend.glslang, backend.device, backend.fromPixelProgram, [], output);
+        backend.glslang, backend.device, backend.fromPixelProgram, [], output,
+        true);
   });
   backend.fromPixelProgram.setWebGPUBinary(bindGroupLayout, pipeline);
-
-  backend.fromPixelProgram.updateOutputShape(outShape);
 
   backend.queue.copyImageBitmapToTexture(
       {imageBitmap, origin: {x: 0, y: 0}}, {
--- a/tfjs-backend-webgpu/src/kernels/FromPixels_utils/from_pixels_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/FromPixels_utils/from_pixels_webgpu.ts
@@ -34,7 +34,7 @@ export class FromPixelsProgram implements WebGPUProgram {
   bindGroupLayout: GPUBindGroupLayout;
 
   uniform: GPUBuffer;
-  lastUniformData = [0, 0];
+  lastUniformData: number[] = [];
 
   inputTexture: GPUTexture = null;
   lastPixelSize = {width: 0, height: 0};
@@ -68,19 +68,22 @@ export class FromPixelsProgram implements WebGPUProgram {
     layout(set = 0, binding = 2) uniform Meta {
       int size;
       int numChannels;
-    } outShape;
+      ivec2 shapeStride;
+    };
+
+    ivec3 getCoordsFromFlatIndex(int flatIndexBase);
 
     void main() {
-      int flatIndexBase = int(gl_GlobalInvocationID.x) * outShape.numChannels;
+      int flatIndexBase = int(gl_GlobalInvocationID.x) * numChannels;
       ivec3 coords = getCoordsFromFlatIndex(flatIndexBase);
       int texR = coords[0];
       int texC = coords[1];
       int depth = coords[2];
       vec4 values = imageLoad(srcImage, ivec2(texC, texR));
-      for(int i = 0; i < outShape.numChannels; i++) {
+      for(int i = 0; i < numChannels; i++) {
         float value = values[i];
         int flatIndex = flatIndexBase + i;
-        if (flatIndex < outShape.size) {
+        if (flatIndex < size) {
           result[flatIndex] = int(floor(255.0 * value));
         }
       }
@@ -101,7 +104,8 @@ export class FromPixelsProgram implements WebGPUProgram {
     // and reuse it always.
     if (!this.uniform) {
       const uniformBuffer = device.createBuffer({
-        size: 8,  // The uniform buffer contains two 4 bytes element always.
+        size: uniformData.length *
+            4,  // The uniform buffer contains two 4 bytes element always.
         usage: GPUBufferUsage.UNIFORM | GPUBufferUsage.COPY_DST
       });
 
@@ -109,18 +113,15 @@ export class FromPixelsProgram implements WebGPUProgram {
     }
 
     // No need to update uniform buffer if no changes.
-    // The initial lastUniformData will have value [0, 0],
-    // which is not a valid numChannels or valid size.
     if (!uniformData ||
-        (uniformData[0] === this.lastUniformData[0] &&
-         uniformData[1] === this.lastUniformData[1])) {
+        ((uniformData.length === this.lastUniformData.length) &&
+         uniformData.every((v, i) => v === this.lastUniformData[i]))) {
       return;
     }
 
     device.queue.writeBuffer(this.uniform, 0, new Uint32Array(uniformData));
 
-    this.lastUniformData[0] = uniformData[0];
-    this.lastUniformData[1] = uniformData[1];
+    this.lastUniformData = uniformData;
   }
 
   makeInputTexture(device: GPUDevice, pixelWidth: number, pixelHeight: number):
--- a/tfjs-backend-webgpu/src/kernels/argminmax_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/argminmax_webgpu.ts
@@ -17,7 +17,7 @@
 
 import {backend_util, util} from '@tensorflow/tfjs-core';
 
-import {getCoordsDataType, getShapeCoords} from '../shader_preprocessor';
+import {getCoordsDataType} from '../shader_preprocessor';
 import {computeDispatch} from '../webgpu_util';
 
 import {WebGPUProgram} from './webgpu_program';
@@ -65,7 +65,7 @@ export class ArgMinMaxProgram implements WebGPUProgram {
         this.dispatchLayout, this.outputShape, this.workGroupSize);
 
     this.inputShape = inputShape;
-    this.shaderKey = `argMinMax_${this.op}_${reduceSize}`;
+    this.shaderKey = `argMinMax${this.op}`;
   }
 
   getUserCode(): string {
@@ -121,9 +121,9 @@ export class ArgMinMaxProgram implements WebGPUProgram {
 
     const indexInputShape = (index: string) => {
       if (this.inputShape.length === 1) {
-        return `${getShapeCoords(this.inputShape)}`;
+        return 'xShape';
       } else {
-        return `${getShapeCoords(this.inputShape)}[${index}]`;
+        return `xShape[${index}]`;
       }
     };
 
--- a/tfjs-backend-webgpu/src/kernels/batchnorm_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/batchnorm_webgpu.ts
@@ -17,7 +17,7 @@
 
 import {backend_util} from '@tensorflow/tfjs-core';
 
-import {getCoordsDataType, getShapeCoords} from '../shader_preprocessor';
+import {getCoordsDataType} from '../shader_preprocessor';
 import {computeDispatch, flatDispatchLayout} from '../webgpu_util';
 
 import {WebGPUProgram} from './webgpu_program';
@@ -83,7 +83,7 @@ export class BatchNormProgram implements WebGPUProgram {
     }
     const userCode = `
       void writeResult(${coordsDataType} coords,float value) {
-        if (coordsInBounds(coords, ${getShapeCoords(this.outputShape)})) {
+        if (coordsInBounds(coords, outShape)) {
           ${setOutput}
         }
       }
--- a/tfjs-backend-webgpu/src/kernels/binary_op_complex_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/binary_op_complex_webgpu.ts
@@ -38,6 +38,7 @@ export class BinaryOpComplexProgram implements WebGPUProgram {
   dispatch: [number, number, number];
   workGroupSize: [number, number, number] = [128, 1, 1];
   op: string;
+  size: number;
 
   constructor(op: string, aShape: number[], bShape: number[]) {
     this.outputShape = backend_util.assertAndGetBroadcastShape(aShape, bShape);
@@ -45,12 +46,12 @@ export class BinaryOpComplexProgram implements WebGPUProgram {
     this.dispatch = computeDispatch(
         this.dispatchLayout, this.outputShape, this.workGroupSize);
 
-    this.shaderKey = `binaryOpComplex${op}`;
+    this.shaderKey = `binaryOpComplex_${op}`;
     this.op = op;
+    this.size = util.sizeFromShape(this.outputShape);
   }
 
   getUserCode(): string {
-    const size = util.sizeFromShape(this.outputShape);
     const userCode = `
       float binaryOpComplex(
           float areal, float aimag, float breal, float bimag) {
@@ -59,7 +60,7 @@ export class BinaryOpComplexProgram implements WebGPUProgram {
 
       void main() {
         int index = int(gl_GlobalInvocationID.x);
-        if(index < ${size}) {
+        if(index < size) {
           float areal = getARealAtOutCoords();
           float aimag = getAImagAtOutCoords();
           float breal = getBRealAtOutCoords();
--- a/tfjs-backend-webgpu/src/kernels/binary_op_shared_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/binary_op_shared_webgpu.ts
@@ -16,8 +16,8 @@
  */
 
 import {backend_util, util} from '@tensorflow/tfjs-core';
-import {getCoordsDataType} from '../shader_preprocessor';
 
+import {getCoordsDataType} from '../shader_preprocessor';
 import {computeDispatch, flatDispatchLayout} from '../webgpu_util';
 
 import {WebGPUProgram} from './webgpu_program';
@@ -33,6 +33,8 @@ export class BinaryOpSharedProgram implements WebGPUProgram {
   useSharedMemoryWithB: boolean;
   lastDimensionSize: number;
   op: string;
+  size: number;
+  sizeFit: boolean;
 
   constructor(
       op: string, aShape: number[], bShape: number[],
@@ -53,14 +55,17 @@ export class BinaryOpSharedProgram implements WebGPUProgram {
     this.dispatch = computeDispatch(
         this.dispatchLayout, this.outputShape, this.workGroupSize,
         [this.workPerThread, 1, 1]);
-    this.shaderKey = `binaryShared_${op}`;
     this.useSharedMemoryWithB = useSharedMemoryWithB;
     this.op = op;
+    this.size = util.sizeFromShape(this.outputShape);
+    this.sizeFit =
+        this.size % (this.workGroupSize[0] * this.workPerThread) === 0;
+    this.shaderKey =
+        `binaryShared_${op}_${this.lastDimensionSize}_${this.sizeFit}`;
   }
 
   getUserCode(): string {
     const type = getCoordsDataType(this.outputShape.length);
-    const size = util.sizeFromShape(this.outputShape);
     const sharedIndexSnippet = this.lastDimensionSize > 1 ?
         `coords[${this.outputShape.length - 1}]` :
         '0';
@@ -69,13 +74,13 @@ export class BinaryOpSharedProgram implements WebGPUProgram {
          float b = sharedBuf[${sharedIndexSnippet}];` :
         `float a = sharedBuf[${sharedIndexSnippet}];
          float b = getBAtOutCoords(coords);`;
-    const sizeFit = size % (this.workGroupSize[0] * this.workPerThread) === 0;
-    const writeDataSnippet = sizeFit ?
+
+    const writeDataSnippet = this.sizeFit ?
         `${type} coords = getCoordsFromFlatIndex(flatIndex);
 
          ${accessDataSnippet}
          setOutput(flatIndex, binaryOperation(a, b));` :
-        `if(flatIndex < ${size}) {
+        `if(flatIndex < size) {
             ${type} coords = getCoordsFromFlatIndex(flatIndex);
 
             ${accessDataSnippet}
--- a/tfjs-backend-webgpu/src/kernels/binary_op_vec4_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/binary_op_vec4_webgpu.ts
@@ -31,6 +31,8 @@ export class BinaryOpVec4Program implements WebGPUProgram {
   workGroupSize: [number, number, number];
   isVec4 = true;
   op: string;
+  size: number;
+  fitShape: boolean;
 
   constructor(op: string, aShape: number[], bShape: number[]) {
     // TODO(jiajia.qin@intel.com): Heuristically select a good work group size.
@@ -42,14 +44,14 @@ export class BinaryOpVec4Program implements WebGPUProgram {
         this.dispatchLayout, this.outputShape, this.workGroupSize,
         [this.workPerThread, 1, 1]);
     this.op = op;
-    this.shaderKey = `binaryVec4_${op}`;
+    this.fitShape = this.size % this.workGroupSize[0] === 0;
+    this.shaderKey = `binaryVec4_${op}_${this.fitShape}`;
+    this.size = util.sizeFromShape(this.outputShape) / this.workPerThread;
   }
 
   getUserCode(): string {
-    const size = util.sizeFromShape(this.outputShape) / this.workPerThread;
-    const fitShape = size % this.workGroupSize[0] === 0;
     let userCode: string;
-    if (fitShape) {
+    if (this.fitShape) {
       userCode = `
       vec4 binaryOperation(vec4 a, vec4 b) {
         ${this.op}
@@ -70,7 +72,7 @@ export class BinaryOpVec4Program implements WebGPUProgram {
 
       void main() {
         int index = int(gl_GlobalInvocationID.x);
-        if (index < ${size})
+        if (index < size)
         {
           vec4 a = A[index];
           vec4 b = B[index];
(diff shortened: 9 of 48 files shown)


axinging:
@qjia7 @xhcao  @haoyunfeix @gyagp, PTAL

qjia7:
Offline talked with Xing. The mainly issue is that there are still lots of shader variants. We should further understand why some inputShapes or outputShape are still needed in shaderKey and try to remove them to reduce the shader variant.

It will be good to give us a data how much time is reduced for the warm-up time with this PR.

@axinging Please ping us again when it's ready for review.

axinging:
@qjia7 @xhcao @haoyunfeix @gyagp, PTAL

tfjs-backend-webgpu/src/kernels/transpose_webgpu.ts
```diff
@@ -43,7 +43,6 @@ export class TransposeProgram implements WebGPUProgram {
         this.dispatchLayout, this.outputShape, this.workGroupSize,
         [this.workPerThread, 1, 1]);
 
-    this.aShape = aShape;
     this.newDim = newDim;
     this.shaderKey = `transpose_${newDim}`;
```
qjia7, line 47: Will this.outputShape affect the shaderKey?

tfjs-backend-webgpu/src/shader_preprocessor.ts
```diff
@@ -222,33 +251,33 @@ function getSetOutputSnippet(
...
+
   if (outRank >= 2) {
     const dims = ['d0', 'd1', 'd2', 'd3'].slice(0, outRank);
     const type = getCoordsDataType(outRank);
 
     if (isVec4) {
       snippet += `
       void setOutput(${dims.map(d => `int ${d}`).join(', ')}, vec4 value) {
-        int flatIndex = getFlatIndex(${type}(${dims.join(', ')}), ${
-          getShapeCoords(outShape)});
```
qjia7, line 233: Don't forget to remove the definition of `getShapeCoords` since it's not needed anymore.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -482,18 +482,86 @@ export class WebGPUBackend extends KernelBackend {
...
+        currentOffset += d.length + padding;
+      });
+    }
+
+    const uniformDataLengths: number[] = [];
+    const uniforms: GPUBindingResource[] = [];
+
+    let hasIntProgramUniforms = false;
+    let hasFloatProgramUniforms = false;
+    if (programUniformsData instanceof Int32Array) {
```
qjia7, line 547: I don't like merging shapes uniform and program uniform together. It will make the code more complicated and not easy to read. Just separate them into two uniform blocks and you don't need to care about data type Int32Arry or Float32Array.

tfjs-backend-webgpu/src/kernels/clip_webgpu.ts
```diff
@@ -29,6 +29,7 @@ export class ClipProgram implements WebGPUProgram {
   dispatch: [number, number, number];
   workPerThread = 1;
   workGroupSize: [number, number, number] = [64, 1, 1];
+  needsShapesUniforms = true;
```
qjia7, line 32: Forget to change shaderKey?

qjia7 (commented):
This patch's purpose is to store tensor input/output shapes into uniform block so that we don't need to use lots of constant value in shader and reduce the shader variant.

Ideally, with this change, `inShape` and `outShape` shouldn't be shown in shaderKey again. Maybe in some shaders, they still relay on shape.rank or outShape's size. In such kind of shader, we should use inShape.length or outShape.length rather than `inShape` and `outShape` in shaderKey. For ${size} constant, you should try to use `coordsInBounds(coord, outShape)` instead of `${size}` to reduce the dependence on constant value.

So in this patch, I hope the new added part in each operator's shaderKey is `*Shape.lengh` and some other constant values if we missed before. If `*Shape` must be needed, please give a common above it to tell why.

tfjs-backend-webgpu/src/kernels/FromPixels_utils/from_pixels_webgpu.ts
```diff
@@ -52,11 +53,11 @@ export class FromPixelsProgram implements WebGPUProgram {
     this.dispatch = computeDispatch(
         this.dispatchLayout, this.outputShape, this.workGroupSize,
         [this.workPerThread, 1, 1]);
+    this.shaderKey = `fromPixels${outputShape}`;
```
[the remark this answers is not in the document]
axinging: getCoordsFromFlatIndex needs outputShape.

tfjs-backend-webgpu/src/kernels/FromPixelsImageBitmap.ts
```diff
@@ -35,26 +35,25 @@ export function fromPixelsImageBitmap(args: {
   const output = backend.makeTensorInfo(outShape, 'int32');
   if (!backend.fromPixelProgram) {
     backend.fromPixelProgram = new FromPixelsProgram(outShape);
+  } else {
+    backend.fromPixelProgram.updateOutputShape(outShape);
```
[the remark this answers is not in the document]
axinging: backend.fromPixelProgram.updateOutputShape(outShape) is always called in original version, I just changed it so the shader key can be correct.

tfjs-backend-webgpu/src/kernels/argminmax_webgpu.ts
```diff
@@ -65,7 +66,8 @@ export class ArgMinMaxProgram implements WebGPUProgram {
         this.dispatchLayout, this.outputShape, this.workGroupSize);
 
     this.inputShape = inputShape;
-    this.shaderKey = `argMinMax_${this.op}_${reduceSize}`;
+    this.shaderKey =
+        `argMinMax${this.op}${reduceSize}${inputShape.length}${outputShape}`;
```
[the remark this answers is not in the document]
axinging: getOutputCoords deps outputShape

tfjs-backend-webgpu/src/kernels/batchnorm_webgpu.ts
```diff
@@ -57,7 +58,8 @@ export class BatchNormProgram implements WebGPUProgram {
     this.offsetShape = offsetShape;
     this.scaleShape = scaleShape;
     this.varianceEpsilon = varianceEpsilon;
-    this.shaderKey = `batchNorm_${varianceEpsilon}`;
+    this.shaderKey = `batchNorm_${varianceEpsilon}_${xShape}_${
+        meanShape.length}_${varianceShape.length}`;
```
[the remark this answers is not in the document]
axinging: getOutputCoords needs outputShape

tfjs-backend-webgpu/src/kernels/binary_op_complex_webgpu.ts
```diff
@@ -45,7 +46,8 @@ export class BinaryOpComplexProgram implements WebGPUProgram {
     this.dispatch = computeDispatch(
         this.dispatchLayout, this.outputShape, this.workGroupSize);
 
-    this.shaderKey = `binaryOpComplex${op}`;
+    this.shaderKey = `binaryOpComplex_${op}_${aShape.length}_${bShape.length}_${
+        this.outputShape}`;
```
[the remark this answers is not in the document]
axinging: getAAtOutCoords and getBAtOutCoords require length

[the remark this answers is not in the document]
axinging: I will add a TODO for this.

tfjs-backend-webgpu/src/kernels/binary_op_vec4_webgpu.ts
```diff
@@ -42,7 +43,7 @@ export class BinaryOpVec4Program implements WebGPUProgram {
         this.dispatchLayout, this.outputShape, this.workGroupSize,
         [this.workPerThread, 1, 1]);
     this.op = op;
-    this.shaderKey = `binaryVec4_${op}`;
+    this.shaderKey = `binaryVec4_${op}_${this.outputShape}`;
```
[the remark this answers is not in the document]
axinging: Add this in TODO.

tfjs-backend-webgpu/src/kernels/binary_op_vec4_webgpu.ts
```diff
@@ -29,6 +29,7 @@ export class BinaryOpVec4Program implements WebGPUProgram {
   variableNames = ['A', 'B'];
   workPerThread = 4;
   workGroupSize: [number, number, number];
+  needsShapesUniforms = true;
```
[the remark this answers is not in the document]
axinging: Wee will remove ${size} in TODO, so I suggest true.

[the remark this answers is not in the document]
axinging: Wee will remove ${size} in TODO, so I suggest true.

tfjs-backend-webgpu/src/kernels/depthwise_conv2d_webgpu.ts
```diff
@@ -58,7 +58,10 @@ export class DepthwiseConv2DProgram implements WebGPUProgram {
     this.activation = activation;
     this.hasPreluActivation = hasPreluActivation;
 
-    this.shaderKey = `depthwise_${activation}`;
+    this.shaderKey = `depthwise_${activation}_${
+        this.convInfo.outChannels /
+        this.convInfo.inChannels}_${convInfo.inShape.length}_${
+        convInfo.filterShape.length}_${convInfo.outShape}`;
```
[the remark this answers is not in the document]
axinging: getX, getW, getOutputCoords. OutShape can be removed in next TODO.

tfjs-backend-webgpu/src/shader_preprocessor.ts
axinging: Done

tfjs-backend-webgpu/src/kernels/matmul_packed_vec4_webgpu.ts
```diff
@@ -139,20 +141,26 @@ export class MatMulPackedVec4Program implements WebGPUProgram {
...
     }
 
     this.workPerThread = rowPerThread;
     this.aShape = aShape;
     this.addBias = addBias;
     this.activation = activation;
     this.hasPreluActivationWeights = hasPreluActivationWeights;
-    this.shaderKey = `matMulPackedVec4_${rowPerThread}_${activation}`;
+    this.shaderKey =
+        `matMulPackedVec4_${rowPerThread}_${activation}_${shapeKey}`;
```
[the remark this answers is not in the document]
axinging: Use shapeKey instead. Infact, I think bias and weights requires output shape for it's shader key. Possible reason is due to test case missing. Add a todo for this.

tfjs-backend-webgpu/src/kernels/matmul_packed_webgpu.ts
```diff
@@ -182,7 +187,7 @@ export class MatMulPackedProgram implements WebGPUProgram {
     this.activation = activation;
     this.hasPreluActivationWeights = hasPreluActivationWeights;
     this.shaderKey = `matMulPacked_${this.workPerThread}_${transposeA}_${
-        transposeB}_${activation}`;
+        transposeB}_${activation}_${shapeKey}`;
```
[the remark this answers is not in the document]
axinging: See shapekey

tfjs-backend-webgpu/src/kernels/pool2d_webgpu.ts
```diff
@@ -44,7 +44,8 @@ export class Pool2DProgram implements WebGPUProgram {
         this.dispatchLayout, this.outputShape, this.workGroupSize,
         [1, this.workPerThread, 1]);
 
-    this.shaderKey = `pool2D_${poolType}`;
+    this.shaderKey =
+        `pool2D_${poolType}_${convInfo.inShape.length}_${convInfo.outShape}`;
```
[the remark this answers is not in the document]
axinging: It calls getOutputCoords(deps output shape) and getX(deps  in shape length)

tfjs-backend-webgpu/src/kernels/pool_filtersizeone_webgpu.ts
```diff
@@ -38,7 +38,8 @@ export class PoolWithFilterSizeEqualsOneProgram implements WebGPUProgram {
     this.dispatch = computeDispatch(
         this.dispatchLayout, this.outputShape, this.workGroupSize);
 
-    this.shaderKey = 'poolWithFilterSizeEqualsOne';
+    this.shaderKey = `poolWithFilterSizeEqualsOne_${convInfo.inShape.length}_${
+        convInfo.outShape}`;
```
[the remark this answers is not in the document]
axinging: getOutputCoords and getX

tfjs-backend-webgpu/src/kernels/resize_bilinear_webgpu.ts
```diff
@@ -40,9 +39,9 @@ export class ResizeBilinearProgram implements WebGPUProgram {
     this.dispatch = computeDispatch(
         this.dispatchLayout, this.outputShape, this.workGroupSize);
 
-    this.inputShape = inputShape;
     this.alignCorners = alignCorners;
-    this.shaderKey = `resizeBilinear_${alignCorners}`;
+    this.shaderKey = `resizeBilinear_${alignCorners}_${inputShape.length}_${
+        this.outputShape}`;
```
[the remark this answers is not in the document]
axinging: getOutputCoords deps outShape
getX deps xShape.length.

adjustHeight and adjustWidth dep alignCorners and out shape. So I think current key is good enough.

tfjs-backend-webgpu/src/kernels/transpose_webgpu.ts
axinging: getCoordsFromFlatIndex and size require out shape (or new DIM)

axinging:
@qjia7 @xhcao @haoyunfeix @gyagp, PTAL

tfjs-backend-webgpu/src/kernels/transpose_webgpu.ts
qjia7: Then it should be `newDim.length` insteadof `newDim`?

tfjs-backend-webgpu/src/kernels/fill_webgpu.ts
```diff
@@ -35,18 +36,18 @@ export class FillProgram implements WebGPUProgram {
         this.dispatchLayout, this.outputShape, this.workGroupSize,
         [this.workPerThread, 1, 1]);
 
-    this.shaderKey = 'fill';
+    this.shaderKey = `fill_${this.outputShape.length}`;
```
qjia7, line 39: Why do you need `this.outputShape.length`?

tfjs-backend-webgpu/src/kernels/AvgPool.ts
```diff
@@ -45,15 +45,16 @@ export function avgPool(
...
+    {type: 'int32', data: [convInfo.strideHeight, convInfo.strideWidth]},
+    {type: 'int32', data: [convInfo.dilationHeight, convInfo.dilationWidth]},
+    {type: 'int32', data: [convInfo.inHeight, convInfo.inWidth]}, {
+      type: 'int32',
+      data: [convInfo.effectiveFilterHeight, convInfo.effectiveFilterWidth]
+    }
   ];
-  const uniformData = new Int32Array(dimensions);
-  return backend.runWebGPUProgram(program, [x], x.dtype, uniformData);
+
```
qjia7, line 56: Can it be written like below? Similar to other places.
```
  const dimensions = [
    convInfo.padInfo.top, convInfo.padInfo.left,      // Padding.
    convInfo.strideHeight, convInfo.strideWidth,      // Stride.
    convInfo.dilationHeight, convInfo.dilationWidth,  // Dilation.
    convInfo.inHeight, convInfo.inWidth,              // Conv dims.
    convInfo.effectiveFilterHeight,
    convInfo.effectiveFilterWidth  // Filter dims.
  ];
  const uniformData = {type: 'int32', data: dimensions}

**Review by qjia7 — COMMENTED** [review]
Overall comments:
1. Remove `disableShapesUniforms`. Currently, only two programs set this flag to true. And it seems that they can be removed. And it also brings many overhead in the preprocessor file. So just remove it.
2. In your current logic, it seems that you still depends on the inputShape.length and outShape.length. So for this PR, I suggest keep `inputShape.length` and `inputShape.length` in the common shader key part. For example, change `shapes.join(',')` -> `shapesLength.join(',')` in `makeShaderKey`  so that you don't need to change the current `shaderKey` in each program.

With above changes, I think this PR will be simplified a lot. Thanks.

File: tfjs-backend-webgpu/src/kernels/AvgPool.ts (hunk 19 above)
Comment (axinging) [reply tfjs-backend-webgpu/src/kernels/AvgPool.ts:56]: This is grouped as shader type: float/int, vec2/ivec2, vec3/ivec3, vec4/ivec4

File: tfjs-backend-webgpu/src/kernels/binary_op_webgpu.ts [hunk 20]
```diff
@@ -33,22 +32,24 @@ export class BinaryOpProgram implements WebGPUProgram {
...
+    this.sizeFit = this.size % workGroupSizeX === 0;
     this.shapesFit = util.arraysEqual(aShape, bShape) && this.sizeFit;
     this.workPerThread = this.sizeFit || this.shapesFit ? 1 : 2;
 
     this.dispatch = computeDispatch(
         this.dispatchLayout, this.outputShape, this.workGroupSize,
         [this.workPerThread, 1, 1]);
-    this.shaderKey = `binary_${op}`;
+    this.shaderKey = `binary_${op}_${aShape}_${bShape}_${this.outputShape}_${
+        this.sizeFit}_${this.shapesFit}`;
```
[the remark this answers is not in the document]
axinging: I need some investigation to under why this is required. Currently I will add a TODO for this, hopefuly we can remove these shapes..

tfjs-backend-webgpu/src/kernels/clip_webgpu.ts
```diff
@@ -41,18 +42,19 @@ export class ClipProgram implements WebGPUProgram {
...
     const type = getCoordsDataType(this.outputShape.length);
-    const size = util.sizeFromShape(this.outputShape);
     const userCode = `
       void main() {
         int index = int(gl_GlobalInvocationID.x);
         for(int i = 0; i < ${this.workPerThread}; i++) {
           int flatIndex = index * ${this.workPerThread} + i;
-          if(flatIndex < ${size}) {
+          if(flatIndex < size) {
             ${type} coords = getCoordsFromFlatIndex(flatIndex);
```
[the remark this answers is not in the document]
axinging: I thin this be another PR.

tfjs-backend-webgpu/src/kernels/mirror_pad_webgpu.ts
```diff
@@ -48,11 +49,11 @@ export class MirrorPadProgram implements WebGPUProgram {
     this.paddings = paddings;
     this.offset = mode === 'reflect' ? 0 : 1;
     this.shaderKey = `mirrorPad_${mode}_${paddings}`;
```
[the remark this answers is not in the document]
axinging: Yes, I marked this in my shader key table. But all case passed, it seems we need a shader key test case.

tfjs-backend-webgpu/src/kernels/transpose_webgpu.ts
axinging: Will results in four fail, Add todo for this.
FAILED transpose test-webgpu {"WEBGPU_CPU_FORWARD":false} 3D [r, c, d] => [d, c, r]
debug.js:15 FAILED transpose test-webgpu {"WEBGPU_CPU_FORWARD":false} 3D [r, c, d] => [r, d, c]
debug.js:15 FAILED transpose test-webgpu {"WEBGPU_CPU_FORWARD":false} 4D [r, c, d, e] => [c, r, e, d]
debug.js:15 FAILED transpose test-webgpu {"WEBGPU_CPU_FORWARD":false} 4D [r, c, d, e] => [e, r, c, d]
debug.js:15 FAILED transpose test-webgpu {"WEBGPU_CPU_FORWARD":false} 4D [r, c, d, e] => [d, c, e, r]

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -483,18 +483,134 @@ export class WebGPUBackend extends KernelBackend {
...
+          break;
+        case 3:
+          baseAlignment = 4;
+          break;
+        case 4:
+          baseAlignment = 4;
+          break;
+        default:
+          util.assert(false, () => `Unsupported ${d.data.length}D shape`);
+      }
```
xhcao, line 571: Could replace this switch branch as shown below?
util.assert(false, () => `Unsupported ${d.data.length}D shape`);
const baseAlignment = d.data.length == 0 ? 1 : (d.data.length == 3 : 4 : d.data.length);

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -483,18 +483,134 @@ export class WebGPUBackend extends KernelBackend {
...
+    let uniformsWithType = bufferShapes.map(d => {
+      return {type: 'int32', data: d};
+    });
+    const strides = util.computeStrides(output.shape);
+    uniformsWithType.push({type: 'int32', data: strides});
+    if (program.size != null) {
+      uniformsWithType.push({type: 'int32', data: [program.size]});
+    }
+    if (programUniforms) {
+      uniformsWithType = [...uniformsWithType, ...programUniforms];
```
xhcao, line 605: Do we merge these two uniform blocks together here? If yes, why? Their attributes are different.
If we do not merge them together, do we only need to handle `int` type in numberOrArrayToDataView function? And do this uniform block data could be packed into Array<number>, instead of Array<{type: string; data: number[]}>? We do not need to encode and decode it.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -483,18 +483,134 @@ export class WebGPUBackend extends KernelBackend {
...
+              dataViewIndex * BYTES_PER_ELEMENT, arrayData, true);
+          dataViewIndex++;
+        }
+      }
+    });
+
+    return uniformDataView;
+  }
+
+  private computePadding(uniformsWithType:
```
xhcao, line 539: Because we had already added padding in this function. Is it better to rename it as "addPaddingToFollowStd140"?

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
```diff
@@ -77,66 +80,68 @@ export class ReduceProgram implements WebGPUProgram {
...
+         ${
+        this.reduceType === 'sum' || this.reduceType === 'mean' ||
+                this.reduceType === 'prod' ?
+            `if(currentSize > 1) bestValue=${initValue};` :
+            ''}
+       }
+       if (gl_LocalInvocationID.x == 0) {
+         ${outputSnippet}
+       }
+     `;
```
xhcao, line 114: Are these codes only format issue? If so, could we upload another CL only fix format issue, if so, it will be more readable for this CL.

axinging: This is done by the Visual studio code automatically.

tfjs-backend-webgpu/src/shader_preprocessor.ts
```diff
@@ -294,21 +300,23 @@ function getSamplerFromInInfo(inInfo: InputInfo, isVec4: boolean): string {
...
     float ${funcName}(${inputs}) {
       return float(${texName}[getFlatIndex(${type}(${dims.join(',')}),
-        ${getShapeCoords(inInfo.shape)})]);
+        ${shapeStr})]);
     }
-  `;
+   `;
 }
 
 function getSamplerAtOutputCoords(
```
qjia7, line 322: For `const broadcastDims = backend_util.getBroadcastDims(inInfo.shape, outShape);`, `broadcastDims` is also a constant value and not impacted by inRank or outRank. It should be also part of the shader key.

tfjs-backend-webgpu/src/backend_webgpu.ts
axinging: This will put all uniforms inside one uniform block. These uniforms data are small, so I think no need to create more than one slot.
These uniforms are from two different  sources (automaticall generated: shapes, shape strides, program size; and program defined). For one uniform buffer case, if we want to got correct padding, we need get all uniforms first, then compute the padding.

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
```diff
@@ -52,7 +54,8 @@ export class ReduceProgram implements WebGPUProgram {
         this.dispatchLayout, this.outputShape, this.workGroupSize);
 
     this.reduceType = reduceType;
-    this.shaderKey = `reduce_${reduceType}`;
+    this.shaderKey =
+        `reduce_${reduceType}_${outputDtype}_${this.outputShape.length}`;
```
qjia7, line 58: `${outputDtype}_${this.outputShape.length}` is already in the common part?

tfjs-backend-webgpu/src/kernels/pad_webgpu.ts
```diff
@@ -44,12 +45,12 @@ export class PadProgram implements WebGPUProgram {
 
     this.xShape = xShape;
     this.paddings = paddings;
-    this.shaderKey = `pad_${paddings}`;
+    this.shaderKey = `pad_${paddings}_${xShape}`;
```
qjia7, line 48: Why is `${xShape}` needed?

tfjs-backend-webgpu/src/kernels/im2col_webgpu.ts
```diff
@@ -44,11 +45,11 @@ export class Im2ColProgram implements WebGPUProgram {
         [this.workPerThread, 1, 1]);
     this.inputShape = inputShape;
     this.convInfo = convInfo;
-    this.shaderKey = `im2col_${convInfo}`;
+    this.shaderKey = `im2col${convInfo.inShape}`;
```
qjia7, line 48: There are too many constant values in this shader. I don't think your change is correct.

tfjs-backend-webgpu/src/kernels/crop_and_resize_webgpu.ts
```diff
@@ -42,8 +41,8 @@ export class CropAndResizeProgram implements WebGPUProgram {
     this.dispatch = computeDispatch(
         this.dispatchLayout, this.outputShape, this.workGroupSize);
 
-    this.shaderKey =
-        `cropAndResize_${method}_${cropSize}_${extrapolationValue}`;
+    this.shaderKey = `cropAndResize_${method}_${cropSize}_${
+        extrapolationValue}_${imageShape}_${boxShape}_${this.outputShape}`;
```
qjia7, line 45: Why do you still need this since they are all in the uniforms?

tfjs-backend-webgpu/src/kernels/binary_op_webgpu.ts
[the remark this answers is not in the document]
qjia7: It's because `broadcastDims` is still used as a constant value, but you didn't add it to the shader key. See my comment in preprocessor.

tfjs-backend-webgpu/src/kernels/batchnorm_webgpu.ts
```diff
@@ -57,7 +57,7 @@ export class BatchNormProgram implements WebGPUProgram {
     this.offsetShape = offsetShape;
     this.scaleShape = scaleShape;
     this.varianceEpsilon = varianceEpsilon;
-    this.shaderKey = `batchNorm_${varianceEpsilon}`;
+    this.shaderKey = `batchNorm_${varianceEpsilon}_${xShape.length}`;
```
qjia7, line 60: `${xShape.length}` no needed?

tfjs-backend-webgpu/src/kernels/AvgPool.ts
qjia7: I am still confused that why we can't do like that. Will take with you offline.

tfjs-backend-webgpu/src/backend_webgpu.ts
axinging: Thanks @xhcao! I prefer the original, it's easily understood, and easy to support 5D/6D.

tfjs-backend-webgpu/src/kernels/FromPixelsImageBitmap.ts
```diff
@@ -30,11 +30,14 @@ export function fromPixelsImageBitmap(args: {
 
   const outShape = [imageBitmap.height, imageBitmap.width, numChannels];
   const size = util.sizeFromShape(outShape);
-  const uniformData: [number, number] = [size, numChannels];
+  const strides = util.computeStrides(outShape);
+  const uniformData = [size, numChannels, ...strides];
```
qjia7, line 34: Why not just reuse the common `size` and `shapeStrides` in WebGPUProgram? And just keep `numChannels` as the only uniform for this program.

tfjs-backend-webgpu/src/shader_preprocessor.ts
```diff
@@ -431,10 +441,9 @@ function generateGetOutputCoords(
...
         int(gl_GlobalInvocationID[${i}]);`;
     } else {
-      const strides =
-          symbolicallyComputeStrides(arr, `${getShapeCoords(outShape)}`);
+      const strides = symbolicallyComputeStrides(arr, 'outShape');
       gatherDimensionsStr += `int index${i} =
-        int(gl_GlobalInvocationID[${i}]);`;
+          int(gl_GlobalInvocationID[${i}]);`;
       for (let j = 0; j < strides.length; j++) {
         gatherDimensionsStr += `int d${arr[j]} = index${i} / ${strides[j]};`;
```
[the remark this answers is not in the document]
axinging: This is symbol strides, not the shape stride. One example is:
"int index0 =
          int(gl_GlobalInvocationID[0]);int d0 = index0 / (outShape[2] * outShape[1]);index0 -= d0 * (outShape[2] * outShape[1]);int d1 = index0 / outShape[2];int d2 = index0 - d1 * outShape[2];"

tfjs-backend-webgpu/src/shader_preprocessor.ts
```diff
@@ -80,7 +65,17 @@ export interface InputInfo {
 
 export function makeShader(
     inputInfo: InputInfo[], outputData: {dtype: DataType, shape: number[]},
-    program: ProgramParams): string {
+    program: ProgramParams, isFromPixel = false): string {
```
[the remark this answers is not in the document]
axinging: FromPixel  uses different uniforms and bind group layout.

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
axinging: Great, thanks jiajia!

tfjs-backend-webgpu/src/shader_preprocessor.ts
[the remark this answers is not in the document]
qjia7: Oh, yes. You are right. Thanks for pointing it out.

tfjs-backend-webgpu/src/kernels/batchnorm_webgpu.ts
axinging: Done

tfjs-backend-webgpu/src/kernels/crop_and_resize_webgpu.ts
axinging: We need some extra efforts to remove this from shaderKey.

tfjs-backend-webgpu/src/kernels/FromPixelsImageBitmap.ts
axinging: Frompixel calls recordFromPixelsCommands, not runProgram. If we want to use size and shapeStrides, we need to copy a lot of code into recordFromPixelsCommands.

tfjs-backend-webgpu/src/kernels/batchnorm_webgpu.ts
axinging: Done

tfjs-backend-webgpu/src/kernels/im2col_webgpu.ts
axinging: I think the goal  for this PR is to use shape/size as uniform. There are some left const in program, suggest we do this in another pr

tfjs-backend-webgpu/src/kernels/crop_and_resize_webgpu.ts
axinging: There are so many const in this program, imageHeight.  I need more understand of this, suggest do this in another PR.

tfjs-backend-webgpu/src/shader_preprocessor.ts
axinging: done

tfjs-backend-webgpu/src/backend_webgpu.ts
axinging: Currently there is only one kind of padding is computed. Mayebe we can rename it when we have more.

tfjs-backend-webgpu/src/kernels/pad_webgpu.ts
axinging: xShape is used by const start and end.

tfjs-backend-webgpu/src/kernels/webgpu_program.ts
```diff
@@ -40,6 +43,7 @@ export interface WebGPUProgram {
   // the group.
   workGroupSize?: [number, number, number];
   isVec4?: boolean;
+  size?: number;
```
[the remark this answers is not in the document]
axinging: The this.outputShape used for size computation can be only known  when program constructor is called. So I think current design is better.

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
```diff
@@ -27,14 +28,15 @@ export class ReduceProgram implements WebGPUProgram {
   dispatch: [number, number, number];
   workGroupSize: [number, number, number];
   variableNames = ['x'];
+  uniforms = 'ivec2 reduceShape;';
```
qjia7, line 31: The `reduceShape` is the original 'this.inputShape'? So why not keep the original name?

tfjs-backend-webgpu/src/kernels/AvgPool.ts
qjia7: Got it. Thanks.

tfjs-backend-webgpu/src/kernels/FromPixelsImageBitmap.ts
qjia7: Yes, I remember that. It's ok for me to keep it for now.

tfjs-backend-webgpu/src/kernels/binary_op_shared_webgpu.ts
```diff
@@ -56,11 +57,11 @@ export class BinaryOpSharedProgram implements WebGPUProgram {
     this.shaderKey = `binaryShared_${op}`;
```
qjia7, line 57: Please add `this.lastDimensionSize` as part of shader key since it's used as the shared memory size. And the `sizeFit` should also in the shader.

tfjs-backend-webgpu/src/kernels/im2col_webgpu.ts
qjia7: I mean `convInfo` is not  saved as uniforms in this program. In your change, it will result incorrect result.

tfjs-backend-webgpu/src/kernels/unary_op_webgpu.ts
```diff
@@ -87,11 +88,11 @@ export class UnaryOpProgram implements WebGPUProgram {
         [this.workPerThread, 1, 1]);
     this.op = op;
     this.shaderKey = `unary_${op}`;
```
qjia7, line 90: Add `fit` to the shader key.

tfjs-backend-webgpu/src/kernels/webgpu_program.ts
```diff
@@ -31,6 +31,8 @@ export interface WebGPUProgram {
   dispatch: [number, number, number];
   variableNames: string[];
   uniforms?: string;
+  // When true, shapes are used as const, otherwise are stored in uniform
+  // buffer.
```
qjia7, line 35: Deprecated comments?

tfjs-backend-webgpu/src/shader_preprocessor.ts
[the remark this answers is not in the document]
qjia7: Thanks for the reminder.

qjia7 (commented):
Looks better. Close to send it out for review:)

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
axinging: They are different

tfjs-backend-webgpu/src/kernels/webgpu_program.ts
axinging: done

tfjs-backend-webgpu/src/kernels/unary_op_webgpu.ts
axinging: done

tfjs-backend-webgpu/src/kernels/im2col_webgpu.ts
axinging: ${convInfo} is an object. I will put all members used as const in shader key.

tfjs-backend-webgpu/src/kernels/binary_op_shared_webgpu.ts
axinging: done

tfjs-backend-webgpu/src/kernels/binary_op_shared_webgpu.ts
```diff
@@ -53,14 +54,14 @@ export class BinaryOpSharedProgram implements WebGPUProgram {
     this.dispatch = computeDispatch(
         this.dispatchLayout, this.outputShape, this.workGroupSize,
         [this.workPerThread, 1, 1]);
-    this.shaderKey = `binaryShared_${op}`;
+    this.shaderKey = `binaryShared_${op}_${this.lastDimensionSize}`;
```
qjia7, line 57: Still missing `sizeFit`?

qjia7 (commented):
Overall, I think we have resolved all the issues that we reach an agreement. Let's add more reviewers to look at this PR. @kainino0x @lina128 @jinjingforever  @pyu10055 

A main issue left for correctness is that some programs(for example, conv2d/matmul) `shaderKey` are missing `fit` part, which is used for avoiding out of bounds checking if possible.

lina128 (approved):
LGTM!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4819#-:-MYIV0RHD8BmRZM08Zz4:b-85i648)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @jinjingforever, @kainino0x, @pyu10055, @qjia7, and @xhcao)

tfjs-backend-webgpu/src/kernels/binary_op_shared_webgpu.ts
axinging: Thanks Jiajia, you are right, fit is needed here (And matmul/conv2d).  However this is not introduced by this PR and the goal of this PR is to uniformilize shape, and the change is not trivial, I prefer to fix this in a followup PR.

qjia7: Hi Xing, in fact, it's introduced by this PR. Previously, the inputShape/OutputShape is as part of shader key. So you don't need to consider the fit issue. However, in this patch, you removed the shapes form shader key. Then you must add the `fit` to the shader key. Otherwise, your change will bring potential bugs.

axinging: Oh, yes,  I will update it.

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
qjia7: If they are different, I notice that the original shader is using `this.inputShape`. But you are changing it to `reduceShape`. Is it expected?

tfjs-backend-webgpu/src/kernels/webgpu_program.ts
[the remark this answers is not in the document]
qjia7: Just to clarify that your reason is not valid. You can put the size as the program `uniforms` variable using `uniforms = 'int size';` in the needed programs. The program uniforms data and the program are together to pass the `runWebGPUProgram`. So you can always get the program.outputShape to calculate the size.

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
axinging: It is expected. In shader,  when enabled uniform, the inputShape is mapped to xShape. 
And the reduceShape(ts) mapped to reduceShape(shader).

    const [outputShape, reduceShape] =
        backend_util.computeOutAndReduceShapes(this.inputShape, [1]); // this.inputShape,=> xShape.

qjia7:
There is one issue that I want to discuss with all reviewers and the PR owner.
In this PR, except for putting inputs/output shapes into uniforms, it also puts output `size` and output `strides` into the uniform. However, the `size` and `strides` can also be computed in the shader, and the uniforms can be removed for `size` and `strides`.
For memory bound, I think putting these calculation into shader is better.
For compute bound, I think putting these calculation into cpu and uploading to uniform may be better.

 Per my understanding, most of kernels are memory bound. Maybe putting the calculation into shader is better? But I am not sure. So I want to hear all your opinions about this. 

@axinging If it's convenient for you, can you collect some data under these two situations? It's also possible that there is no big difference for these two.

lina128 (commented):
If there is not big difference in terms of speed, I vote for putting the calculation into shader so that shape is single source of truth.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4819#-:-MYNGkW45256pHZWw0HR:b8qtl6z)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @jinjingforever, @kainino0x, @pyu10055, @qjia7, and @xhcao)

axinging:
The key logic compute stride is as below:
```
export function computeStrides(shape: number[]): number[] {
  const rank = shape.length;
  if (rank < 2) {
    return [];
  }

  const strides = new Array(rank - 1);
  strides[rank - 2] = shape[rank - 1];
  for (let i = rank - 3; i >= 0; --i) {
    strides[i] = strides[i + 1] * shape[i + 1];
  }
  return strides;
}
```
This sure can be done in shader or CPU.  
If done in CPU, shader need to access this by uniform (For 4d, this will three int32). And think about that we already have other data in the uniform buffer(such as in/out shapes), For most case, those uniform data can be fetched into one or two cache line, for example, when running yarn test, the uniform byte length range from 12-120. So, I think three int32(worst case) will not add too much memory IO.

If done in GPU(shader), every thread instance will compute same computeStrides above, this will have bad influence on perf and power. 

Anyway, during my test, I didn't found any obvious performance regression when compute stride in CPU, so I think current compute stride in CPU is not bad idea at perf and power side.  Maybe we can try compute stride in GPU in another PR.

qjia7:
> The key logic compute strides is as below, this sure can be done in GPU(shader) or CPU(This PR proposed):
> 
> ```
> export function computeStrides(shape: number[]): number[] {
>   const rank = shape.length;
>   if (rank < 2) {
>     return [];
>   }
> 
>   const strides = new Array(rank - 1);
>   strides[rank - 2] = shape[rank - 1];
>   for (let i = rank - 3; i >= 0; --i) {
>     strides[i] = strides[i + 1] * shape[i + 1];
>   }
>   return strides;
> }
> ```

There is already an implementation `symbolicallyComputeStrides` in `shader_util.ts`. And `generateGetOutputCoords` is using it.

> If done in GPU(shader), every thread instance will compute same computeStrides above, this will have bad influence on power.

I think it also has no obvious perf/power impact for GPU since all threads are running parallel ;)
> 
> Anyway, during my test, I didn't found any obvious performance regression when compute stride in CPU, so I think current compute stride in CPU is not bad idea at perf and power side. Maybe we can try compute stride in GPU in another PR.
> 
It' just weird for me that for `generateGetOutputCoords`, we use `symbolicallyComputeStrides` in shader, for `generateGetCoordsFromFlatIndex`, we use the value gotten from `computeStrides`. I hope these two can be consistent.

kainino0x:
I think you should pick whatever performs well, and if they're the same whatever is easiest to maintain.

My intuition is about the performance is: I wouldn't expect there to be much memory bandwidth overhead to a few bytes of uniforms, as the same value is shared by all threads. Meanwhile I do expect there would be a small power/performance cost to redundant computations.

Do you know which workloads are compute vs memory bound? I suspect almost all kernels are memory-bound (at least for part of their execution time), though additional uniforms probably wouldn't make anything worse because uniform loads can be issued earlier than storage loads.

qjia7:
> My intuition is about the performance is: I wouldn't expect there to be much memory bandwidth overhead to a few bytes of uniforms, as the same value is shared by all threads. Meanwhile I do expect there would be a small power/performance cost to redundant computations.
> 
Hi Kai, I did a quick test for the performance on supported models on CFL. The result is very close and overlap on several sets. So I can't say which is better based on the existed test. Good to know that uniforms won't be a concern for memory bandwidth. I am ok to keep the current change. @axinging please make sure uniform strides are used in all places in the following PR to further clean the code.

> Do you know which workloads are compute vs memory bound? I suspect almost all kernels are memory-bound (at least for part of their execution time), though additional uniforms probably wouldn't make anything worse because uniform loads can be issued earlier than storage loads.

Agree with you that almost all or all kernels are memory-bound in TFJS WebGPU.

@kainino0x Do you have other comments for this PR? I would like to merge it after getting your approval. Thanks.

kainino0x:
Sorry I don't think I will have time to review fully, and definitely not on
Monday. I skimmed it and I trust your review, so if it's okay with you, I
think you can merge without my review.

qjia7:
> Sorry I don't think I will have time to review fully, and definitely not on Monday. I skimmed it and I trust your review, so if it's okay with you, I think you can merge without my review.

Got it. Landing it now. If any issues pop up or any comments are added later, we can fix them with follow-up PRs. Thanks.

pyu10055:
@qjia7 This is great discussion, if possible, can we setup a meeting to discuss your thoughts on migrating webgl backend to storing uniform input/output shapes to reduce shader compilation.

qjia7:
> @qjia7 This is great discussion, if possible, can we setup a meeting to discuss your thoughts on migrating webgl backend to storing uniform input/output shapes to reduce shader compilation.

Sure. Feel free to setup a meeting and forward the calendar to me.

PR #4831, opened 2021-03-18
e2e: Use env().set instead of env().setFlags

qjia7:
setFlags will reset all flags with the given list and override other flags values.
So we should use set instead of setFlags function.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4831)

### DIFF
--- a/e2e/benchmarks/benchmark_util.js
+++ b/e2e/benchmarks/benchmark_util.js
@@ -475,9 +475,8 @@ async function setEnvFlags(flagConfig) {
               TUNABLE_FLAG_VALUE_RANGE_MAP[flag]}], while ${flagConfig[flag]}` +
           ' is found.');
     }
+    tf.env().set(flag, flagConfig[flag]);
   }
-
-  tf.env().setFlags(flagConfig);
 
   // `WASM_HAS_SIMD_SUPPORT` and `WEBGL_VERSION` are also evaluated when
   // initializing backends, not only inferring.


qjia7:
@axinging  @haoyunfeix @gyagp, PTAL

haoyunfeix:
I thought it's been designed in this way, to constrain users to only set tunable flags (the keys of `TUNABLE_FLAG_TYPE_MAP`). If this blocks we testing some new flags we could add flag to TUNABLE_FLAG_TYPE_MAP.

pyu10055 (commented):
I agree with Yunfei, the idea is for benchmark tool, it will reset the flags to a fix state, so we are sure that what is being used for the benchmark, not trying to incremental change flags.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4831#-:-MW5PQVi7yaKbxJVvgG9:bmiixsd)* status: 0 of 1 approvals obtained

qjia7:
> I agree with Yunfei, the idea is for benchmark tool, it will reset the flags to a fix state, so we are sure that what is being used for the benchmark, not trying to incremental change flags.
> 
The issue is that webgpu is a real async implementation. So it will meet error if you use promise in a sync function in some situation, see #4303, #4492.  So to resolve that issue, we set `CHECK_COMPUTATION_FOR_ERRORS` to false for webgpu. However,  if we use `env().setFlags` in e2e, it will reset  `CHECK_COMPUTATION_FOR_ERRORS`  to true for webgpu and result in a lot of errors again.
`CHECK_COMPUTATION_FOR_ERRORS` is a tunable value for other backends, but for webgpu, currently it must be false. So that's why I suggest use `set` instead of `setFlags` so that it doesn't affect other flags' value which are not in `TUNABLE_FLAG_VALUE_RANGE_MAP`.

My understanding is that `TUNABLE_FLAG_VALUE_RANGE_MAP` just lists the flags that you can tune and the corresponding values that you can set in this benchmark. However, it shouldn't affect other flags that are not in this list.

pyu10055 (commented):
Got it, thank you for explaining. I think in the function getTunableRange(flag) method, you can limit the ranges programmatically, which means you can make CHECK_COMPUTATION_FOR_ERRORS can only be false for WebGPU backends.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4831#-:-MWC4Ywk2mk8xPcqSPup:blfvvk5)* status: 0 of 1 approvals obtained

Issue #2111, opened 2019-09-29
add `meshgrid` function


silky:
would be nice to have an equiv of `tf.meshgrid` in python tensorflow.

rthadur:
@silky  can you please tell us your use case ?

silky:
```
  x = tf.linspace(-1, 1, w);
  y = tf.linspace(-1, 1, h);

  const xx = tf.matMul( tf.ones  ([h, 1]), x.reshape([1, w]) )
  const yy = tf.matMul( y.reshape([h, 1]), tf.ones  ([1, w]) );
```

vs. `xx, yy = tf.meshgrid(x, y)`

hoechenberger:
Hello @rthadur, I see you closed this issue. Does this mean the feature was implemented? Or that it is *not* going to be implemented, after all? Or that there is a good alternative to achieve the desired behavior? Thank you very much!

hoechenberger:
Friendly ping @rthadur :)

DCtheTall:
@rthadur I am interested in trying to implement this, if you're still accepting contributions 😄

rthadur:
@DCtheTall please go ahead , thank you.

DCtheTall:
@rthadur I handled the 2D case since JS doesn't let you have kwargs after variable-length argument parameters like Python and to keep the scope of the PR simple at first.

I added unit tests based on [NumPy's unit tests](https://github.com/numpy/numpy/blob/cb557b79fa0ce467c881830f8e8e042c484ccfaa/numpy/lib/tests/test_function_base.py#L2222) since the Python TensorFlow tests just compare the output to NumPy's implementation.

PR #4855, opened 2021-03-23
Add tf.meshgrid support for 2D case only

DCtheTall:
Fixes [#2111](https://github.com/tensorflow/tfjs/issues/2111)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4855)


mattsoulanille (changes requested):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4855)* status: 0 of 1 approvals obtained (waiting on @DCtheTall)

___
*[tfjs-core/src/ops/meshgrid.ts, line 26 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdvBzG1HPEnuNRQxVe:-MWdvBzG1HPEnuNRQxVf:bqsbdd) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L26)):*
> ```XML
> import {sizeFromShape} from '../util_base';
> 
> export function meshgrid<T extends Tensor>(
> ```

Please add a jsdoc comment so this shows up on our documentation site. It looks like it should go under [Operations -> Slicing and Joining](https://js.tensorflow.org/api/latest/#Operations-Slicing%20and%20Joining), which can be done by adding `@doc {heading: 'Operations', subheading: 'Slicing and Joining'}` to the jsdoc comment (feel free to add it to a different section if that doesn't seem right). See `gather_nd.ts` or any other op for an example. 
___
*[tfjs-core/src/ops/meshgrid.ts, line 27 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWe-uJn44lYA1vg7MRw:-MWe-uJn44lYA1vg7MRx:bi52i8f) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L27)):*
> ```XML
> 
> export function meshgrid<T extends Tensor>(
>     x?: T|TensorLike, y?: T|TensorLike, indexing = 'xy'): T[] {
> ```

I think we should solve the question of how to support passing any number of arguments even if the implementation only currently supports two. This looks like it can be [solved](https://www.typescriptlang.org/play?#code/PTAEFkHsGMGtQC4FMB2BnSAnRBPADkgFAL5KgAqqG2AvKCgK4C2ARkpgNoC6A3MaRSpYAMgEtYZOpXRZQAH3rM2mPvwIQkaABYBzTKIAmAeTwJRkdKDoBvQqHuhRKA0gAeTnQH4AXKADkrjh+8v6iAFZ+hAC+hGpk4Jq6+gYAgpg6aADqoghaJmYWaAA85KBuyM5ogjKYAHxWoHYOHAB0bRXUaL4AFORy0tRiEgCU3AA0oJCm5ui+Cdp6hvkzaLxxGgvJaRklZa4VBlUDWPV0G0mG21k5edOFJfUKvf1CmENIo1yEAGYMKNAFFCgJiJRYGXblVCHarUWrdNotACG6S65zBVwew185G4oFsDkc31A3WRGRaABtUDpclYaHQAAzDPFNAn2TBIBAMTBA7h8AkxFmgSkIRCvVHPY5vcQfXmC4WTO6zNHJZaFEJ-FzfJxIAx8hzQQoi8mItAIbYNUloDiWilUmkAWlAAEY1gSQKAAMJaJBwQlCk1m9KOKqI0U1FqgACSRJQkAQE1yqEcIqYDFNoDYyqWirQLUF7oAcnGyLlRFVoN7fWX6HHQAbMOyAaBvrJE6B2YjyYhvmFc4LRETuiQCJAicbTea6XQ-IxWOxgnIFOPAzpHOgEIj-khR6A0phETgmfjWfYOlgqnRLaATcS+pL3p89fYomVyWgyMeTyBWu0xRMpoCqwWii15VPMFypCi2S5Kq6APE+rJnpgF7XiiLRoOSojQEg3T0hMNqUig1JaKAjpOkyN4Sq8D6yie9gASswEZNaaGEcRpHOlwoFZsYOYIQKbpgCCmyGI4TB4JSIIoBugIInmgrspy3KgBwHBOhMABMEwAMxcBMHAACwTAArBMABsXBcTeOJrAKwkQd06mgFpKnaRMRmgMZel4k4LjuERvgBEEUTDHw9lgt0amaTp3keRwpmgGZEwAOxcKFQA) with variadic tuple types, but that solution [does not work in TypeScript v3.6.2](https://www.typescriptlang.org/play?ts=3.6.3#code/PTAEFkHsGMGtQC4FMB2BnSAnRBPADkgFAL5KgAqqG2AvKCgK4C2ARkpgNoC6A3MaRSpYAMgEtYZOpXRZQAH3rM2mPvwIQkaABYBzTKIAmAeTwJRkdKDoBvQqHuhRKA0gAeTnQH4AXKADkrjh+8v6iAFZ+hAC+hGpk4Jq6+gYAgpg6aADqoghaJmYWaAA85KBuyM5ogjKYAHxWoHYOHAB0bRXUaL4AFORy0tRiEgCU3AA0oJCm5ui+Cdp6hvkzaLxxGgvJaRklZa4VBlUDWPV0G0mG21k5edOFJfUKvf1CmENIo1yEAGYMKNAFFCgJiJRYGXblVCHarUWrdNotACG6S65zBVwew185G4oFsDkc31A3WRGRaABtUDpclYaHQAAzDPFNAn2TBIBAMTBA7h8AkxFmgSkIRCvVHPY5vcQfXmC4WTO6zNHJZaFEJ-FzfJxIAx8hzQQoi8mItAIbYNUloDiWilUmkAWlAAEY1gSQKAAMJaJBwQlCk1m9KOKqI0U1FqgACSRJQkAQE1yqEcIqYDFNoDYyqWirQLUF7oAcnGyLlRFVoN7fWX6HHQAbMOyAaBvrJE6B2YjyYhvmFc4LRETuiQCJAicbTea6XQ-IxWOxgnIFOPAzpHOgEIj-khR6A0phETgmfjWfYOlgqnRLaATcS+pL3p89fYomVyWgyMeTyBWu0xRMpoCqwWii15VPMFypCi2S5Kq6APE+rJnpgF7XiiLRoOSojQEg3T0hMNqUig1JaKAjpOkyN4Sq8D6yie9gASswEZNaaGEcRpHOlwoFZsYOYIQKbpgCCmyGI4TB4JSIIoBugIInmgrspy3KgBwHBOhMABMEwAMxcBMHAACwTAArBMABsXBcTeOJrAKwkQd06mgFpKnaRMRmgMZel4k4LjuERvgBEEUTDHw9lgt0amaTp3keRwpmgGZEwAOxcKFQA) (we're on 3.5.3). 

Since we're planning on updating TypeScript anyway, these types should be fine for now. We can convert them to use a variadic tuple when we update TypeScript and when we add support for arbitrarily many input tensors.

The one change I would request is that we store the python implementation's keyword args, like `indexing`, in an object so we can add more / omit them as needed.
___
*[tfjs-core/src/ops/meshgrid.ts, line 32 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdswweBy7A0AuA9KZg:-MWdswweBy7A0AuA9KZh:bg4cwc8) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L32)):*
> ```XML
>         `${indexing} is not a valid third argument to meshgrid`);
>   }
>   if (!x) {
> ```

`number` is a subtype of `TensorLike`,  so `meshgrid(0)` will return `[]` instead of `[0]`. Probably replace this condition with `x === undefined`. It would probably also be good to add a test for this case.
___
*[tfjs-core/src/ops/meshgrid.ts, line 41 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdtiWqC1YQqV6whwEm:-MWdtiWrDotomkTh9kXB:bkkw74) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L41)):*
> ```XML
>   }
> 
>   if (!y) {
> ```

Same here as with `if (!x) {` above.
___
*[tfjs-core/src/ops/meshgrid.ts, line 50 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdufl13_ELTMmJcAOD:-MWdufl13_ELTMmJcAOE:bgro1q4) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L50)):*
> ```XML
>   }
> 
>   const dtype = $x.dtype;
> ```

Should we preserve the dtype for each of the inputs independent of one another? If I correctly understand what meshgrid does, it's only re-displaying / copying the contents of the inputs into a different view, so we should be able to preserve each input's dtype.

On the other hand, the python version throws an error if you mix dtypes of arguments.
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdxkbwB2kdNjUpw0cn:-MWdxkbwB2kdNjUpw0co:bodlgh1) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L28)):*
> ```XML
>   });
> });
> ```

A test case for `tf.meshgrid(0)` and `tf.meshgrid([some tensor], 0)` would help catch a bug I found above.
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 86 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWe-_7O-VSmlN448QtX:-MWe-_7O-VSmlN448QtY:b-gujzum) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L86)):*
> ```XML
>   });
> 
>   it('Should use the input tensor dtype', async () => {
> ```

If it makes sense to preserve the dtype for each input as I mentioned above, this test will need to be updated.
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 98 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdyxSM2dcbozWRO7w9:-MWdyxSM2dcbozWRO7wA:b-uisc8c) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L98)):*
> ```
>     expect(() => tf.meshgrid(3 as {} as Tensor)).toThrow();
>     expect(() => tf.meshgrid([1], 3 as {} as Tensor)).toThrow();
> ```

Should these fail? It looks like they work in python, and it looks like they should work with your implementation as it's currently written. Maybe I'm missing something here.

DCtheTall (commented):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4855)* status: 0 of 1 approvals obtained (waiting on @mattsoulanille)

___
*[tfjs-core/src/ops/meshgrid.ts, line 26 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdvBzG1HPEnuNRQxVe:-MWz8zYaAGZgF_Piw5HW:b-aowwae) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L26)):*

Please add a jsdoc comment so this shows up on our documentation site. It looks like it should go under [Operations -> Slicing and Joining](https://js.tensorflow.org/api/latest/#Operations-Slicing%20and%20Joining), which can be done by adding `@doc {heading: 'Operations', subheading: 'Slicing and Joining'}` to the jsdoc comment (feel free to add it to a different section if that doesn't seem right). See `gather_nd.ts` or any other op for an example. 
</blockquote></details>

Done. Totally forgot about that, whoops!
___
*[tfjs-core/src/ops/meshgrid.ts, line 27 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWe-uJn44lYA1vg7MRw:-MWz1aKY74eWqVldGo5W:b-tdyhui) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L27)):*
> Since we're planning on updating TypeScript anyway, these types should be fine for now. We can convert them to use a variadic tuple when we update TypeScript and when we add support for arbitrarily many input tensors.

I agree once variadic tuple types are supported then we should extend this code to support the N-D case, but for now the 1-/2-D case is good.

> The one change I would request is that we store the python implementation's keyword args, like indexing, in an object so we can add more / omit them as needed.

Done.
___
*[tfjs-core/src/ops/meshgrid.ts, line 32 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdswweBy7A0AuA9KZg:-MWz-5yG9aZpbymbnDBx:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L32)):*

`number` is a subtype of `TensorLike`,  so `meshgrid(0)` will return `[]` instead of `[0]`. Probably replace this condition with `x === undefined`. It would probably also be good to add a test for this case.
</blockquote></details>

Done.
___
*[tfjs-core/src/ops/meshgrid.ts, line 41 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdtiWqC1YQqV6whwEm:-MWz-57B1Qcx6VdtTmxS:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L41)):*

Same here as with `if (!x) {` above.
</blockquote></details>

Done.
___
*[tfjs-core/src/ops/meshgrid.ts, line 50 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdufl13_ELTMmJcAOD:-MWz-AYT8fV3BXaRQJkR:b9uyfxj) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L50)):*
> On the other hand, the python version throws an error if you mix dtypes of arguments.

Is this done in one of the ops that meshgrid uses internally? I am using this as a reference and I don't see that: https://github.com/tensorflow/tensorflow/blob/v2.4.1/tensorflow/python/ops/array_ops.py#L3476-L3553
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdxkbwB2kdNjUpw0cn:-MWz84BcAwDtCXkaujQ8:be4spzf) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L28)):*

A test case for `tf.meshgrid(0)` and `tf.meshgrid([some tensor], 0)` would help catch a bug I found above.
</blockquote></details>

Ack'd, see my reply to that comment.
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 86 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWe-_7O-VSmlN448QtX:-MWz92Ny0tbkw_9h7jzN:brj97f9) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L86)):*

If it makes sense to preserve the dtype for each input as I mentioned above, this test will need to be updated.
</blockquote></details>

Done (for now).
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 98 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdyxSM2dcbozWRO7w9:-MWz6x0xEFLfW3jWgfJF:b-hcjkke) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L98)):*
> Should these fail?

So the TensorFlow Python implementation does not check the rank of the tensors to make sure they are gt zero, but I am not sure that is intended behavior.

TensorFlow's Python tests compare the output against NumPy's implementation: https://github.com/tensorflow/tensorflow/blob/32f8688c1c108ab2df794fe4ceb9e058641359b0/tensorflow/python/kernel_tests/array_ops_test.py#L552

NumPy explicitly disallows scalars as arguments, and TF does not test against this in the Python anywhere AFAIK.

So I am thinking this might actually be a bug in the TF Python code, but wanted to get your thoughts.

mattsoulanille (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4855)* status: 0 of 1 approvals obtained (waiting on @DCtheTall)

___
*[tfjs-core/src/ops/meshgrid.ts, line 26 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdvBzG1HPEnuNRQxVe:-MX2cO2A4cM9gs4FrEGG:bat0qry) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L26)):*
<details><summary><i>Previously, DCtheTall (Dylan Cutler) wrote…</i></summary><blockquote>

Done. Totally forgot about that, whoops!
</blockquote></details>

No worries! Thanks for adding it.
___
*[tfjs-core/src/ops/meshgrid.ts, line 50 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdufl13_ELTMmJcAOD:-MX2_ApV7ofT02uYoOGg:b-qpgqjy) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L50)):*
When I try to create a meshgrid of `[1, 2, 3]` and `[4.0, 5.0, 6.0]` in the python implementation, I get 
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: cannot compute Mul as input #1(zero-based) was expected to be a float tensor but is a int32 tensor [Op:Mul]
```

so I'm guessing the [matmul](https://github.com/tensorflow/tensorflow/blob/v2.4.1/tensorflow/python/ops/array_ops.py#L3553) at the end is throwing the error I'm seeing.

This may be a bug in tf's meshgrid, since it seems to work in numpy:
```
>>> np.meshgrid([1, 2, 3], [4.0, 5.0, 6.0])
[array([[1, 2, 3],
       [1, 2, 3],
       [1, 2, 3]]), array([[4., 4., 4.],
       [5., 5., 5.],
       [6., 6., 6.]])]
```
It also works for tensors
```
>>> a = tf.constant([1,2,3])
>>> b = tf.constant([4,5,6], dtype='float32')
>>> np.meshgrid(a,b)
[array([[1, 2, 3],
       [1, 2, 3],
       [1, 2, 3]], dtype=int32), array([[4., 4., 4.],
       [5., 5., 5.],
       [6., 6., 6.]], dtype=float32)]
```

Given that that tf's implementation is tested against the numpy implementation, I think it's fine to mirror numpy instead of tf in this case (and preserve dtype of inputs).
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdxkbwB2kdNjUpw0cn:-MX2pstu4lC89EFoH8Kz:b-k61jwn) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L28)):*
<details><summary><i>Previously, DCtheTall (Dylan Cutler) wrote…</i></summary><blockquote>

Ack'd, see my reply to that comment.
</blockquote></details>

Keeping this open until we resolve my new comment on the 'invalid args' test.
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 98 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdyxSM2dcbozWRO7w9:-MX2f15q7a04yaSWpfcF:bht7xpu) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L98)):*
I haven't looked at the numpy implementation, and maybe I've misunderstood what you're saying, but I'm able to run `np.meshgrid` with scalars.

```
>>> np.version.version
'1.19.5'
>>> np.meshgrid(1,2)
[array([[1]]), array([[2]])]
>>> np.meshgrid([1], 3)
[array([[1]]), array([[3]])]
```
Do these test cases throw errors for you?


The [numpy docs](https://numpy.org/doc/stable/reference/generated/numpy.meshgrid.html) say meshgrid takes a number of `array_like` parameters, and I guess scalars fit into that type, at least in practice? It's kind of hard to tell since the type doesn't seem to be defined anywhere. Going off of [this stackoverflow](https://stackoverflow.com/questions/40378427/numpy-formal-definition-of-array-like-objects), it looks like most anything is allowed. I was even able to call `meshgrid` on two dictionaries:
```
>>> np.meshgrid({'x': 'hi'}, {'y', 'bye'})
[array([[{'x': 'hi'}]], dtype=object), array([[{'y', 'bye'}]], dtype=object)]
```
We of course don't have to support this case.

Given that `np.meshgrid` seems to work with scalars, I think it's probably a good idea to support them here as well. I think numpy just treats them as 1d arrays with a single element. Our [TensorLike](https://github.com/tensorflow/tfjs/blob/70f6eba60823209f2e213156493295583bb8a7a8/tfjs-core/src/types.ts#L137) type also includes scalars, so we won't have to change the type of the function.

What do you think?

DCtheTall (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4855)* status: 0 of 1 approvals obtained (waiting on @mattsoulanille)

___
*[tfjs-core/src/ops/meshgrid.ts, line 50 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdufl13_ELTMmJcAOD:-MX4RZbw-JT1o0IXIuTO:b-in1f1i) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L50)):*

When I try to create a meshgrid of `[1, 2, 3]` and `[4.0, 5.0, 6.0]` in the python implementation, I get 
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: cannot compute Mul as input #1(zero-based) was expected to be a float tensor but is a int32 tensor [Op:Mul]
```

so I'm guessing the [matmul](https://github.com/tensorflow/tensorflow/blob/v2.4.1/tensorflow/python/ops/array_ops.py#L3553) at the end is throwing the error I'm seeing.

This may be a bug in tf's meshgrid, since it seems to work in numpy:
```
>>> np.meshgrid([1, 2, 3], [4.0, 5.0, 6.0])
[array([[1, 2, 3],
       [1, 2, 3],
       [1, 2, 3]]), array([[4., 4., 4.],
       [5., 5., 5.],
       [6., 6., 6.]])]
```
It also works for tensors
```
>>> a = tf.constant([1,2,3])
>>> b = tf.constant([4,5,6], dtype='float32')
>>> np.meshgrid(a,b)
[array([[1, 2, 3],
       [1, 2, 3],
       [1, 2, 3]], dtype=int32), array([[4., 4., 4.],
       [5., 5., 5.],
       [6., 6., 6.]], dtype=float32)]
```

Given that that tf's implementation is tested against the numpy implementation, I think it's fine to mirror numpy instead of tf in this case (and preserve dtype of inputs).
</blockquote></details>

Ah, I see. In that case I think we should preserve the dtype here to mirror NumPy and I can file a bug against Python TensorFlow to point out the inconsistency.
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdxkbwB2kdNjUpw0cn:-MX4Rhyq6sm-i45D-4dB:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L28)):*

Keeping this open until we resolve my new comment on the 'invalid args' test.
</blockquote></details>

Done.
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 98 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdyxSM2dcbozWRO7w9:-MX4Rpz65ngGPVhA1pO2:b-wgutcy) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L98)):*

I haven't looked at the numpy implementation, and maybe I've misunderstood what you're saying, but I'm able to run `np.meshgrid` with scalars.

```
>>> np.version.version
'1.19.5'
>>> np.meshgrid(1,2)
[array([[1]]), array([[2]])]
>>> np.meshgrid([1], 3)
[array([[1]]), array([[3]])]
```
Do these test cases throw errors for you?


The [numpy docs](https://numpy.org/doc/stable/reference/generated/numpy.meshgrid.html) say meshgrid takes a number of `array_like` parameters, and I guess scalars fit into that type, at least in practice? It's kind of hard to tell since the type doesn't seem to be defined anywhere. Going off of [this stackoverflow](https://stackoverflow.com/questions/40378427/numpy-formal-definition-of-array-like-objects), it looks like most anything is allowed. I was even able to call `meshgrid` on two dictionaries:
```
>>> np.meshgrid({'x': 'hi'}, {'y', 'bye'})
[array([[{'x': 'hi'}]], dtype=object), array([[{'y', 'bye'}]], dtype=object)]
```
We of course don't have to support this case.

Given that `np.meshgrid` seems to work with scalars, I think it's probably a good idea to support them here as well. I think numpy just treats them as 1d arrays with a single element. Our [TensorLike](https://github.com/tensorflow/tfjs/blob/70f6eba60823209f2e213156493295583bb8a7a8/tfjs-core/src/types.ts#L137) type also includes scalars, so we won't have to change the type of the function.

What do you think?
</blockquote></details>

Oops, I realize now it was the function in tfjs I was using as a model for this one that didn't allow scalars (dumb mistake...).

I fixed my JS implementation to support scalars and added a unit test to cover this case.

mattsoulanille (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4855)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[tfjs-core/src/ops/meshgrid.ts, line 50 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdufl13_ELTMmJcAOD:-MXCrP5s5R2croOyEcAM:b-mpjepr) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid.ts#L50)):*
<details><summary><i>Previously, DCtheTall (Dylan Cutler) wrote…</i></summary><blockquote>

Ah, I see. In that case I think we should preserve the dtype here to mirror NumPy and I can file a bug against Python TensorFlow to point out the inconsistency.
</blockquote></details>

That sounds good. Thanks!
___
*[tfjs-core/src/ops/meshgrid_test.ts, line 98 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4855#-MWdyxSM2dcbozWRO7w9:-MXCreyRD0DG1YwJouV_:b-ucfknh) ([raw file](https://github.com/tensorflow/tfjs/blob/27a6e206ff4d5101de669d0024373cb4213d6dc5/tfjs-core/src/ops/meshgrid_test.ts#L98)):*
<details><summary><i>Previously, DCtheTall (Dylan Cutler) wrote…</i></summary><blockquote>

Oops, I realize now it was the function in tfjs I was using as a model for this one that didn't allow scalars (dumb mistake...).

I fixed my JS implementation to support scalars and added a unit test to cover this case.
</blockquote></details>

Thanks!

PR #4856, opened 2021-03-23
Build tfjs-core and tfjs-backend-cpu with Bazel

mattsoulanille:
This PR adds Bazel BUILD targets for tfjs-core and tfjs-backend-cpu. The following commands can invoke the new Bazel targets. Yarn rules have not been added for them since Bazel not the default way to build yet.
- `yarn bazel build //tfjs-core:tfjs-core_lib` compiles tfjs-core.
- `yarn bazel test //tfjs-core:tfjs-core_test` tests tfjs-core with a headless browser.
- `yarn bazel run //tfjs-core:tfjs-core_test` launches tfjs-core tests in a browser and waits for human interaction.
- `yarn bazel test //tfjs-core:browserstack_tfjs-core_test` tests tfjs-core on BrowserStack. Only Chrome on MacOS is enabled at the moment. This test requires that `BROWSERSTACK_USERNAME` and `BROWSERSTACK_KEY` are set in your environment.
- `yarn bazel run //tfjs-core:browserstack_tfjs-core_test` (theoretically) launches tfjs-core tests in a BrowserStack session for debugging and waits for you to connect before running the tests. I haven't been able to test this yet since I can't seem to get an interactive session to start (probably due to network issues).

Currently, these build targets use esbuild for bundling since it was the easiest to set up. We will probably use Rollup for some cases e.g. to support es5. Targets for publishing to npm (i.e. `pkg_npm`) are not yet written.

This PR also sets up Bazel CI tests. CI tests use a Bazel query to find what targets should be tested, so any Bazel target with `tag = ["ci"]` will be tested in CI. Bazel remote caching is now working, so we run all the tests on every commit and let Bazel figure out what actually needs to be run.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4856)


mattsoulanille:
The remote build cache is [working](https://console.cloud.google.com/cloud-build/builds/25e5b628-29c0-4ca9-8c86-db270e2c4118?project=834911136599):
```
Step #3 - "bazel-tests": INFO: Elapsed time: 46.652s, Critical Path: 35.49s
Step #3 - "bazel-tests": INFO: 46 processes: 6 remote cache hit, 35 internal, 3 local, 2 worker.
Step #3 - "bazel-tests": INFO: Build completed successfully, 46 total actions
Step #3 - "bazel-tests": //tfjs-core:browserstack_tfjs-core_test                         (cached) PASSED in 0.2s
```

mattsoulanille:
Sorry this PR is so long. Getting the minimum Bazel build / test / ci setup working for `tfjs-core` took a lot of changes. Future Bazel PRs won't be so huge.

mattsoulanille:
@pyu10055 I'm not really sure who to assign this PR to, so I assigned it to you. Feel free to add reviewers or reassign. Thanks!

pyu10055 (changes requested):
Thank you Matt, this is simply beautiful! I tried locally, and it is super smooth. couple high level questions:
1. currently we are running both the bazel test suite at the root level and individual cloudbuild.yml test, we will eventually eliminate the cloudbuild.yml in individual directories?
2. It would be good to have markdown doc for build files that explain the targets and their purpose, and good to provide a migration document that layout the steps for migrate other packages, so others can help, and for other repos too.
3. Is all tests running in the single step after migrating to bazel? can that take advantage of cloud machine with multiple cores?
4. how different is the bazel build file and blaze build file? can we use copybara to import this files to g3 and expect all the test run properly? (I imaging dependencies could be a problem)

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4856#-:-MXtAfjX0FTUpuzUpExw:b-w2ok6n)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @mattsoulanille)

___
*[.bazelignore, line 18 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MXt3f9T9KuThD0Xifdo:-MXt3f9T9KuThD0Xifdp:b-pw2a03) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/.bazelignore#L18)):*
> ```
> tfjs-react-native/node_modules
> tfjs-tflite/node_modules
> tfjs-vis/node_modules
> ```

extra spaces at the end
___
*[package.json, line 24 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MXt42JY23s_smYWKwf6:-MXt42JY23s_smYWKwf7:b-raoub2) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/package.json#L24)):*
> ```JSON
>     "jasmine-core": "~3.1.0",
>     "js-yaml": "^3.14.0",
>     "karma": "5.0.0",
> ```

can we use the latest karma 6.x, I think there might be security vulnerabilities with earlier versions

jinjingforever (approved):
> Reviewed 10 of 24 files at r1, 8 of 17 files at r2, 13 of 13 files at r3.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4856)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128 and @mattsoulanille)

*[a discussion](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MY5w_L78yvrOzlopd6b:-MY5w_L78yvrOzlopd6c:b29anen) (no related file):*
Thanks for all the hard work for this! This looks super awesome! 

(+1 on the markdown doc Ping mentioned)

---

lina128 (approved):
This is awesome!!! Thank you Matt!!!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4856#-:-MY6zEcs83V0EDctFTZm:b-jd1eoj)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @mattsoulanille)

___
*[karma.conf.js, line 1 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MY6GVqu7gR8bi_hIWIi:-MY6GVqu7gR8bi_hIWIj:b-mi2sa7) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/karma.conf.js#L1)):*
> ```JavaScript
> const browserstackConfig = {
> ```

Add license.
___
*[tfjs-core/BUILD.bazel, line 49 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MY6xYVI-nZoSRFwUQC1:-MY6xYVI-nZoSRFwUQC2:bgn5cnq) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/tfjs-core/BUILD.bazel#L49)):*
> ```Starlark
> 
> ts_library(
>     name = "tfjs-core_test_lib",
> ```

Curious, can we have a build file for ops directory, so that only changes in ops trigger op tests?
___
*[tfjs-core/BUILD.bazel, line 80 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MY6xKDpDvwx_7sR0JTi:-MY6xKDpDvwx_7sR0JTj:bnko4e5) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/tfjs-core/BUILD.bazel#L80)):*
> ```Starlark
>         ":tfjs-core_lib",
>         ":tfjs-core_test_lib",
>         "//tfjs-backend-cpu",
> ```

Now that we can express the dependency here, can we remove cpu as a dev-dependency from package.json?
___
*[tfjs-core/src/kernel_registry_test.ts, line 27 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MY6wuM4BIiTau7XC77T:-MY6wuM4BIiTau7XC77U:b-ph6m2b) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/tfjs-core/src/kernel_registry_test.ts#L27)):*
> ```XML
>   it('register a kernel and call it', () => {
>     // Make sure the backend is loaded. Perhaps tf.getBackend
>     // should call tf.backend to make sure the backend is loaded?
> ```

Is this a comment or question?

mattsoulanille (commented):
Thanks for the review! Here are some answers to Ping's questions:

1. When we fully switch to Bazel, we will (in most cases) stop running specific cloudbuild steps in each package's directory. The cloudbuild step that runs Bazel will run all the tests. Ideally, we'll be able to remove each package's `cloudbuild.yaml` file, but there may be some that run commands that would be difficult to get Bazel to run. 
2. That's a good idea. I'll write that up.
3. Each test target will be run in a different thread, and Bazel will run as many at once as there are CPU cores available. Within a test target, however, Bazel can't provide any parallelization. That means we get great parallelization for things like the WASM CPU tests, which has 15 test targets, but we get no parallelization for tests that use a single `setup_test.ts` entrypoint and use `jasmine_util.ts` to choose what tests to run.

    If we move the logic of `setup_test.ts` and `jasmine_util.ts` into a `.bzl` file, we can programmatically create test targets that Bazel can run in parallel, but that (I think) will use a browserstack instance for each test target we create (there might be a way to avoid this, but I need to do more research). For local development, there may be a way to set up karma so it can reload just a single test without reloading the entire page, but I'll have to do more research on that too.c

4. Right now, they're somewhat different since we're using ts_project instead of ts_library (to make sure build artifacts end up in `dist/` since that's where dependencies expect them to be and where the package.json expects to find them). When we fully switch over to building with Bazel, it will be easier to adapt the rules so they can be copied and used internally (e.g. use ts_library). Running tests may be more difficult since the internal versions of the test rules look slightly different, but they may still work. I can talk more about this over chat or on a Google doc.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4856#-:-MY5lnWU5Pft9hdpK2r6:btgp145)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @pyu10055)

___
*[.bazelignore, line 18 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MXt3f9T9KuThD0Xifdo:-MY7Bi504qvykoQpj8Ti:b-5c8hjz) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/.bazelignore#L18)):*

extra spaces at the end
</blockquote></details>

Should be fixed. I added a newline.
___
*[karma.conf.js, line 1 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MY6GVqu7gR8bi_hIWIi:-MY7BgCq9v0gVXMwqgWP:bhmhk5s) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/karma.conf.js#L1)):*

Add license.
</blockquote></details>

Nice catch! Done.
___
*[package.json, line 24 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MXt42JY23s_smYWKwf6:-MY6Ed_u06bWma2taPPV:b4ecnpb) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/package.json#L24)):*

can we use the latest karma 6.x, I think there might be security vulnerabilities with earlier versions
</blockquote></details>

[This issue](https://github.com/bazelbuild/rules_nodejs/issues/2093) with karma_web_test prevents us from upgrading karma. It looks like it could be an easy fix, so I might try fixing it. 
___
*[tfjs-core/BUILD.bazel, line 49 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MY6xYVI-nZoSRFwUQC1:-MY7Blxw9hOw6TiyzTnv:b-ftdbj3) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/tfjs-core/BUILD.bazel#L49)):*

Curious, can we have a build file for ops directory, so that only changes in ops trigger op tests?
</blockquote></details>

We could do that if we used `ts_library`, but then the outputs would not end up in `dist`, which we sometimes import from (e.g. jasmine_util). After we have fully transitioned to Bazel, we can replace the single `ts_project` target with multiple `ts_library` targets for a more granular build and change the `package.json` files and import statements to accurately reflect where `ts_library` stores its outputs (it stores them where the source files are, but in the workspace's `dist/bin` directory). 

We'd also have to refactor how we launch tests. Right now, we launch them all with a single `setup_test.ts` entrypoint that decides what tests are run. We would have to split that logic up so the ops tests can be run separately from the others, but that should be possible.
___
*[tfjs-core/BUILD.bazel, line 80 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MY6xKDpDvwx_7sR0JTi:-MY7Di2N8bqctJIP3X2Q:b-a9mt8t) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/tfjs-core/BUILD.bazel#L80)):*

Now that we can express the dependency here, can we remove cpu as a dev-dependency from package.json?
</blockquote></details>

Not yet. We're maintaining the original build system's functionality as we write the new Bazel build system, so we can't remove anything it relies on yet.
___
*[tfjs-core/src/kernel_registry_test.ts, line 27 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MY6wuM4BIiTau7XC77T:-MY7ErNp-ugLDosrW8UM:b-hztrxv) ([raw file](https://github.com/tensorflow/tfjs/blob/d827f5e201a7cfecdfa6eee11a0d46f558ea316f/tfjs-core/src/kernel_registry_test.ts#L27)):*

Is this a comment or question?
</blockquote></details>

It's a question. Oops. Should have commented that on the code review instead of in a comment.

I've noticed that sometimes when this test calls [tf.getBackend()](https://github.com/tensorflow/tfjs/blob/023a7a63bab0881f5b558ba7f758794350734085/tfjs-core/src/globals.ts#L312-L314), which returns `ENGINE.backendName`, isn't defined yet since [tf.backend()](https://github.com/tensorflow/tfjs/blob/023a7a63bab0881f5b558ba7f758794350734085/tfjs-core/src/globals.ts#L372-L374), which calls the `ENGINE.backend` getter, has not been called yet to load it.

I'm wondering if `tf.getBackend()` should call `tf.backend()` to make sure it's loaded, or perhaps `ENGINE.backendName` should be a getter like `ENGINE.backend`.

pyu10055 (approved):
thank you for writing the detailed migration guide, one more question:
Seems the deps are declared both in the package.json as well as the bazel build file, will this be the case going forward? can bazel generate the package.json for release?

overall, LGTM and awesome work!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4856#-:-MYRMjxK-_pPmuj4pV4w:b-suwsht)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @pyu10055)

mattsoulanille (commented):
As far as I've found, there's no Bazel rule to generate the package.json. We could write our own, but we can't rely on a fully Bazel-generated package.json anyway since we'll want to still be able to run yarn scripts (a Bazel-generated package.json would end up in //dist/bin/tfjs-core/package.json). This might unfortunately have to be automated outside of Bazel since Bazel rules can't affect the source tree.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4856#-:-MYRWk1_3jl2K_xCaTOr:b-pnnmsg)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @pyu10055)

*[a discussion](https://reviewable.io/reviews/tensorflow/tfjs/4856#-MY5w_L78yvrOzlopd6b:-MYRXF23A5tkZYjrzFKL:boyexb3) (no related file):*

Thanks for all the hard work for this! This looks super awesome! 

(+1 on the markdown doc Ping mentioned)
</blockquote></details>

Done. Thanks!

---

PR #4640, opened 2021-02-05
Enable CPU kernel profiling for browser does not support timer query

pyu10055:
Most mobile browsers do not support timer query, enable CPU based kernel profiling as fallback mechanism.

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4640)

PR #4862, opened 2021-03-25
webgpu: fix profiling error

xhcao:
This patch works around the profiling error when the device
doesn't support timestamp-query extension.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4862)


tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -574,6 +587,12 @@ export class WebGPUBackend extends KernelBackend {
           name: program.constructor.name,
           query: this.getQueryTime(this.querySet)
         });
```
qjia7, line 589: To simplify the code, maybe we can just remove the check of `this.supportTimeQuery` and do it like below:
```
     if (shouldTimeProgram) {
        this.activeTimers.push({
          name: program.constructor.name,
          query: this.getQueryTime(this.querySet)
        });
      }
```
And in getQueryTime, we just keep the original code
```
  async getQueryTime(query: GPUQuerySet): Promise<number> {
    if (this.supportTimeQuery) {
      return this.getTimeFromQuerySet(query);
    } else {
      return 0;
    }
  }
```
In this way, we don't need to add any cpuTimer and just return 0 if the user enables the profile mode with time query unsupported.

How do you think?

tfjs-backend-webgpu/src/index.ts
```diff
@@ -42,6 +42,10 @@ registerBackend('webgpu', async () => {
 
   if (supportTimeQuery) {
     deviceDescriptor = {extensions: ['timestamp-query' as const ]};
+  } else {
+    console.warn(`The webgpu backend only supports to asynchronously read ` +
+        `data from GPU side, but the device doesn't support timestamp-query ` +
+        `extension, you will get incorrect time if you profile operators.`);
```
qjia7, line 48: How about giving a warning like below:
```
This device doesn't support timestamp-query extension. Zero will shown for the kernel time when profiling mode is enabled. Using performance.now is not workable for webgpu since it doesn't support synchronously to read data from GPU.

xhcao: I am not sure what's more appropriate to throw this warning. Does someone else have comments?

qjia7:
LGTM.

@pyu10055 @lina128 @jinjingforever take a look, thanks.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4862)* status: 0 of 1 approvals obtained (waiting on @qjia7 and @xhcao)

___
*[tfjs-backend-webgpu/src/backend_webgpu.ts, line 404 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4862#-MWkKAtX2_qLQbIO29DT:-MWkKAtX2_qLQbIO29DU:bimfrlb) ([raw file](https://github.com/tensorflow/tfjs/blob/a3d7639d87ebbf159570a5480e71795c5cf688e3/tfjs-backend-webgpu/src/backend_webgpu.ts#L404)):*
> ```XML
>     res['getExtraProfileInfo'] = () =>
>         kernelMs.map((d, i) => ({name: flattenedActiveTimerNames[i], ms: d}))
>             .map(d => `${d.name}: ${d.ms}`)
> ```

When supportTimeQuery is false, d is likely null or d.ms is 0. Can you do something like in this PR? https://github.com/tensorflow/tfjs/pull/4640/files

When the extension is not available, it times the gpu execution plus download to cpu time. Not ideal, but still informational.

xhcao:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4862)_ status: 0 of 1 approvals obtained (waiting on @qjia7 and @xhcao)
> 
> _[tfjs-backend-webgpu/src/backend_webgpu.ts, line 404 at r3](https://reviewable.io/reviews/tensorflow/tfjs/4862#-MWkKAtX2_qLQbIO29DT:-MWkKAtX2_qLQbIO29DU:bimfrlb) ([raw file](https://github.com/tensorflow/tfjs/blob/a3d7639d87ebbf159570a5480e71795c5cf688e3/tfjs-backend-webgpu/src/backend_webgpu.ts#L404)):_
> 
> > ```
> >     res['getExtraProfileInfo'] = () =>
> >         kernelMs.map((d, i) => ({name: flattenedActiveTimerNames[i], ms: d}))
> >             .map(d => `${d.name}: ${d.ms}`)
> > ```
> 
> When supportTimeQuery is false, d is likely null or d.ms is 0. Can you do something like in this PR? https://github.com/tensorflow/tfjs/pull/4640/files
> 
> When the extension is not available, it times the gpu execution plus download to cpu time. Not ideal, but still informational.

There is no `dataSync` method on webgpu backend, it only supports download to cpu asynchronously, if we follow webgl backend, it times the GPU execution, download to cpu time, plus system rescheduling time, and the rescheduling time is very large.

xhcao:
For example, if there are N operators in model, and the pairs of [s1, e1], [s2, e2], …, [sN, eN] are the beginning time stamp and the ending time stamp.
On WebGL, https://github.com/tensorflow/tfjs/blob/97cde443aebe04152f873c464a79436efab2bb67/tfjs-core/src/profiler.ts#L53, use `dataSync` to get result, the time axis is defined,  ‘ s1, e1, s2, e2, …, sN, eN’, so we could compute the correct spending time for operators, which are e1-s1, e2-s2, …, sN-eN.
On WebGPU, we could only use the async method `getBufferData`, the time axis is undefined, It may be ‘s1, s2, …, sN, …, e1, e2, …, eN’, but e1-s1, e2-s2, …, eN-sN are not the correct spending time for operators.

lina128 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4862)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @qjia7, and @xhcao)

___
*[tfjs-backend-webgpu/src/index.ts, line 48 at r2](https://reviewable.io/reviews/tensorflow/tfjs/4862#-MWc_Z089FnT5H4KSdDL-r2-48:-MXBFTD56L_faEpUWxW-:b5d2lqb) ([raw file](https://github.com/tensorflow/tfjs/blob/be787471efd434a9ad77c77ab7c4eff7d214b8f0/tfjs-backend-webgpu/src/index.ts#L48)):*
<details><summary><i>Previously, xhcao wrote…</i></summary><blockquote>

I am not sure what's more appropriate to throw this warning. Does someone else have comments?
</blockquote></details>

Sounds good to me.

PR #4888, opened 2021-03-31
webgpu: Batch submitted command encoders

qjia7:
PERF

This patch uses WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE flag instead of
the old WEBGPU_IMMEDIATE_EXECUTION_ENABLED flag. When
WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE is one, it will become to the old
immediate mode.

queue.submit has a fixed cpu overhead. So it's not efficient to call it
too many times. It's also not a good balance if we just submit once for
each frame. Here we use an empirical value 15, which is observed over 5%
performance improvement for mobilenet/resnet50 on different platforms.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4888)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -564,7 +564,8 @@ export class WebGPUBackend extends KernelBackend {
       this.uniformDisposalQueue.push(uniformInfo);
     }
 
-    if (env().get('WEBGPU_IMMEDIATE_EXECUTION_ENABLED')) {
+    if (env().get('WEBGPU_COMMAND_ENCODER_COUNT_IN_QUEUE') as number ===
+        this.commandQueue.length) {
       this.submitQueue();
     }
 
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -564,8 +564,8 @@ export class WebGPUBackend extends KernelBackend {
       this.uniformDisposalQueue.push(uniformInfo);
     }
 
-    if (env().get('WEBGPU_COMMAND_ENCODER_COUNT_IN_QUEUE') as number ===
-        this.commandQueue.length) {
+    if (env().get('WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE') as
+        number <= this.commandQueue.length) {
       this.submitQueue();
     }
 
--- a/tfjs-backend-webgpu/src/backend_webgpu_test.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu_test.ts
@@ -66,40 +66,9 @@ describeWebGPU('backend webgpu cpu forwarding turned on', () => {
 });
 
 describeWebGPU('backend webgpu', () => {
-  it('should not leak memory when command encoder count is larger than one',
-     async () => {
-       const savedFlag = tf.env().get('WEBGPU_COMMAND_ENCODER_COUNT_IN_QUEUE');
-       tf.env().set('WEBGPU_COMMAND_ENCODER_COUNT_IN_QUEUE', 15);
-       const a = tf.tensor2d([2, 4, 6, 8], [2, 2]);
-       const b = tf.tensor2d([0.5, 0.5, 0.5, 0.5], [2, 2]);
-
-       const c = tf.mul(a, b);
-
-       const startNumBytes = tf.memory().numBytes;
-       const startNumTensors = tf.memory().numTensors;
-       const startNumBytesInGPU =
-           (tf.memory() as WebGPUMemoryInfo).numBytesInGPU;
-
-       const f = tf.tensor2d([1, 2, 3, 4, 5, 6], [2, 3]);
-       const d = tf.matMul(c, f);
-
-       const dData = await d.data();
-       const endNumBytes = tf.memory().numBytes;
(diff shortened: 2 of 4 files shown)


tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -564,7 +564,8 @@ export class WebGPUBackend extends KernelBackend {
       this.uniformDisposalQueue.push(uniformInfo);
     }
 
-    if (env().get('WEBGPU_IMMEDIATE_EXECUTION_ENABLED')) {
+    if (env().get('WEBGPU_COMMAND_ENCODER_COUNT_IN_QUEUE') as number ===
```
kainino0x, line 567: Probably should be `>=` just in case multiple command buffers end up waiting for any reason, or the WEBGPU_COMMAND_ENCODER_COUNT_IN_QUEUE is set to 0.

qjia7:
@kainino0x @lina128 @jinjingforever Please take a look. Thanks.

cc @haoyunfeix @axinging @xhcao @gyagp

tfjs-backend-webgpu/src/backend_webgpu.ts
kainino0x, line 567: nit: The fact that these are command encoders and not command buffers is _I think_ an accident of history and not strictly necessarily true in the future. I'd pick a more targeted name like
`WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE`
or something

kainino0x (approved):

kainino0x:
This is an interesting result. Your explanation:

> queue.submit has a fixed cpu overhead.

makes sense to me.

However, and I know we've talked about this several times in the distant past, that overhead shouldn't matter if more work can be accumulated in a single command encoder. Why do we need to submit many small command buffers instead of fewer bigger ones?

qjia7:
> However, and I know we've talked about this several times in the distant past, that overhead shouldn't matter if more work can be accumulated in a single command encoder. Why do we need to submit many small command buffers instead of fewer bigger ones?

I have another patch #4776 doing the thing like you said. The perf impact is similar like this one (but my test is limited). So I think the main issue is queue.submit is called too many times. Since both ways can reduce submit count and the code change is less compared #4776, I choose the current way. But in fact, maybe we need to do it in the two levels in future. One is do more work in a single command encoder. The other is to have several command encoders in a queue. My reference is dx12-dos-and-donts https://developer.nvidia.com/dx12-dos-and-donts `Try to aim at a reasonable number of command lists in the range of 15-30 or below. Try to bundle those CLs into 5-10 ExecuteCommandLists() calls per frame.`

What's your suggestion?

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: Done

qjia7: Done

qjia7:
Please take another look, thanks.

lina128 (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4888#-:-MXBE6Up1zk8NrnEkKig:bb6ywmx)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @jinjingforever)

kainino0x (approved):
> What's your suggestion?

My suggestion is similar to D3D12 but of course our submit calls (and our createCommandEncoder/beginComputePass/endPass/finish calls!) are slower.

AFAIK, there is no benefit to smaller command buffers, and I don't think there's any _need_ for many small command buffers in TF.js; it just requires some work to refactor - instead of configuring the commandbuffers-per-submit we would control the ops-per-submit (or even the min number of milliseconds between submits or something).

qjia7:
> AFAIK, there is no benefit to smaller command buffers, and I don't think there's any _need_ for many small command buffers in TF.js; it just requires some work to refactor - instead of configuring the commandbuffers-per-submit we would control the ops-per-submit (or even the min number of milliseconds between submits or something).

Thanks, Kai. Maybe commandbuffers-per-submit and ops-per-commandbuffer are both needed. Will land this patch first. Then I will add the control for ops-per-commandbuffer.

commandbuffers-per-submit is useful if we are processing more than one frame, then they can be executed parallelly.

kainino0x:
> commandbuffers-per-submit is useful if we are processing more than one frame, then they can be executed parallelly.

Do you mean in parallel on the GPU? I don't _think_ separate command buffers gain any parallelism on the GPU - at least, from my understanding of Vulkan it should be able to parallelize different passes in a single command buffer (barriers permitting) as well. But not 100% sure about this.

qjia7:
> Do you mean in parallel on the GPU? I don't _think_ separate command buffers gain any parallelism on the GPU - at least, from my understanding of Vulkan it should be able to parallelize different passes in a single command buffer (barriers permitting) as well. But not 100% sure about this.

I just checked d3d12 doc. And find below sentences in https://docs.microsoft.com/en-us/windows/win32/direct3d12/executing-and-synchronizing-command-lists#executing-command-lists
`Applications can submit command lists to any command queue from multiple threads. The runtime will perform the work of serializing these requests in the order of submission.`

Does `serializing these requests in the order of submission` mean command buffers won't be executed parallelly? Anyway, we just sent message to Rafael. Hope we can get clear answer.

PR #4895, opened 2021-04-01
Make tfjs-core tests more robust

mattsoulanille:
Some of the tests in tfjs-core `spyOn` properties of imported ECMAScript modules, which is not supported by [the spec](https://262.ecma-international.org/11.0/#sec-module-namespace-exotic-objects-set-p-v-receiver). This PR refactors the tests to avoid this practice, and makes some other changes to tests for increased robustness:

In `globals_test.ts`, we assume `DEPRECATION_WARNINGS_ENABLED` is true when testing if it logs a warning to the console, but that may not be the case if a prior test changed it. We already test that it's set by default in `flags_test.ts`, so this PR explicitly sets it to true for that test.

In a few of the optimizer tests, we expect an exact number of tensors to be in memory. This PR makes those checks relative to the number that were in memory at the start of the test.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4895)


lina128 (approved):
Thank you Matt!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4895#-:-MXcAPnsEP0aLMJqFMzI:b-t3m4wf)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @mattsoulanille)

___
*[tfjs-core/src/globals_test.ts, line 29 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4895#-MXcADT-48UMn5bxVu1d:-MXcADT-48UMn5bxVu1e:b4nc80j) ([raw file](https://github.com/tensorflow/tfjs/blob/f548c8d536f8e754533a57c40e1acf74035895c6/tfjs-core/src/globals_test.ts#L29)):*
> ```XML
>   it('deprecationWarn warns', () => {
>     // flags_test.ts verifies deprecation warnings are on by default.
>     tf.env().set('DEPRECATION_WARNINGS_ENABLED', true);
> ```

Should we set it to its before state at the end of this test?

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4895)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

___
*[tfjs-core/src/globals_test.ts, line 29 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4895#-MXcADT-48UMn5bxVu1d:-MXhuYyTEYM6puz73BLy:b-mwzdv1) ([raw file](https://github.com/tensorflow/tfjs/blob/2da67fc2d59a556d3e9adaa0de0fa3fe751d1db9/tfjs-core/src/globals_test.ts#L29)):*

Should we set it to its before state at the end of this test?
</blockquote></details>

Yep. Thanks! Done.

PR #4897, opened 2021-04-02
webgpu: batch several compute passes into one command encoder

qjia7:
This will make different passes parallel on GPU if possible.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4897)


qjia7:
@kainino0x Please take a look. This one batches several compute passes into one command buffer.

cc @hao @xhcao @xing @gyagp

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -559,7 +567,7 @@ export class WebGPUBackend extends KernelBackend {
     }
 
     if (env().get('WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE') as
```
kainino0x, line 569: side thought: I think this is hit quite often. Would it help to cache this `env().get()` call instead of calling it every time?

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -524,7 +526,14 @@ export class WebGPUBackend extends KernelBackend {
         this.device, bindGroupLayout, inputs.map(t => this.tensorToBinding(t)),
         this.tensorToBinding(output), uniforms);
 
-    const encoder = this.device.createCommandEncoder();
+    let encoder: GPUCommandEncoder;
+    if (this.commandQueue.length > 0) {
```
kainino0x, line 530: nit: Does this mean `commandQueue` is now always 0 or 1 in size? Could it just be `currentCommandEncoder: GPUCommandEncoder | undefined`?

kainino0x (approved):
LGTM (regardless of the comments below)

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: >Does this mean commandQueue is now always 0 or 1 in size?

Yes, I think it is.

>Could it just be currentCommandEncoder: GPUCommandEncoder | undefined?

Sorry, I am not fully understand your comment. Do you want me to change it like below:
```
let currentCommandEncoder: GPUCommandEncoder | undefined = this.commandQueue[0];
if (!currentCommandEncoder)
{
      currentCommandEncoder = this.device.createCommandEncoder();
      this.commandQueue.push(encoder);
}

File: tfjs-backend-webgpu/src/backend_webgpu.ts (hunk 1 above)
Comment (qjia7) [reply tfjs-backend-webgpu/src/backend_webgpu.ts:569]: WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE may be changed by the user. So it seems that there is not a good place to cache it.

File: tfjs-backend-webgpu/src/backend_webgpu.ts (hunk 2 above)
Comment (kainino0x) [reply tfjs-backend-webgpu/src/backend_webgpu.ts:530]: What I mean is, replace the `commandQueue: GPUCommandEncoder[]` with a `currentCommandEncoder: GPUCommandEncoder | undefined`, which is a simpler way to store 0-or-1 of something.

Anyway, not a blocking issue. Sorry for the delay in replying.

File: tfjs-backend-webgpu/src/backend_webgpu.ts (hunk 1 above)
Comment (kainino0x) [reply tfjs-backend-webgpu/src/backend_webgpu.ts:569]: Got it.

File: tfjs-backend-webgpu/src/kernels/FromPixels_utils/from_pixels_webgpu.ts [hunk 3]
```diff
@@ -142,38 +142,6 @@ export class FromPixelsProgram implements WebGPUProgram {
...
-        {
-          binding: 2,
-          resource: {
-            buffer: this.uniform,
-          }
-        }
-      ],
-    });
-
-    const commandEncoder = device.createCommandEncoder({});
```
qjia7, line 168: I have to move `generateEncoder` to `backend_webgpu.ts` due to encoder needed to be created if backend.currentCommandEncoder is null. However, if we pass currentCommandEncoder to this function, it will create a new encoder instead of the original currentCommandEncoder. And it also doesn't make sense to bring backend to this file. So I move it to backend_webgpu.ts.

qjia7 (commented):
Hi Kai, please take another look. I am also ok if you want me to split the last commit into a separate PR.

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: Done

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -285,10 +285,11 @@ export class WebGPUBackend extends KernelBackend {
     const staging = this.acquireBuffer(
         info.bufferInfo.byteSize,
         GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ);
-    const encoder = this.device.createCommandEncoder();
-    encoder.copyBufferToBuffer(
+    if (!this.currentCommandEncoder) {
+      this.currentCommandEncoder = this.device.createCommandEncoder();
+    }
```
kainino0x, line 290: optional nit: These three lines, used 4 times, could be wrapped in a `getOrCreateCommandEncoder()` method.

kainino0x (approved):
LGTM!

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: Done. Use `ensureCommandEncoderReady` so that we don't need to return an encoder.

kainino0x:
👍 Please feel free to merge

PR #4908, opened 2021-04-07
[webgl]Remove cpu backend warning.

lina128:
After modularization, we no longer forward small tensors to cpu backend, instead, we import cpu impls to each backend. This change removes the warning to register cpu backend, and also changes the logic to not checking cpu backend availability. Instead, the logic is only based on cpu_forward flag and tensor states.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4908)

### DIFF
--- a/tfjs-backend-webgl/src/backend_webgl.ts
+++ b/tfjs-backend-webgl/src/backend_webgl.ts
@@ -141,7 +141,6 @@ export class MathBackendWebGL extends KernelBackend {
   private gpgpuCreatedLocally: boolean;
   private numMBBeforeWarning: number;
   private warnedAboutMemory = false;
-  private warnedAboutCPUBackend = false;
 
   constructor(gpgpu?: GPGPUContext) {
     super();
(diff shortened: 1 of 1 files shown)


pyu10055 (approved):
Thank you!

tfjs-backend-webgl/src/backend_webgl.ts
```diff
@@ -623,18 +622,6 @@ export class MathBackendWebGL extends KernelBackend {
     return this.texData.get(dataId);
   }
 
-  private getCPUBackend(): KernelBackend|null {
-    if (!env().getBool('WEBGL_CPU_FORWARD')) {
-      return null;
-    }
-
-    if (this.cpuBackend == null) {
-      this.cpuBackend = engine().findBackend('cpu');
```
qjia7, line 632: Q: When to set `this.cpuBackend` since it is still used in below code? Otherwise, we may meet memory leak issue.
```
  numDataIds() {
    return this.texData.numDataIds() +
        (this.cpuBackend ? this.cpuBackend.numDataIds() : 0) -
        this.pendingDeletes;
  }

lina128: Nice catch, thank you Jiajia. I just removed it in this PR: https://github.com/tensorflow/tfjs/pull/4920

PR #4912, opened 2021-04-08
Add essential types for all supported tflite task libraries

jinjingforever:
Also, the tfweb js client issue I mentioned should be solved. I realized that I shouldn't export types from d.ts file itself. This wouldn't work because d.ts file is basically a reference for the ts compiler and its content won't appear in the final binary. Instead, I extract the root type to a separate file, and export that file instead. 

Then, at the client side (e.g. from another package that uses this package), I can import the root type, and easily construct the js client there. In this way, the code duplication is kept to a minimum. The ground truth stays in this package.

I created another PR (not finalized yet) in the tfjs-models repo for the tasks API: https://github.com/tensorflow/tfjs-models/pull/639. You can see how the tfweb js client type is being used there. Also, please see how the task library types defined here are used in that PR to create the initial tasks API (please feel free to leave tasks API related comments there). Thanks!

(Also, I needed to revert the commonjs plugin version to the older one, because the new version doesn't bundle the tfweb js client code for some reason. Will investigate more after IO)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4912)


pyu10055 (approved):
manual declaration works if the external definition are stable, for example TF graphDef proto. With TFLite task API, it is adding more task gradually, how we can stay up to date with the changes?
The source of truth for us is the generated tf_web wasm and js files, in longer term might be good to find an automated solution that can generate the types with the js files.

Can you also add cloudbuild.yml file for the project, so the PR can be tested with the CI. thanks!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4912#-:-MXm8ali4GRsX0KYAFJt:b-kqv9e4)* status: :shipit: complete! 1 of 1 approvals obtained

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4912#-:-MXm9yddDstynwAL_7Jj:bdxoczg)* status: 0 of 1 approvals obtained

jinjingforever (commented):
Totally agree. It is definitely possible to generate the types file from our specs, just needs more time to work on that. We can discuss more about this next week. My current thinking is that we probably want to work on open-sourcing the toolchain first, then types auto-gen to save some potential duplicated work. We can discuss more about this offline.

I will work on the CI and let you know when I am done. Thanks!   

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4912#-:-MXmBdPLEjlOy7lZw3Vm:b-is8p5q)* status: 0 of 1 approvals obtained

jinjingforever:
I updated the PR to add it to CI. PTAL. Thank you Ping!

pyu10055 (approved):
> Reviewed 13 of 13 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4912)* status: :shipit: complete! 1 of 1 approvals obtained

PR #4916, opened 2021-04-08
Address mutability edge case in tf.Tensor docs

### LINKED ISSUE #4905: After tf.Tensor creation, its value can be changed by input TypedArray (not immutable)
<em>Please make sure that this is a bug. As per our
[GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md),
we only address code/doc bugs, performance issues, feature requests and
build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): macOS 11.2.3
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device: N/A
- TensorFlow.js installed from (npm or script link): <script src="https://cdn.jsdelivr.net/npm/@tensorflow/tfjs@3.2.0"> </script>
- TensorFlow.js version (use command below): 3.2.0
- Browser version: Google Chrome | 89.0.4389.114
- Tensorflow.js Converter Version: N/A

```js
const array = new Float32Array([1, 2, 3, 4]);
const tensor = tf.tensor(array, [2, 2]);
tensor.print();
// Tensor
//    [[1, 2],
//     [3, 4]]
array[0] = 2;
tensor.print();
// Tensor
//    [[2, 2],
//     [3, 4]]
```

As the [doc](https://js.tensorflow.org/api/latest/#class:Tensor) says, `tf.Tensor` should be immutable. So change `array` should not change the value of `tensor`. It is expected that:
```js
const array = new Float32Array([1, 2, 3, 4]);
const tensor = tf.tensor(array, [2, 2]);
tensor.print();
// Tensor
//    [[1, 2],
//     [3, 4]]
array[0] = 2;
tensor.print();
// Tensor
//    [[1, 2],
//     [3, 4]]
```

Provide a reproducible test case that is the bare minimum necessary to generate
the problem. If possible, please share a link to Colab/CodePen/any notebook.

mattsoulanille:
Tensors are meant to be immutable, but for performance reasons, we don't always copy the data used to create them. If that underlying data is changed the tensor will be changed as well. This PR updates the docs to reflect this limitation.

Closes #4905 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4916)

### DIFF
--- a/tfjs-core/src/tensor.ts
+++ b/tfjs-core/src/tensor.ts
@@ -236,6 +236,12 @@ export declare namespace Tensor {}
  * A `tf.Tensor` object represents an immutable, multidimensional array of
  * numbers that has a shape and a data type.
  *
+ * For performance reasons, functions that create tensors do not necessarily
+ * perform a copy of the data passed to them (e.g. if the data is passed as a
+ * `Float32Array`), and changes to that data will change the tensor. This is not
+ * a feature and is not supported. To avoid this behavior, use the tensor before
+ * changing the input data or manually copy it with `tf.add(yourTensor, 0)`.
+ *
  * See `tf.tensor` for details on how to create a `tf.Tensor`.
  *
  * @doc {heading: 'Tensors', subheading: 'Classes'}
--- a/tfjs-core/src/tensor.ts
+++ b/tfjs-core/src/tensor.ts
@@ -240,7 +240,7 @@ export declare namespace Tensor {}
  * perform a copy of the data passed to them (e.g. if the data is passed as a
  * `Float32Array`), and changes to that data will change the tensor. This is not
  * a feature and is not supported. To avoid this behavior, use the tensor before
- * changing the input data or manually copy it with `tf.add(yourTensor, 0)`.
+ * changing the input data or create a copy with `copy = tf.add(yourTensor, 0)`.
  *
  * See `tf.tensor` for details on how to create a `tf.Tensor`.
  *
--- a/tfjs-core/src/tensor.ts
+++ b/tfjs-core/src/tensor.ts
@@ -238,7 +238,7 @@ export declare namespace Tensor {}
  *
  * For performance reasons, functions that create tensors do not necessarily
  * perform a copy of the data passed to them (e.g. if the data is passed as a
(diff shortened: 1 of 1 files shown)


tfjs-core/src/tensor.ts
```diff
@@ -236,6 +236,12 @@ export declare namespace Tensor {}
  * A `tf.Tensor` object represents an immutable, multidimensional array of
  * numbers that has a shape and a data type.
  *
+ * For performance reasons, functions that create tensors do not necessarily
+ * perform a copy of the data passed to them (e.g. if the data is passed as a
+ * `Float32Array`), and changes to the data will change the tensor. This is not
+ * a feature and is not supported. To avoid this behavior, use the tensor before
+ * changing the input data or create a copy with `copy = tf.add(yourTensor, 0)`.
```
pyu10055, line 243: will `tf.clone(x)` be a better suggestion? since it does not involve computation. 
But I am not sure if the result would also suffer the mutability issue.

mattsoulanille: `tf.clone(x)` does not work. I just tested it, and it creates a reference to the original tensor's data instead of performing a copy.

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4916)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @mattsoulanille)

___
*[tfjs-core/src/tensor.ts, line 243 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4916#-MXnaB3Tn4DMf5VDYe_2-r1-243:-MXripUw7cpAOx8ZT3ll:b5oj69n) ([raw file](https://github.com/tensorflow/tfjs/blob/41d6b3052362a78f7f8b33d1babeff6bcedd87f4/tfjs-core/src/tensor.ts#L243)):*

`tf.clone(x)` does not work. I just tested it, and it creates a reference to the original tensor's data instead of performing a copy.
</blockquote></details>

got it, thanks for verifying.

PR #4920, opened 2021-04-11
[webgl]Remove cpu backend.

lina128:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4920)

### DIFF
--- a/tfjs-backend-webgl/src/backend_webgl.ts
+++ b/tfjs-backend-webgl/src/backend_webgl.ts
@@ -128,7 +128,6 @@ export class MathBackendWebGL extends KernelBackend {
   private uploadWaitMs = 0;
   // Accumulated time spent (including blocking in downloading data from webgl.
   private downloadWaitMs = 0;
-  private cpuBackend: KernelBackend;
 
   // record the last manual GL Flush time.
   private lastGlFlushTime = 0;
@@ -167,9 +166,7 @@ export class MathBackendWebGL extends KernelBackend {
   }
 
   numDataIds() {
-    return this.texData.numDataIds() +
-        (this.cpuBackend ? this.cpuBackend.numDataIds() : 0) -
-        this.pendingDeletes;
+    return this.texData.numDataIds() - this.pendingDeletes;
   }
 
   write(values: BackendValues, shape: number[], dtype: DataType): DataId {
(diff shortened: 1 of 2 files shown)


lina128:
cc @qjia7

tfjs-backend-webgl/src/backend_webgl.ts
```diff
@@ -167,9 +166,7 @@ export class MathBackendWebGL extends KernelBackend {
   }
 
   numDataIds() {
-    return this.texData.numDataIds() +
-        (this.cpuBackend ? this.cpuBackend.numDataIds() : 0) -
```
qjia7, line 171: Hi Na, I am wondering if it is possible that `cpuBackend.numDataIds()` is not zero if some operators are using cpu. In this case, will `numDataIds()` be correct for webgl? If it doesn't exist, your change looks good for me. Thanks.

lina128 (commented):
Hi Jiajia, good question. For webgl backend, after modularization, the operators that use cpu will no longer forward the data to cpu backend, instead, it simply uses the vanilla Javascript implementation imported from cpu package, e.g.: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/kernels/Abs.ts#L36

Because data never leaves webgl backend, we don't need to count cpu backend anymore. 

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4920#-:-MY35wHX7lwue3DrjGW5:b9tx1l5)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

qjia7:
> Hi Jiajia, good question. For webgl backend, after modularization, the operators that use cpu will no longer forward the data to cpu backend, instead, it simply uses the vanilla Javascript implementation imported from cpu package, e.g.: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/kernels/Abs.ts#L36
> 
> Because data never leaves webgl backend, we don't need to count cpu backend anymore.
> 

Make sense for me. Thanks for the explaination.

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4920)* status: :shipit: complete! 1 of 1 approvals obtained

PR #4938, opened 2021-04-15
Add missing headings to core operations docs

MaximeKjaer:
These two operations were missing from the [latest API docs](https://js.tensorflow.org/api/latest/) because they were missing an `@doc` annotation

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4938)


MaximeKjaer:
@googlebot I signed it!

pyu10055 (approved):
Thanks!

PR #4950, opened 2021-04-17
Feat/add mish

WenheLI:
This PR adds support for a new activation mish and it's used in yolov4 & v5. A native implementation in tfjs would be great!

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4950)


pyu10055 (approved):
Thanks!

tfjs-layers/src/activations_test.ts
```diff
@@ -333,3 +333,36 @@ describeMathCPUAndGPU('swish activation', () => {
     expectNoLeakedTensors(() => swish(initX), 1);
   });
 });
+
+describeMathCPUAndGPU('mish activation', () => {
+  const mish = new Mish().apply;
+  // Setup: Array with initial values.
+  // Execute: Swish on the last dimension.
```
pyu10055, line 340: s/Swish/Mish

WenheLI:
Seems like we need to also export this class in `exports_layers`. Will open another PR to export it

PR #4942, opened 2021-04-16
[wasm] Add flag WASM_THREAD_POOL_SIZE 

axinging:
This flag indicates how many threads can be used in multi-thread mode, it helps us understand the performance difference between wasm and webgpu/webgl.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4942)

PR #4957, opened 2021-04-19
[wasm] Fix wasm backend not starting any threads

### LINKED ISSUE #4796: Threaded SIMD tests for the wasm backend crash the browser
<em>Please make sure that this is a bug. As per our
[GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md),
we only address code/doc bugs, performance issues, feature requests and
build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): 3.2
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Debian rodete
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device:
- TensorFlow.js installed from (npm or script link):
- TensorFlow.js version (use command below):
- Browser version: Chrome ~88
- Tensorflow.js Converter Version:

Threaded SIMD tests for the wasm backend create a new webworker for each test, crashing the page with hundreds of webworkers.

Each test does not have its own webworker, or webworkers are cleaned up after the test is finished.

Provide a reproducible test case that is the bare minimum necessary to generate
the problem. If possible, please share a link to Colab/CodePen/any notebook.

### LINKED ISSUE #4932: tfjs 3.4.0 regression: wasm causes browser hang
tfjs 3.4.0 was just released and unfortunately wasm module is somewhat broken  

to be precise, `tfjs-backend-wasm.wasm` and `tfjs-backend-wasm-simd.wasm` seem to work (although haven't tested in depth)  
what's definitely broken is `tfjs-backend-wasm-threaded-simd.wasm`

and symptoms are simple - complete browser hang,  
up to a point where browser window is nonresponsive to close requests and requires a hard kill

to reproduce: 
- enable `WebAssembly SIMD support` in <chrome://flags>
- enable `WebAssembly threads support` in <chrome://flags>
- set backend to `wasm`
- run any model inference

this used to work in all recent versions of tfjs

environment: tfjs 3.4.0 on windows 10 with chrome 89 or edge 89

### LINKED ISSUE #4934: Benchmark tool freeze with tfjs@3.4.0
The benchmark tool freeze (no response) with the latest tfjs@3.4.0 release. https://tensorflow.github.io/tfjs/e2e/benchmarks/local-benchmark/index.html

mattsoulanille:
Without the linker option `PTHREAD_POOL_SIZE` set, the wasm backend does not create any webworkers and spins forever waiting for them to be created (crashing the browser). This PR sets `PTHREAD_POOL_SIZE` to `Math.min(4, Math.max(1, (navigator.hardwareConcurrency || 1) / 2))`, meaning it will use half of the logical cores available up to a max of 4 (or only 1 if `navigator.hardwareConcurrency` is not available). See [this comment](https://github.com/emscripten-core/emscripten/pull/10263#issuecomment-582010187) for details on why this works. This option apparently supersedes the value passed to [`pthreadpool_create` in `backend.cc`](https://github.com/tensorflow/tfjs/blob/c9dfebddfa34f531d2f0c363b6ea574a9fe9745d/tfjs-backend-wasm/src/cc/backend.cc#L65-L66), but its current setting should match the value computed in the `.cc` file. Fixes #4932 

This PR also fixes a bug where webworkers created by the wasm backend were not removed when `dispose` is called. Fixes #4796.

Fixes #4934 as well.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4957)


mattsoulanille:
The [thread count computed in `backend.cc`](https://github.com/tensorflow/tfjs/blob/c9dfebddfa34f531d2f0c363b6ea574a9fe9745d/tfjs-backend-wasm/src/cc/backend.cc#L65-L66) could perhaps be removed and replaced with `pthreadpool *threadpool = pthreadpool_create(1);`, but that might make it create N webworkers but only use 1. I haven't tested this yet.

tfjs-backend-wasm/src/backend_wasm.ts
```diff
@@ -19,6 +19,7 @@ import './flags_wasm';
 import {backend_util, BackendTimingInfo, DataStorage, DataType, deprecationWarn, engine, env, KernelBackend, TensorInfo, util} from '@tensorflow/tfjs-core';
 
 import {BackendWasmModule, WasmFactoryConfig} from '../wasm-out/tfjs-backend-wasm';
+import {BackendWasmThreadedSimdModule} from '../wasm-out/tfjs-backend-wasm-threaded-simd';
```
lina128, line 22: Does the Simd module need a separate instance too or not?

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4957)* status: 0 of 1 approvals obtained (waiting on @mattsoulanille)

___
*[tfjs-backend-wasm/src/backend_wasm.ts, line 22 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4957#-MYgMqvujaX4WL0Zsr7T-r1-22:-MYgvPEc4CwaFcICXsKV:b2o3leg) ([raw file](https://github.com/tensorflow/tfjs/blob/f2760123bf8af5a5e082d7b01e35fd1ecd564e3c/tfjs-backend-wasm/src/backend_wasm.ts#L22)):*

The simd module doesn't need a separate interface because its API is the same as the non-simd API. The threaded simd module, however, does need a separate interface because it now has the `PThread.terminateAllThreads` function exposed (via the type declarations).
</blockquote></details>

Got it, 
___
*[tfjs-backend-wasm/src/backend_wasm.ts, line 164 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4957#-MYgw8cJDVEuZWcLTOmP:-MYgw8cJDVEuZWcLTOmQ:b-ctlodp) ([raw file](https://github.com/tensorflow/tfjs/blob/f2760123bf8af5a5e082d7b01e35fd1ecd564e3c/tfjs-backend-wasm/src/backend_wasm.ts#L164)):*
> ```XML
>     this.wasm.tfjs.dispose();
>     if ('PThread' in this.wasm) {
>       this.wasm.PThread.terminateAllThreads();
> ```

Is PThread in both wasm types? Is it better to use a type guard, like:
```
function isWasmThreaded(object: BackendWasmModule|BackendWasmThreadedSimdModule): object is BackendWasmThreadedSimdModule {
    return object != null && 'PThread' in object;
}

if (isWasmThreaded(this.wasm)) {
    this.wasm.PThread.terminateAllThreads();
}

```

tfjs-backend-wasm/src/cc/BUILD
```diff
@@ -37,6 +37,32 @@ cc_binary(
...
+        "-s EXPORTED_FUNCTIONS='[\"_malloc\", \"_free\"]'",
+        "-s EXTRA_EXPORTED_RUNTIME_METHODS='[\"cwrap\"]'",
+        "-s MODULARIZE=1",
+        "-s EXPORT_NAME=WasmBackendModuleThreadedSimd",
+        "-s MALLOC=emmalloc",
+        "-s USE_PTHREADS=1",
+        "-s PROXY_TO_PTHREAD=1",
+        # Many x86-64 processors have 2 threads per core, so we divide by 2.
+        "-s PTHREAD_POOL_SIZE=" +
+        "'Math.min(4, Math.max(1, (navigator.hardwareConcurrency || 1) / 2))'",
```
qjia7, line 58: Is it doable that we register a flag `WASM_PTHREAD_POOL_SIZE` in `flags_wasm.ts` and initialized with `'Math.min(4, Math.max(1, (navigator.hardwareConcurrency || 1) / 2))'`. Then in here and `backend.cc`, we both use the flag value?

mattsoulanille (commented):
I think I've found the root cause of this bug, and I'm pretty sure it's caused by the`[PROXY_TO_PTHREADS` link option](https://emscripten.org/docs/porting/pthreads.html#additional-flags) not working. During my testing, I noticed that emscripten was running on the main thread, and that the browser freeze was due to the `_emscripten_futex_wait` function running a busy wait loop. This matches the behavior described in the second paragraph of [Blocking on the Main Browser Thread](https://emscripten.org/docs/porting/pthreads.html#blocking-on-the-main-browser-thread).

I'll look into why `PROXY_TO_PTHREADS` isn't working as expected.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4957#-:-MYjw_bf-dc0djf6RWoL:bwfb97w)* status: 0 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @qjia7)

___
*[tfjs-backend-wasm/src/backend_wasm.ts, line 164 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4957#-MYgw8cJDVEuZWcLTOmP:-MYjsjqu7BLrWpmA5asD:b-uag6xr) ([raw file](https://github.com/tensorflow/tfjs/blob/f2760123bf8af5a5e082d7b01e35fd1ecd564e3c/tfjs-backend-wasm/src/backend_wasm.ts#L164)):*

Is PThread in both wasm types? Is it better to use a type guard, like:
```
function isWasmThreaded(object: BackendWasmModule|BackendWasmThreadedSimdModule): object is BackendWasmThreadedSimdModule {
    return object != null && 'PThread' in object;
}

if (isWasmThreaded(this.wasm)) {
    this.wasm.PThread.terminateAllThreads();
}

```
</blockquote></details>

PThread is only in the `BackendWasmThreadedSimdModule` type (and it only actually appears in the JS object when the simd threaded build is run). We could use a type guard if that's the recommended style, but tsserver knows that `this.wasm` is a `BackendWasmThreadedSimdModule` after I check `if ('PThread' in this.wasm)`, so I don't think a type guard is necessary.
___
*[tfjs-backend-wasm/src/cc/BUILD, line 58 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4957#-MYgMqvujaX4WL0Zsr7U-r1-58:-MYjtdXJD8IE_9THAxVA:b-ti16na) ([raw file](https://github.com/tensorflow/tfjs/blob/f2760123bf8af5a5e082d7b01e35fd1ecd564e3c/tfjs-backend-wasm/src/cc/BUILD#L58)):*

Is it doable that we register a flag `WASM_PTHREAD_POOL_SIZE` in `flags_wasm.ts` and initialized with `'Math.min(4, Math.max(1, (navigator.hardwareConcurrency || 1) / 2))'`. Then in here and `backend.cc`, we both use the flag value? 
</blockquote></details>

This is a good idea, but I think it will have to be done in a separate PR after this fix is out. Feel free to create a feature request.

lina128 (approved):
Thank you Matt. This is great! LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4957#-:-MYug72OAGxJ3h82xwmb:b-3u9i52)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @qjia7)

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/4957)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @qjia7)

___
*[tfjs-backend-wasm/src/cc/BUILD, line 58 at r1](https://reviewable.io/reviews/tensorflow/tfjs/4957#-MYgMqvujaX4WL0Zsr7U-r1-58:-MYuhch94FBP0V2uVI2l:biradkq) ([raw file](https://github.com/tensorflow/tfjs/blob/f2760123bf8af5a5e082d7b01e35fd1ecd564e3c/tfjs-backend-wasm/src/cc/BUILD#L58)):*

This is a good idea, but I think it will have to be done in a separate PR after this fix is out. Feel free to create a feature request.
</blockquote></details>

Hi Jiajia, the flag you proposed is different than this one, https://github.com/tensorflow/tfjs/pull/4942, right? For both flags, do you expect them to be tunable?

mattsoulanille:
Fixes #4934:

Wasm Threaded Simd

qjia7:
> Hi Jiajia, the flag you proposed is different than this one, #4942, right? For both flags, do you expect them to be tunable?

Yes. #4942 is just a get method. I hope we can control the threads number or pool size. Do you know the relationship between the `PTHREAD_POOL_SIZE` and the thread number in `pthreadpool_create`? @lina128 @mattsoulanille

gyagp:
They should be same. 
From https://emscripten.org/docs/porting/pthreads.html:
"
Optionally, pass the linker flag -s PTHREAD_POOL_SIZE=<integer> to specify a predefined pool of web workers to populate at page preRun time before application main() is called. This is important because if the workers do not already exist then we may need to wait for the next browser event iteration for certain things, see below.

When pthread_create() is called, if we need to create a new Web Worker, then that requires returning the main event loop. That is, you cannot call pthread_create and then keep running code synchronously that expects the worker to start running - it will only run after you return to the event loop. This is a violation of POSIX behavior and will break common code which creates a thread and immediately joins it or otherwise synchronously waits to observe an effect such as a memory write. There are several solutions to this:
Return to the main event loop (for example, use emscripten_set_main_loop, or Asyncify).
Use the linker flag -s PTHREAD_POOL_SIZE=<integer>. Using a pool creates the Web Workers before main is called, so they can just be used when pthread_create is called.
Use the linker flag -s PROXY_TO_PTHREAD, which will run main() on a worker for you. When doing so, pthread_create is proxied to the main browser thread, where it can return to the main event loop as needed.
"

PR #5001, opened 2021-04-28
[webgpu] Reuse GPUBindGroupLayout and GPUPipelineLayout

axinging:
Currently, different models create different size of GPUBindGroupLayouts and GPUPipelineLayouts. For example, bodypix(ResNet50) needs about 36 GPUBindGroupLayouts and GPUPipelineLayouts, mobilenet_v2 needs about 14.
With this change, only 7 GPUBindGroupLayouts and GPUPipelineLayouts is created. 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5001)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -27,13 +27,17 @@ import {BinaryOpType, getBinaryOpString} from './kernels/binary_ops';
 import {FromPixelsProgram} from './kernels/FromPixels_utils/from_pixels_webgpu';
 import * as unary_op from './kernels/unary_op_webgpu';
 import * as webgpu_program from './kernels/webgpu_program';
-import {WebGPUBinary} from './kernels/webgpu_program';
 import * as webgpu_util from './webgpu_util';
 
 export interface WebGPUMemoryInfo extends backend_util.MemoryInfo {
   numBytesInGPU: number;
   numBytesAllocatedInGPU: number;
   unreliable: boolean;
+}
+
+interface WebGPULayout {
+  bindGroupLayout: GPUBindGroupLayout;
+  pipelineLayout: GPUPipelineLayout;
 }
 
 type BufferInfo = {
@@ -80,6 +84,7 @@ export class WebGPUBackend extends KernelBackend {
   currentCommandEncoder: GPUCommandEncoder;
   tensorMap: DataStorage<TensorBufferInfo>;
   fromPixelProgram: FromPixelsProgram;
+  fromPixelLayout: WebGPULayout;
   supportTimeQuery: boolean;
 
   private static nextDataId = 0;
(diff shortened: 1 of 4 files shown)


axinging:
@qjia7 @xhcao @haoyunfeix @gyagp, PTAL

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -623,9 +692,19 @@ export class WebGPUBackend extends KernelBackend {
     const broadcastDimsKey = broadcastDims.map(d => d.join('_')).join(';');
     const key = webgpu_program.makeShaderKey(
         program, bufferShapes, bufferTypes, broadcastDimsKey);
-    const {bindGroupLayout, pipeline} = this.getAndSavePipeline(key, () => {
+    if (program.variableNames.length > 5) {
+      throw new Error(
+          `The length of bindGroupLayouts ` +
+          `${program.variableNames.length + 1}` +
+          `is not supported! Change maxBindGroupLayouts to support it.`);
```
qjia7, line 699: Personally, I don't like this change. It makes it non-extensible (for example, `program.variableNames.length` is larger than 5 or we want to use more than one uniform blocks in future). And the benefit is less (I suppose it only saves the first time to create a shader's bind group layout). Is creating bind group layouts a cpu bottleneck for the warmup time? 

@kainino0x take a look. Thanks.

kainino0x: If you do want to pre-generate pipeline layouts (I'm not sure whether this is necessary), couldn't it just be done lazily instead of having an error here? i.e. have a method like `getCachedOrCreatePipelineLayout`

I would be somewhat surprised if there was much overhead to creating layout objects, though I don't really know. Internally, IIRC, we cache and deduplicate them. Creation of bind groups is probably more expensive (and could also be potentially reused if that's not done already).

axinging: I think getCachedOrCreatePipelineLayout is a good idea,  updated, PTAL.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -120,6 +123,7 @@ export class WebGPUBackend extends KernelBackend {
         count: 2,
       });
     }
+    this.createLayouts();
```
qjia7, line 126: Maybe here we can just call
`[this.fromPixelBindGroupLayout, this.fromPixelPipelineLayout] =
        this.createTextureLayout();`
And `createLayouts()` can be removed. We can do it lazily using `getCachedOrCreateLayout` in runWebGPUProgram.

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: >I would be somewhat surprised if there was much overhead to creating layout objects, though I don't really know.

I think the overhead is not much. But the current logic does reduce the number of creating group layout. From this point, the current change looks good to me.

>Creation of bind groups is probably more expensive (and could also be potentially reused if that's not done already).

Yes, you are right. Creation of bind groups is more expensive. But I am not clear about the method of reusing creation of bind groups. Unlike bind group layout, for each bind group, the buffer will change. That's the hard part to cache the bind group.

kainino0x: > Unlike bind group layout, for each bind group, the buffer will change. That's the hard part to cache the bind group.

I think there would need to be some heuristics to use the same buffers. For example if running the same model many times (e.g. on a video) it would be nice if we were able to use the same buffers and same bind groups, every frame in the same positions in the model. However making that work seems quite nontrivial.

kainino0x (approved):
LGTM with @qjia7's approval

qjia7 (approved):

PR #5075, opened 2021-05-17
Add WebGPU support for e2e benchmarks

haoyunfeix:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5075)


haoyunfeix:
@qjia7 @axinging @xhcao @gyagp PTAL

e2e/benchmarks/local-benchmark/index.html
```diff
@@ -178,6 +179,8 @@ <h2>TensorFlow.js Model Benchmark</h2>
 
         // load model and run inference
         try {
+          // TODO: Remove tf.ready() when resolve issue on WebGPU backend.
```
axinging, line 182: LGTM. nit: Do we have any details about the issue? Otherwise we can just remove this TODO?

e2e/benchmarks/local-benchmark/index.html
```diff
@@ -93,6 +93,7 @@ <h2>TensorFlow.js Model Benchmark</h2>
   <script src="https://unpkg.com/@tensorflow/tfjs-converter@latest/dist/tf-converter.js"></script>
   <script src="https://unpkg.com/@tensorflow/tfjs-backend-wasm@latest/dist/tf-backend-wasm.js"></script>
   <script src="https://unpkg.com/@tensorflow/tfjs-automl@latest/dist/tf-automl.js"></script>
+  <script src="https://unpkg.com/@tensorflow/tfjs-backend-webgpu@latest/dist/tf-backend-webgpu.min.js"></script>
```
qjia7, line 96: Only add it when webgpu backend is supported. You can add it like below in the script body:
```
    if (isBackendSupported('webgpu'))
    {
      var script = document.createElement("script");
      script.setAttribute("src", "https://unpkg.com/@tensorflow/tfjs-backend-webgpu@latest/dist/tf-backend-webgpu.min.js");
      document.getElementsByTagName("body")[0].appendChild(script);
    }

File: e2e/benchmarks/local-benchmark/index.html [hunk 3]
```diff
@@ -655,10 +661,19 @@ <h2>TensorFlow.js Model Benchmark</h2>
       return false;
     }
 
+    function isBackendSupported() {
+      if (state.backend === 'webgpu' && typeof navigator !== 'undefined' && !navigator.gpu){
+        showMsg(`These samples should run in Chrome Canary behind the flag "--enable-unsafe-webgpu".`);
+        throw new Error(`WebGPU is not supported.`);
+      }
+      showMsg(null);
+    }
```
qjia7, line 670: Put it in `index.js` and define it like below:
```
function isBackendSupported(backendName) {
  if (backendName === 'webgpu' && typeof navigator !== 'undefined' &&
      !navigator.gpu) {
    return false;
  } else {
    return true;
  }
}

File: e2e/benchmarks/local-benchmark/index.js [hunk 4]
```diff
@@ -28,7 +28,8 @@ const BACKEND_FLAGS_MAP = {
     'WEBGL_FORCE_F16_TEXTURES', 'WEBGL_RENDER_FLOAT32_CAPABLE',
     'WEBGL_FLUSH_THRESHOLD', 'WEBGL_PACK_DEPTHWISECONV',
     'CHECK_COMPUTATION_FOR_ERRORS'
-  ]
+  ],
+  webgpu: ['CHECK_COMPUTATION_FOR_ERRORS', 'WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE'],
```
qjia7, line 32: 'CHECK_COMPUTATION_FOR_ERRORS' is not needed for webgpu. And only add `webgpu: ['CHECK_COMPUTATION_FOR_ERRORS', 'WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE'` when `isBackendSupported('webgpu')` is true.

e2e/benchmarks/local-benchmark/index.html
qjia7: At the beginning of `testCorrectness` and `runBenchmark`, insert below code so that we can early return:
```
      if(!isBackendSupported(state.backend))
      {
              return;
      }
```

And when switch backend, do it like below:
```
backendsController.onChange(async backend => {
        if (!isBackendSupported(state.backend))
        {
          showMsg(
        `WebGPU is not supported. Please use Chrome Canary browser with flag "--enable-unsafe-webgpu" enabled.`);
          return;
        } else {
           showMsg(null);
       }
```
And in other places, you don't need to call isBackendSupported.

e2e/benchmarks/local-benchmark/index.js
```diff
@@ -42,6 +43,7 @@ const TUNABLE_FLAG_NAME_MAP = {
   WEBGL_FLUSH_THRESHOLD: 'GL flush wait time(ms)',
   WEBGL_PACK_DEPTHWISECONV: 'Packed depthwise Conv2d',
   CHECK_COMPUTATION_FOR_ERRORS: 'Check each op result',
+  WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE: 'deferred submit batch size',
```
qjia7, line 46: Same here.

e2e/benchmarks/local-benchmark/index.html
qjia7: Please put `await tf.ready();` before ` await tf.setBackend` and comment that tf.ready() is needed for webgpu.

e2e/benchmarks/local-benchmark/index.html
axinging: I think simply dynamically add a script tag doesn't gurantee that user can use the tag immediately, because 
chrome needs extra time to make this script ready. So we need some logic done here: https://github.com/tensorflow/tfjs/pull/5094

haoyunfeix:
@qjia7 @axinging @xhcao @gyagp PTAL

Improvement: 
* dynamically load external/internal js files and track in promises (download files in parallel and trigger "onPageLoad" when resources all ready)
*  No need to use tf.ready() before tf.setBackend('webgpu'), maybe it comes from the benefit of using src.onload promise to wait for registryBackend method completed.

e2e/benchmarks/local-benchmark/index.html
```diff
@@ -797,7 +850,20 @@ <h2>TensorFlow.js Model Benchmark</h2>
       updateGUIFromURLState();
     }
 
-    onPageLoad();
+    /**
+     * Check if specific backend is supported in current browser.
+     * @param {string} backendName
+     */
+    function isBackendSupported(backendName) {
+      if (backendName === 'webgpu' && typeof navigator !== 'undefined' &&
```
gyagp, line 858: How about typeof navigator == 'undefined'?

qjia7 (approved):
Looks better. Thanks.

@pyu10055 @lina128 @jinjingforever @mattsoulanille Please take a look, thanks.

And please help to release a new npm package for webgpu. There are some APIs   changed and unsupported any more in latest chrome canary. So if some users want to try the webgpu backend, we need to release a new one based the latest code of tfjs-webgpu.

e2e/benchmarks/local-benchmark/index.html
haoyunfeix: Browser will always has navigator, and it's a read-only object, so remove condition above.

haoyunfeix:
@lina128 @pyu10055 @jinjingforever PTAL Thanks!

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5075)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @gyagp, @haoyunfeix, and @qjia7)

___
*[e2e/benchmarks/local-benchmark/index.html, line 96 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpb-r1-96:-Mb8z4rM6Qz5lEI-yBba:b-j1ipm7) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L96)):*

I think simply dynamically add a script tag doesn't gurantee that user can use the tag immediately, because 
chrome needs extra time to make this script ready. So we need some logic done here: https://github.com/tensorflow/tfjs/pull/5094
</blockquote></details>

I think we can keep the original script tags and simply add a new script tag for webgpu. The backend dropdown list initialization should have logic to only show available backends. If this logic is not there, we can add it.

mattsoulanille:
I'm working on releasing a new npm package for webgpu to unblock this PR, but I'm getting some test failures on Chrome Canary 93.0.4530.0 on a 2015 Macbook Pro with an i7-5557U. Specifically, I'm seeing failures in `unstack`, `stridedSlice`, `slice2d`, and batched `matmul` where the outputs do not match the expected values. I'm also seeing `TypeError: Failed to execute 'writeBuffer' on 'GPUQueue': Overload resolution failed.` for string comparisons in `less`, `lessEqual`, `greater`, and `greaterEqual`. 

Do you see the same errors when you run tests, or is there an issue with my setup? [Here are the full logs](https://gist.github.com/mattsoulanille/54e95fba8fbfbad1b5a3c0fa591acdfb).

qjia7:
> I'm seeing failures in `unstack`, `stridedSlice`, `slice2d`, and batched `matmul` where the outputs do not match the expected values. 

We didn't track the mac status. But on windows, these cases are all pass. I will check if I can reproduce it on mac and feedback to you.

> I'm also seeing `TypeError: Failed to execute 'writeBuffer' on 'GPUQueue': Overload resolution failed.` for string comparisons in `less`, `lessEqual`, `greater`, and `greaterEqual`.

These failures will be fixed by #5149

qjia7:
> > I'm seeing failures in `unstack`, `stridedSlice`, `slice2d`, and batched `matmul` where the outputs do not match the expected values.
> 
> We didn't track the mac status. But on windows, these cases are all pass. I will check if I can reproduce it on mac and feedback to you.
> 

@mattsoulanille  `unstack`, `stridedSlice`, `slice2d` should be fixed by #5164. `matmul` seems like a driver bug. I can reproduce it on an old windows driver 26.20.100.7642. But after upgrading it to 27.20.100.9316. It disappears. Can you try to upgrade your macos to see if it still exists?

mattsoulanille:
I applied both PRs locally and re-ran the tests on the same machine. Now the only failure is the batched matmul one. I can't update the Mac I ran the tests on to fix the driver issue causing the matmul failure, but once #5149 and  #5164 are merged, I'll run the tests on an up-to-date Windows machine and publish WebGPU to npm when they pass.

mattsoulanille:
[Tests passed on Windows](https://gist.github.com/mattsoulanille/bd4d5319fe548c14877ed4a4aaeaaedc). I'll publish a new version of WebGPU to npm.

mattsoulanille:
[Published 0.0.1-alpha.7](https://www.npmjs.com/package/@tensorflow/tfjs-backend-webgpu/v/0.0.1-alpha.7). I'll tag the release shortly.

haoyunfeix (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5075)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @gyagp, @haoyunfeix, and @qjia7)

___
*[e2e/benchmarks/local-benchmark/index.html, line 96 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpb-r1-96:-MbkGE7KDAV84ULRqU5M:bqp17g4) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L96)):*

I think we can keep the original script tags and simply add a new script tag for webgpu. The backend dropdown list initialization should have logic to only show available backends. If this logic is not there, we can add it.
</blockquote></details>

`script tag` equals to `static import`, and static import will directly download js file from CDN and execute it immediately after downloaded. There is no chance to stop the download and execution during html page loading, so we'll get error in console in browsers which does not support webgpu.
My suggestion is to use `dynamic import` instead, use `await import('http://xxx/tfjs-backend-webgpu.min.js')` which can be put into a if {...} block, when browser not support we skip import the library file.
Also I added dropdown initialization logic, PTAL thanks.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5075)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @gyagp, @haoyunfeix, and @qjia7)

___
*[e2e/benchmarks/local-benchmark/index.html, line 96 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpb-r1-96:-MblWPfGE8tI-3PVhn81:b5z1mqr) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L96)):*
<details><summary><i>Previously, haoyunfeix (Hao Yunfei) wrote…</i></summary><blockquote>

`script tag` equals to `static import`, and static import will directly download js file from CDN and execute it immediately after downloaded. There is no chance to stop the download and execution during html page loading, so we'll get error in console in browsers which does not support webgpu.
My suggestion is to use `dynamic import` instead, use `await import('http://xxx/tfjs-backend-webgpu.min.js')` which can be put into a if {...} block, when browser not support we skip import the library file.
Also I added dropdown initialization logic, PTAL thanks.
</blockquote></details>

Thank you for the explanation @haoyunfeix. Is the message in console error or warning? If it's warning, that's fine. If it's error, it's more problematic, basically if a user adds tfjs-backend-webgpu as a dependency and if the browser doesn't support it, it will break their app, is that right? Where in the code does it throw error? Is it possible to handle the error more gracefully?

haoyunfeix (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5075)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @gyagp, @haoyunfeix, and @qjia7)

___
*[e2e/benchmarks/local-benchmark/index.html, line 96 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpb-r1-96:-McFVN1U1oCn7wVqdj8F:bd2y2en) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L96)):*

Thank you for the explanation @haoyunfeix. Is the message in console error or warning? If it's warning, that's fine. If it's error, it's more problematic, basically if a user adds tfjs-backend-webgpu as a dependency and if the browser doesn't support it, it will break their app, is that right? Where in the code does it throw error? Is it possible to handle the error more gracefully?
</blockquote></details>

Thanks for your advice, it returns error when browser doesn't support WebGPU, https://github.com/tensorflow/tfjs/pull/5218 will fix this problem, it added a check before initial WebGPU backend.
By the way, I thought kept 'webgpu' in dropdown list is also an option, changes in this PR could remind and guide user how to try WebGPU if WebGPU is not supported in current browser and won't throw error. Is this OK for you?

lina128 (approved):
LGTM. Please address all the comments before merge.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5075#-:-McFmZP_B8PPvbRWNcuf:bjj8374)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @axinging, @gyagp, @haoyunfeix, and @qjia7)

___
*[e2e/benchmarks/local-benchmark/index.html, line 96 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpb-r1-96:-McFlnPZCF2dziuSNBuD:bnfsj7n) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L96)):*
<details><summary><i>Previously, haoyunfeix (Hao Yunfei) wrote…</i></summary><blockquote>

Thanks for your advice, it returns error when browser doesn't support WebGPU, https://github.com/tensorflow/tfjs/pull/5218 will fix this problem, it added a check before initial WebGPU backend.
By the way, I thought kept 'webgpu' in dropdown list is also an option, changes in this PR could remind and guide user how to try WebGPU if WebGPU is not supported in current browser and won't throw error. Is this OK for you?
</blockquote></details>

SG. Thank you!
___
*[e2e/benchmarks/local-benchmark/index.html, line 182 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpb-r1-182:-McFlzT5EADNUbKUXH1c:bxn0aix) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L182)):*

Please put `await tf.ready();` before ` await tf.setBackend` and comment that tf.ready() is needed for webgpu.
</blockquote></details>

+1, please provide explanation of the special handling or link to an issue that describes it.
___
*[e2e/benchmarks/local-benchmark/index.html, line 667 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-McFmSEY1hvMnKVecZnp:-McFmSEY1hvMnKVecZnq:bxf4xda) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L667)):*
> ```
> throw new Error(`WebGPU is not supported.`);
> ```
We should just show msg without throwing error, right?

haoyunfeix (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5075)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @axinging, @gyagp, @lina128, and @qjia7)

___
*[e2e/benchmarks/local-benchmark/index.html, line 96 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpb-r1-96:-McHgaW728b852q_2vg-:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L96)):*

SG. Thank you!
</blockquote></details>

Done.
___
*[e2e/benchmarks/local-benchmark/index.html, line 182 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpb-r1-182:-McHpv6D10aoQW12w5Wk:br0td2w) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L182)):*

+1, please provide explanation of the special handling or link to an issue that describes it.
</blockquote></details>

Removed this change as it doesn't reproduce in latest code.
___
*[e2e/benchmarks/local-benchmark/index.html, line 667 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-McFmSEY1hvMnKVecZnp:-McHrq2T3J2On0eMqrNt:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L667)):*

> ```
> throw new Error(`WebGPU is not supported.`);
> ```
We should just show msg without throwing error, right?
</blockquote></details>

Done.
___
*[e2e/benchmarks/local-benchmark/index.html, line 670 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpb-r1-670:-McHqGGV6Uz0BqWg8bmv:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.html#L670)):*

At the beginning of `testCorrectness` and `runBenchmark`, insert below code so that we can early return:
```
      if(!isBackendSupported(state.backend))
      {
              return;
      }
```

And when switch backend, do it like below:
```
backendsController.onChange(async backend => {
        if (!isBackendSupported(state.backend))
        {
          showMsg(
        `WebGPU is not supported. Please use Chrome Canary browser with flag "--enable-unsafe-webgpu" enabled.`);
          return;
        } else {
           showMsg(null);
       }
```
And in other places, you don't need to call isBackendSupported.
</blockquote></details>

Done.
___
*[e2e/benchmarks/local-benchmark/index.html, line 858 at r2](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpb-r2-858:-McHrnjh9-TKvBanHZfi:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/fd4fd3fe9652cd65f19b0a9998d7ec9bd945d025/e2e/benchmarks/local-benchmark/index.html#L858)):*
<details><summary><i>Previously, haoyunfeix (Hao Yunfei) wrote…</i></summary><blockquote>

Browser will always has navigator, and it's a read-only object, so remove condition above.
</blockquote></details>

Done.
___
*[e2e/benchmarks/local-benchmark/index.js, line 32 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpc-r1-32:-McHrwMJA9OufMxBfhXa:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.js#L32)):*

'CHECK_COMPUTATION_FOR_ERRORS' is not needed for webgpu. And only add `webgpu: ['CHECK_COMPUTATION_FOR_ERRORS', 'WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE'` when `isBackendSupported('webgpu')` is true.
</blockquote></details>

Done.
___
*[e2e/benchmarks/local-benchmark/index.js, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5075#-M_sCt7j953XXgun7Qpc-r1-46:-McHrtBg7KS8TgjJRZuC:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/e2610d9d1fcb5e3c0dfded5e6772478a84636aaf/e2e/benchmarks/local-benchmark/index.js#L46)):*

Same here.
</blockquote></details>

Done.

PR #5094, opened 2021-05-19
e2e: add option to load custom tfjs

xhcao:

### DIFF
--- a/e2e/benchmarks/local-benchmark/index.html
+++ b/e2e/benchmarks/local-benchmark/index.html
@@ -86,13 +86,6 @@ limitations under the License.
       <tbody></tbody>
     </table>
   </div>
-  <script src="https://unpkg.com/@tensorflow/tfjs-core@latest/dist/tf-core.js"></script>
-  <script src="https://unpkg.com/@tensorflow/tfjs-backend-cpu@latest/dist/tf-backend-cpu.js"></script>
-  <script src="https://unpkg.com/@tensorflow/tfjs-backend-webgl@latest/dist/tf-backend-webgl.js"></script>
-  <script src="https://unpkg.com/@tensorflow/tfjs-layers@latest/dist/tf-layers.js"></script>
-  <script src="https://unpkg.com/@tensorflow/tfjs-converter@latest/dist/tf-converter.js"></script>
-  <script src="https://unpkg.com/@tensorflow/tfjs-backend-wasm@latest/dist/tf-backend-wasm.js"></script>
-  <script src="https://unpkg.com/@tensorflow/tfjs-automl@latest/dist/tf-automl.js"></script>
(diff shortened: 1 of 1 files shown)


e2e/benchmarks/local-benchmark/index.html
```diff
@@ -110,11 +103,70 @@ <h2>TensorFlow.js Model Benchmark</h2>
     let warmupTimes = 1;
     // Default do not run any task.
     let task = '';
+
+    function loadScript(url, callback) {
+      var script = document.createElement("script")
+      script.type = "text/javascript";
+
+      if (script.readyState) { //IE
```
gyagp, line 111: Maybe we can drop the IE support?

e2e/benchmarks/local-benchmark/index.html
```diff
@@ -110,11 +103,70 @@ <h2>TensorFlow.js Model Benchmark</h2>
...
+        script.onload = function() {
+          callback();
+        };
+      }
+
+      script.src = url;
+      document.getElementsByTagName("head")[0].appendChild(script);
+    }
+
+    var unpkgUrlsIndex = 0;
```
gyagp, line 129: We need to support the hybrid mode, that is some urls are from unpkg, while the others are from self-built. Then we can have a parameter like "localDist=core,cpu,webgpu" (customtfjs is not a good name for me) to use the ones we build from source code.

e2e/benchmarks/local-benchmark/index.html
```diff
@@ -110,11 +103,70 @@ <h2>TensorFlow.js Model Benchmark</h2>
...
+    var unpkgUrls = ["https://unpkg.com/@tensorflow/tfjs-core@latest/dist/tf-core.js",
+      "https://unpkg.com/@tensorflow/tfjs-backend-cpu@latest/dist/tf-backend-cpu.js",
```
gyagp, line 131: We can generate the link more dynamically. Don't hard code "https://unpkg.com/@tensorflow", “@latest” and so on again and again.

xhcao:
This patch had changed to load tfjs related JS files synchronously, it may be slower that loading JS file.
Please take a look, thank you. @gyagp @haoyunfeix @qjia7 @axinging 
Another question: Is there wrong that loads tfjs related JS files after tfjs-models related JS files.

e2e/benchmarks/local-benchmark/index.html
```diff
@@ -124,6 +152,28 @@ <h2>TensorFlow.js Model Benchmark</h2>
       if (params.has('task')) {
         task = params.get('task');
       }
+      if (params.has('core')) {
```
gyagp, line 155: These names should be values of a key, rather than different keys.

Issue #4947, opened 2021-04-16
Delete called on a Tensor not referenced with Node.js worker_threads


sudo-carson:
Using `tf.tidy` in a Node.js worker thread sometimes causes `Error: Delete called on a Tensor not referenced` if two instances of the thread are running at the same time.

- OS: Windows 10 20H2
- TensorFlow.js version: v3.3.0
- Node version: v14.16.0

```js
// master.js
const { Worker } = require('worker_threads');

const w1 = new Worker('worker.js');
const w2 = new Worker('worker.js');

// worker.js
const { tidy, scalar } = require('@tensorflow/tfjs-node');

tidy(() => {
    const s = scalar(3.14);
});
```

```
Error: Delete called on a Tensor not referenced (tensor_id: 2392)
    at Object.<anonymous> (<anonymous>)
    at NodeJSKernelBackend.disposeData (node_modules\@tensorflow\tfjs-node\dist\nodejs_kernel_backend.js:264:30)
    at Engine.disposeTensor (node_modules\@tensorflow\tfjs-core\dist\tf-core.node.js:3381:26)
    at Tensor.dispose (node_modules\@tensorflow\tfjs-core\dist\tf-core.node.js:2361:21)
    at Engine.scopedRun (node_modules\@tensorflow\tfjs-core\dist\tf-core.node.js:3017:13)
    at Engine.tidy (node_modules\@tensorflow\tfjs-core\dist\tf-core.node.js:3001:21)
    at Object.tidy (node_modules\@tensorflow\tfjs-core\dist\tf-core.node.js:8744:19)
```

rthadur:
related PR has been merged , @bmcdorman thanks for the contribution.

PR #5108, opened 2021-05-21
Make tfjs-node create a new TFJSBackend for each environment

bmcdorman:
This fixes issue #4947.

`tfjs-node` assumes a single global `TFJSBackend` for all `napi_env`s. This assumption holds for most applications, but not when embedding node in C++ with multiple environments or when using web workers. The user observes typical non-deterministic behavior with random crashes and weird error messages.

To resolve this, a map is created that holds a `TFJSBackend` for each `napi_env`.

Cleanup wasn't handled by the previous global implementation, so it's also not handled by this implementation. This may be undesirable. I also did this quickly to fix an issue I was having with little regard for style, DRYness, or other considerations, so hope it's a useful starting point for a project developer at least.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5108)


pyu10055 (changes requested):
Thank you very much for the contribution. Our team is not specialized in the node.js development, sorry that we missed the clean up logic for the backend instance.  We would love to hear more about your thoughts on the backend clean up, how we should hook up with the node napi_env tear down.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5108#-:-MaGXMJi-o8IWN6zuSbW:b-7tvy8q)* status: 0 of 1 approvals obtained (waiting on @bmcdorman)

___
*[tfjs-node/binding/tfjs_binding.cc, line 79 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5108#-MaGX250DknGpCPXAwgi:-MaGX250DknGpCPXAwgj:b-o7b5k6) ([raw file](https://github.com/tensorflow/tfjs/blob/3b659774792427e0d715e7268bd4ab8c41fc7bd6/tfjs-node/binding/tfjs_binding.cc#L79)):*
<details><summary><i alt="summary">Quoted 7 lines of code…</i></summary>

> ```
>   TFJSBackend *backend = nullptr;
>   {
>     std::lock_guard<std::mutex> lock(gBackendsMut);
>     const auto it = gBackends.find(env);
>     if (it == gBackends.cend()) NAPI_THROW_ERROR(env, "Backend not found for environment");
>     backend = it->second;
>   }
> ```
</details>


I will be good to extract this logic into a separate method to reduce code duplication.

bmcdorman:
It looks like there's actually a better way to go about this with Node's [Environment Lifecycle API](https://nodejs.org/api/n-api.html#n_api_environment_life_cycle_apis). That'll also make cleanup trivial to implement. I'll take another stab at it in the next couple days.

Thanks for the feedback!

bmcdorman:
Okay, cleanup has been added and we're now using the Node Environment Lifecycle API. Removing the lock also sped things up a little for me in my testing (with 3 environments hammering tfjs-node). I think it's good to go.

pyu10055 (approved):
> Environment Lifecycle API
Looks great, thanks!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5108#-:-MaUWiuWE8aBCFCa30Vi:bsy0inx)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @bmcdorman)

PR #5148, opened 2021-05-31
webgpu: Fix the reduce shader key

qjia7:
BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5148)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
@@ -15,7 +15,7 @@
  * =============================================================================
  */
 
-import {backend_util, DataType, util} from '@tensorflow/tfjs-core';
+import {backend_util, DataType} from '@tensorflow/tfjs-core';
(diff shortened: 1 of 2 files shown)


tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
```diff
@@ -78,7 +76,7 @@ export class ReduceProgram implements WebGPUProgram {
     }
 
     const outputSnippet = this.reduceType === 'mean' ?
-        `setOutput(flatOutputIndex, bestValue / ${this.reduceSize});` :
```
qjia7, line 81: This is the place where we forgot to use uniform instead of the constant value.  Or this value should be as part of the shader key. Otherwise, the shader key can correctly reflect different shader variants.

qjia7 (commented):
@lina128 @jinjingforever Please take a look, thanks.

cc @axinging

jinjingforever (approved):

PR #5209, opened 2021-06-11
webgpu: Optimize depthwise conv2d

qjia7:
PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5209)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/DepthwiseConv2dNative.ts
+++ b/tfjs-backend-webgpu/src/kernels/DepthwiseConv2dNative.ts
@@ -18,6 +18,7 @@
 import {backend_util, DepthwiseConv2dNative, DepthwiseConv2dNativeAttrs, DepthwiseConv2dNativeInputs, KernelConfig, KernelFunc} from '@tensorflow/tfjs-core';
 
 import {WebGPUBackend} from '../backend_webgpu';
+import {DepthwiseConv2D3x3Program} from './depthwise_conv2d_3x3_webgpu';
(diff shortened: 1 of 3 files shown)


qjia7:
@kainino0x @lina128 @jinjingforever Please take a look, thanks.

In our last meeting, I ever said that the perf of depthwiseConv2d was almost 2x slower than webgl on Intel GPU. With this change, the perf is close between webgpu and webgl.
hand_detector:



DepthwiseConv2dNative | 13.88
-- | --




becomes



DepthwiseConv2dNative | 8.28
-- | --






This optimization is mainly for filter size 3x3 and stride size is 1x1. It's widely used for all of we tested models.
For such kind of depthwise conv2d, due to the stride size is only 1x1, there will be many data are repeated access in each channel tile. If we only calculate one output data in one invocation, we need to access 3x3 data in `x` and 3x3 data in `filter`. So to get a 4x4 output data, we need (3x3+3x3)* (4x4) = 288 times memory access. However, if we calculate 4x4 data in one invocation, we only need to access (3x6)*4 data in `x` and (3x3)*4 data in `filter`. So it needs (3x6 + 3x3) * 4 = 108 times, which is only half of previous one.

lina128 (approved):
Thank you for the explanation and great perf improvement!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5209#-:-Mbw8T0q-rwA6PX_DKGN:b-teetxf)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever and @kainino0x)

PR #5217, opened 2021-06-15
Update react-native monorepo dependencies to 3.7.0

mattsoulanille:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5217)


pyu10055 (commented):
looks like the tests are failing for react native, seems the values of the output are shifted.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5217#-:-McLfIVH0EZabAKkc_Ve:b-o3psx5)* status: 0 of 1 approvals obtained

mattsoulanille:
It looks like react native webgl fails on calculating image thresholds via [otsu's method](https://en.wikipedia.org/wiki/Otsu%27s_method#:~:text=In%20computer%20vision%20and%20image,two%20classes%2C%20foreground%20and%20background.). Image threshold ops were added about two months ago, but we haven't seen test failures in CI. This surprised me, so I took a look at how we test tfjs-react-native.

tfjs-react-native isn't part of the monorepo, so it doesn't use `link:../` dependencies to link to other TensorflowJS packages. It has a `link` command in its package.json, but its never used by CI. Instead, it builds against whatever versions are specified in its package.json. 

tfjs-react-native runs tests with an example app `integration_rn59`. It has several commands in its package.json to copy monorepo dependencies to `node_modules`, but it only copies the `dist` directory and not the whole package (this is likely a bug that by chance hasn't caused any issues yet). When react-native tests are run in CI, they copy the tfjs-react-native package into the test app but [don't actually copy any other dependencies](https://github.com/tensorflow/tfjs/blob/50a5932c0fc75d3c3888c484ce6e1362fad73152/tfjs-react-native/integration_rn59/scripts/test-ci.sh#L28-L31). This is consistent with `/scripts/package_dependencies.json`, which lists tfjs-react-native as having no dependencies. 

This bug is only now showing up because we have been running tfjs-react-native against 3.1.0 monorepo dependencies.

mattsoulanille:
I visualized the test results and for some reason, the otsu threshold test output matches the expected output for binary. This might be a coincidence.
threshold

mattsoulanille:
Misclicked the "close PR" button.

I disabled the Otsu tests and created a bug to track them.

pyu10055 (approved):
> Reviewed 3 of 5 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5217)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #5237, opened 2021-06-20
Kernel 'Linspace" not registered for backend 'undefined'


asilvas:
**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Windows 10
- TensorFlow.js installed from (npm or script link): 3.7.0
- TensorFlow.js version (use command below): 3.7.0
- Browser version: 3.7.0

**Describe the current behavior**

```
> tf.linspace(0, 9, 10)
Uncaught Error: Kernel 'LinSpace' not registered for backend 'undefined'
```

In browser docs example or locally in node

Might be connected to a recent commit but nothing obvious to me so haven't submitted PR yet: https://github.com/tensorflow/tfjs/commit/293d357e884cb903fc9c8a4ae9c48eb130f05e3b#diff-f085c223aa4f2e310b69b8f405f28107ae81921dbf1468392683f4e83fb9a0d7

**Standalone code to reproduce the issue**

```
tf.linspace(0, 9, 10)
```

rthadur:
@jinjingforever I was able to reproduce the same , please check.

jinjingforever:
Thank you for the report @asilvas! The PR above should fix this issue (it will be in the next release). For now, a workaround is wrapping the code around `tf.ready`:

```js
tf.ready(()=>{
  tf.linspace(0, 9, 10);
  // ...
});
```

PR #5251, opened 2021-06-24
Add the missing backend initialization in runKernel

jinjingforever:
This is related to #5237 

**Background**

Most of our kernels will create/convert tensors before calling `ENGINE.runKernel` ([example](https://github.com/tensorflow/tfjs/blob/fixrunkernel/tfjs-core/src/ops/greater.ts#L45-L46)). In those cases, `convertToTensor` will eventually call `ENGINE.makeTensor` where `this.backend` will be [referenced](https://github.com/tensorflow/tfjs/blob/fixrunkernel/tfjs-core/src/engine.ts#L808). `this.backend` has a side effect that will [initialize the backend](https://github.com/tensorflow/tfjs/blob/fixrunkernel/tfjs-core/src/engine.ts#L216). This design allows our backend to be lazily initialized. The similar lazy backend initialization also happens in `runKernelFunc`. See this [code block](https://github.com/tensorflow/tfjs/blob/fixrunkernel/tfjs-core/src/engine.ts#L607-L614).

**The problem**

Some kernels do **not** create/convert tensors before calling `ENGINE.runKernel`. For example, [linspace](https://github.com/tensorflow/tfjs/blob/fixrunkernel/tfjs-core/src/ops/linspace.ts#L41), [fill](https://github.com/tensorflow/tfjs/blob/fixrunkernel/tfjs-core/src/ops/fill.ts#L42), etc. As a result, when they call `ENGINE.runKernel`, an error will be thrown when calling `getKernel` [here](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/engine.ts#L538), because `this.backendName` is not set since the backend has not been initialized.

**The fix**

I copied the code block about lazily initializing backend from `runKernelFunc` to `runKernel` which should fix this problem (tested locally). This was probably missed during the kernel modularization process when we switched from `runKernelFunc` to `runKernel`.   

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5251)

### DIFF
--- a/tfjs-core/src/engine.ts
+++ b/tfjs-core/src/engine.ts
@@ -535,6 +535,14 @@ export class Engine implements TensorTracker, DataMover {
    */
   runKernel<T extends Tensor|Tensor[]>(
       kernelName: string, inputs: NamedTensorMap, attrs?: NamedAttrMap): T {
+    if (this.backendName == null) {
+      // backend has not been initialized yet (backend initialization is lazy
+      // can be deferred until an op/ kernel is run).
+      // The below getter has side effects that will try to initialize the
+      // backend and set properties like this.backendName
+      // tslint:disable-next-line: no-unused-expression
+      this.backend;
+    }
     const hasKernel = getKernel(kernelName, this.backendName) != null;
     if (!hasKernel) {
       throw new Error(`Kernel '${kernelName}' not registered for backend '${


pyu10055 (approved):
LGTM, thanks

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5251#-:-Md-63oBDMUmAe1X6Bc0:b2pkvq)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

Issue #4844, opened 2021-03-20
When in "webgl" backend and image width is odd, the second run of tf.conv2d() may be wrong.


ColorfulCakeChen:
**System information**

- OS Platform and Distribution: Chrome OS, Android
- Mobile device (the issue happens on mobile device, too): ASUS ZenPhone Max Pro
- TensorFlow.js installed from: script link
- TensorFlow.js version: 3.2.0
- Browser version:  89.0.4389.82 (Stable) (64 bits)

**Describe the current behavior**

- When tf.gtBackend() is "webgl".
  - If the width of the input is even, the result of tf.conv2d() is correct.
  - If the width of the input is odd, the result of first time tf.conv2d() is correct. But the result of  second times run may be wrong.
- When tf.gtBackend() is "wasm" or "cpu".
  - The above problem seems not existed.

**Describe the expected behavior**
- No matter which backend, what kinds image width (even or odd), how many times to run tf.conv2d(), the result should be same and corrrect.

**Standalone code to reproduce the issue**
- https://gist.github.com/ColorfulCakeChen/e3c7e6ce1be9c6c0f5f1a6b209b9dd02
- (The code could be pasted into any example console box of Tensorflow.js API (https://js.tensorflow.org/api/3.2.0/) web page to run.)

rthadur:
I was able to reproduce the same in MAC book pro.
cc @lina128

ColorfulCakeChen:
fd

ColorfulCakeChen:
Thank you all efforts for this. However, re-try the test codes in the gist (https://gist.github.com/ColorfulCakeChen/e3c7e6ce1be9c6c0f5f1a6b209b9dd02) in 3.8.0 tensorflow.js (https://js.tensorflow.org/api/3.8.0/) example console box. The problem seems still exists.

lina128:
Hi @ColorfulCakeChen, the fix has not been released yet. It will be released in 3.9.0.

ColorfulCakeChen:
> Hi @ColorfulCakeChen, the fix has not been released yet. It will be released in 3.9.0.

Oh. Thanks. It's my fault too impatient. XD

PR #5264, opened 2021-06-30
[webgl]Fix conv2d when width is odd.

lina128:
Fix https://github.com/tensorflow/tfjs/issues/4844.

Background:
This bug is not recently introduced, it seems it has been existing since the beginning. The algorithm that special handle packed x with odd number of cols. The algorithm assumes xTexData.shape's height, width and inChannels are same as x.shape. However, they can be different, in which case, the texShape is wrong, so the result is wrong. Imagine the input image x has a shape [3, 3, 1]. When we call conv2D for the first time, it is not packed, so we use the algorithm in the if condition in `conv2dByMatMul`, x is reshaped to [1, 9, 1]. When the texture is uploaded to GPU, it is packed, xTexData.shape is [1, 9, 1] and xTexData.texShape is [10, 2]. When we call conv2D for the second time, x is packed, so we use the algorithm in the else condition. x is shallowly reshaped to [1, 12, 1], and xTexData.shape is shallowly mutated to [1, 10, 1], texShape is still [10, 2], which result in 2 data missing. Imagine a case where x is already packed. Its xTexData.shape will be [3, 3, 1] and xTexData.texShape will be [12, 2]. For the else condition, x is shallowly reshaped to [3, 4, 1], and xTexData.shape is shallowly mutated to [3, 4, 1], texShape is still [12, 2], no data is lost.

Given that the algorithm in the else condition assumes shape match between x and xTexData, the fix is to check this assumption and only use the algorithm when shapes match.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5264)

### DIFF
--- a/tfjs-backend-webgl/src/kernels/Conv2D_impl.ts
+++ b/tfjs-backend-webgl/src/kernels/Conv2D_impl.ts
@@ -23,7 +23,7 @@ import {mapActivationToShaderProgram} from '../kernel_utils/kernel_funcs_utils';
 import {MatMulPackedProgram} from '../mulmat_packed_gpu';
 import * as webgl_util from '../webgl_util';
 
-import {batchMatMulImpl} from './BatchMatMul_impl';
+import {batchMatMulImpl, MATMUL_SHARED_DIM_THRESHOLD} from './BatchMatMul_impl';
 import {identity} from './Identity';
 import {reshape} from './Reshape';
 
@@ -55,6 +55,9 @@ export function conv2dByMatMul({
   // result from 2D to 4D.
   const xShape = x.shape;
   const xTexData = backend.texData.get(x.dataId);
+  const sharedMatMulDim = convInfo.inChannels;
+  const outerShapeX = xShape[0] * xShape[1] * xShape[2];
+  const outerShapeFilter = convInfo.outChannels;
   const isChannelsLast = convInfo.dataFormat === 'channelsLast';
   const transposeA = false;
   const transposeB = false;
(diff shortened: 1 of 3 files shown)


pyu10055 (commented):
The changes look good, but I have a high level question, what is the relationship relationship between odd cols failure vs shape difference of x and xTexData?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5264#-:-MdXSKgDCILU84BunjEw:br3svd6)* status: 0 of 1 approvals obtained (waiting on @jinjingforever)

lina128 (commented):
odd cols failure happens in the else condition in the conv2dByMatMul method. The algorithm manually mutates x and xTexData's shape without actually doing reshape. The mutation assumes and only works when when x and xTexData's shapes are the same, otherwise, the xTexData.shape, xTexData.texShape, and x.shape are all messed up, see my explanation in the PR description. This will require an actual reshape to fix it. WebGL's reshape for packed data runs a packedReshape program. Now, why texShape and shape mismatch will result in data missing, I don't know exactly in which step it happens, it's hard to inspect GPU textures step by step, I'm guessing data is lost when calling `acquireTexture(texShape, ...)`, which calls `createPackedMatrixTexture` with calculated width and height from texShape.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5264#-:-MdY50af7yxTaIG0ZhL3:bbfpmpl)* status: 0 of 1 approvals obtained (waiting on @jinjingforever)

pyu10055 (commented):
Thank you for the in-depth analysis, LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5264#-:-Mdlr2QU0zmsouFhYZtd:b-87yiop)* status: 0 of 1 approvals obtained (waiting on @jinjingforever)

pyu10055 (approved):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5264)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever)

pyu10055 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5264#-:-MdlrFpI8nJE6CcNwe5I:b-jbsf8z)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever)

PR #5272, opened 2021-07-01
[E2E] Wait for backend initialization complete before benchmark testing

haoyunfeix:
Factory function for WebGPU backend is an async function https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/src/index.ts#L29.
So before using any other functions like tf.tensor() we should add `await tf.ready()` to make sure WebGPU backend entirely initialized.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5272)

### DIFF
--- a/e2e/benchmarks/local-benchmark/index.html
+++ b/e2e/benchmarks/local-benchmark/index.html
@@ -168,6 +168,7 @@ limitations under the License.
         });
       },
       testCorrectness: async () => {
+        await tf.ready();
         if (!tf.engine().backendNames().includes(state.backend)) {
           return;
         }
@@ -442,7 +443,6 @@ limitations under the License.
     }
 
     async function loadModelAndRecordTime() {
-      await tf.ready();
       updateStateFromURLState();
       const benchmark = benchmarks[state.benchmark];
       state.modelType = benchmark['type'];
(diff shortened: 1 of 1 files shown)


haoyunfeix:
@qjia7 @axinging @xhcao @gyagp, PTAL

qjia7 (approved):
@pyu10055 @lina128 @jinjingforever Please take a look, thanks.

It's easy to trigger the error that 'Backend 'webgpu' has not yet been initialized.' without this change.  When we start to run the benchmarks from url parameters, for example 'https://xxx/e2e/benchmarks/local-benchmark/?task=correctness&warmup=50&run=50&benchmark=AutoML%20Image&backend=webgpu'. In this case, `backendsController.onChange` is an async function, which won't block the main thread. So `correctnessButton.domElement.click()` will be triggered to call loading model. However, `await tf.setBackend` in `backendsController.onChange` may be still in progress. Then we will get the error that `Uncaught (in promise) Error: Backend 'webgpu' has not yet been initialized. Make sure to await tf.ready() or await tf.setBackend() before calling other methods
    at Engine.get [as backend] (engine.ts:210)
`

jinjingforever (approved):

pyu10055 (approved):

haoyunfeix:
Rebased. 
@jinjingforever @pyu10055 Please help merge this PR, Thanks.

PR #5286, opened 2021-07-01
Unpack input in TopK WebGL kernel

ahmedsabie:
This prevents each shader from having to unpack the input again due to lazy unpacking

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5286)


lina128 (commented):
I think the current mechanism can handle this case:

The GPGPUProgram defaults packedInputs and packedOutput to false: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/gpgpu_math.ts#L31-L33
When an input is packed, and the program expects an unpacked input, it will do the unpacking here: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/backend_webgl.ts#L823

Maybe I missed anything?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5286#-:-MdYxR9m58RbarjcAmVr:bphd4yd)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

ahmedsabie:
The issue isn't that its unpacking, it's that it unpacks for every shader so it's repeating computation

lina128 (commented):
Do you mean, because TopK call the shaders repeatedly in a loop, and so it is doing unpacking and packing in each iteration?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5286#-:-MdZ7I2s7rorucMqxBcP:b-tmab5n)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

ahmedsabie:
Yes, i think in other kernels shaders aren't usually repeated and if they are it will be on new input not the original input, so they don't have that issue

lina128 (approved):
Make sense. In that case, should we do the reshape after the unpack, because reshape of unpacked tensor is not expensive. Otherwise, LGTM. Also curious how much perf improvement with this change?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5286#-:-MdZAC9k03WbtpbrOtQE:ba9gbjc)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5286)* status: :shipit: complete! 2 of 1 approvals obtained

ahmedsabie:
> Make sense. In that case, should we do the reshape after the unpack, because reshape of unpacked tensor is not expensive. Otherwise, LGTM. Also curious how much perf improvement with this change?
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5286#-:-MdZAC9k03WbtpbrOtQE:ba9gbjc)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

Great catch, will swap the reshape. The change was about a ~20% improvement on packed inputs (0 for unpacked since there's no issues with packing there)

Issue #5313, opened 2021-07-12
Outdated Documentation


djbreen7:
The documentation appears to be outdated.

The main readme states:
> ⚠️ We recently released TensorFlow.js 2.0. If you have been using TensorFlow.js via a script tag without specifying a version and see an error saying no backends are found, then you should read our release notes for instructions on how to upgrade.

npmjs shows that v2.0.0 was one year ago. The current "latest" version is 3.7.0, which was published last month.
- https://www.npmjs.com/package/@tensorflow/tfjs/v/3.7.0
- https://www.npmjs.com/package/@tensorflow/tfjs/v/2.0.0

The concern is that developers are likely using the README for v2 code while using v3.

Additionally, [WINDOWS_TROUBLESHOOTING](https://github.com/tensorflow/tfjs/blob/master/tfjs-node/WINDOWS_TROUBLESHOOTING.md) is outdated.

> Currently, node-gyp requires Python 2.x to work properly. If Python 3.x is installed, you will see build failures. Also double check your python version python --version and update the Windows $PATH as needed.

That statement is inaccurate. See https://github.com/nodejs/node-gyp/issues/1977#issuecomment-567131298
> You do not need Python 2.7 in order to run node-gyp. **_Python 3 has been supported since version 5.0.5_**. If you have verified that your issue is python related, as in that node-gyp is trying to run a script which contains errors according to the python 3 interpreter, then try the following:
...

rthadur:
@djbreen7 thanks for checking this , I have submitted 2 separate PRs to remove above lines

PR #5326, opened 2021-07-13
Remove warning message

rthadur:
Cleaning up README as users are complaining , please check here https://github.com/tensorflow/tfjs/issues/5313#event-5008116652

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5326)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,11 +3,6 @@
 TensorFlow.js is an open-source hardware-accelerated JavaScript library for
 training and deploying machine learning models.
 
-> :warning: We recently released **TensorFlow.js 2.0**. If you have been using TensorFlow.js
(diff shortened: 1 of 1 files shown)


pyu10055 (approved):

mattsoulanille (approved):
I agree that we should remove this.

rules_nodejs has a [changelog document](https://github.com/bazelbuild/rules_nodejs/blob/stable/CHANGELOG.md) that details the changes between versions and links to [migration instructions](https://github.com/bazelbuild/rules_nodejs/wiki#migrating-to-30). Not sure if this would add value in our case since users can look through the releases, but it may be worth considering.

PR #5344, opened 2021-07-19
benchmarks: add WEBGL_USE_SHAPES_UNIFORMS for auto test

qjia7:
With this change, it's convenient to test WEBGL_USE_SHAPES_UNIFORMS with
url parameter like https://XXX/local-benchmark/?backend=webgl&WEBGL_USE_SHAPES_UNIFORMS=true.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5344)


qjia7:
@pyu10055 @lina128 @jinjingforever Please take a look, thanks.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5344)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @pyu10055, and @qjia7)

___
*[e2e/benchmarks/local-benchmark/index.html, line 819 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5344#-MezaKnbCIvp9onzyOCh:-MezaKnbCIvp9onzyOCi:b-3dw7qg) ([raw file](https://github.com/tensorflow/tfjs/blob/b88b2f7e1dc3c894b07e1f5bfa20f760942910bb/e2e/benchmarks/local-benchmark/index.html#L819)):*
> ```
> updateGUIFlagsValueFromURLState();
> ```
Why do you need to call this function onChange? Isn't url param only read at initial page load?
___
*[e2e/benchmarks/local-benchmark/index.js, line 204 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5344#-Meza0DN2hZtPoB0tRwC:-Meza0DP5nlGXmxiu4pT:b-9lwkmp) ([raw file](https://github.com/tensorflow/tfjs/blob/b88b2f7e1dc3c894b07e1f5bfa20f760942910bb/e2e/benchmarks/local-benchmark/index.js#L204)):*
> ```
> || flag === 'WEBGL_USE_SHAPES_UNIFORMS'
> ```
Why does it need special handling?

qjia7 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5344)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @pyu10055)

___
*[e2e/benchmarks/local-benchmark/index.html, line 819 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5344#-MezaKnbCIvp9onzyOCh:-Mf0EehVEllkj1TkJf4U:b-ocozqw) ([raw file](https://github.com/tensorflow/tfjs/blob/b88b2f7e1dc3c894b07e1f5bfa20f760942910bb/e2e/benchmarks/local-benchmark/index.html#L819)):*

> ```
> updateGUIFlagsValueFromURLState();
> ```
Why do you need to call this function onChange? Isn't url param only read at initial page load?
</blockquote></details>

The tunableFlagsControllers are changeable based on `state.backend`. `showFlagSettingsXXX` only return the current backend's tunable flags. So at initial page load, it will get the `wasm`'s tunableFlagsControllers, which don't include  webgl's `WEBGL_USE_SHAPES_UNIFORMS`. Another way is we get all backends's tunable flags controllers at initial load. Then we don't need to depend on the backend onChange.
___
*[e2e/benchmarks/local-benchmark/index.js, line 204 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5344#-Meza0DN2hZtPoB0tRwC:-Mf0GbJ09WWL0094yAyI:b-n8ef6d) ([raw file](https://github.com/tensorflow/tfjs/blob/b88b2f7e1dc3c894b07e1f5bfa20f760942910bb/e2e/benchmarks/local-benchmark/index.js#L204)):*

> ```
> || flag === 'WEBGL_USE_SHAPES_UNIFORMS'
> ```
Why does it need special handling?
</blockquote></details>

This flag's default value is false. In the original check, it will be treated as a non-tunable flag.  Currently, all bool flags, which have the `false` default value, will be treated as the non-tunable flags. So I have to do the special handling. Maybe we should remove this limitation. And only do the non-tunable process for wasm multi-thread and SIMD.

qjia7 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5344)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @pyu10055, and @qjia7)

___
*[e2e/benchmarks/local-benchmark/index.html, line 819 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5344#-MezaKnbCIvp9onzyOCh:-Mf5a4A6EgGZkeCyEGB_:b-i3qpnq) ([raw file](https://github.com/tensorflow/tfjs/blob/b88b2f7e1dc3c894b07e1f5bfa20f760942910bb/e2e/benchmarks/local-benchmark/index.html#L819)):*

If at initial page, load, wasm is the default backend, then we don't need the webgl's flags. And then after initial load, any flag change will set isFlagChanged to true, which will invoke setEnvFlags. This flow works well for other flags, why do we need to change it? Does the new flag WEBGL_USE_UNIFORM_SHAPE different in any way from other webgl flags?
</blockquote></details>

What we do here is for recognizing url parameter and doing the corresponding change. We should make sure that after initial load, the url parameter 'WEBGL_USE_UNIFORM_SHAPE' is also set. But without the change here, how to trigger webgl's flag change?

qjia7:
@lina128 @pyu10055 Change it as we talked. Please take another look, thanks.

lina128 (approved):
Thank you so much Jiajia, this is much cleaner! Great work! LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5344#-:-Mf8vH6e6hkHv1LUk1Hk:bcrn9v0)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @pyu10055, and @qjia7)

PR #5379, opened 2021-07-26
Fix `yarn build` command under tfjs-backend-webgl folder

haoyunfeix:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5379)


haoyunfeix:
I submit this PR because when I execute `yarn && yarn build` in tfjs-backend-webgl folder, "bazel is not found" error occurs.

But I tried this on another Windows machine, `yarn` in other folder (e.g. tfjs-core) will install bazel.exe to `C:\tools` and add to system PATH, so command `bazel build` could execute correctly. 

I don't know why there is difference between two machines but that is not a blocker for me for present.

qjia7:
I met the similar error `bazel is not found`. So I manfully download a bazel.exe and add it into the `PATH` env. Then I can proceed. 
@haoyunfeix if the PR is ready for review, please change the state from `Draft` to `Active`.

haoyunfeix:
@ahmedsabie PTAL. Do you think this is an issue?

mattsoulanille (approved):
LGTM. Thanks for catching this!

PR #5384, opened 2021-07-27
[e2e] Fix object compare

axinging:
The predict results of most models are objects, instead of arrays. Current implementation assumes that the predict results are arrays, which results in some fake pass. For example, when tint changed the shared memory limits, most models should fail the conformance test. However, these cases are passed when we compare these results as arrays.

BUG: https://github.com/tensorflow/tfjs/issues/5397

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5384)

### DIFF
--- a/e2e/benchmarks/local-benchmark/util.js
+++ b/e2e/benchmarks/local-benchmark/util.js
@@ -66,3 +66,102 @@ function queryTimerIsEnabled() {
   return _tfengine.ENV.getNumber(
              'WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_VERSION') > 0;
 }
+
+function areClose(a, e, epsilon) {
+  if (!isFinite(a) && !isFinite(e)) {
+    return true;
+  }
+  if (isNaN(a) || isNaN(e) || Math.abs(a - e) > epsilon) {
+    return false;
+  }
+  return true;
+}
+
+function expectObjectsPredicate(actual, expected, epsilon, predicate) {
+  let actualKeys = Object.getOwnPropertyNames(actual);
+  let expectedKeys = Object.getOwnPropertyNames(expected);
+  if (actualKeys.length != expectedKeys.length) {
+    throw new Error(`Actual length ${
+        actualKeys.length} not equal Expected length ${expectedKeys.length}`);
+  }
+  for (let i = 0; i < actualKeys.length; i++) {
+    let key = actualKeys[i];
+    let isObject = typeof (actual[key]) === 'object' &&
+        typeof (expected[key]) === 'object';
+    let isArray = tf.util.isTypedArray(actual[key]) &&
+        tf.util.isTypedArray(expected[key]);
+    if (isArray) {
+      expectArraysClose(actual[key], expected[key], epsilon, key);
+    } else if (!isObject && (!predicate(actual[key], expected[key]))) {
+      throw new Error(`Objects differ: actual[${key}] = ${
+          JSON.stringify(actual[key])}, expected[${key}] = ${
+          JSON.stringify(expected[key])}!`);
+    } else if (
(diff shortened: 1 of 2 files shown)


axinging:
@qjia7 @haoyunfeix @xhcao @gyagp, PTAL

e2e/benchmarks/local-benchmark/util.js
```diff
@@ -66,3 +66,102 @@ function queryTimerIsEnabled() {
...
+    }
+  }
+}
+
+function expectArraysClose(actual, expected, epsilon, key) {
+  if (epsilon == null) {
+    epsilon = tf.test_util.testEpsilon();
+  }
+
+  if (key == 'data') {
```
qjia7, line 156: Can we make sure that if key is `data`, the value must be `0|1`?

qjia7 (commented):
Good finding, Xing.
Would like @pyu10055 @lina128 @jinjingforever take a look, thanks.

e2e/benchmarks/local-benchmark/util.js
axinging: This path is specially for bodypix, which has a special member data in its predict results:
`export type SemanticPersonSegmentation = {
  data: Uint8Array,
  width: number,
  height: number,
  allPoses: Pose[],
};`
The current value for this data member is 0|1. But I am not sure if the same rule applies for the other models.

lina128 (commented):
Thank you, Xing, this is great!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5384#-:-MfnCd9ZALxANdxOD_QH:b-46vdxe)* status: 0 of 1 approvals obtained (waiting on @axinging, @jinjingforever, @pyu10055, and @qjia7)

___
*[e2e/benchmarks/local-benchmark/util.js, line 102 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5384#-MfnAmzx2_bEcSWthKAD:-MfnAmzx2_bEcSWthKAE:b-bxfquy) ([raw file](https://github.com/tensorflow/tfjs/blob/1fc1899b53c246c8fe50d12e5c830074db1fdaac/e2e/benchmarks/local-benchmark/util.js#L102)):*
<details><summary><i alt="summary">Quoted 14 lines of code…</i></summary>

> ```
>     if (isArray) {
>       expectArraysClose(actual[key], expected[key], epsilon, key);
>     } else if (!isObject && (!predicate(actual[key], expected[key]))) {
>       throw new Error(`Objects differ: actual[${key}] = ${
>           JSON.stringify(actual[key])}, expected[${key}] = ${
>           JSON.stringify(expected[key])}!`);
>     } else if (
>         isObject &&
>         !expectObjectsPredicate(
>             actual[key], expected[key], epsilon, predicate)) {
>       throw new Error(`Objects differ: actual[${key}] = ${
>           JSON.stringify(actual[key])}, expected[${key}] = ${
>           JSON.stringify(expected[key])}!`);
>     }
> ```
</details>

Can we refactor the conditions to be more readable?
```
if (isArray) {
  expectArraysClose(...);
} else if(isObject) {
  expectObjectsPredicate(...); // if you want to log extra context for error, surround it with a try ... catch ...
} else {
  predicate(...);
}
```

axinging:
@lina128 , Updated, PTAL

lina128 (approved):
Thank you so much, great stuff! LGTM.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5384#-:-MfpmYMA4xgYG9J9iqmF:b-s8x2m5)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @jinjingforever, @pyu10055, and @qjia7)

PR #5451, opened 2021-08-10
Make Bazel npm outputs match current packages

### LINKED ISSUE #5382: Webpack fails when using Bazel's `.mjs` output files
Webpack fails to bundle when run on Bazel-created `.mjs` outputs.

Webpack successfully bundles when given Bazel outputs.

In the [webpack_test](https://github.com/mattsoulanille/tfjs/tree/webpack_test/e2e/webpack_test) branch, I've created a simple webpack bundle that depends on the Bazel-compiled outputs of `tfjs-core` and `tfjs-backend-cpu`. To reproduce the bug, run the following in that directory:
```sh
yarn
yarn build-deps
yarn build
```

```
$ yarn build
yarn run v1.22.10
$ webpack
assets by status 673 bytes [cached] 1 asset
orphan modules 4.55 KiB [orphan] 2 modules
runtime modules 274 bytes 1 module
./app.js + 2 modules 5.37 KiB [built] [code generated]

WARNING in ./app.js 22:10-21
export 'tensor1d' (imported as 'tf') was not found in '@tensorflow/tfjs-core' (module has no exports)

ERROR in ../../link-package/node_modules/@tensorflow/tfjs-backend-cpu/dist/index.mjs 18:0-23
Module not found: Error: Can't resolve './base' in '/home/msoulanille/tfjs/link-package/node_modules/@tensorflow/tfjs-backend-cpu/dist'
Did you mean 'base.js'?
BREAKING CHANGE: The request './base' failed to resolve only because it was resolved as fully specified
(probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ./app.js 19:0-38

...
```

We are likely misusing the `.mjs` extension by not including the file extension in relative imports.

A possible workaround is to add the following to the webpack config when bundling tfjs, but we still need to fix our outputs to work without this: https://github.com/graphql/graphql-js/issues/2721#issuecomment-723008284
```js
module: {
  rules: [{
    test: /\.m?js/,
    resolve: {
      fullySpecified: false
    }
  }],
}
```

mattsoulanille:
Make npm packages generated by Bazel match what we currently publish to npm.
```
tfjs-core/
    package.json
    dist/
        index.js # <-- ESModule format
        index.d.ts
        tf-core.node.js # <-- CommonJS node bundle
        ...
```

Also, add a webpack regression test to e2e that makes sure webpack can consume the npm packages we publish.

Additionally, it fixes a bug in version tests that was caused by setting `package_name` of a js_library and `module_name` of a ts_library to the same value. [See this for more details.](https://github.com/angular/angular/blob/4e73b8820acf7368ba874b09567547627573595f/tools/defaults.bzl#L206-L218)

Fixes #5382

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5451)


e2e/webpack_test/app.js
```diff
@@ -1,6 +1,6 @@
 /**
```
mattsoulanille, line 1: This file was not renamed. e2e/webpack_test/app.js is a new file, and version_test.ts was deleted since it's now generated by Bazel.

lina128 (approved):
Thank you for the fix, Matt! It looks much cleaner!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5451#-:-Mgm6my97kkqvF1MvXzC:bk91e7r)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @mattsoulanille)

___
*[tfjs-core/src/BUILD.bazel, line 87 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5451#-Mgm6PuS-HauVrA5syI8:-Mgm6PuS-HauVrA5syI9:b8883t4) ([raw file](https://github.com/tensorflow/tfjs/blob/a944ba907f9c86a6e66ef42dad43b1fcad8dabd8/tfjs-core/src/BUILD.bazel#L87)):*
> ```
> :version_test
> ```
Is it possible to only run version_test in release flow? Similarly, we may also want to run bundle size tests in release or nightly flow. How can we support this use case? Just curious whether it's possible, the actual solution can be in a separate PR.
___
*[tools/make_version_test_file.ts, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5451#-Mgm5i1q-epeePgoKJ24:-Mgm5i1q-epeePgoKJ25:bcfy19x) ([raw file](https://github.com/tensorflow/tfjs/blob/a944ba907f9c86a6e66ef42dad43b1fcad8dabd8/tools/make_version_test_file.ts#L69)):*
> ```
> core
> ```
Consider parameterize this part too to accomodate other version tests, such as:
converter: https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/src/version_test.ts
wasm: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/version_test.ts

mattsoulanille (commented):
Thanks for the review!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5451#-:-Mgq4i3SByUxTNP4lIRF:b-9qdoz3)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

___
*[tfjs-core/src/BUILD.bazel, line 87 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5451#-Mgm6PuS-HauVrA5syI8:-Mgq1rPt-pshW62rtrBs:b-xc33rn) ([raw file](https://github.com/tensorflow/tfjs/blob/a944ba907f9c86a6e66ef42dad43b1fcad8dabd8/tfjs-core/src/BUILD.bazel#L87)):*

> ```
> :version_test
> ```
Is it possible to only run version_test in release flow? Similarly, we may also want to run bundle size tests in release or nightly flow. How can we support this use case? Just curious whether it's possible, the actual solution can be in a separate PR.
</blockquote></details>

That's an interesting question. For the version test, it's more difficult to test only during release because it is added to the `tests.ts` file due to being a dependency of the rule that generates it, and there isn't a good way to change dependencies based on development vs release (as far as I know).

The bundle size rules are also tricky since they run in the current PR's branch and the master branch (by cloning a copy of it). The first part is easy with Bazel. We can just write a test rule that checks the output size of the bundles. The second part, comparing the size to the master branch, is more difficult since it requires running tests in a completely different Bazel workspace which has the exact same name as the current workspace. This is likely not easy to do with Bazel, and may actually be impossible. This script may need to remain as a shell script.
___
*[tools/make_version_test_file.ts, line 69 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5451#-Mgm5i1q-epeePgoKJ24:-Mgq23Os9xvTKycvV6gz:b576m1t) ([raw file](https://github.com/tensorflow/tfjs/blob/a944ba907f9c86a6e66ef42dad43b1fcad8dabd8/tools/make_version_test_file.ts#L69)):*

> ```
> core
> ```
Consider parameterize this part too to accomodate other version tests, such as:
converter: https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/src/version_test.ts
wasm: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/version_test.ts
</blockquote></details>

That's a good point. We may want to pass a template file to the make_version_test_file rule to allow different packages to use it with their own version test logic, although maybe the pattern is similar enough to implement with code generation. I think this is best addressed when we convert another package that has a version test.

PR #5505, opened 2021-08-19
[webgpu] update warning message to guide user enable timestamp-query …

haoyunfeix:
…extension

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5505)

### DIFF
--- a/tfjs-backend-webgpu/src/index.ts
+++ b/tfjs-backend-webgpu/src/index.ts
@@ -47,7 +47,9 @@ if (device_util.isBrowser() && isWebGPUSupported()) {
     } else {
       console.warn(
           `This device doesn't support timestamp-query extension. ` +
-          `Zero will shown for the kernel time when profiling mode is` +
+          `Start Chrome browser with flag ` +
+          `--disable-dawn-features=disallow_unsafe_apis then try again. ` +
+          `Or zero will shown for the kernel time when profiling mode is` +
           `enabled. Using performance.now is not workable for webgpu since` +
           `it doesn't support synchronously to read data from GPU.`);
     }


qjia7 (approved):
LGTM.
@kainino0x @lina128 take a look, thanks.

tfjs-backend-webgpu/src/index.ts
```diff
@@ -47,7 +47,9 @@ if (device_util.isBrowser() && isWebGPUSupported()) {
     } else {
       console.warn(
           `This device doesn't support timestamp-query extension. ` +
-          `Zero will shown for the kernel time when profiling mode is` +
+          `Start Chrome browser with flag ` +
+          `--disable-dawn-features=disallow_unsafe_apis then try again. ` +
+          `Or zero will shown for the kernel time when profiling mode is` +
           `enabled. Using performance.now is not workable for webgpu since` +
           `it doesn't support synchronously to read data from GPU.`);
```
kainino0x, line 54: comment unrelated to this change: I don't see why async means you can't use performance.now. You can get a timestamp before issuing the work, await the work being completed, and get another timestamp.

kainino0x (approved):

tfjs-backend-webgpu/src/index.ts
qjia7: @kainino0x It's limited by the tfjs implementation. You can find the detailed reason here https://github.com/tensorflow/tfjs/pull/4862#issuecomment-809003392 and https://github.com/tensorflow/tfjs/pull/4862#issuecomment-809004617

kainino0x: Thanks for the background, I think I understand the issue is that apps using the tf.js api are calling many ops synchronously so you can't just `await` on `onSubmittedWorkDone` between ops like you could with dataSync.

qjia7: Yes, we can't use `await` in a sync function `profileKernel`. And if we put it in the async `time`, we can't make sure the correct time is collected as comment https://github.com/tensorflow/tfjs/pull/4862#issuecomment-809004617 said.

lina128 (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5505#-:-MhtTrTz7OY-0N71r8M4:bb6ywmx)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @kainino0x)

___
*[tfjs-backend-webgpu/src/index.ts, line 54 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5505#-MhQtNGjNRUGcPIgjE_j-r1-54:-MhtTucd23zfw30LAocS:b-8ceg2l) ([raw file](https://github.com/tensorflow/tfjs/blob/0b1963e2aa956011d0c2b97184747ea8d152daff/tfjs-backend-webgpu/src/index.ts#L54)):*

Yes, we can't use `await` in a sync function `profileKernel`. And if we put it in the async `time`, we can't make sure the correct time is collected as comment https://github.com/tensorflow/tfjs/pull/4862#issuecomment-809004617 said.
</blockquote></details>

Hi Kai, is it possible to have a sync version of data sync for WebGPU? Besides benchmarking, we have some high throughput video pipelines that prefer a sync version.

kainino0x:
> Hi Kai, is it possible to have a sync version of data sync for WebGPU? Besides benchmarking, we have some high throughput video pipelines that prefer a sync version.

We are still discussing it but it's very likely that if we add anything it will be available only in workers.

qjia7:
> > Hi Kai, is it possible to have a sync version of data sync for WebGPU? Besides benchmarking, we have some high throughput video pipelines that prefer a sync version.
> 
> We are still discussing it, but it's very likely that - if we add anything - it will be available only in workers.

@kainino0x  If readBuffer sync is only available in workers, does that mean we need to put the whole webgpu backend in the worker? Or we can put the partial work to the worker, like read buffer data back, and keep the most of the work in the main thread?

kainino0x:
We haven't finished designing multithreaded WebGPU yet so initially you'd have to move the whole thing. But communication between threads is also asynchronous, so even multithreading wouldn't let you make the readback synchronous.

PR #5432, opened 2021-08-04
webgpu: fix dispatch size exceeds the limits

xhcao:
The maximum values of dispatch size and work group size on x/y/z
dimensions are 65535 and 256, if output data is only flatted on x
dimension and exceeds the limits, we should reshape dispatch layout
on x/y or x/y/z dimensions.

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5432)

PR #5519, opened 2021-08-23
[webgpu] Fix exceed dispatch size on WGSL

axinging:
This change is based on the GLSL version: 
https://github.com/tensorflow/tfjs/pull/5432

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5519)

### DIFF
--- a/tfjs-backend-webgpu/src/shader_preprocessor_wgsl.ts
+++ b/tfjs-backend-webgpu/src/shader_preprocessor_wgsl.ts
@@ -68,12 +68,26 @@ export interface InputInfo {
 
 export function getWorkGroupSizeStringWgsl(
     workGroupSize: [number, number, number]): string {
-  if (workGroupSize == null) {
-    return '';
-  }
   return `
   [[stage(compute), workgroup_size(${workGroupSize[0]}, ${workGroupSize[1]}, ${
       workGroupSize[2]})]]
+`;
+}
+
+export function getGlobalIndexStringWgsl(
+    workGroupSize: [number, number, number]): string {
+  return `
+  let index = getGlobalIndex(globalId, localId, vec3<u32>(${
+      workGroupSize[0]}u, ${workGroupSize[1]}u, ${workGroupSize[2]}u));
+`;
+}
+
+export function getMainHeaderStringWgsl(
+    workGroupSize: [number, number, number]) {
+  return `
+  [[stage(compute), workgroup_size(${workGroupSize[0]}, ${workGroupSize[1]}, ${
+      workGroupSize[2]})]]
+  fn main([[builtin(local_invocation_id)]] localId : vec3<u32>, [[builtin(global_invocation_id)]] globalId : vec3<u32>)
 `;
 }
 
@@ -118,7 +132,7 @@ export function makeShader(
   if (program.size != null) {
     uniformDeclaration += 'size : u32; ';
   }
-
+  uniformDeclaration += 'dispatchSize : vec3<u32>; ';
   if (program.uniformsWgsl) {
     uniformDeclaration += program.uniformsWgsl;
   }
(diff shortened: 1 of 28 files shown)


axinging:
@xhcao @qjia7 @haoyunfeix @gyagp, PTAL

tfjs-backend-webgpu/src/shader_preprocessor_wgsl.ts
```diff
@@ -246,6 +269,21 @@ const SAMPLING_SNIPPETS = `
     return u32(dot(vec4<f32>(coords), vec4<f32>(
         f32(shape.y) * f32(shape.z) * f32(shape.w), f32(shape.z) * f32(shape.w), f32(shape.w), 1.0)));
   }
+
+  // Only used when the y/z dimension of workgroup size is 1.
+  fn getGlobalIndex(globalId : vec3<u32>, localId : vec3<u32>, workGroupSize : vec3<u32>) -> u32 {
+    let workGroupID = (globalId - localId)/vec3<u32>(
```
qjia7, line 275: nit: Please move `let workGroupID = ` under the `if` block.

qjia7 (approved):
LGTM with one nit, thanks.
@kainino0x @lina128 take a look.

tfjs-backend-webgpu/src/shader_preprocessor_wgsl.ts
axinging: Done

tfjs-backend-webgpu/src/shader_preprocessor_wgsl.ts
```diff
@@ -535,8 +576,10 @@ export function generateGetOutputCoords(
   const outRank = outShape.length;
   if (x.length === outRank) {
     const dtype = getCoordsDataTypeWgsl(outRank);
-    const snippet = `fn getOutputCoords(globalId : vec3<u32>) -> ${dtype}{
-      return getCoordsFromFlatIndex(u32(globalId.x));
+    const snippet =
+        `fn getOutputCoords(globalId : vec3<u32>, globalIndex : u32) -> ${
```
qjia7, line 580: A second thought, do you think is it better to define `getOutputCoords` like below?
```
fn getOutputCoords(globalId : vec3<u32>, localId : vec3<u32>) -> ${dtype}{
  return getCoordsFromFlatIndex(getGlobalIndex(globalId, localId));
}
```
In this case, we don't need to calculate `globalIndex` in all programs main body and pass it here. We only use it when needed like `x.length === outRank` is satified.

And for `getGlobalIndex`, `workGroupSize` is not necessary since we can get it from program.workGroupsize in this file.

axinging: This maty result in getGlobalIndex is called many times, example:
https://github.com/tensorflow/tfjs/pull/5519/files#diff-df04afa736a2348237666850c1f5f40b6f17e2bb47c421198302fbf7abb72e65R135

axinging: Yes, I think we can do some futher optimization about this. However, how about we land this first? Thi swill block my next two PRs. 
> That's a good point. Maybe we should provide a new func `getXXXFromFlatIndex(int index)` instead of `getXXXAtOutCoords()` when it's flat dispatch layout to avoid repeating calculations. Then all issues can be resolved.

@kainino0x  @lina128, PTAL, thanks!

kainino0x (approved):
no further comments, doing the getXXXFromFlatIndex change in a later PR seems best to me unless it's introducing a new issue in this PR.

lina128 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5519)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @axinging and @kainino0x)

___
*[tfjs-backend-webgpu/src/shader_preprocessor_wgsl.ts, line 580 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5519#-MhlLO33qNFqfPKtuAPK-r1-580:-MhtS7LbEi5JGR3Ijrzv:bace1gb) ([raw file](https://github.com/tensorflow/tfjs/blob/faa5fcbf171e2bd7931e9883465d45abdae0ec7c/tfjs-backend-webgpu/src/shader_preprocessor_wgsl.ts#L580)):*

Yes, I think we can do some futher optimization about this. However, how about we land this first? Thi swill block my next two PRs. 
> That's a good point. Maybe we should provide a new func `getXXXFromFlatIndex(int index)` instead of `getXXXAtOutCoords()` when it's flat dispatch layout to avoid repeating calculations. Then all issues can be resolved.

@kainino0x  @lina128, PTAL, thanks!

</blockquote></details>

The refactor sounds good. Can be in followup PR, just add a TODO here.

tfjs-backend-webgpu/src/shader_preprocessor_wgsl.ts
```diff
@@ -69,14 +69,37 @@ export interface InputInfo {
...
+  let index = getGlobalIndex(globalId, localId, vec3<u32>(${
+      workGroupSize[0]}u, ${workGroupSize[1]}u, ${workGroupSize[2]}u));
+`;
+}
+
+export function getMainHeaderStringWgsl(
+    workGroupSize: [number, number, number]) {
+  if (workGroupSize == null) {
+    throw new Error('Work group size is null!');
+  }
```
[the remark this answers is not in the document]
axinging: Thanks, null check is removed.

axinging:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5519)_ status: :shipit: complete! 3 of 1 approvals obtained (waiting on @axinging and @kainino0x)
> 
> _[tfjs-backend-webgpu/src/shader_preprocessor_wgsl.ts, line 580 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5519#-MhlLO33qNFqfPKtuAPK-r1-580:-MhtS7LbEi5JGR3Ijrzv:bace1gb) ([raw file](https://github.com/tensorflow/tfjs/blob/faa5fcbf171e2bd7931e9883465d45abdae0ec7c/tfjs-backend-webgpu/src/shader_preprocessor_wgsl.ts#L580)):_
> 
> Previously, axinging (Xu Xing) wrote…
> The refactor sounds good. Can be in followup PR, just add a TODO here.

TODO added.

PR #5524, opened 2021-08-23
improving error message when input or dataset output shape does not match the inputShape

sourabh112:
Changed the error message as mentioned in #1100 
also tested locally with yarn build, format, lint.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5524)

### DIFF
--- a/tfjs-backend-webgl/karma.conf.js
+++ b/tfjs-backend-webgl/karma.conf.js
@@ -58,13 +58,13 @@ const devConfig = {
   ],
   preprocessors: {'**/*.ts': ['karma-typescript']},
   karmaTypescriptConfig,
-  reporters: ['dots', 'karma-typescript']
+  reporters: ['dots', 'karma-typescript'],
+  singleRun: true
 };
 
 const browserstackConfig = {
   ...devConfig,
   hostname: 'bs-local.com',
-  singleRun: true,
   port: 9886
 };
 
--- a/tfjs-backend-webgl/scripts/test-ci.sh
+++ b/tfjs-backend-webgl/scripts/test-ci.sh
@@ -22,8 +22,7 @@ if [ "$NIGHTLY" = true ]; then
   node ../scripts/run_flaky.js "yarn run-browserstack --browsers='win_10_chrome,bs_android_9' --testEnv webgl2"
   node ../scripts/run_flaky.js "yarn run-browserstack --browsers=bs_chrome_mac --testEnv webgl2 --flags '{"\""WEBGL_PACK"\"": false}'"
   node ../scripts/run_flaky.js "yarn run-browserstack --browsers=bs_chrome_mac --testEnv webgl2 --flags '{"\""WEBGL_CPU_FORWARD"\"": true}'"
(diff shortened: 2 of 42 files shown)


sourabh112:
@googlebot I signed it

sourabh112:
@googlebot I signed it!

jinjingforever:
This PR looks good! Could you please add a test for the error message [here](https://github.com/tensorflow/tfjs/blob/master/tfjs-layers/src/engine/training_test.ts#L73)? Thank you!

sourabh112:
@jinjingforever Can you please provide some resources about how these tests work? or an example/related PR in which I can check how to add a new test.

sourabh112:
@jinjingforever So, I read about the jasmine testing framework. And I think it's a misunderstanding. I didn't add a new error type but just improved the existing error. As for adding tests in the training_test.ts file, they are already there.
https://github.com/tensorflow/tfjs/blob/6a80f8a936fdc52b54852be0fc54c359c96c01bc/tfjs-layers/src/engine/training_test.ts#L112-L115

If I'm missing something here, thus causing this misunderstanding. Please correct me.

jinjingforever:
Hi @sourabh112 Sorry I didn't make myself clear. What I meant was that it would be useful to have a test case (or "spec" in jasmine term, aka the `it` function) that can trigger the error that you improved in this PR, so that we can make sure that the error does get thrown when `standardizeInputData` function is called with certain inputs, and the error message content matches what we expect.

The test cases from L100 to L123 are all for testing errors for different inputs. https://github.com/tensorflow/tfjs/blob/6a80f8a936fdc52b54852be0fc54c359c96c01bc/tfjs-layers/src/engine/training_test.ts#L100-L123

For example, the spec you mentioned in your comment tests the error message here: https://github.com/tensorflow/tfjs/blob/4e3bf73cc2f1bdfa4e6ed81f703173193e30772b/tfjs-layers/src/engine/training.ts#L138-L141

For this PR, you would need to add another one for the particular error that you edited in the PR. Something like:

```ts
it('A description for your test', () => {
  expect(() => standardizeInputData(...).toThrowError(/.../);
});
```  

This test case should've existed before but I guess we didn't add it at the time. So thank you for adding it along with this PR:) 

To test locally, you can try running `yarn` then `yarn test` in the `tfjs-layers` directory. You can also change your spec from using `it` to `fit` so that it will be the only test to run when running `yarn test` (more details [here](https://jasmine.github.io/2.1/focused_specs.html)). Don't forget to change it back when committing. Using `fit` in this case is probably better because `yarn test` will show some unrelated test failures somewhere else and `fit` helps you focus.

Please feel free to let me know any questions. Thank you!

sourabh112:
Thanks for such a detailed explanation. Now I understand What I need to do, I will update the tests accordingly and make the commit.

sourabh112:
@jinjingforever I have added a test for shape mismatch error. Please review it.

jinjingforever (approved):
Looks great! Thank you!

PR #5528, opened 2021-08-24
webgl: Use uniforms for PadV2

qjia7:


qjia7:
@pyu10055 @lina128 please take a look, thanks.

We are adding more and more `customUniforms` for each program. Unlike shapes uniforms behind a flags, custom uniforms take effect immediately. We may lose the chance to measure the performance impact. Do you think is it necessary to also add a flag for custom uniforms? But the additional flag will bring more complexity for the code. What's your opinion?

lina128 (commented):
Can each PR with customUniforms change include benchmark result?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5528#-:-MhtpI4yBtUl6YtZc0La:b4c2z4x)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

pyu10055 (commented):
Thank you for the refactoring, I agree that having the per-op benchmark result with the PR would be useful.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5528#-:-MhuG5F60i54ZIAaB-9H:bszjyai)* status: 0 of 1 approvals obtained

qjia7:
@pyu10055 @lina128 I ran a round of benchmarks. It's hard to see the perf impact of custom uniforms for padv2. Or I may say it won't have obvious perf impact for the performance. Please take another look. Thanks.

lina128 (approved):
LGTM

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5528)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @qjia7)

___
*[tfjs-backend-webgl/src/pad_packed_gpu.ts, line 44 at r3](https://reviewable.io/reviews/tensorflow/tfjs/5528#-MpCyITM6Okmjtm-8SL_:-MpCyITM6Okmjtm-8SLa:b-7v74q2) ([raw file](https://github.com/tensorflow/tfjs/blob/1167276554d33a4a8ca35792b2329d22a1b7b34d/tfjs-backend-webgl/src/pad_packed_gpu.ts#L44)):*
> ```XML
>     });
>     const start = paddings.map((_, i) => `pad${i}[0]`).join(',');
>     // If the |xShape| is squeezed, we can't use it to calculate |end|.
> ```

why the xShape would be squeezed only for shapeUniform case?

qjia7 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5528)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

___
*[tfjs-backend-webgl/src/pad_packed_gpu.ts, line 44 at r3](https://reviewable.io/reviews/tensorflow/tfjs/5528#-MpCyITM6Okmjtm-8SL_:-MpEZzbL1uKN41Z6V8aT:bp5p8o4) ([raw file](https://github.com/tensorflow/tfjs/blob/1167276554d33a4a8ca35792b2329d22a1b7b34d/tfjs-backend-webgl/src/pad_packed_gpu.ts#L44)):*

why the xShape would be squeezed only for shapeUniform case?
</blockquote></details>

All the input shapes are uploaded as the `uniformShape` from https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/gpgpu_math.ts#L238. The `uniformShape` may or may not be squeezed. In `getSampler[2|3|4]D/getPackedSampler3D` in `shader_compiler.ts`, you can see under some situations, the shader is using the squeezed shape to do the calculation. And in this file, the original code is using the non-squeezed shape `xShape`. However, the uniform `xShape` may be squeezed. That's why I added this check.

Issue #5452, opened 2021-08-10
(ranged) Sigmoid layer in advanced activation function


bibs2091:
- Are you willing to contribute it (Yes I am willing to):

In tfjs, there is a lot of activation functions as a layer, but what is weirdly missing is layers.sigmoid.

This day I was making a dot product matrix factorization recommender system and the output of the dot layer should be in between 0 and 1, for this layers.sigmoid is needed.

FastAI also offer ranged sigmoid which is very interesting, it squeeze the value into a range ]a,b[ instead of ]1,0[, this can be also used in my case when i need to predict ratings in a certain range (let's say ratings from 1 star to 5 stars).

I can contribute to add this feature.

sourabh112:
@rthadur Thanks for the opportunity and sorry for the late reply. I was making myself comfortable with the code of the advanced activation layer. In order to accommodate the sigmoid layer in advanced activation layers. I just need to make some changes to the` exports_layers.ts` and `advanced_activations.ts` files. I will make a PR hopefully by tomorrow. with these changes with a working example. After that, I will add test in `advanced_activations.ts` and make any changes if needed.

sourabh112:
@rthadur As for the ranged sigmoid function suggested by @bibs2091 might be a difficult task because (As per my knowledge) there is no file related to the ranged sigmoid activation function and I have to implement that from scratch. In doing so, I would need some (Actually a lot in the tfjs-core part where actually these functions start) help. If you allow doing that, then I will do that in the same PR after implementing stuff included in the above comment. 

If you think I'm wrong somewhere please correct me.

bibs2091:
@rthadur the implementation of SigmoidRange is in this pull request https://github.com/tensorflow/tfjs/pull/5548 ,
 kindly review it it and tell me what do you think.

@sourabh112  I am sorry I did not notice that you have been assigned, I already started working on the issue days ago and I wrote in my issue that I am ready to contribute to the idea.

sourabh112:
@bibs2091 No problem, I was just starting to work on it. And after going through your PR. I got to know that I was on quite a bit wrong path. And sorry for not reading the last line of the issue.

gaikwadrahul8:
Hi, @bibs2091 

[...]

PR #5548, opened 2021-08-29
Add SigmoidRange layer

bibs2091:
Hello,

This is an implementation of the SigmoidRange layer mentioned in https://github.com/tensorflow/tfjs/issues/5452#issue-965182223 .

The SigmoidRange takes two parameters  **min**  and **max** and forces the output to be between them. The default value of **min** and **max** are **0** and **1** respectively (a normal sigmoid) 
The SigmoidRange equation goes this way:
`SigmoidRange(x)  = sigmoid(x) * (max - min) + min`

Please review it and tell me if I should add anything

### DIFF
--- a/tfjs-layers/src/exports_layers.ts
+++ b/tfjs-layers/src/exports_layers.ts
@@ -11,7 +11,7 @@
 import {InputLayer, InputLayerArgs} from './engine/input_layer';
 import {Layer, LayerArgs} from './engine/topology';
 import {input} from './exports';
-import {ELU, ELULayerArgs, LeakyReLU, LeakyReLULayerArgs, PReLU, PReLULayerArgs, ReLU, ReLULayerArgs, Softmax, SoftmaxLayerArgs, ThresholdedReLU, ThresholdedReLULayerArgs} from './layers/advanced_activations';
+import {ELU, ELULayerArgs, LeakyReLU, LeakyReLULayerArgs, PReLU, PReLULayerArgs, ReLU, ReLULayerArgs, Softmax, SoftmaxLayerArgs, ThresholdedReLU, ThresholdedReLULayerArgs, SigmoidRange, SigmoidRangeLayerArgs} from './layers/advanced_activations';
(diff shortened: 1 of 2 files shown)


bibs2091:
@googlebot I signed it!

pyu10055 (commented):
The PR looks good, thanks you for the contribution. 
I did some googling, I did not see this layer is defined by either keras or tf.keras
I am trying to understand the use case, can you provide any links to the definition of this activation layers?
Also if you can add tests would be great, thanks

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5548#-:-MiN5DOK1F5JtmuA2ZGP:b-30zye4)* status: 0 of 1 approvals obtained

bibs2091:
@pyu10055 Thank you for taking the time to review my PR.
The SigmoidRange layer is inspired from the framework `fastai` which has more than 21k star in GitHub, you can check for the docs of the layer [here ](https://docs.fast.ai/layers.html#SigmoidRange) and the code [here](https://github.com/fastai/fastai/blob/master/fastai/layers.py#L97).

The layer is used when you want to bound the output of your model in a certain range. For my exact use case, I use it for my recommendation system engine, I know for sure that the ratings are between 1 and 5 and my model can accept float values, and SigmoidRange makes a perfect sense to be used. I think other regression tasks can benefit from that if the output min and max are already known (example: when the model task it to output the axis (x,y) of the nose in an image, the SigmoidRange will assure that the coordinates are bigger than 0 and smaller than the size of the image)

One can argue that the model will learn that value is between min and max, but it will take some epochs to learn that. I ran a simple performance test on it, and it really helped accelerating the training speed in the first epochs:
image

Finally, if you think the feature is not essential enough or the name is too confusing, we can rename the layer into Sigmoid (since there's no sigmoid layer) and this layer optionally can receive two params, min and high, if the user does not set them they are automatically set to 0 and 1 respectively making a the normal sigmoid function.

pyu10055 (commented):
@bibs2091 Thank you for the detailed explanation. I think keep it as SigmoidRange would be better, since it is different from the official Sigmoid definition.
Please let me when you have the tests ready. Thanks!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5548#-:-Mj6P6ji2EaOyzC7WkTl:b4sfycz)* status: 0 of 1 approvals obtained

PR #5594, opened 2021-09-08
webgpu: remove if-else block of getGlobalIndex

xhcao:
There is a little performance loss for if-else block when calling
shader getGlobalIndex function at high frequency, so remove if-else
block.


xhcao:

xhcao:
I had run some models on different platforms, Intel TGL and CFL, AMD RX6900XT, no evident performance loses for the if-else block, so I don't want to upstream this patch. Do you have any comments about it?

PR #5598, opened 2021-09-09
webgpu: Batch several dispatching calls into one pass

qjia7:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5598)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -73,6 +73,7 @@ export class WebGPUBackend extends KernelBackend {
   device: GPUDevice;
   queue: GPUQueue;
   currentCommandEncoder: GPUCommandEncoder;
+  currentComputePass: GPUComputePassEncoder;
   tensorMap: DataStorage<TensorBufferInfo>;
   supportTimeQuery: boolean;
   dummyCanvas: HTMLCanvasElement;
@@ -96,7 +97,7 @@ export class WebGPUBackend extends KernelBackend {
   private activeTimers: TimerNode[];
   private uploadWaitMs = 0;
   private downloadWaitMs = 0;
-  private computePassNumberInEncoder = 0;
+  private dispatchNumberInEncoder = 0;
   private querySet: GPUQuerySet;
   private fromPixelProgram?: FromPixelsProgram;
   private fromPixelImportProgram?: FromPixelsImportProgram;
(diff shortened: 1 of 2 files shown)


qjia7:
@kainino0x  Based on the comment https://github.com/gpuweb/gpuweb/issues/2046#issuecomment-909981545, multiple dispatching calls per pass should be better than one dispatching call per pass. But I tested this PR on TFJS benchmarks, no obvious changes observed both on windows and mac. So I'd like you to review this PR to see if we need this change or if it has any potential issues in this change. Thanks.

kainino0x (approved):
LGTM. It may be that we observe no performance difference because there are implicit memory dependencies between every dispatch, and so the driver isn't able to take advantage of the fact that multiple dispatches are in the same compute pass. It may also be that TF.js workloads bottleneck on something else and so the cost is hidden. I'm not sure. But in any case, this looks like it should be strictly better - at the very least, it should result in a tiny bit less CPU overhead due to WebGPU function calls.

pyu10055 (commented):
a high level question, what determines the amount of batching?

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5598#-:-MjeIlpO5KZ-rmybevz4:b-wtt5v1)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

qjia7:
> a high level question, what determines the amount of batching?
> 
The webgpu flag `WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE`. See  https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/src/backend_webgpu.ts#L809

pyu10055 (commented):
If the batch did not exceed 15, is there a timer that forces the submit?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5598#-:-MjjRA1c79m_YLDYF3T0:b-tzp18w)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

qjia7:
> If the batch did not exceed 15, is there a timer that forces the submit?
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5598#-:-MjjRA1c79m_YLDYF3T0:b-tzp18w)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

When tf.data() is called, it will be forced to submit. See https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/src/backend_webgpu.ts#L351

PR #5602, opened 2021-09-10
[webgpu] Use i32 as dimension type in WGSL

axinging:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5602)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/argminmax_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/argminmax_webgpu.ts
@@ -31,7 +31,7 @@ export class ArgMinMaxProgram implements WebGPUProgram {
   workGroupSize: [number, number, number];
   variableNames = ['x'];
   uniforms = 'int axis;';
-  uniformsWgsl = 'axis : u32;';
+  uniformsWgsl = 'axis : i32;';
   inputShape: number[];
   reductionFactor: number;
   op: string;
@@ -270,22 +270,22 @@ export class ArgMinMaxProgram implements WebGPUProgram {
       fn getInputCoordInfo(globalId : vec3<u32>, globalIndex : u32) -> vec2<u32>{
         let outputCoords : ${
         outputCoordsType} = getOutputCoords(globalId, globalIndex);
-        var i = ${this.outputShape.length - 1}u;
+        var i = ${this.outputShape.length - 1};
 
         var stride = 1u;
         var inputStride = 1u;
         var offset = 0u;
 
-        for (var r = 1u; r <= ${this.inputShape.length}u; r = r + 1u) {
-          let length = ${indexInputShape(`${this.inputShape.length}u - r`)};
-          if (${this.inputShape.length}u - r == uniforms.axis) {
+        for (var r = 1; r <= ${this.inputShape.length}; r = r + 1) {
+          let length = ${indexInputShape(`${this.inputShape.length} - r`)};
+          if (${this.inputShape.length} - r == uniforms.axis) {
             inputStride = stride;
           } else {
             offset = offset + ${
         indexOutputCoords('outputCoords', 'i')} * stride;
-            i = i - 1u;
+            i = i - 1;
           }
-          stride = stride * length;
+          stride = stride * u32(length);
         }
 
         return vec2<u32>(offset, inputStride);
--- a/tfjs-backend-webgpu/src/kernels/matmul_packed_vec4_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/matmul_packed_vec4_webgpu.ts
@@ -194,7 +194,7 @@ export function makeMatMulPackedVec4SourceWgsl(
 
     let globalRow = globalId.y * RowPerThread;
     let globalCol = globalId.x;
-    let numTiles = (uniforms.dimInner - 1u) / TileInner + 1u;
+    let numTiles = u32(i32(uniforms.dimInner) - 1) / TileInner + 1u;
 
     var acc: array<vec4<f32>, ${tileInfo.RowPerThread}>;
     var ACached : vec4<f32>;
@@ -261,7 +261,7 @@ export function makeMatMulVectorVec4SourceWgsl(
     let globalCol = globalId.x;
     let globalRow = globalId.y;
 
-    let numTiles = (uniforms.dimInner - 1u) / tileSize + 1u;
+    let numTiles = u32(i32(uniforms.dimInner) - 1) / tileSize + 1u;
 
     // Without this initialization strange values show up in acc.
     var acc = vec4<f32>(0.0);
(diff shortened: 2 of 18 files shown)


axinging:
@qjia7 @haoyunfeix @xhcao @gyagp , PTAL.

tfjs-backend-webgpu/src/kernels/strided_slice_webgpu.ts
```diff
@@ -91,17 +91,17 @@ export class StridedSliceProgram implements WebGPUProgram {
...
       let outputAxis = 0;
       newCoords =
           this.outputShape
               .map((_, i) => {
                 outputAxis++;
                 return this.outputShape.length === 1 ?
-                    `coords * uniforms.strides[${i}] + uniforms.begin[${i}]` :
-                    `coords[${outputAxis - 1}] * uniforms.strides[${
-                        i}] + uniforms.begin[${i}]`;
+                    `coords * u32(uniforms.strides[${i}]) + u32(uniforms.begin[${i}])` :
```
qjia7, line 102: For https://js.tensorflow.org/api/latest/#stridedSlice, we can see that `a stride can be negative, which causes a reverse slice.`. But here you are converting it to `u32`. Is it a problem?

tfjs-backend-webgpu/src/kernels/slice_webgpu.ts
```diff
@@ -47,7 +47,7 @@ export class SliceProgram implements WebGPUProgram {
 
     this.start = start;
     this.uniforms = `${getCoordsDataType(start.length)} start; `;
-    this.uniformsWgsl = `start : ${getCoordsDataTypeWgsl(start.length)}; `;
+    this.uniformsWgsl = `start : ${getCoordsDataTypeWgsl(start.length, 'i32')}; `;
```
qjia7, line 50: It seems that u32 is better.

tfjs-backend-webgpu/src/kernels/reduce_webgpu.ts
```diff
@@ -30,7 +30,7 @@ export class ReduceProgram implements WebGPUProgram {
   workGroupSize: [number, number, number];
   variableNames = ['x'];
   uniforms = 'int reduceSize;';
-  uniformsWgsl = 'reduceSize : u32;';
+  uniformsWgsl = 'reduceSize : i32;';
```
qjia7, line 33: It's not necessary to change `reduceSize` type from `u32` to `i32`？
And it's also weird to do below change:
```
       fn DIV_CEIL(a : u32, b : u32) -> u32 {
        return u32(i32(a) - 1) / b + 1u;
       }
```
if a = 0u, in anyway, it is not correct unless you change a/b's type to `i32`

tfjs-backend-webgpu/src/kernels/pad_webgpu.ts
```diff
@@ -99,10 +99,11 @@ export class PadProgram implements WebGPUProgram {
     const rank = this.xShape.length;
     const type = getCoordsDataTypeWgsl(rank);
     // The length of paddings are same with the rank of the input tensor.
-    const start = this.xShape.map((_, i) => `uniforms.pad${i}[0]`).join(',');
+    const start =
+        this.xShape.map((_, i) => `u32(uniforms.pad${i}[0])`).join(',');
```
qjia7, line 103: Is it possible that `pad[i]` is a negative value? But you are converting it to `u32`. Can you add a unit test for this kind of situation?

tfjs-backend-webgpu/src/kernels/matmul_packed_vec4_webgpu.ts
```diff
@@ -194,7 +194,7 @@ export function makeMatMulPackedVec4SourceWgsl(
 
     let globalRow = globalId.y * RowPerThread;
     let globalCol = globalId.x;
-    let numTiles = (uniforms.dimInner - 1u) / TileInner + 1u;
+    let numTiles = u32(i32(uniforms.dimInner) - 1) / TileInner + 1u;
```
qjia7, line 197: Is it better to use `let numTiles = (i32(uniforms.dimInner) - 1) / ${tileInner} + 1;`? It's just error-prone to change a i32 to u32.

qjia7:
@pyu10055 @lina I have a bad feeling about the current change to mix `u32` and `i32` together. It's error-prone and not good for maintenance. Maybe it's time to reconsider that whether we should unify all types to `i32`. What's your opinion on this?

tfjs-backend-webgpu/src/kernels/argminmax_webgpu.ts
```diff
@@ -31,7 +31,7 @@ export class ArgMinMaxProgram implements WebGPUProgram {
   workGroupSize: [number, number, number];
   variableNames = ['x'];
   uniforms = 'int axis;';
-  uniformsWgsl = 'axis : u32;';
+  uniformsWgsl = 'axis : i32;';
```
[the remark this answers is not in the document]
pyu10055: the axis could be between -rank(input), rank(input), but tfjs-core should handle the conversion of the axis to unsigned value.

lina128 (commented):
Along your suggestion Jiajia, I wonder if we should also use i32 for shapes too? Basically eliminate u32 everywhere. I checked TF's TensorShape class, they seem to use i32 for dimensions. 

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5602#-:-Mj_IOfy1lU-zL1VUpl-:b-rpd9qr)* status: 0 of 1 approvals obtained (waiting on @axinging, @pyu10055, and @qjia7)

pyu10055 (commented):
are there any performance difference between i32 and u32? 
I think u32 would be simpler for the shader, since the branching logic ( for values < 0 ) should be taken cared in js than the shader.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5602#-:-Mj_Jzsw0w8WLpQDkf4e:b-63u55y)* status: 0 of 1 approvals obtained (waiting on @axinging and @qjia7)

lina128 (commented):
I believe that's the original motivation to use u32. But then it can cause hard-to-debug bugs because of the implicit conversion for values < 0, in some intermediate steps, dimension can be negative, we still need to explicitly check the value at the end and set to 0 for negative values. Also Jiajia said the performance is the same between i32 and u32.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5602#-:-Mj_MIkYAe4810Zd64_B:b-rhd60e)* status: 0 of 1 approvals obtained (waiting on @axinging and @qjia7)

qjia7:
@axinging As we discussed in the meeting, we all agree to use i32 in all places, including the shapes, coords and uniforms. Please update this PR to follow this rule. Thanks.

PR #5610, opened 2021-09-13
[webgpu] Create FromPixles program as required

axinging:
Two goals for this PR:
1. When fromPixel isn't used, such as using tensor as input, no need to construct fromPixels programs. 
2. In current design, we can not get URLSearchParams related flags from WebGPUBackend's constructor(This scenario is from e2e). If doing so, will result in cycle dependence:
```
setFlags(from URLSearchParams, such as WEBGPU_USE_GLSL) call tf(from loadScripts). This means we need call loadScripts before setFlags.

loadScript will initialize WebGPU backend and new FromPixlesProgram. FromPixlesProgram needs to check flag WEBGPU_USE_GLSL. This means we need call setFlags before loadScripts.
```
To fix this cycle dependence,  one method is to avoid call new FromPixlesProgram when initialize WebGPU Backend, this is the second goal of this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5610)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -100,8 +100,8 @@ export class WebGPUBackend extends KernelBackend {
   private downloadWaitMs = 0;
   private computePassNumberInEncoder = 0;
   private querySet: GPUQuerySet;
-  private fromPixelProgram:
-      {copyExternal: FromPixelsProgram, import: FromPixelsImportProgram};
+  private fromPixelProgram?: FromPixelsProgram;
+  private fromPixelImportProgram?: FromPixelsImportProgram;
 
   constructor(device: GPUDevice, glslang: Glslang, supportTimeQuery = false) {
     super();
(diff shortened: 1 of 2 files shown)


axinging:
@qjia7 @shaoboyan @haoyunfeix @xhcao @gyagp , PTAL

shaoboyan:
@axinging Would you mind to describe a bit about why doing this?

axinging:
> @axinging Would you mind to describe a bit about why doing this?
1. When there isn't fromPixel, such as Tensor as input, no need to construct fromPixels programs. 
2. In current design, we can not get URLSearchParams related flags from WebGPUBackend's constructor. If doing so, will result in cycle dep:
```
setFlags(from URLSearchParams, such as WEBGPU_USE_GLSL) call tf(From loadScripts). This means we need loadScripts before setFlags.
loadScript will initialize WebGPU Backend and new FromPixlesProgram. FromPixlesProgram needs to check flag WEBGPU_USE_GLSL. This means we need setFlags before loadScripts.
```
So to fix this,  one method is to avoid call new FromPixlesProgram when initialize WebGPU Backend, this is the second goal of this PR.


> @axinging Would you mind to describe a bit about why doing this?

I add some comments in the PR description.

shaoboyan:
@axinging  Ah, got it.  +1 from me.

qjia7 (approved):
LGTM

lina128 (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5610#-:-MjWGwvUAHhk1zxFFhOZ:bb6ywmx)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @jinjingforever)

PR #5620, opened 2021-09-14
Use system python and node:16 for CI

mattsoulanille:
This PR makes Bazel use the system's available python2 and python3 instead of building from source. It also switches many steps to node:16 because node:10's python3 is at 3.5, which does not support format strings that are needed by esbuild's toolchain.

A few steps do not yet use node:16:
1. e2e tests still use the release docker because they require some of the dependencies it installs.
2. tfjs-node[-gpu] still use node:10 because they need to build for n-api v7 for the e2e tests.
3. tfjs-backend-wasm tests use node:10 because they fail with 'out of wasm memory' on node:16.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5620)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -228,51 +228,52 @@ else
 fi
 """
 
-http_archive(
-    name = "python3_interpreter",
-    build_file_content = """
-exports_files(["python3_bin"])
-filegroup(
-    name = "files",
-    srcs = glob(["bazel_install_py3/**"], exclude = ["**/* *"]),
-    visibility = ["//visibility:public"],
-)
-""",
-    patch_cmds = [
-        "mkdir $(pwd)/bazel_install_py3",
-        _py3_configure,
-        "make",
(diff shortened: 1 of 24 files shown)


mattsoulanille:
[Nightly run](https://console.cloud.google.com/cloud-build/builds/6ee5ea91-736c-4352-93a4-eca187772ef9;step=65?project=learnjs-174218), since not all parts of this change may be caught in daily.

pyu10055 (commented):
Does this work for local development environment?  Removing the bazel local python version would fail for local dev environment unless they have the proper setup.
Is there a way to only disable the bazel python install for CI?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5620#-:-Mj_Om101ktlNZ-Edou5:b-yxnkhf)* status: 0 of 1 approvals obtained

Issue #5656, opened 2021-09-23
When we use the WebAssembly backend, `tf.fill()` will throw an error if the `dtype` parameter is not explicitly given


tiandeyu-cs:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Linux Ubuntu 18.04 LTS
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device: N/A
- TensorFlow.js installed from (npm or script link): <script> link
- TensorFlow.js version (use command below): commit cb283fde98 (Date: 2021-09-23 16:54:08 UTC+08)
- Browser version: Google Chrome 92
- Tensorflow.js Converter Version: N/A

**Describe the current behavior**
When we uses the WebAssembly backend, calling `tf.fill()` function without explicitly giving the `dtype` parameter will make it throw an error.

**Describe the expected behavior**
As [the official document says](https://js.tensorflow.org/api/latest/#fill), the `dtype` parameter should default to `float` if it is not explicitly given.

**Standalone code to reproduce the issue**
Provide a reproducible test case that is the bare minimum necessary to generate
the problem. If possible, please share a link to Colab/CodePen/any notebook.

The error can be reproduced by the following steps. First compile the source code and obtain the `tf-core.min.js` and `tf-backend-wasm.min.js` files, along with the WebAssembly module files (i.e. the `.wasm` files). Then put these library files and the following `main.html` file under the same folder. Finally run a Web server and open `main.html` in a browser and the error will show in the JavaScript console.

The source code of `main.html`:
```HTML
<html>
	<head>
		<script src="tf-core.min.js"></script>
		<script src="tf-backend-wasm.min.js"></script>
	</head>
	<body>
		<script>
			async function main(){
				await tf.setBackend("wasm");
				tf.fill([10], 0);
			}
			main();
		</script>
	</body>
</html>
```

**Other info / logs** Include any logs or source code that would be helpful to
diagnose the problem. If including tracebacks, please include the full
traceback. Large logs and files should be attached.

Here is the stack trace shown in JS console:
```
Uncaught (in promise) Error: Unknown dtype undefined
    at Object._ [as bytesPerElement] (util_base.ts:520)
    at a.move (backend_wasm.ts:82)
    at a.write (backend_wasm.ts:54)
    at a.makeOutput (backend_wasm.ts:183)
[...]

PR #5657, opened 2021-09-23
[wasm] add the default dtype for tf.fill (#5656)

tiandeyu-github:
Fix the bug #5656

### DIFF
--- a/tfjs-backend-wasm/src/kernels/Fill.ts
+++ b/tfjs-backend-wasm/src/kernels/Fill.ts
@@ -17,12 +17,16 @@
 
 import {KernelConfig, KernelFunc} from '@tensorflow/tfjs-core';
 import {Fill, FillAttrs} from '@tensorflow/tfjs-core';
+import {util} from '@tensorflow/tfjs-core';
 
 import {BackendWasm} from '../backend_wasm';
 
 export function fill(args: {attrs: FillAttrs, backend: BackendWasm}) {
-  const {attrs: {shape, value, dtype}, backend} = args;
-  const out = backend.makeOutput(shape, dtype);
+  const {attrs, backend} = args;
+  const {shape, value, dtype} = attrs;
+  const $dtype = dtype || util.inferDtype(value);
+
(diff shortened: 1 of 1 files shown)


tiandeyu-github:
@googlebot I signed it!

pyu10055 (approved):
Thank you for fixing this.

jinjingforever (commented):
Thank you for the fix! 

Looks like the WASM backend is not actually testing Fill because it is not included [here](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/setup_test.ts).

Could you please help add Fill there and make sure its tests still pass? Thank you!

tiandeyu-github:
I'm willing to help, but I do not understand the meaning of the test file, and I do not know how to actually run these tests. 

Do you have any introduction or tutorial resources that can help me to understand the test configuration? Thank you😄

jinjingforever:
Hi @tiandeyu-github, thank you! For this, you need to add ` {startsWith: 'fill '},` (note the extra space after "fill") to the end of the `TEST_FILTERS` array [here](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/setup_test.ts#L34). This will ask the WASM backend to run tests whose names start with "fill " which are defined [here](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/fill_test.ts).

```diff
@@ -433,6 +433,7 @@ const TEST_FILTERS: TestFilter[] = [
   {include: 'image.transform'},
   {include: 'batchToSpaceND'},
   {include: 'spaceToBatchND'},
+  {startsWith: 'fill '},
 ];
```

After that, under the `tfjs-backend-wasm` directory, run:

```bash
$ yarn
$ yarn test
```

This should eventually start a chrome which will run all the wasm tests. I tried it in the master branch and it complains about `Unknown dtype undefined` in all the "fill" tests, which should be fixed by your PR.

(If running tests locally on your machine doesn't work due to certain environment setup issues, you could also try pushing it here and check on the cloud build result)

Please give it a try and let me know any problems. Thanks!

PR #5662, opened 2021-09-24
[wasm] Fix kernels in wasm backend about how int32 inputs are handled

### LINKED ISSUE #5641: wasm backend produces incorrect results for int32 tensors
i was wondering why my app produces results with lower precision when using `wasm` backend and finally narrowed it down to `wasm` handling of **int32** tensors

*environment: tfjs 3.9.0 on ubuntu 21.04*
*note: ive tried wasm with and without simd support*

below is a very simple reproduction  
(and i suggest to add something like this to automated tests in the future)  

note that `tf.sum()` is just easiest way to reproduce, this issue is accross the board in tfjs

```js
const data = Array.from(imageData.data); // in my case data is imageData array so each value is just 0..255, but can be any dataset
data.length = 1024 * 1024 * 4; // lets crop array to specific size, just for test
let sumJS = 0;
for (let i = 0; i < data.length; i++) sumJS += data[i];
const tI32 = tf.tensor(data, [data.length], 'int32');
const tF32 = tf.tensor(data, [data.length], 'float32');
console.log({ backend: tf.getBackend(), arrayLength: data.length, jsSum: sumJS, tfSumI32: tf.sum(tI32).dataSync()[0], tfSumF32: tf.sum(tF32).dataSync()[0] });
```

this is ok output when using `tfjs-node` as both js and tf produce same sum:
```js
{
  backend: 'tensorflow',
  arrayLength: 4194304,
  jsSum: 1000114465,
  tfSumI32: 1000114465,
  tfSumF32: 1000114432 // sum of float is slightly off, but close
}
```

but when using `wasm` backend, its completely **broken**:
```js
{
  backend: 'wasm',
  arrayLength: 4194304,
  jsSum: 1000114465,
  tfSumI32: 66326844, // completely broken
  tfSumF32: 1023692544 // slightly off as well, but close
}
```

and this is not a case of `int32` overflow as its reproducible with much lower values as well:
(it seems that error occur with tensors with more than 64k values)

```js 
{
  backend: 'wasm',
  arrayLength: 65536,
  jsSum: 16042946,
  tfSumI32: 16042946, // this is correct
  tfSumF32: 16042946
}
```

```js
{
  backend: 'wasm',
  arrayLength: 131072,
  jsSum: 31840717,
  tfSumI32: 24302074, // this is incorrect!!!
  tfSumF32: 31826932
}
```

and those numbers are faaaar off any `int32` limits

and yes, i've checked that tensors get created correctly in both cases by downloading all values from it using `dataSync()` and comparing to original array

jinjingforever:
I went through all the kernels in the wasm backend, and found that:

- Some kernels don't have the correct implementations for int32 inputs. For example, float pointers are used to access int32 memory content. I added the correct int32 implementations as well as the tests for int32 inputs. They are:
  
  Abs, Exp, LeakyRelu, Max, Min, Neg, Relu, Relu6, Reverse, Round, SelectV2, Square, Step, StridedSlice, Sum,

- Some kernels are not supposed to accept int32 inputs. They don't make sense in real life situations, and I also cross-checked the official TF documents. For them, I added tests to make sure errors will be thrown for int32 inputs. They are:

  Ceil, Conv2d, Cos, Cosh, DepthwiseConv2d, Elu, FlipLeftRight, Floor, FusedConv2d, FusedDepthwiseConv2d, Log, MaxPool, NonMaxSuppression, RotateWithOffset, Rsqrt, Sigmoid, Sin, Softmax, Sqrt, Tan, Tanh, Transform,

- I also updated Prelu and LeakyRelu in webgl/webgpu backend to output float32 tensors (which is how they work in TF).

Fixes #5641 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5662)


jinjingforever (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5662#-:-MkhPD0CENQ2Jtq-8nhx:b-896fix)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

___
*[tfjs-backend-wasm/src/cc/unary.h, line 27 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5662#-MkdbeT9-s9AkSfpLuGU:-MkhKyo87g4Snczbraf5:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/b5127ac10f4c878fa5dcae789e76764f8cd8378f/tfjs-backend-wasm/src/cc/unary.h#L27)):*

name it unary_f32 to be more specific?
</blockquote></details>

Done.
___
*[tfjs-backend-wasm/src/cc/kernels/Exp.cc, line 40 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5662#-Mkddnpu4RKa3GHK3PLh:-MkhLnO7D-N-5n20FO1c:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/b5127ac10f4c878fa5dcae789e76764f8cd8378f/tfjs-backend-wasm/src/cc/kernels/Exp.cc#L40)):*

exp should return float, please update other backends that return int values.
</blockquote></details>

Done.
___
*[tfjs-backend-wasm/src/cc/kernels/Sum.cc, line 73 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5662#-Mkdg_KlBITPMUO59f1q:-MkhMGOh2KV5wJDVWyXC:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/b5127ac10f4c878fa5dcae789e76764f8cd8378f/tfjs-backend-wasm/src/cc/kernels/Sum.cc#L73)):*

sum of boolean does not make sense.
</blockquote></details>

Done.
___
*[tfjs-core/src/ops/depth_to_space_test.ts, line 76 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5662#-MkdhgR0Aaa0sm4vxZ9o:-MkhMRADAp7Q57QxV1z2:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/b5127ac10f4c878fa5dcae789e76764f8cd8378f/tfjs-core/src/ops/depth_to_space_test.ts#L76)):*

check for error message?
</blockquote></details>

Done.
___
*[tfjs-core/src/ops/exp_test.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5662#-MkdhnGJCJ6Cy9-5bQui:-MkhMSIqEVPOFkPo1PKH:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/b5127ac10f4c878fa5dcae789e76764f8cd8378f/tfjs-core/src/ops/exp_test.ts#L34)):*

exp should return float tensor
</blockquote></details>

Done.

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5662)* status: :shipit: complete! 1 of 1 approvals obtained

vladmandic:
there's a leftover line: 

```js
console.log('reduceshape', reduceShape);`
```

PR #5667, opened 2021-09-27
WebGPU: Optimize ScatterNd

qjia7:
PERF

The total time of ScatterNd becomes 22.68 ms from 126.76 ms in
USE-batchsize 30 model in benchmarks on CFL.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5667)


qjia7:
@axinging @xhcao @haoyunfeix PTAL, thanks.

qjia7:
If we don't need to sum duplicated index value (assume there is no duplicated index), we can change the algorithm to analyze input `updates` instead of output. In this case, we can totally remove the `for` loop in shader. The time of scatterNd in USE can be further reduced to less than 1 ms.

@pyu10055 @lina128  Do you know whether tfjs-core knows there are duplicated indices or we need to implement it in backend?

lina128 (approved):
Hi Jiajia, Ping and I discussed, here's our suggestion:
1. Check duplicate in ts code, it's a trade-off between time reduced in optimized GPU path vs. time to download indices to CPU to check. Need to benchmark with different data size to decide whether worth it. 2. If WebGPU can implement Unique op (reference CUDA implementation) in GPU, then can check duplicate in GPU without download, just check whether size(unique(indices)) === size(indices). However, this introduces one more shader to enable the optimization path,  it's also a trade-off, so should also benchmark with different data size to decide whether worth it. 3. Other ways of optimization, if you can sort indices, you may be able to optimize the loop.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5667#-:-MkdKSrLAUrbmozXbTxo:bhr6c0o)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055 and @qjia7)

___
*[tfjs-backend-webgpu/src/kernels/scatter_webgpu.ts, line 32 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5667#-MkcwgVP4km5j5qHIYoM:-MkcwgVQ1SiwsZS7pL4-:b8iuuc2) ([raw file](https://github.com/tensorflow/tfjs/blob/34ef8aabedb6f60fd8ca6aceaf994f25982c6084/tfjs-backend-webgpu/src/kernels/scatter_webgpu.ts#L32)):*
> ```
> 4
> ```
Curious, how is 4 decided?

qjia7 (commented):
Thanks for your valuable suggestion.
I will try 1) `atomicAdd` . Unfortunately, it only supports `uint`/`int` type. I need to check whether we can implement a similar one for float. 2) implement unique op in gpu 3) others. 

Will land this PR first. And replace it once we finish the further optimization.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5667#-:-Mke2rjK8PiXjpNgGEx2:b-t1cjzr)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[tfjs-backend-webgpu/src/kernels/scatter_webgpu.ts, line 32 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5667#-MkcwgVP4km5j5qHIYoM:-Mke1zI21yHilOFEMGaa:b-d75dz) ([raw file](https://github.com/tensorflow/tfjs/blob/9cc85f98d6c6d4bc4d4d1545aecb103c8b04fc5b/tfjs-backend-webgpu/src/kernels/scatter_webgpu.ts#L32)):*

> ```
> 4
> ```
Curious, how is 4 decided?
</blockquote></details>

It's an empirical value. I think it should be larger if the output tensor is large. Temporarily set it as 4 for simplification.

PR #5683, opened 2021-09-30
[e2e] Fix pose detection correctness test

axinging:
For pose-detection, the returned value is predict result * imageSize.
This means the error will be magnified by imageSize, which results in
comparasion fail. So here uses relative comparision.

### DIFF
--- a/e2e/benchmarks/local-benchmark/util.js
+++ b/e2e/benchmarks/local-benchmark/util.js
@@ -71,8 +71,14 @@ function areClose(a, e, epsilon) {
   if (!isFinite(a) && !isFinite(e)) {
     return true;
   }
-  if (isNaN(a) || isNaN(e) || Math.abs(a - e) > epsilon) {
-    return false;
+  if (Math.abs(a) >= 1) {
+    if (isNaN(a) || isNaN(e) || Math.abs(a - e) / Math.abs(a) > 0.01) {
+      return false;
+    }
+  } else {
+    if (isNaN(a) || isNaN(e) || Math.abs(a - e) > epsilon) {
+      return false;
+    }
   }
   return true;
 }
@@ -145,6 +151,7 @@ function expectArraysPredicateFuzzy(actual, expected, predicate, errorRate) {
   }
 }
 
+// TODO: support relative comparison for array.
 function expectArraysClose(actual, expected, epsilon, key) {
   if (epsilon == null) {
     epsilon = tf.test_util.testEpsilon();


mattsoulanille (commented):
I like this new approach. Thanks! I have just a few comments on it.

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5683#-:-Mkrkk9b9JGEKvzVgAYk:b7t4lq5)* status: 0 of 1 approvals obtained (waiting on @axinging)

___
*[e2e/benchmarks/local-benchmark/util.js, line 75 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5683#-MkrkEKQ6dwSvVFJiKnZ:-MkrkEKQ6dwSvVFJiKn_:b-wvp7r2) ([raw file](https://github.com/tensorflow/tfjs/blob/abf51f49c387ee9776411b4b3df2d23c73c867a0/e2e/benchmarks/local-benchmark/util.js#L75)):*
> ```
> 0.01)
> ```
Nit: Should this be hard-coded?
___
*[e2e/benchmarks/local-benchmark/util.js, line 75 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5683#-MkrkJdx0dEi_lEg0VX0:-MkrkJdx0dEi_lEg0VX1:b-qitxsr) ([raw file](https://github.com/tensorflow/tfjs/blob/abf51f49c387ee9776411b4b3df2d23c73c867a0/e2e/benchmarks/local-benchmark/util.js#L75)):*
> ```
> Math.abs(a - e) / Math.abs(a)
> ```
Should we sort `a` and `b` to make sure `areClose(x, y) === areClose(y, x)` in all cases. For example, currently, `areClose(99.995, 101) === false' but `areClose(101, 99.995) === true`.
___
*[e2e/benchmarks/local-benchmark/util.js, line 79 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5683#-Mkri01E9ipNsCz4q6eT:-Mkri01E9ipNsCz4q6eU:b-3qqhlz) ([raw file](https://github.com/tensorflow/tfjs/blob/abf51f49c387ee9776411b4b3df2d23c73c867a0/e2e/benchmarks/local-benchmark/util.js#L79)):*
> ```
>     if (isNaN(a) || isNaN(e) || Math.abs(a - e) > epsilon) {
>       return false;
> ```
I initially thought we should remove this method and only use the new one, but this method works well when numbers are very close to zero and differ from each other by the numerical precision of the backend. The other method would fail in this case since it would compare e.g. `0.00001` with `0.00002` and find that they're different.

axinging:
> I like this new approach. Thanks! I have just a few comments on it.
> 
> > Reviewed 1 of 1 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5683#-:-Mkrkk9b9JGEKvzVgAYk:b7t4lq5)_ status: 0 of 1 approvals obtained (waiting on @axinging)
> 
> _[e2e/benchmarks/local-benchmark/util.js, line 75 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5683#-MkrkEKQ6dwSvVFJiKnZ:-MkrkEKQ6dwSvVFJiKn_:b-wvp7r2) ([raw file](https://github.com/tensorflow/tfjs/blob/abf51f49c387ee9776411b4b3df2d23c73c867a0/e2e/benchmarks/local-benchmark/util.js#L75)):_
> 
> > ```
> > 0.01)
> > ```
> 
> Nit: Should this be hard-coded?

You mean not hard-coded? I change it to RELATIVE_EPSILON.

> 
> _[e2e/benchmarks/local-benchmark/util.js, line 75 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5683#-MkrkJdx0dEi_lEg0VX0:-MkrkJdx0dEi_lEg0VX1:b-qitxsr) ([raw file](https://github.com/tensorflow/tfjs/blob/abf51f49c387ee9776411b4b3df2d23c73c867a0/e2e/benchmarks/local-benchmark/util.js#L75)):_
> 
> > ```
> > Math.abs(a - e) / Math.abs(a)
> > ```
> 
> Should we sort `a` and `b` to make sure `areClose(x, y) === areClose(y, x)` in all cases. For example, currently, `areClose(99.995, 101) === false' but `areClose(101, 99.995) === true`.

Yes, this is great, updated, thanks a lot!

> 
> _[e2e/benchmarks/local-benchmark/util.js, line 79 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5683#-Mkri01E9ipNsCz4q6eT:-Mkri01E9ipNsCz4q6eU:b-3qqhlz) ([raw file](https://github.com/tensorflow/tfjs/blob/abf51f49c387ee9776411b4b3df2d23c73c867a0/e2e/benchmarks/local-benchmark/util.js#L79)):_
> 
> > ```
> >     if (isNaN(a) || isNaN(e) || Math.abs(a - e) > epsilon) {
> >       return false;
> > ```
> 
> I initially thought we should remove this method and only use the new one, but this method works well when numbers are very close to zero and differ from each other by the numerical precision of the backend. The other method would fail in this case since it would compare e.g. `0.00001` with `0.00002` and find that they're different.

Yes, we can not remove this yet.

axinging:
@mattsoulanille , updated, PTAL, I also  add another epsilonOfBigNumber to make it more strict.

axinging:
Thanks @mattsoulanille 

@pyu10055 @lina128 @jinjingforever , PTAL

@qjia7 @haoyunfeix @xhcao @gyagp

PR #5706, opened 2021-10-09
[webgpu] Use different global index for flat and non-flat dispatch

axinging:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5706)


axinging:
@qjia7 @haoyunfeix  @xhcao @gyagp PTAL

PR #5712, opened 2021-10-11
[e2e] Fix USE - batchsize 1 conformance fail

axinging:
For e2e conformance test, the inputs for different backends should be the same. However, for USE - batchsize 1, every time predictFunc is called, nextIdx wil be changed, and so the input changed. This means the expected backend and the actual backend have different inputs. This PR ensures the input for different backends have the same inputs.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5712)


axinging:
@pyu10055  @lina128 , PTAL

PR #5790, opened 2021-10-28
Lint Bazel packages with a central lint script

mattsoulanille:
Lint Bazel-built packages with a central `yarn lint` script at the root of the repository. This change also removes several cloudbuild files that are currently only used for linting.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5790)


mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5790)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

___
*[BAZEL_MIGRATION.md, line 407 at r2](https://reviewable.io/reviews/tensorflow/tfjs/5790#-Mn8X4uq5-kpRn3ucKs-:-MnRss9P6dLRxMq11ju7:b-bxi1uc) ([raw file](https://github.com/tensorflow/tfjs/blob/03b052d41d7c5d8a3191f1c55117c24a2e9af40a/BAZEL_MIGRATION.md#L407)):*

how about the tsconfig.json file at the root directory, will that need to be updated?
</blockquote></details>

The normal `tsconfig.json` file does not need to be updated when `tsconfig_tslint.json` is updated. The `tsconfig_tslint.json` file is only used by tslint. It tells tslint the project structure that is contained in the Bazel  BUILD files so it can correctly check rules that depend on types across our packages (we do this differently in google3, but I think it would be difficult to dreplicate).
___
*[scripts/package_dependencies.json, line 3 at r2](https://reviewable.io/reviews/tensorflow/tfjs/5790#-Mn8XMPDCyOdF3703HAp:-MnRtkk45rs_xQ3p3r3M:b-mti6o7) ([raw file](https://github.com/tensorflow/tfjs/blob/03b052d41d7c5d8a3191f1c55117c24a2e9af40a/scripts/package_dependencies.json#L3)):*

why the deps config needs to be updated?
</blockquote></details>

I removed several `cloudbuild.yml` files from packages that were only running lint scripts. With this change, those packages are linted by the global lint script, so they had no more cloudbuild steps to run. `package_dependencies.json` exists only to track dependencies for the for the cloudbuild generator script, so those packages that no longer have a `cloudbuild.yml` file should be removed from it.
___
*[tfjs-core/src/log.ts, line 18 at r2](https://reviewable.io/reviews/tensorflow/tfjs/5790#-Mn8Y3hUCwmiznDmp50Q:-MnRudZZ6ZJlEagKuDIz:bg4lr1b) ([raw file](https://github.com/tensorflow/tfjs/blob/03b052d41d7c5d8a3191f1c55117c24a2e9af40a/tfjs-core/src/log.ts#L18)):*

this is fixed by Jing here https://github.com/tensorflow/tfjs/pull/5783, maybe you can try to sync again.
</blockquote></details>

Thanks! There seems to be another one caused by #5784. base_side_effects.ts <-> platform_node.ts. I removed it and the tests still pass.

pyu10055 (approved):

PR #5811, opened 2021-11-04
[e2e] Enable DEBUG flag

axinging:
Add DEBUG=true in the url will turn on debug mode.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5811)


axinging:
Hi, @lina128 @pyu10055 @jinjingforever. currently there are at least two flags related to debug (KEEP_INTERMEDIATE_TENSORS inside tfjs-converter, DEBUG inside tfjs-core). 
How about merge those two bool type flags to a number type flag, name it DEBUG: number?

cc @qjia7  @xhcao @gyagp @haoyunfeix PTAL.

pyu10055 (commented):
Hi @axinging Thank you for sharing your thoughts. Since the flags are defined by different packages and have quite different meaning, I think it is a good idea to combine the debug flags at benchmark UI level, meaning the tuneable flag for DEBUG can have multiple values to indicate level of debugging. The UI can show the description as debug verbose level (High, medium, low, none), which could corresponding to combination of those flags.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5811#-:-MnlfVJX3BwxzCr4_X1m:b-kbuf9u)* status: 0 of 1 approvals obtained

axinging:
> Hi @axinging Thank you for sharing your thoughts. Since the flags are defined by different packages and have quite different meaning, I think it is a good idea to combine the debug flags at benchmark UI level, meaning the tuneable flag for DEBUG can have multiple values to indicate level of debugging. The UI can show the description as debug verbose level (High, medium, low, none), which could corresponding to combination of those flags.
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5811#-:-MnlfVJX3BwxzCr4_X1m:b-kbuf9u)_ status: 0 of 1 approvals obtained


Thanks @pyu10055! 
After more investigation, if we combine those flags in the e2e UI level, it will make the logic a bit complicated.
For example,  below is the related part of UI:
```
{
  CHECK_COMPUTATION_FOR_ERRORS: false
  DEBUG: false
  IS_TEST: false
  KEEP_INTERMEDIATE_TENSORS: true
  TENSORLIKE_CHECK_SHAPE_CONSISTENCY: false
  WASM_HAS_MULTITHREAD_SUPPORT: false
  WASM_HAS_SIMD_SUPPORT: true
  WEBGL_CPU_FORWARD: true
  WEBGL_FLUSH_THRESHOLD: -1
  WEBGL_FORCE_F16_TEXTURES: false
  WEBGL_PACK: true
  WEBGL_PACK_DEPTHWISECONV: true
  WEBGL_RENDER_FLOAT32_CAPABLE: true
  WEBGL_USE_SHAPES_UNIFORMS: false
  WEBGL_VERSION: 2
  WEBGPU_CPU_FORWARD: true
  WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE: 15
}
```
And all the values are from tf.env().features (In showEnvironment). So if we only change this in the UI level, we will make below logic more complicated.
``` 
    async function showEnvironment() {
      await tf.time(() => tf.add(tf.tensor1d([1]), tf.tensor1d([1])).data());
      const sortedFeatures = Object.keys(tf.env().features).sort().map(
        name => `  ${name}: ${tf.env().features[name]}`);
      envDiv.innerHTML += `<br/>{<br>${sortedFeatures.join('<br>')}<br>}`;
    }
```

Based on changing UI will make it more complicated, changing in the framework will break the boundary between core and converter.  How about we simply adopt the changes proposed by this PR?

pyu10055 (commented):
Thanks, from the code the DEBUG flag seems to be UI only feature, while KEEP_INTERMEDIATE_TENSORS is not, I am wondering if KEEP_INTERMEDIATE_TENSORS flag should be shown, should let the DEBUG flag controls it and dispose the tensors immediately after, otherwise it would have large mem leak if users are not careful. 

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5811#-:-MoEuNzp7XT_mwEK6d2v:b-xeat32)* status: 0 of 1 approvals obtained

axinging:
> Thanks, from the code the DEBUG flag seems to be UI only feature, while KEEP_INTERMEDIATE_TENSORS is not, I am wondering if KEEP_INTERMEDIATE_TENSORS flag should be shown, should let the DEBUG flag controls it and dispose the tensors immediately after, otherwise it would have large mem leak if users are not careful.
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5811#-:-MoEuNzp7XT_mwEK6d2v:b-xeat32)_ status: 0 of 1 approvals obtained

In this PR, DEBUG and KEEP_INTERMEDIATE_TENSORS  are independent.

PR #5812, opened 2021-11-04
[webgpu] fix type dismatch in matmul_packed_webgpu

haoyunfeix:
BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5812)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/matmul_packed_webgpu.ts
+++ b/tfjs-backend-webgpu/src/kernels/matmul_packed_webgpu.ts
@@ -271,16 +271,16 @@ export class MatMulPackedProgram implements WebGPUProgram {
 
     if (this.transposeA === false) {
       sampleA = this.fitA ?
-          `return A.numbers[batch * batchASize + row * uniforms.dimInner + col];` :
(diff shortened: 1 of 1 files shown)


gyagp:
Title should be mismatch instead of dismatch.

qjia7:
We meet more and more such kind of type casting error. The previous one is #5767. The root case is that we don't check the data's type and just write it as a `f32` data. And our helper functions all force casting all type to `f32`. However, our inputs accept `f32` and `i32`. It makes the data casting a mess. We need to do below cleanup:
either 1) treat all data as f32 and only accept `f32` in shader like the webgl one.
or 2) choose the correct data type to do the calculation. This means if the inputs are `i32`, then all helper functions and main body are using `i32`. Otherwise, all helper functions and main body are using `f32`. No type casting happens in the shader. 

@gyagp @xhcao @axinging What's your opinion?
cc @lina128 @pyu10055

gyagp:
Synced with Jiajia, we will try first approach.

lina128:
First approach sounds good to me. If we are going to support quantization in the future, it may be a different set of functions any way.

PR #5813, opened 2021-11-04
[WebGPU] Implement shared shader lib

axinging:
The goal is to remove SHADER_PREFIX and make shader small.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5813)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/shader_lib.ts
+++ b/tfjs-backend-webgpu/src/kernels/shader_lib.ts
@@ -1,0 +1,73 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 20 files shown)


tfjs-backend-webgpu/src/kernels/shader_lib.ts
```diff
@@ -0,0 +1,59 @@
...
+    for (var i = 0u; i < 4u; i = i + 1u) {
+      if (isNanCustom(val[i])) {
+        res[i] = 1.0;
+      } else {
+        res[i] = 0.0;
+      }
+    }
+    return res;
+  }`;
+}
```
gyagp, line 59: Essentially these are just strings. We need to define finer granularity ones instead of a macro one used in the current code. Just put these sharable strings in a file, and import them specifically when we need them.

Issue #5689, opened 2021-10-03
WebGPU Performance Issues


vladmandic:
i just tried new `tfjs-backend-webgpu` *0.0.1-alpha.8* on `tfjs` *3.9.0*   
*environment: chrome 96 canary on windows 11*

first, great job on adding tons of new ops - from perspective of supported kernel ops, `webgpu` is becoming usable!

however, switch to WGSL is anything but useful so far - it comes as a **major performance degredation**  

overall, `webgpu` has  gotten slower than `webgl`  
(and `webgl` itself has become significantly slower since `tfjs` 3.4.0 - this is discussed separately in several open issues)  

not to mention that new work that has gone into `webgl` to make it manageable (enable uniforms) has no effect on `webgpu`

comparing warmup times  
(fyi, my app by default uses 8 simple models running in parallel - total models size is actually tiny, below 30mb):

- `webgl` *(default settings)*
  > **14 sec** (double the value with uniforms enabled)
- `webgl` with `WEBGL_PACK_DEPTHWISECONV=false` and `WEBGL_USE_SHAPES_UNIFORMS=true`
  > **7 sec** (pretty good)
- `webgpu` *(default settings)*
  > **25 sec** (this is incredibily slow)
- `webgpu` with `WEBGPU_USE_GLSL=true`
  > **15 sec** (already slower than webgl)
- `wasm` (no real warmup, included for refrerence only)
  > **2 sec**

*imo, when developing new backend, goal should be that its better than the previous one - not just that it passes unit tests
if `webgpu` is not significantly improved, it will be a d.o.a. once released*

cc @qjia7 and @xhcao due to work on webgpu  
cc @pyu10055 as assignee on webgl performance degradation issue

gyagp:
@vladmandic Thanks for the good comments and data, as always! 
[...]

PR #5815, opened 2021-11-05
[WIP][WebGPU] Use createComputePipelineAsync to parallelly compile the shader

qjia7:
This PR is in draft. There are still some errors needed to be fixed.

BUG #5689

The main idea is that when execute each operator, it just uploads data to gpu and save all useful information(the promise of pipeline, bindgroup) to the corresponding program. And push this program to an `activePrograms` array. When the user call `tensor.data`, do the actual execution (dispatch the compute shader) and get the data. At last, clear `activePrograms` array.  In this process, we can utilize the `createComputePipelineAsync` API to trigger the parallel compilation. And in `tensor.data`, we can await all pipeline promises resolved, and then use the pipeline and dispatch.


qjia7:
cc @lina128

qjia7:
@lina128 @pyu10055 I have finished the prototype to use parallel shader compilation. And the warmup is >2x faster than before. I'd like to discuss with you how to integrate it and keep consistency in webgl and webgpu.
I roughly look at #5826. I think both ways can reach what we want. And both of them can be implemented in webgl and webgpu. So the main difference is how to expose this feature and use it.
I think the advantage of this PR is that it combines the `Compilation round` and `Warm up upload and download round` in one pass compared with webgl. And make the usage simpler. But it depends on `tensor.data` is called. I don't know whether there is a situation that the data is never read back. 
What's your opinions on this?

cc @axinging @xhcao @haoyunfeix

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -356,6 +358,31 @@ export class WebGPUBackend extends KernelBackend {
       // Data is on the CPU.
       return info.values;
     }
+
+    const parallelCompilation =
+        env().getBool('WEBGPU_PARALLEL_COMPILATION_PASS') && !this.activeTimers;
```
axinging, line 363: During the lifecycle of a model predict, env().getBool('WEBGPU_PARALLEL_COMPILATION_PASS') should be the same.
So we should make sure this is the same as the one get inside runWebGPUProgram

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -770,43 +807,34 @@ export class WebGPUBackend extends KernelBackend {
...
-          this.device, program, pipelineLayout, inputsData, output);
-    });
-
     const shouldTimeProgram = this.activeTimers != null;
+    // Currently only support parallel compilation for non-profiling mode.
+    const parallelCompilation =
+        env().getBool('WEBGPU_PARALLEL_COMPILATION_PASS') && !shouldTimeProgram;
+    if (parallelCompilation) {
+      program.pipeline = this.getAndSaveAsyncPipeline(program.shaderKey, () => {
+        return webgpu_program.compileProgramAsync(
```
axinging, line 823: Should fromPixel also call compileProgramAsync?

vladmandic:
Has this PR stalled? I was looking forward to this improvement a lot.

qjia7:
> Has this PR stalled? I was looking forward to this improvement a lot.

The team is considering providing a graph execution rather than eager execution for graph models underlying. With this support, it's more easier to support parallel compilation and some other optimizations. It's a big change. So we are figuring out all details and possible issues. Hope it can be finished in Q1. @lina128 is working on this.

vladmandic:
Thanks for the update @qjia7

PR #5825, opened 2021-11-08
add check for valid gl context when reading isContextLost

vladmandic:
Safari with WebGL v1 (v1 is still default although WebGL v2 was added in Safari over 3 years ago)  
destroys GL context on `contextLost` event, so reading `isContextLost()` afterwards results in error:  

> exception: null is not an object (evaluating 'gl.isContextLost')

This is a rare scenario that only occurs if `contextLost` event is captured
(if its ignored, error will still happen but its non-fatal)

Chrome, Firefox or Safari with WebGL v2 enabled are not affected

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5825)

### DIFF
--- a/tfjs-backend-webgl/src/canvas_util.ts
+++ b/tfjs-backend-webgl/src/canvas_util.ts
@@ -47,7 +47,7 @@ export function getWebGLContext(webGLVersion: number): WebGLRenderingContext {
     }
   }
   const gl = contexts[webGLVersion];
-  if (!gl || gl === null || gl.isContextLost()) {
+  if (gl == null || gl.isContextLost()) {
     delete contexts[webGLVersion];
     return getWebGLContext(webGLVersion);
   }
--- a/tfjs-backend-webgl/src/canvas_util.ts
+++ b/tfjs-backend-webgl/src/canvas_util.ts
@@ -47,7 +47,7 @@ export function getWebGLContext(webGLVersion: number): WebGLRenderingContext {
     }
   }
   const gl = contexts[webGLVersion];
-  if (gl.isContextLost()) {
+  if (!gl || gl === null || gl.isContextLost()) {
     delete contexts[webGLVersion];
     return getWebGLContext(webGLVersion);
   }


pyu10055 (changes requested):
thanks for your continuous contribution!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5825#-:-Mo-lxZ_2TuL6Umk19zw:b-evqnx8)* status: 0 of 1 approvals obtained (waiting on @vladmandic)

___
*[tfjs-backend-webgl/src/canvas_util.ts, line 50 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5825#-Mo-liBF8uPSo5t97ayu:-Mo-liBG4-D4AoSy6kEe:b-c19r0i) ([raw file](https://github.com/tensorflow/tfjs/blob/1ff07bd7be291d990d6f190b3c414599f653dedb/tfjs-backend-webgl/src/canvas_util.ts#L50)):*
> ```XML
>   }
>   const gl = contexts[webGLVersion];
>   if (!gl || gl === null || gl.isContextLost()) {
> ```

can you help me to understand how the gl become null?
And I think the proper check for null is just
`gl == null`

vladmandic:
> can you help me to understand how the gl become null?

magic of safari with webgl 1.0 :(

if a user app does something like:
```js
const ctx = canvas.getContext('webgl2', webGLattr) as WebGL2RenderingContext;
if (ctx ) canvas.addEventListener('webglcontextlost', (e) => { console.error(e); }
```

with `webgl` 1.0, `ctx ` should not be set since it explicitly requested `webgl2`, but it is  
and once event handler gets triggered, `tfjs` will throw an exception on `if (gl.isContextLost())`  
even if `gl` inside `tfjs` has nothing to do with `ctx` in user app - basically, safari decides to destroy any instance of gl context  

> And I think the proper check for null is just `gl == null`

`!gl` part guards against `undefined` which is not the same as explicitly set to null  
not strictly needed here, but i find it a good practice before checking any object property

pyu10055 (commented):
Got it, thanks for explaining. 

gl == null check for both undefined and null
https://stackoverflow.com/questions/28975896/is-there-a-way-to-check-for-both-null-and-undefined

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5825#-:-Mo-tp-S8k5o3HMuSqb-:b6ihf4e)* status: 0 of 1 approvals obtained (waiting on @vladmandic)

vladmandic:
@pyu10055 github has this marked as `1 change requested`, but i made that change seconds before - so now its in limbo - can you check?

vladmandic (commented):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5825)* status: 0 of 1 approvals obtained (waiting on @pyu10055 and @vladmandic)

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5825)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @vladmandic)

PR #5833, opened 2021-11-11
[webgpu] Get device supported limits info from WebGPU API

haoyunfeix:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5833)


haoyunfeix:
@qjia7 @axinging @xhcao @gyagp PTAL

tfjs-backend-webgpu/src/device_limits.ts
```diff
@@ -15,12 +15,9 @@
  * =============================================================================
  */
 
-export const MAX_COMPUTE_WORKGROUP_INVOCATIONS = 256;
+export let maxComputeWorkgroupsPerDimension: number;
```
qjia7, line 18: keep the initial value? `export let maxComputeWorkgroupsPerDimension = 256;`

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -15,9 +15,7 @@
  * =============================================================================
  */
 import {DataType, util} from '@tensorflow/tfjs-core';
-
-import {MAX_COMPUTE_PER_DIMENSION_DISPATCH_SIZE} from './constants';
-
+import * as device_limits from './device_limits';
```
qjia7, line 18: I prefer you directly use `import maxComputeWorkgroupsPerDimension ...`

tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -184,3 +186,17 @@ export interface WebGPULayout {
   bindGroupLayout: GPUBindGroupLayout;
   pipelineLayout: GPUPipelineLayout;
 }
+
+export type requiredLimitsNames =|'maxTextureDimension1D'|
```
qjia7, line 190: Is it better to move it to `device_limits.ts`? And only keep the names that we use for now. For example, `export type requiredLimitsNames = 'maxComputeWorkgroupsPerDimension';  We can add more in future as needed.

tfjs-backend-webgpu/src/device_limits.ts
haoyunfeix: Done

tfjs-backend-webgpu/src/webgpu_util.ts
haoyunfeix: Done.

tfjs-backend-webgpu/src/webgpu_util.ts
haoyunfeix: Done.

tfjs-backend-webgpu/src/device_limits.ts
```diff
@@ -15,12 +15,11 @@
  * =============================================================================
  */
 
-export const MAX_COMPUTE_WORKGROUP_INVOCATIONS = 256;
+export let maxComputeWorkgroupsPerDimension = 256;
```
qjia7, line 18: I think the default `maxComputeWorkgroupsPerDimension` is 65535 according to the spec https://www.w3.org/TR/webgpu/#limits. You can link this address above the line as the annotation.

qjia7: Please also expose `maxComputeWorkgroupStorageSize` since we want to use it to do more optimizations.

tfjs-backend-webgpu/src/index.ts
```diff
@@ -52,6 +62,7 @@ if (device_util.isBrowser() && isWebGPUSupported()) {
           `it doesn't support synchronously to read data from GPU.`);
     }
     const device: GPUDevice = await adapter.requestDevice(deviceDescriptor);
+    setDeviceLimits(device.limits);
```
qjia7, line 65: Can you do a little refactor here for WebGPUBackend which only accept an optional parameter `device` and move above logic to the constructor of `WebGPUBackend` if no `device` is passed? And do `setDeviceLimits` in `WebGPUBackend` rather than before a new `WebGPUBackend` generated.

tfjs-backend-webgpu/src/device_limits.ts
haoyunfeix: Done.

tfjs-backend-webgpu/src/index.ts
haoyunfeix: `setDeviceLimits` moved to constructor of WebGPUBackend, PTAL

tfjs-backend-webgpu/src/device_limits.ts
```diff
@@ -15,12 +15,13 @@
  * =============================================================================
  */
 
-export const MAX_COMPUTE_WORKGROUP_INVOCATIONS = 256;
+export let maxComputeWorkgroupsPerDimension = 65535;
+export let maxComputeWorkgroupStorageSize = 16352;
+export let maxStorageBufferBindingSize = 134217728;
```
qjia7, line 20: Maybe remove maxStorageBufferBindingSize? It's not possible to exceed this value?

qjia7 (approved):
LGTM with one nit.

tfjs-backend-webgpu/src/device_limits.ts
gyagp: I don't like the idea here to put these as global variables and create another file for them. They can be a new deviceLimits object in WebGPUBackend. In index.ts, can we put all the creation of adapter and device into WebGPUBackend init?
In addition, we can put them in a dict so that we can iterate them to initialize. The initial values here are not necessary.

tfjs-backend-webgpu/src/device_limits.ts
```diff
@@ -15,12 +15,13 @@
...
-
-export const MAX_COMPUTE_WORKGROUP_SIZE_Y = 256;
-
-export const MAX_COMPUTE_WORKGROUP_SIZE_Z = 256;
+export function setDeviceLimits(deviceLimits: GPUSupportedLimits) {
+  maxComputeWorkgroupsPerDimension =
+      deviceLimits.maxComputeWorkgroupsPerDimension;
+  maxComputeWorkgroupStorageSize = deviceLimits.maxComputeWorkgroupStorageSize;
+  maxStorageBufferBindingSize = deviceLimits.maxStorageBufferBindingSize;
+}
```
gyagp, line 27: As mentioned above, I prefer to iterate a dict and initialize.

tfjs-backend-webgpu/src/index.ts
```diff
@@ -37,11 +37,17 @@ if (isWebGPUSupported()) {
...
     const adapter = await navigator.gpu.requestAdapter(gpuDescriptor);
-    let deviceDescriptor: GPUDeviceDescriptor = {};
+    const adapterLimits = adapter.limits;
+    const deviceDescriptor: GPUDeviceDescriptor = {};
     const supportTimeQuery = adapter.features.has('timestamp-query');
+    deviceDescriptor.requiredLimits = {
+      'maxComputeWorkgroupStorageSize':
+          adapterLimits.maxComputeWorkgroupStorageSize,
+      'maxStorageBufferBindingSize': adapterLimits.maxStorageBufferBindingSize,
+    };
```
gyagp, line 47: Only two limits are initialized here,  and we declare 3. So an iteration of dict can avoid such kind of mistake.

tfjs-backend-webgpu/src/device_limits.ts
haoyunfeix: > In addition, we can put them in a dict so that we can iterate them to initialize. The initial values here are not necessary.

Thanks for suggestion! I used adapter.limits to initialize a device, so these limits info could store in a `device` parameter of a backend instance, no need another file anymore, we could call tf.backend().device.limits where ever we want. PTAL.

> In index.ts, can we put all the creation of adapter and device into WebGPUBackend init?

Maybe not, a constructor is not possible to directly return a Promise, so async function `requestAdapter` and `requestDevice` cannot be called from a `new WebGPUBackend()`.

haoyunfeix:
After sync with Jiajia, computeDispatch() divided into two parts, one part as origin is based on layout of output dimensions and workGroupSize, another part moved to backend_webgpu for reshaping dispatch, make sure not to exceed device limits.

@qjia7 @gyagp @xhcao @axinging PTAL

gyagp (approved):
LGTM

PR #5837, opened 2021-11-12
webgpu: Unify Unary and Binary kernels

gyagp:
Currently the unary and binary kernels are implemented in many separated
files, with too much redundenence and inconsistence. This PR is to
simplify the implementation, so that new unary and binary kernels can be
added more easily.


gyagp:
@axinging @xhcao @haoyunfeix @qjia7, PTAL!

qjia7:
cc @pyu10055 @lina128
I haven't started to look at it in detail. But I'd like to hear Ping and Na's high level review comments about this refactor. Not sure whether you ever considered such kind of refactor on other backends.

This refactor definitely saves lots of duplicated code. The only bad thing I see is  we can't just through file name to find the corresponding kernel implementation. But I don't know if there are other considerations to put one kernel as one file.

lina128 (commented):
For WebGL, we do have unification to reduce duplication, basically we use binaryImpl and unaryImpl for the common part. The different kernel files (add.ts, multiple.ts, etc.) are mainly a wrapper. We keep the kernel file names to be equivalent to TF kernel names in ops.pbtxt file, so that (1) we can build custom bundle that only include certain kernels for a TF model therefore significantly reduce the bundle size (this is existing feature) (2) directly call the kernels from tfjs-converter runtime without interop through tfjs-core's op APIs (this is still a future vision). Just something for your consideration for WebGPU design.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5837#-:-MoLavL2AryySY7Ypa8W:b6tz1fr)* status: 0 of 1 approvals obtained

gyagp:
@lina128 Thanks for the insights! 
For 1), I agree to build a custom bundle including only necessary kernels will significantly reduce the bundle size. But specific for unary and binary kernels, the diff among kernels are just shader snippets. All we may save are just these short strings, with a cost whole bundle size if we release a bundle with all the kernels supported. To have unnecessary duplicated code here and there will introduce more inconsistency (I already found some in this PR) and increase the overall maintenance cost. 
For 2), I'm not sure how this will be implemented. But it seems to me we don't necessarily enforce a standalone file a specific kernel impl. 

The intention of this PR is to remove the duplicated code in unary and binary impl so that they can be consistent and easy to maintain. The change will break the symmetry with other backends, but I hope you may be open to allow WebGPU to experiment some kind of innovation in its implementation.

tfjs-backend-webgpu/src/kernels/Unary.ts
```diff
@@ -0,0 +1,249 @@
...
+          setOutputFlat(index, unaryKernel(a));
+        }
+      }
+      `;
+  }
+}
+
+type UnaryConfig = {
+  kernelName: string,
+  implCPU?: ImplCPU,
```
qjia7, line 164: If declare `implCPU?: ImplCPU` to `implCPU? ImplCPU | negImplCPU` and do similar in `kernelInfo`, does it work to merge `neg` to `unaryFunc`?

tfjs-backend-webgpu/src/kernels/Unary.ts
```diff
@@ -0,0 +1,249 @@
...
+    backendName: 'webgpu',
+    kernelFunc: neg,
+  });
+}
+
+export function getShaderUnary(kernelName: string) {
+  return shaders[kernelName];
+}
+
+export const exp = unaryFunc({
```
qjia7, line 239: Why do you still need to export `unaryFunc`?

tfjs-backend-webgpu/src/kernels/Binary.ts
```diff
@@ -0,0 +1,488 @@
...
+  const { inputs, backend } = args;
+  const { x, alpha } = inputs;
+
+  const program = new BinaryProgram(Prelu, x.shape, alpha.shape);
+  return backend.runWebGPUProgram(program, [x, alpha], 'float32');
+}
+
+export function registerKernelsBinary() {
+  for (const kernelName in kernelInfo) {
+    if ([Complex, Prelu].includes(kernelName)) {
```
qjia7, line 442: Why does `Prelu` need to be registered separately?

Issue #5856, opened 2021-11-16
results of `nonMaxSuppression` are incorrect for non-clamped boxes


vladmandic:
some models return non-clamped outputs (example is very popular `blazeface` detector) and in that case, running `nonMaxSuppression` will produce incorrect results *if* box coordinates are bellow 0.

below are example input image and two boxes that are returned as over `iouThreshold`  

![in](https://user-images.githubusercontent.com/57876960/142005818-9d62f380-5fa2-43d1-8bdf-aa64e05dc3ca.jpg)
![out](https://user-images.githubusercontent.com/57876960/142005826-a289e6bf-86f1-421a-940d-f2e82baa6dd4.jpg)

reducing `iouThreshold` to any value does not help as `nonMaxSuppression` calculates it as `0` while its clear that first box is almost completely overlapped with second box

capturing values inside `tfjs-core/src/backends/non_max_suppression_impl.ts:intersectOverUnion`:

```json
{
  "iCoord": { "0": 115.93872833251953, "1": 42.05706787109375, "2": 172.44512939453125, "3": 98.56344604492188 },
  "jCoord": { "0": -25.357688903808594, "1": -12.717769622802734, "2": 97.00607299804688, "3": 109.66532897949219 },
  "areaI": 3192.9720676520374,
  "areaJ": 14975.256338182517,
  "intersectionYmin": 115.93872833251953,
  "intersectionYmax": 97.00607299804688,
  "intersectionXmin": 42.05706787109375,
  "intersectionXmax": 98.56344604492188,
  "intersectionArea": 0
}
```

shows that `intersectionArea` is incorrectly calculated

environment: tfjs 3.11
backend or os are not relevant since `nms` implementation is in `tfjs-core`

vladmandic:
ignore, my fault for incorrect box transformation before nms.

PR #5857, opened 2021-11-16
fix nms calculation for negative coordinates

vladmandic:
quick solution for #5856 although a better overall implementation of `intersectionOverUnion` would be nice


vladmandic:
ignore, incorrect box transofrmation before nms on my side

PR #5917, opened 2021-12-02
[benchmark] Support collecting and showing per-op stats (latency for now) for tflite models in benchmark tool

### LINKED ISSUE #5829: Support TFLite models in e2e benchmark
**System information**
- TensorFlow.js version (you are using): 3.11
- Are you willing to contribute it (Yes/No): Yes

Thanks to the [tfjs-tflite API](https://github.com/tensorflow/tfjs/tree/master/tfjs-tflite), we are able to run TFLite models on the Web, and people may want to evaluate the performance. So I suggest to support running TFLite models in e2e local benchmark by using tfjs-tflite API.

we can support [MobileNet V2](https://tfhub.dev/tensorflow/lite-model/mobilenet_v2_1.0_224/1/default/1) as start point. 

**Will this change the current api? How?**
 No

**Who will benefit with this feature?** 
People who want to benchmark the TFLite models / tfjs-tflite API on the Web.

**Any Other info.**
No

cc/ @pyu10055, @jinjingforever, @huningxin

jinjingforever:
This PR uses the latest tfjs-tflite package that supports getting per-op latency stats, and integrate it with the benchmark tool to show it in the kernel table.

One caveat is that for tflite models, the profiling can only be enabled when loading the model, which is different from tfjs models where we could just wrap the `predict` call in `tf.profile`. To make this work, the tflite model will be loaded first (with enableProfiling=false) during the "total run time" profiling phase, then loaded again (with enableProfiling=true) during the "kernel ops" profiling phase.

The result looks like [this](https://user-images.githubusercontent.com/8752427/144487541-b4b61ccc-5a43-40ed-95de-c48d064c4329.png)

Fixes #5829 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5917)

### DIFF
--- a/e2e/benchmarks/benchmark_util.js
+++ b/e2e/benchmarks/benchmark_util.js
@@ -335,10 +335,12 @@ async function downloadValuesFromTensorContainer(tensorContainer) {
  * @param model An instance of tf.GraphModel or tf.LayersModel for profiling
  *     memory usage in the inference process.
  * @param input The input tensor container for model inference.
+ * @param isTflite Whether a TFLite model is being profiled or not.
  */
-async function profileModelInference(model, input) {
-  const predict = getPredictFnForModel(model, input);
-  return profileInference(predict);
+async function profileModelInference(model, input, isTflite = false) {
+  const predict = isTflite ? () => tfliteModel.predict(input) :
+                             getPredictFnForModel(model, input);
+  return profileInference(predict, isTflite);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5917)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever)

___
*[e2e/benchmarks/local-benchmark/index.html, line 585 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5917#-Mpx3-P68pcENvXWcC00:-Mpx3-P7BWejZ6AQh6Kv:bcryz3j) ([raw file](https://github.com/tensorflow/tfjs/blob/e2a0dc61b5e5b0746e079e8afdec883beb850df5/e2e/benchmarks/local-benchmark/index.html#L585)):*
> ```HTML
>     async function profileMemoryAndKernelTime() {
>       await showMsg('Profile memory and kernels');
>       const start = performance.now();
> ```

should the reloading of tflite model before the start of the timing?

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5917)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

___
*[e2e/benchmarks/local-benchmark/index.html, line 585 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5917#-Mpx3-P68pcENvXWcC00:-Mpx3ulr5ukudk-BBnxm:bhmhk5s) ([raw file](https://github.com/tensorflow/tfjs/blob/e2a0dc61b5e5b0746e079e8afdec883beb850df5/e2e/benchmarks/local-benchmark/index.html#L585)):*

should the reloading of tflite model before the start of the timing?
</blockquote></details>

Nice catch! Done.

PR #5927, opened 2021-12-08
webgpu: fix NaN warning

xhcao:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5927)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/binary_op_util.ts
+++ b/tfjs-backend-webgpu/src/kernels/binary_op_util.ts
@@ -66,16 +66,16 @@ const CHECK_NAN_SNIPPET = `
   if (isNanCustom(b)) { return b; }
   `;
 const CHECK_NAN_SNIPPET_VEC4 = `
-  if (isNaN.r > 0.) {
+  if (isNaN.r) {
     resultTemp.r = uniforms.NAN;
   }
-  if (isNaN.g > 0.) {
+  if (isNaN.g) {
     resultTemp.g = uniforms.NAN;
   }
-  if (isNaN.b > 0.) {
+  if (isNaN.b) {
     resultTemp.b = uniforms.NAN;
   }
-  if (isNaN.a > 0.) {
+  if (isNaN.a) {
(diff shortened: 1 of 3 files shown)


xhcao:

tfjs-backend-webgpu/src/kernels/binary_op_util.ts
```diff
@@ -143,7 +143,7 @@ const POW_VEC4 = `
   if (isExpZero.a) {
     resultTemp.a = 1.0;
   }
-  let isNaN = vec4<f32>(a < vec4<f32>(0.0)) * vec4<f32>(floor(b) < b);
+  let isNaN = a < vec4<f32>(0.0) & floor(b) < b;
```
qjia7, line 146: Why the check of `isNaN` is different with L161?

gyagp:
Can we change the name to isnan instead of isNanCustom?

tfjs-backend-webgpu/src/kernels/binary_op_util.ts
xhcao: `a` and `b` are the different parameters for `pow` and `minumin/maxmium`, I think these two operators have different behaviors for NaN.

xhcao (commented):
@gyagp Do you mean `isNan` or `isnan`? If `isNan`, it is a built-in function, it will be a compiling error when running tfjs on old chrome, or adding this built-in function back.
I think `isNanCustom` is good name.

xhcao:
@gyagp It seems that WGSL do NOT support macro declaration, so cannot define an macro as GLSL.  Is it OK that use the name `isNanCustom` currently?

gyagp:
> @gyagp It seems that WGSL do NOT support macro declaration, so cannot define an macro as GLSL. Is it OK that use the name `isNanCustom` currently?

ok for now.

gyagp (approved):
LGTM

gyagp:
@lina128 @pyu10055 FYI

PR #5928, opened 2021-12-08
webgpu: Use mapAsync instead of writeBuffer for uploading

qjia7:
PERF
BUG: #5689

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5928)

### DIFF
--- a/tfjs-backend-webgpu/src/buffer_manager.ts
+++ b/tfjs-backend-webgpu/src/buffer_manager.ts
@@ -88,9 +88,15 @@ export class BufferManager {
 
   releaseUploadBuffer(
       buffer: GPUBuffer, byteSize: number, usage: GPUBufferUsageFlags) {
-    buffer.mapAsync(GPUMapMode.WRITE).then(() => {
-      this.releaseBuffer(buffer, byteSize, usage);
-    });
+    buffer.mapAsync(GPUMapMode.WRITE)
+        .then(
+            () => {
+              this.releaseBuffer(buffer, byteSize, usage);
+            },
+            (err) => {
+              // Do nothing;
+              console.log('jiajia:' + err);
+            })
   }
 
   getNumUsedBuffers(): number {
--- a/tfjs-backend-webgpu/src/buffer_manager.ts
+++ b/tfjs-backend-webgpu/src/buffer_manager.ts
@@ -26,7 +26,12 @@ export class BufferManager {
 
   constructor(private device: GPUDevice) {}
 
-  acquireBuffer(byteSize: number, usage: GPUBufferUsageFlags) {
+  acquireUploadBuffer(byteSize: number, usage: GPUBufferUsageFlags) {
+    return this.acquireBuffer(byteSize, usage, true);
+  }
+
+  acquireBuffer(
+      byteSize: number, usage: GPUBufferUsageFlags, mappedAtCreation = false) {
     const key = getBufferKey(byteSize, usage);
     if (!this.freeBuffers.has(key)) {
       this.freeBuffers.set(key, []);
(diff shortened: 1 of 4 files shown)


tfjs-backend-webgpu/src/buffer_manager.ts
```diff
@@ -80,6 +86,13 @@ export class BufferManager {
     this.numBytesUsed -= byteSize;
   }
 
+  releaseUploadBuffer(
+      buffer: GPUBuffer, byteSize: number, usage: GPUBufferUsageFlags) {
+    buffer.mapAsync(GPUMapMode.WRITE).then(() => {
```
qjia7, line 91: This PR can solve the bad performance of `writeBuffer` when uploading large data set. See the [bug](https://github.com/tensorflow/tfjs/issues/5689#issuecomment-958641027). With this PR, webgpu can get better performance than webgl tested with https://vladmandic.github.io/tfjs-utils/src/latency-issue.html. 
However, it brings `mapAsync(GPUMapMode.WRITE).then` in the sync function, which results that `Buffer is destroyed before the mapping is resolved` error for some cases in the `yarn test` which is within a `tidy` function. It seems that with this change, it will break the tidy function. One possible solution is when the buffer is destroyed, just mark it needed to be destroyed and destroy it when the mapping is resolved. But I am not sure if it's an acceptable change or do you have better suggestion? I'd hear your opinions before I do more changes. @kainino0x @pyu10055 @lina128

kainino0x: Hm, I didn't even realize that was an error. The spec isn't very well-defined here yet but I think it should be valid to destroy a buffer that's in the "mapping pending" state. (The current spec sort of implicitly allows it.)

Filed https://crbug.com/dawn/1234

kainino0x: In the mean time, I think your suggested solution is good.

> One possible solution is when the buffer is destroyed, just mark it needed to be destroyed and destroy it when the mapping is resolved. But I am not sure if it's an acceptable change or do you have better suggestion? I'd hear your opinions before I do more changes.

kainino0x (approved):

tfjs-backend-webgpu/src/buffer_manager.ts
Kangz: The promise is rejected with `Buffer is destroyed before the mapping is resolved` if you destroy a buffer while the mapping is pending. However since the buffer has already been passed to `tidy`, doesn't that mean that you already don't need it anymore? In that case you could just ignore the promise rejection.

qjia7: @Kangz Thanks for your suggestion. That's the simplest change. It works!

pyu10055 (approved):

qjia7:
Correct some tests expectation due to explicit staging buffer management. Meanwhile, remove some tests for `numBytesInGPU` due to that some `releaseBuffer` calling is in  a promise `then` method. It's hard to get the exact number when multiple async tests are running together.

@pyu10055 @kainino0x Please take another look, thank.

tfjs-backend-webgpu/src/buffer_manager.ts
```diff
@@ -98,7 +116,7 @@ export class BufferManager {
   }
 
   dispose() {
-    if (this.freeBuffers == null && this.usedBuffers == null) {
+    if (this.freeBuffers.size === 0 && this.usedBuffers.size === 0) {
       return;
     }
```
kainino0x, line 121: This early-return probably isn't needed anymore because it seems it was just there to type-guard the rest of the function against `null`, but now the rest of the function should behave just fine without the early-return.

kainino0x (approved):
lgtm

tfjs-backend-webgpu/src/buffer_manager.ts
qjia7: Done

PR #5930, opened 2021-12-09
[webgpu] Remove dtype

axinging:


axinging:

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -562,8 +562,13 @@ export class WebGPUBackend extends KernelBackend {
     info.bufferInfo.buffer = this.acquireBuffer(info.bufferInfo.byteSize);
 
     if (info.values) {
+      let gpuValues = info.values;
```
xhcao, line 565: Could we remove `gpuValues` here?  Could we directly use `info.values = Float32Array.from(info.values);` below?

axinging: For most case, it is Float32Array, so no need to do this for Float32Array.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -562,8 +562,13 @@ export class WebGPUBackend extends KernelBackend {
...
+      if (info.values instanceof Int32Array) {
+        gpuValues = Float32Array.from(info.values);
```
qjia7, line 567: I notice that there will be 2 extra copying when the data is `Int32Array`. One is uploading an int32 tensor, the other is downloading an int32 tensor. Can these two copying somehow be avoided?
If not, I am thinking another way that we keep the original data and upload them directly to GPU. And in shader, we treat them all as float to do the calculation. Is that better?

PR #5953, opened 2021-12-17
Add an API to keep data on GPU: dataToGPU

lina128:
This PR adds a convenient API to read data out as a texture, so that it can directly be rendered on canvas or used by downstream shaders.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953)

### DIFF
--- a/tfjs-backend-webgl/src/backend_webgl.ts
+++ b/tfjs-backend-webgl/src/backend_webgl.ts
@@ -112,9 +112,6 @@ export class MathBackendWebGL extends KernelBackend {
   }
   // Maps data ids that have a pending read operation, to list of subscribers.
   private pendingRead = new WeakMap<DataId, Array<(arr: TypedArray) => void>>();
-  // Maps data ids that have a pending read operation, to list of subscribers.
-  // private pendingReadTexture =
-  //     new WeakMap<DataId, Array<(arr: TextureMetadata) => void>>();
   // List of data ids that are scheduled for disposal, but are waiting on a
   // pending read operation.
   private pendingDisposal = new WeakSet<DataId>();
@@ -393,9 +390,9 @@ export class MathBackendWebGL extends KernelBackend {
   /**
    * Read tensor to a new texture that is densely packed for ease of use.
    * @param dataId The source tensor.
-   * @param customTexShape Optional. If set, will use the user defined texture
+   * @param options
+   *     customTexShape: Optional. If set, will use the user defined texture
    *     shape to create the texture.
-
    */
   readToGPU(dataId: DataId, options?: DataToGPUWebGLOption): GPUResource {
     const texData = this.texData.get(dataId);
@@ -403,7 +400,7 @@ export class MathBackendWebGL extends KernelBackend {
 
     if (dtype === 'complex64') {
       throw new Error('Does not support reading texture for complex64 dtype.');
-    };
+    }
 
     // The presence of `slice` indicates this tensor is a shallow slice of a
     // different tensor, and is using that original tensor's texture. Run
@@ -436,8 +433,8 @@ export class MathBackendWebGL extends KernelBackend {
 
     return {
       dataId: tmpDownloadTarget.dataId,
-          ...this.texData.get(tmpDownloadTarget.dataId).texture
-    }
+      ...this.texData.get(tmpDownloadTarget.dataId).texture
+    };
   }
 
   bufferSync<R extends Rank>(t: TensorInfo): TensorBuffer<R> {
@@ -806,7 +803,8 @@ export class MathBackendWebGL extends KernelBackend {
     return {dataId: output.dataId, shape: afterShape, dtype: output.dtype};
   }
 
-  public decode(dataId: DataId, customTexShape?: [number, number]): TensorInfo {
+  private decode(dataId: DataId, customTexShape?: [number, number]):
+      TensorInfo {
     const texData = this.texData.get(dataId);
     const {isPacked, shape, dtype} = texData;
     const shapeAs3D =
--- a/tfjs-backend-webgl/src/backend_webgl.ts
+++ b/tfjs-backend-webgl/src/backend_webgl.ts
@@ -433,7 +433,6 @@ export class MathBackendWebGL extends KernelBackend {
 
     return {
       dataId: tmpDownloadTarget.dataId,
-      dtype,
       ...this.texData.get(tmpDownloadTarget.dataId).texture
     };
   }
@@ -471,7 +470,7 @@ export class MathBackendWebGL extends KernelBackend {
     }
   }
 
-  public getValuesFromTexture(dataId: DataId): Float32Array {
+  private getValuesFromTexture(dataId: DataId): Float32Array {
     const {shape, dtype, isPacked} = this.texData.get(dataId);
     const size = util.sizeFromShape(shape);
     if (env().getBool('WEBGL_DOWNLOAD_FLOAT_ENABLED')) {
--- a/tfjs-backend-webgl/src/backend_webgl.ts
+++ b/tfjs-backend-webgl/src/backend_webgl.ts
@@ -433,6 +433,7 @@ export class MathBackendWebGL extends KernelBackend {
 
     return {
       dataId: tmpDownloadTarget.dataId,
+      dtype,
       ...this.texData.get(tmpDownloadTarget.dataId).texture
     };
   }
@@ -470,7 +471,7 @@ export class MathBackendWebGL extends KernelBackend {
     }
   }
 
-  private getValuesFromTexture(dataId: DataId): Float32Array {
+  public getValuesFromTexture(dataId: DataId): Float32Array {
     const {shape, dtype, isPacked} = this.texData.get(dataId);
     const size = util.sizeFromShape(shape);
     if (env().getBool('WEBGL_DOWNLOAD_FLOAT_ENABLED')) {
--- a/tfjs-backend-webgl/src/backend_webgl.ts
+++ b/tfjs-backend-webgl/src/backend_webgl.ts
@@ -431,11 +431,11 @@ export class MathBackendWebGL extends KernelBackend {
     const tmpTarget = this.decode(dataId, options.customTexShape);
 
     // Make engine track this tensor, so that we can dispose it later.
-    engine().makeTensorFromDataId(
+    const tensorRef = engine().makeTensorFromDataId(
         tmpTarget.dataId, tmpTarget.shape, tmpTarget.dtype);
 
     const tmpData = this.texData.get(tmpTarget.dataId);
-    return {dataId: tmpTarget.dataId, ...tmpData.texture};
+    return {tensorRef, ...tmpData.texture};
   }
 
   bufferSync<R extends Rank>(t: TensorInfo): TensorBuffer<R> {
--- a/tfjs-backend-webgl/src/backend_webgl.ts
+++ b/tfjs-backend-webgl/src/backend_webgl.ts
@@ -394,7 +394,7 @@ export class MathBackendWebGL extends KernelBackend {
    *     customTexShape: Optional. If set, will use the user defined texture
    *     shape to create the texture.
    */
-  readToGPU(dataId: DataId, options?: DataToGPUWebGLOption): GPUResource {
+  readToGPU(dataId: DataId, options: DataToGPUWebGLOption = {}): GPUResource {
     const texData = this.texData.get(dataId);
     const {values, shape, slice, dtype, isPacked, texture} = texData;
 
@@ -807,6 +807,15 @@ export class MathBackendWebGL extends KernelBackend {
       TensorInfo {
     const texData = this.texData.get(dataId);
     const {isPacked, shape, dtype} = texData;
+    if (customTexShape != null) {
+      const size = util.sizeFromShape(shape);
+      const texSize = customTexShape[0] * customTexShape[1] * 4;
+      util.assert(
+          size <= texSize,
+          () => 'customTexShape is too small. ' +
+              'Row by column by 4 should be equal or larger than the ' +
+              'size of the tensor data.');
+    }
     const shapeAs3D =
         webgl_util.getShapeAs3D(shape) as [number, number, number];
     let program;
--- a/tfjs-backend-webgl/src/backend_webgl.ts
+++ b/tfjs-backend-webgl/src/backend_webgl.ts
@@ -112,6 +112,9 @@ export class MathBackendWebGL extends KernelBackend {
   }
   // Maps data ids that have a pending read operation, to list of subscribers.
   private pendingRead = new WeakMap<DataId, Array<(arr: TypedArray) => void>>();
+  // Maps data ids that have a pending read operation, to list of subscribers.
+  // private pendingReadTexture =
+  //     new WeakMap<DataId, Array<(arr: TextureMetadata) => void>>();
   // List of data ids that are scheduled for disposal, but are waiting on a
   // pending read operation.
   private pendingDisposal = new WeakSet<DataId>();
@@ -386,6 +389,74 @@ export class MathBackendWebGL extends KernelBackend {
     }
     return dTypeVals;
   }
+
+  // async readToTexture(dataId: DataId): Promise<TextureMetadata> {
+  //   if (this.pendingReadTexture.has(dataId)) {
+  //     const subscribers = this.pendingReadTexture.get(dataId);
+  //     return new Promise<TextureMetadata>(resolve =>
+  //     subscribers.push(resolve));
+  //   }
+  //   const texData = this.texData.get(dataId);
+  //   const {values, shape, slice, dtype, isPacked} = texData;
+
+  //   if (dtype !== 'complex64') {
+  //     throw new Error('Does not support reading texture for complex64
+  //     dtype.');
+  //   };
+
+  //   // The presence of `slice` indicates this tensor is a shallow slice of a
+  //   // different tensor, and is using that original tensor's texture. Run
+  //   // `clone` in order to copy that texture and read from it.
+  //   if (slice != null) {
+  //     let program;
+  //     if (isPacked) {
+  //       program = new UnaryOpPackedProgram(shape, unary_op.CLONE);
+  //     } else {
+  //       program = new UnaryOpProgram(shape, unary_op.CLONE);
+  //     }
+  //     const res =
+  //         this.runWebGLProgram(program, [{dataId, shape, dtype}], dtype);
+  //     const data = this.readToTexture(res.dataId);
+  //     this.disposeIntermediateTensorInfo(res);
+  //     return data;
+  //   }
+
+  //   let tmpDownloadTarget: TensorInfo;
+
+  //   // Decode the texture so that it is stored densely (using four channels).
+  //   tmpDownloadTarget = this.decode(dataId);
+  //   const tmpData = this.texData.get(tmpDownloadTarget.dataId);
+
+  //   this.pendingReadTexture.set(dataId, []);
+
+  //   if (dtype !== 'complex64') {
+  //     // Create a fence and wait for it to resolve.
+  //     await this.gpgpu.createAndWaitForFence();
+  //   }
+
+  //   if (tmpDownloadTarget != null) {
+  //     this.disposeIntermediateTensorInfo(tmpDownloadTarget);
+  //   }
+  //   if (buffer != null) {
+  //     const gl = this.gpgpu.gl;
+  //     webgl_util.callAndCheck(gl, () => gl.deleteBuffer(buffer));
+  //   }
+  //   const dTypeVals = this.convertAndCacheOnCPU(dataId, vals);
+
+  //   const subscribers = this.pendingRead.get(dataId);
+  //   this.pendingRead.delete(dataId);
+
+  //   // Notify all pending reads.
+  //   subscribers.forEach(resolve => resolve(dTypeVals));
+  //   if (this.pendingDisposal.has(dataId)) {
+  //     this.pendingDisposal.delete(dataId);
+  //     if (this.disposeData(dataId)) {
+  //       engine().removeDataId(dataId, this);
+  //     }
+  //     this.pendingDeletes--;
+  //   }
+  //   return dTypeVals;
+  // }
 
   bufferSync<R extends Rank>(t: TensorInfo): TensorBuffer<R> {
     const data = this.readSync(t.dataId);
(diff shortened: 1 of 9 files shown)


danwexler:
Very excited to see this new function. Will this be in v13?

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953)* status: 0 of 1 approvals obtained (waiting on @lina128)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 809 at r3](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MrG5b6X7YjFF7GAzLHr:-MrG5b6X7YjFF7GAzLHs:bqv55rp) ([raw file](https://github.com/tensorflow/tfjs/blob/773a3ba42f79276daebaf286fb7edd16c18b96c1/tfjs-backend-webgl/src/backend_webgl.ts#L809)):*
> ```XML
>           size <= texSize,
>           () => 'customTexShape is too small. ' +
>               'Row by column by 4 should be equal or larger than the ' +
> ```

Row * Column * 4



_Code quote:_
```XML
Row by column by 4
```
___
*[tfjs-backend-webgl/src/backend_webgl_test.ts, line 752 at r3](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MrG6VxF7FmQ2NYcpoOM:-MrG6VxF7FmQ2NYcpoON:b-xlbyea) ([raw file](https://github.com/tensorflow/tfjs/blob/773a3ba42f79276daebaf286fb7edd16c18b96c1/tfjs-backend-webgl/src/backend_webgl_test.ts#L752)):*
> ```XML
> });
> 
> // TODO(lina128): Debug issue in CI. Code runs fine in Safari WebGL1 in local.
> ```

is this failing due to texture is float16? Since the difference is only the ENVS, we delete this section when we figure out CI issue for WebGL1.
___
*[tfjs-core/src/tensor.ts, line 369 at r3](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MrG7DOs67HL8ZjkWx9h:-MrG7DOs67HL8ZjkWx9i:b-jixedt) ([raw file](https://github.com/tensorflow/tfjs/blob/773a3ba42f79276daebaf286fb7edd16c18b96c1/tfjs-core/src/tensor.ts#L369)):*
> ```XML
> 
>   /**
>    * Synchronously copy the tensor's data to a new GPU resource. Comparing to
> ```

This is async copy in terms of JS, since it is runs a webGL program, the JS return does not guarantee completion of shader execution. 
But for WebGL this is sync, since the GPU queue is synchronized.
___
*[tfjs-backend-webgl/src/gpgpu_util.ts, line 227 at r2](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MrG82nq9wTWE9elMjUM:-MrG82nq9wTWE9elMjUN:bkm7k5n) ([raw file](https://github.com/tensorflow/tfjs/blob/fc28b9534f1b0c10c967730033873eca15f3ae74/tfjs-backend-webgl/src/gpgpu_util.ts#L227)):*
> ```XML
>               gl.TEXTURE_2D, 0, 0, 0, pixels.width, pixels.height, gl.RGBA,
>               gl.UNSIGNED_BYTE, (pixels as PixelData).data));
>       gl.flush();
> ```

`gl.flush()` I think this line should be removed, I guess I added by accident, this should not be needed.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 809 at r3](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MrG5b6X7YjFF7GAzLHr:-MrOfeTM2YO4f1nl0FIY:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/773a3ba42f79276daebaf286fb7edd16c18b96c1/tfjs-backend-webgl/src/backend_webgl.ts#L809)):*

Row * Column * 4
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/backend_webgl_test.ts, line 752 at r3](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MrG6VxF7FmQ2NYcpoOM:-MrOffZOCF0nIxp1dzKI:bt2vr1b) ([raw file](https://github.com/tensorflow/tfjs/blob/773a3ba42f79276daebaf286fb7edd16c18b96c1/tfjs-backend-webgl/src/backend_webgl_test.ts#L752)):*

is this failing due to texture is float16? Since the difference is only the ENVS, we delete this section when we figure out CI issue for WebGL1.
</blockquote></details>

Fixed.
___
*[tfjs-core/src/tensor.ts, line 369 at r3](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MrG7DOs67HL8ZjkWx9h:-MrOfr7wDERbFerxNogC:bohak22) ([raw file](https://github.com/tensorflow/tfjs/blob/773a3ba42f79276daebaf286fb7edd16c18b96c1/tfjs-core/src/tensor.ts#L369)):*

This is async copy in terms of JS, since it is runs a webGL program, the JS return does not guarantee completion of shader execution. 
But for WebGL this is sync, since the GPU queue is synchronized.
</blockquote></details>

Removed the confusing wording.
___
*[tfjs-backend-webgl/src/gpgpu_util.ts, line 227 at r2](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MrG82nq9wTWE9elMjUM:-MrOg2Vg3Z9HASQ_lnlF:b2oyid1) ([raw file](https://github.com/tensorflow/tfjs/blob/fc28b9534f1b0c10c967730033873eca15f3ae74/tfjs-backend-webgl/src/gpgpu_util.ts#L227)):*

`gl.flush()` I think this line should be removed, I guess I added by accident, this should not be needed.
</blockquote></details>

Will fix in a separate PR.

lina128:
> Very excited to see this new function. Will this be in v13?

Yes

lina128:
> Very excited to see this new function. Will this be in v13?

@danwexler , yes.

tfjs-backend-webgl/src/backend_webgl.ts
```diff
@@ -380,6 +380,53 @@ export class MathBackendWebGL extends KernelBackend {
...
+        program = new UnaryOpProgram(shape, unary_op.CLONE);
+      }
+      const res =
+          this.runWebGLProgram(program, [{dataId, shape, dtype}], dtype);
+      const gpuResouorce = this.readToGPU(res, options);
+      this.disposeIntermediateTensorInfo(res);
+      return gpuResouorce;
+    }
+
+    if (values != null && texture == null) {
```
qjia7, line 415: If `values` is not null, can we just upload the data to a texture and return the texture?

lina128 (commented):
Hi Jiajia, yes, GPUResource is still limited in current tfjs context. We suggest them to pass the canvas to use when creating webgl backend. Here's the other PR that allows user to pass a canvas: https://github.com/tensorflow/tfjs/pull/5983

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953#-:-MsfMMsU7V3i-XBKJnT8:bf7jcij)* status: 0 of 1 approvals obtained (waiting on @pyu10055 and @qjia7)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 415 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-Mr4kKIqKX6cHxR6oNDx-r5-415:-MsfN-l75kkxxt4yhe2X:b-u4ajb3) ([raw file](https://github.com/tensorflow/tfjs/blob/6363510006e79110edbe36bda20f69c5e99f00ac/tfjs-backend-webgl/src/backend_webgl.ts#L415)):*

If `values` is not null, can we just upload the data to a texture and return the texture?
</blockquote></details>

I have thought about this, it's hard to say which way is more convenient to users, so we'll see. Normally, if values are not null, it's either a small amount of data or the values haven't been uploaded yet, in both cases, users may have their preferred way of using the data, not necessarily using TFJS formatted data on GPU, so I prefer to leave it to user for now. However, if there's such request to provide TFJS formatted texture in the future, we can definitely support it.

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953)* status: 0 of 1 approvals obtained (waiting on @lina128, @pyu10055, and @qjia7)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 415 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-Mr4kKIqKX6cHxR6oNDx-r5-415:-MsfRCBu107ANljo3mxI:b-27euwa) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-backend-webgl/src/backend_webgl.ts#L415)):*

I have thought about this, it's hard to say which way is more convenient to users, so we'll see. Normally, if values are not null, it's either a small amount of data or the values haven't been uploaded yet, in both cases, users may have their preferred way of using the data, not necessarily using TFJS formatted data on GPU, so I prefer to leave it to user for now. However, if there's such request to provide TFJS formatted texture in the future, we can definitely support it.
</blockquote></details>

might be good to throw error instead of returning null. given that we don't support reading on CPU data to GPU. 
___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 423 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MsfRpAs3PoBL127eJyB:-MsfRpAs3PoBL127eJyC:b-rs81y1) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-backend-webgl/src/backend_webgl.ts#L423)):*
> ```XML
> 
>     // Make engine track this tensor, so that we can dispose it later.
>     engine().makeTensorFromDataId(
> ```

how user can track this tensor and dispose later?
___
*[tfjs-backend-webgl/src/backend_webgl_test.ts, line 682 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MsfS7o16AbsqcJJWlKm:-MsfS7o16AbsqcJJWlKn:bbj4r8g) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-backend-webgl/src/backend_webgl_test.ts#L682)):*
> ```XML
> });
> 
> describeWithFlags('keeping data on gpu ', WEBGL2_ENVS, () => {
> ```

can you add tests for memory leak check?
___
*[tfjs-core/src/tensor.ts, line 392 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MsfSNxz3hkWJCO6Iszh:-MsfSNxz3hkWJCO6Iszi:b5uqcm3) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-core/src/tensor.ts#L392)):*
> ```XML
>    * @doc {heading: 'Tensors', subheading: 'Classes'}
>    */
>   dataToGPU(options?: DataToGPUOptions): GPUResource {
> ```

should the GPUResource also contains reference to the tensor? so call can dispose it later?

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953)* status: 0 of 1 approvals obtained (waiting on @lina128, @pyu10055, and @qjia7)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 423 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MsfRpAs3PoBL127eJyB:-MsfZQj-3tkvsLgJ2cFq:bv4zlq1) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-backend-webgl/src/backend_webgl.ts#L423)):*

how user can track this tensor and dispose later?
</blockquote></details>

The GPUResource has dataId, which is the only thing needed when disposing tensor. User can dispose it the same way as tensor, for typescript user, they may need to cast GPUResource to tensor first. So something like: `tf.dispose(gpuResource as {} as Tensor)`.
___
*[tfjs-core/src/tensor.ts, line 392 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MsfSNxz3hkWJCO6Iszh:-Msf_0Aq1-GAD7g5chN-:b-l9ld1p) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-core/src/tensor.ts#L392)):*

should the GPUResource also contains reference to the tensor? so call can dispose it later? 
</blockquote></details>

It has reference to the tensor, dataId.

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953)* status: 0 of 1 approvals obtained (waiting on @lina128, @pyu10055, and @qjia7)

___
*[tfjs-core/src/tensor.ts, line 392 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MsfSNxz3hkWJCO6Iszh:-MsgK4fgD0QqESHWixfY:bowu2as) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-core/src/tensor.ts#L392)):*

It has reference to the tensor, dataId.
</blockquote></details>

can tidy automatically dispose GPUResource?

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953)* status: 0 of 1 approvals obtained (waiting on @pyu10055 and @qjia7)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 415 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-Mr4kKIqKX6cHxR6oNDx-r5-415:-MsgVQlT8M3w-MC5n_Oq:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-backend-webgl/src/backend_webgl.ts#L415)):*

might be good to throw error instead of returning null. given that we don't support reading on CPU data to GPU. 
</blockquote></details>

Done.
___
*[tfjs-backend-webgl/src/backend_webgl_test.ts, line 682 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MsfS7o16AbsqcJJWlKm:-MsgXMlm5ox88ewC5F98:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-backend-webgl/src/backend_webgl_test.ts#L682)):*

can you add tests for memory leak check?
</blockquote></details>

Done.
___
*[tfjs-core/src/tensor.ts, line 392 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MsfSNxz3hkWJCO6Iszh:-MsgYUENDzourV2RgGyn:bx9x2nr) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-core/src/tensor.ts#L392)):*

can tidy automatically dispose GPUResource?
</blockquote></details>

It should behave the same as Tensor, added a test to verify.

tfjs-core/src/tensor.ts
```diff
@@ -353,6 +365,35 @@ export class Tensor<R extends Rank = Rank> {
...
+   *        dataId: Object,
+   *        texture: WebGLTexture,
+   *        texShape: [number, number] // [height, width]
+   *     }
+   *     Remember to dispose the GPUResource after it is used by
+   *     `tf.dispose(GPUResource as {} as Tensor)`.
+   *
+   * @doc {heading: 'Tensors', subheading: 'Classes'}
+   */
+  dataToGPU(options?: DataToGPUOptions): GPUResource {
```
qjia7, line 392: Just curious, why support `customTexShape`? Is there any special requirement for this?

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953)* status: 0 of 1 approvals obtained (waiting on @pyu10055 and @qjia7)

___
*[tfjs-backend-webgl/src/backend_webgl.ts, line 423 at r5](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MsfRpAs3PoBL127eJyB:-MsiDVIIBfUOpWaPZJDJ:b-xaj32r) ([raw file](https://github.com/tensorflow/tfjs/blob/5b7727a232d7ce912f66970e1332751a0e87608c/tfjs-backend-webgl/src/backend_webgl.ts#L423)):*

The GPUResource has dataId, which is the only thing needed when disposing tensor. User can dispose it the same way as tensor, for typescript user, they may need to cast GPUResource to tensor first. So something like: `tf.dispose(gpuResource as {} as Tensor)`.
</blockquote></details>

Hi Ping, you are right, it's better to track this tensor instead of just use dataId, I added the memory test. Previously, the tensor cannot be disposed properly because GPUResource is not recognized as instanceOf Tensor in one place in engine. After the latest change, it should work. Also see the memory test.
___
*[tfjs-core/src/tensor.ts, line 392 at r6](https://reviewable.io/reviews/tensorflow/tfjs/5953#-Mr4kKIqKX6cHxR6oNE3-r6-392:-MsiE47i4oXuAG5zaNOV:b-rcrz1) ([raw file](https://github.com/tensorflow/tfjs/blob/2b2a5ae9468f0ea7509214291b2779d3a05578f4/tfjs-core/src/tensor.ts#L392)):*

Just curious, why support `customTexShape`? Is there any special requirement for this?
</blockquote></details>

The main use case is if user set the texShape to be the same as the canvas shape, then they can directly render the result on canvas. If they don't specify the texShape, we try to make a squarish texture.

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @pyu10055, and @qjia7)

___
*[tfjs-backend-webgl/src/backend_webgl_test.ts, line 802 at r7](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MskpvsCCau1ekJyWC4q:-MskpvsDDB0qiPiAKHWa:b2kvmkm) ([raw file](https://github.com/tensorflow/tfjs/blob/edc71e97569c4b5f7db84e9adef7c3a174b4e058/tfjs-backend-webgl/src/backend_webgl_test.ts#L802)):*
> ```XML
>       const b = tf.add(a, 0);
>       b.dataToGPU();
>       return b
> ```

missing `;`

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5953)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055 and @qjia7)

___
*[tfjs-backend-webgl/src/backend_webgl_test.ts, line 802 at r7](https://reviewable.io/reviews/tensorflow/tfjs/5953#-MskpvsCCau1ekJyWC4q:-MsksLAr9jxHx0rVvFtg:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/edc71e97569c4b5f7db84e9adef7c3a174b4e058/tfjs-backend-webgl/src/backend_webgl_test.ts#L802)):*

missing `;`
</blockquote></details>

Done.

xhcao:
Hi, @lina128 , you had replied Jiajia's as below,
"The main use case is if user set the texShape to be the same as the canvas shape, then they can directly render the result on canvas. If they don't specify the texShape, we try to make a squarish texture."

I have a confusion here, If canvas size is 4*4, and texture size is 2*2, we could directly render this texture to the canvas, and hardware will use texture LINEAR or NEAREST filter sampling to render the content.
And if texture size 2*2, and its content is [1, 2, 3, 4], and call dataToGPU(..., [4,4]) to create a new texture, whose content is [1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3, 4, 4, 4, 4] (Assume it is correct), these content is expected data for users? Are there rules for result when enlarging the texture size.

lina128:
Hi @xhcao , I think for WebGL, if the texture is larger than the data size, the unused space will be 0. We try to store data in dense format, so the content will be [1, 2, 3, 4, 0, ..., 0].

xhcao:
Hi, @lina128 , I had tested unit case on webgl backend,
it('uses user defined texShape [3, 3].', () => {
    const data = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12];
    const a = tf.tensor(data, [1, 3, 4]);
    const b = tf.add(a, 0);
    const texShape = [3, 3] as [number, number];
    const res = b.dataToGPU({customTexShape: texShape});
    expectArraysEqual(res.texShape, texShape);

    const webGLBackend = tf.backend() as MathBackendWebGL;
    const buffer = webGLBackend.gpgpu.createBufferFromTexture(
        res.texture, res.texShape[0], res.texShape[1]);
    const vals = webGLBackend.gpgpu.downloadFloat32MatrixFromBuffer(buffer, 36);
    console.log(vals);
    expectArraysEqual(vals, data);
  });
The result is shown as below, it does not fill unused space with 0, and the code is related in https://github.com/tensorflow/tfjs/blob/5c3a89c3b1c360631c6ec71d770815d1c157009d/tfjs-backend-webgl/src/decode_matrix_gpu.ts#L50
Float32Array(36) [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, buffer: ArrayBuffer(144), byteLength: 144, byteOffset: 0, length: 36, Symbol(Symbol.toStringTag): 'Float32Array']

I am confusion why we need to change the shape. We could not ensure the filling content is correct and is expected by Users. What are rules to define the filling behaviors? I think it is incorrect to fill with 0. For example, if original texture is 2x2, and then call dataToGPU with shape 4x4 to create the new texture, most of the new texture is transparent, I think it is not the expected result.

lina128:
Hi @xhcao, you only take the first 12 elements, which is correct, right? This is the same approach we use for reading data to cpu in WebGL, we create the right size of the buffer and call readPixels: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/gpgpu_util.ts#L271

Data beyond the size doesn't matter whether it is 0 or some other values, we simply ignore them.

xhcao:
@lina128 , Does it mean the data beyond the size of the new created texture must be written again before it could be rendered to canvas? There may be a wrong understanding of dataToGPU usage for me. Could you simply give an usage example of dataToGPU with custom size here. Thank you.

lina128:
Hi @xhcao, I see what you mean. There should be another shader to format how it wants to render if the size doesn't match the canvas size. Can you describe a real use case? It will help me prepare the example. Thanks.

xhcao:
Hi, @lina128, actually, I did not know the scenarios which could gain the performance improvement by using custom size option, so I asked you why we needed the custom size option. 
Now, I may image a scenario as below.
[1]a tensor (shape: 2x4) calling dataToGPU with custom size (4x4) to create a new bigger texture  A(4x4), the first half content of A is equal to the tensor. Then we use another shader to fill the other half content of A, render the texture A to the canvas at last.

We could also use the below steps to implement the above scenario. 
[2]A tensor (shape: 2x4) calling dataToGPU without custom size to create a texture  A(2x4), the content of A is equal to the tensor, then we use another shader to create the texture B(2x4) and fill the content, individually render the texture A and texture B to the first half and second half of the canvas at last.

I am not sure whether [1] is better than [2], but I sort of agree to add the option to dataToGPU api if there is no side effect.

lina128:
Hi @xhcao, thank you for providing the imagined scenarios. The actual scenario of using customSize is more like the other way round, the customSize will fit exactly the data, whereas our default (the more squarish texture) will waste some space.

PR #5954, opened 2021-12-17
[webgpu] refactor setup_test

haoyunfeix:
1. Replace include with startsWith
2. Sort ascending
3. Add tests for exp, batchnorm, log, neg, sigmoid

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5954)

### DIFF
--- a/tfjs-backend-webgpu/src/setup_test.ts
+++ b/tfjs-backend-webgpu/src/setup_test.ts
@@ -26,20 +26,39 @@ import './backend_webgpu_test_registry';
 import {parseTestEnvFromKarmaFlags, setTestEnvs, setupTestFilters, TEST_ENVS, TestFilter} from '@tensorflow/tfjs-core/dist/jasmine_util';
 
 const TEST_FILTERS: TestFilter[] = [
+  // skip test cases include 5D, 6D, gradients webgpu
+  {
+    include: '5D',
+    excludes: [
+      'webgpu '
+    ],
+  },
+  {
+    include: '6D',
+    excludes: [
+      'webgpu '
+    ],
+  },
+  {
+    include: 'gradients webgpu',
+    excludes: [
+      'webgpu '
+    ],
+  },
+
+  // skip specific test cases for supported kernels
   {
     startsWith: 'abs ',
     excludes: [
       'complex64',  // Kernel 'ComplexAbs' not registered.
-      '5D',         // Rank 5 is not yet implemented.
-      '6D',         // Rank 5 is not yet implemented.
-      'gradient',   // zerosLike not yet implemented.
+      'gradient',   // Step kernel not yet implemented.
     ]
   },
   {
     startsWith: 'avgPool ',
     excludes: [
       'gradient',   // Not yet implemented.
-      'avgPool3d',  // Not yet implemented.
+      //'avgPool3d',  // Not yet implemented.
     ]
   },
   {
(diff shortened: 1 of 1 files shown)


haoyunfeix:
Filtered case number: 1821/4128 -> 2515/4128

gyagp:
In current implementation, when we add a new kernel, we need to add a new filter. It's hard to know how many are left. 
Can we change this to exclude rather than include. Then when we add a new kernel, we will remove the relative exclude.

haoyunfeix:
Good point, I added unsupported kernel to the filter in the bottom(https://github.com/tensorflow/tfjs/pull/5954/files#diff-ec4048880b6607b71cc2a79b41df43ac2901ddd196877d4e18c06a14d78ada60R287-R366).
Next time when we add kernels, we could remove relative name in this exclude list simply, but if we want to skip some specific cases(not ready or will support in future), the contributor should also write a filter with `startsWith` and `exclude` like before and put it above the last filter.

gyagp:
LGTM. I assume we still have 2515 cases to run after the new change.

gyagp (approved):

haoyunfeix:
The latest number is 2510/4129, compared to current UT result 1821/4128, we increased 689 cases coverage.
The reason is that I added a filter with `include: webgpu`(pass ALL_ENV to describeWithFlags will add webgpu to tfjs-core cases), in addition to kernel ops test it also include other tests like optimizer, PlatformBrowser ... which we didn't touch before.

tfjs-backend-webgpu/src/setup_test.ts
```diff
@@ -27,65 +27,30 @@ import {parseTestEnvFromKarmaFlags, setTestEnvs, setupTestFilters, TEST_ENVS, Te
...
-    include: 'fused conv2d',
+    startsWith: 'abs ',
     excludes: [
-      'gradient x=[2,3,3,1] f=[2,2,1,1] s=1 p=0',  // conv2dDerInput not yet
-                                                   // implemented
-      'backProp',   // Conv2DBackpropFilter not yet
-                    // implemented
-      'leakyrelu',  // Not yet implemented
+      'complex64',  // Kernel 'ComplexAbs' not registered.
+      '5D',         // Rank 5 is not yet implemented.
```
gyagp, line 33: Can we remove all the '5D' and '6D' here given they're in excludes list?

haoyunfeix: Should not in current filter logic (https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/jasmine_util.ts#L141-L160)
When a filter returns true, other filters behind it will not be executed.
That means the general EXCLUDE filter has no chance to filter `5D` cases, STARTSWITH filters have to handle this itself.

gyagp: We may add below in the top of list so that we can handle all the 5D and 6D cases in the same place.
{
include: '5D',
excludes: [' webgpu ']
}

haoyunfeix: Does your point is to make this as the first filter
{
include: '5D',
excludes: [' webgpu ']
}
but it means to select case which has 5D but isn't related to webgpu, this kind of cases are not our target.

haoyunfeix:
Cleaned 5D, 6D and unnecessary gradients in specific filters. @gyagp PTAL

PR #5970, opened 2021-12-29
[webgpu] Convert bool tensor to Uint8Array

haoyunfeix:
BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5970)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -397,6 +397,10 @@ export class WebGPUBackend extends KernelBackend {
     this.maybeReleaseBuffer(dataId);
 
     info.values = data;
+    // converted bool to Uint8Array.
+    if (info.dtype === 'bool') {
+      return Uint8Array.from(data);
+    }
     return info.values;
   }
 
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -264,11 +264,6 @@ export class WebGPUBackend extends KernelBackend {
     const byteSize =
         util.sizeFromShape(shape) * webgpu_util.GPUBytesPerElement(dtype);
 
-    // bool is stored in Uint8Array, converted it to Int32Array.
-    if (dtype === 'bool' && values instanceof Uint8Array) {
-      values = Int32Array.from(values);
-    }
-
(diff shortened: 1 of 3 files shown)


haoyunfeix:

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -397,6 +397,10 @@ export class WebGPUBackend extends KernelBackend {
     this.maybeReleaseBuffer(dataId);
 
     info.values = data;
+    // converted bool to Uint8Array.
+    if (info.dtype === 'bool') {
+      return Uint8Array.from(data);
```
qjia7, line 402: Please do it in `ArrayBufferToTypedArray` and keep this file unchanged. And I am curious why an extra `Uint8Array.from` is needed since in `ArrayBufferToTypedArray`, you already return a Uint8Array and the data is assigned one by one there.

haoyunfeix: `ArrayBufferToTypedArray` is not reached when tensorBufferInfo.values is not null, `read` function will return values through `convertAndCacheOnCPU`. For bool tensor, it stores as Int32Array in tensorMap(converted by `write` function), so convert it back to Uint8Array here.

qjia7: If the values are always in CPU, maybe we shouldn't change it to int32Array in `write` https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/src/backend_webgpu.ts#L269. And lazily do it in `uploadToGPU`. Then we don't need to do special process here.

haoyunfeix: > `If the values are always in CPU, maybe we shouldn't change it to int32Array in write https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/src/backend_webgpu.ts#L269. And lazily do it in uploadToGPU. Then we don't need to do special process here.`

Agree. I removed `Uint8Array.from` from `write`, as uploadToGPU calls TypedArray.prototype.set() to store values, the two arrays may share the same underlying ArrayBuffer.

haoyunfeix:
@qjia7 PTAL

qjia7 (approved):
LGTM, thanks.

cc @lina128 @pyu10055

PR #5984, opened 2022-01-05
[webgl] GatherV2 fill out of range values with zero

### LINKED ISSUE #5869: tf.gather is needlessly slow, CPU-intensive
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): Yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Windows 11 Home
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device:
- TensorFlow.js installed from (npm or script link): [npm](https://unpkg.com/@tensorflow/tfjs@3.11.0/dist/tf.js)
- TensorFlow.js version (use command below): 3.11.0
- Browser version: Google Chrome 96.0.4664.45
- Tensorflow.js Converter Version: ???

`tf.gather` seems disproportionately slow, taking  33x the wall time of `tf.cumsum` with comparably sized input and output. When viewing performance data, runtime seems dominated by uploading and downloading textures, with much of the CPU time spent looping over indices to merely check that they are within range.

It seems also that `tf.gatherND` does not suffer similar performance problems. So replacing ` tf.gather(x,ix)` with `tf.gatherND(x,ix.expandDims(1))` seems an appropriate workaround when gathering on the first axis.

I expect `tf.gather` to perform gathering on the GPU without synchronous operations. Any index checks (if necessary) should be done on GPU.

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Document</title>
</head>
<body>
    <script crossorigin="anonymous" src=https://unpkg.com/@tensorflow/tfjs@3.11.0/dist/tf.js></script>
    <script>
        window.addEventListener("load",()=>{
            const x = tf.range(0, 10)
            const ix = tf.randomUniform([50000],0,10).cast('int32')
            for (let i=0; i<1000; ++i){
                tf.tidy(()=>{ tf.gather(x,ix);
                tf.cumsum(ix); })
            }
        })
    </script>
</body>
</html>
```

See screenshot and attached Chrome devtools profile.

image
[Profile-20211118T172039.zip](https://github.com/tensorflow/tfjs/files/7566680/Profile-20211118T172039.zip)

mattsoulanille:
GatherV2 currently throws an error when accessing an index out of range, but that requires a copy back to the CPU. This change avoids that copy by filling the out-of-range values with zero, which is more in line with the [C api](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/gather-v2). For debugging purposes, the original behavior can be re-enabled via the DEBUG flag.

Fixes #5869

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)


lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: 0 of 1 approvals obtained (waiting on @mattsoulanille and @pyu10055)

___
*[tfjs-backend-webgl/src/kernels/GatherV2.ts, line 38 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-Msfj-E81QA2eHOWCJDN:-Msfj-E81QA2eHOWCJDO:b-dt6mi6) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/kernels/GatherV2.ts#L38)):*
> ```XML
>   if (env().get('DEBUG')) {
>     // In debug mode, throw error when any index is out of bound.
>     // Otherwise, just fill out of bounds with zeroes.
> ```

Is this the behavior of TF?



_Code quote:_
```XML
Otherwise, just fill out of bounds with zeroes.
```

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: 0 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[tfjs-backend-webgl/src/kernels/GatherV2.ts, line 38 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-Msfj-E81QA2eHOWCJDN:-MsfjasEA0JHVZEKcSSN:b-v4dmac) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/kernels/GatherV2.ts#L38)):*
For filling with zeroes, yes. For throwing errors anyway in debug mode, I'm not sure.

From the C++ [GatherV2 docs](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/gather-v2):

> Note that on CPU, if an out of bound index is found, an error is returned. On GPU, if an out of bound index is found, a 0 is stored in the corresponding output value.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: 0 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @pyu10055)

___
*[tfjs-backend-webgl/src/gather_gpu.ts, line 39 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgSRrK5bB9ceKPjK5o:-MsgSRrK5bB9ceKPjK5p:b-bkzme2) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/gather_gpu.ts#L39)):*
> ```XML
>         ${dtype} resRC = getOutputCoords();
>         float index = getIndices(resRC.x, resRC.z);
>         float inBounds = (index >= 0.0) && (index <= float(${aShape[2]})) ? 1.0 : 0.0;
> ```

Should this be '<' or '<=${aShape[2]}-1'?



_Code quote:_
```XML
<=
```
___
*[tfjs-backend-webgl/src/setup_test.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgTMGf-IIJLHTGS7LG:-MsgTMGf-IIJLHTGS7LH:bfatfz9) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/setup_test.ts#L33)):*
> ```XML
>   const toExclude =
>       ['isBrowser: false', 'dilation gradient',
>        'throws when index is out of bound'];
> ```

Why do you need to exclude this test?



_Code quote:_
```XML
throws when index is out of bound
```
___
*[tfjs-backend-webgl/src/webgl_ops_test.ts, line 891 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgTa7b5aOYAmAnC7S9:-MsgTa7b5aOYAmAnC7SA:bxao1y4) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/webgl_ops_test.ts#L891)):*
> ```XML
> });
> 
> describeWithFlags('gather debug', WEBGL_ENVS, async () => {
> ```

Do you need async?



_Code quote:_
```XML
 async
```
___
*[tfjs-backend-webgl/src/webgl_ops_test.ts, line 894 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgT6pR4m0UJbQTJSWn:-MsgT6pR4m0UJbQTJSWo:b-aivg49) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/webgl_ops_test.ts#L894)):*
> ```XML
>   it('throws when index is out of bound if in debug mode', async () => {
>     const t = tf.tensor2d([1, 11, 2, 22], [2, 2]);
>     tf.enableDebugMode();
> ```

Do you need to set DEBUG flag to what it was after this test?



_Code quote:_
```XML
tf.enableDebugMode();
```

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: 0 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[tfjs-backend-webgl/src/gather_gpu.ts, line 39 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgSRrK5bB9ceKPjK5o:-MsgcNRs2aUFOWjWaSmu:bily6mc) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/gather_gpu.ts#L39)):*

Should this be '<' or '<=${aShape[2]}-1'?
</blockquote></details>

Yes. Thanks!
___
*[tfjs-backend-webgl/src/setup_test.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgTMGf-IIJLHTGS7LG:-MsgcRtw7X7la9zPNJPv:b-l3r0vn) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/setup_test.ts#L33)):*

Why do you need to exclude this test?
</blockquote></details>

The new behavior is to return zero instead of throwing an error. If we don't exclude that test, it fails.

However, the new test I wrote for debug mode is also matched by this expression. Thanks for catching that!

Now that the debug mode test is running, I've discovered that we probably can't support this debug mode check. The reason is that in debug mode, [computations are asynchronously checked for errors](https://github.com/tensorflow/tfjs/blob/38f8462fe642011ff1b7bcbb52e018f3451be58b/tfjs-core/src/profiler.ts#L62-L63). This causes an uncaught promise rejection when I throw the 'index out of bounds' error, which I can't catch in my test. 

Do you think I should just remove the debug mode check?
___
*[tfjs-backend-webgl/src/webgl_ops_test.ts, line 891 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgTa7b5aOYAmAnC7S9:-MsgeA7Y2R3p-HVLJzVh:b-knkl1c) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/webgl_ops_test.ts#L891)):*

Do you need async?
</blockquote></details>

No. Thanks!
___
*[tfjs-backend-webgl/src/webgl_ops_test.ts, line 894 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgT6pR4m0UJbQTJSWn:-MsgdmGCCl-jpdwYAjjy:b-nqgk5v) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/webgl_ops_test.ts#L894)):*

Do you need to set DEBUG flag to what it was after this test?
</blockquote></details>

I don't think so, because describeWithFlags calls `ENGINE.reset()` in beforeAll, which clears the flags, and this is the only test in the `describeWithFlags`.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: 0 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @pyu10055)

___
*[tfjs-backend-webgl/src/setup_test.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgTMGf-IIJLHTGS7LG:-MshFTqJ1fIY1CxNz4de:bgcr46o) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/setup_test.ts#L33)):*

The new behavior is to return zero instead of throwing an error. If we don't exclude that test, it fails.

However, the new test I wrote for debug mode is also matched by this expression. Thanks for catching that!

Now that the debug mode test is running, I've discovered that we probably can't support this debug mode check. The reason is that in debug mode, [computations are asynchronously checked for errors](https://github.com/tensorflow/tfjs/blob/38f8462fe642011ff1b7bcbb52e018f3451be58b/tfjs-core/src/profiler.ts#L62-L63). This causes an uncaught promise rejection when I throw the 'index out of bounds' error, which I can't catch in my test. 

Do you think I should just remove the debug mode check?
</blockquote></details>

Why not remove the original out of bound test that's no longer valid anymore, the new debug mode test you added in this PR should work, if it doesn't, there's something wrong. The async error check is used in profile kernels, but we are not using profile function here.



_Code quote:_
```XML
throws when index is out of bound
```

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: 0 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[tfjs-backend-webgl/src/setup_test.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgTMGf-IIJLHTGS7LG:-MskLxSa5rSooCNJ6pT7:bzdo0up) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/setup_test.ts#L33)):*

Why not remove the original out of bound test that's no longer valid anymore, the new debug mode test you added in this PR should work, if it doesn't, there's something wrong. The async error check is used in profile kernels, but we are not using profile function here.
</blockquote></details>

We still need the original out of bound test for the CPU backend, which should throw an error when an index is out of bounds (same as the c++ api). 

In debug mode without profiling set, it seems to profile anyway. [Here's the logic that controls whether to profile or not](https://github.com/tensorflow/tfjs/blob/38f8462fe642011ff1b7bcbb52e018f3451be58b/tfjs-core/src/engine.ts#L702-L705). I've set a breakpoint, and it seems to hit this when calling GatherV2. It looks like it profiles the code if `this.state.profiling` is true or the `DEBUG` flag is set.

I also tried unsetting 'CHECK_COMPUTATION_FOR_ERRORS', since it uses the `output.data()` promise, but that did not fix it.

Now, I'm pretty sure it's caused by the [timer](https://github.com/tensorflow/tfjs/blob/38f8462fe642011ff1b7bcbb52e018f3451be58b/tfjs-core/src/profiler.ts#L48-L49) we use to profile the kernel. [We don't check for a rejected promise in the timer](https://github.com/tensorflow/tfjs/blob/38f8462fe642011ff1b7bcbb52e018f3451be58b/tfjs-core/src/profiler.ts#L72). I think this is actually a bug in the webgl timer, which synchronously executes the op but asynchronously times it. Instead of throwing an error if the op fails, it rejects the promise it returns. Making it throw an error instead of rejecting a promise when the op fails fixes this.

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: 0 of 1 approvals obtained (waiting on @lina128 and @mattsoulanille)

___
*[tfjs-backend-webgl/src/gather_gpu.ts, line 38 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-Mskd-9wDjdHfny_JOB4:-Mskd-9wDjdHfny_JOB5:b-96yybt) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/gather_gpu.ts#L38)):*
> ```XML
>       void main() {
>         ${dtype} resRC = getOutputCoords();
>         float index = getIndices(resRC.x, resRC.z);
> ```

should let the index be `int index = int(getIndices(resRC.x, resRC.z);`
and make the condition below be
`index >= 0 && index < ${aShape[2]}`

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: 0 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[tfjs-backend-webgl/src/gather_gpu.ts, line 38 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-Mskd-9wDjdHfny_JOB4:-MsklEgP7SJkuSxbO0tA:bt69pe6) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/gather_gpu.ts#L38)):*

should let the index be `int index = int(getIndices(resRC.x, resRC.z);`
and make the condition below be
`index >= 0 && index < ${aShape[2]}`
</blockquote></details>

That's better. Thanks!

pyu10055 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: 0 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[tfjs-backend-webgl/src/setup_test.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgTMGf-IIJLHTGS7LG:-MskoL287usveUI-fIBR:bovzz7d) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/setup_test.ts#L33)):*

We still need the original out of bound test for the CPU backend, which should throw an error when an index is out of bounds (same as the c++ api). 

In debug mode without profiling set, it seems to profile anyway. [Here's the logic that controls whether to profile or not](https://github.com/tensorflow/tfjs/blob/38f8462fe642011ff1b7bcbb52e018f3451be58b/tfjs-core/src/engine.ts#L702-L705). I've set a breakpoint, and it seems to hit this when calling GatherV2. It looks like it profiles the code if `this.state.profiling` is true or the `DEBUG` flag is set.

I also tried unsetting 'CHECK_COMPUTATION_FOR_ERRORS', since it uses the `output.data()` promise, but that did not fix it.

Now, I'm pretty sure it's caused by the [timer](https://github.com/tensorflow/tfjs/blob/38f8462fe642011ff1b7bcbb52e018f3451be58b/tfjs-core/src/profiler.ts#L48-L49) we use to profile the kernel. [We don't check for a rejected promise in the timer](https://github.com/tensorflow/tfjs/blob/38f8462fe642011ff1b7bcbb52e018f3451be58b/tfjs-core/src/profiler.ts#L72). I think this is actually a bug in the webgl timer, which synchronously executes the op but asynchronously times it. Instead of throwing an error if the op fails, it rejects the promise it returns. Making it throw an error instead of rejecting a promise when the op fails fixes this.
</blockquote></details>

given that your new tests are only for WebGL envs, can you make the out of bound error test for CPU only, then it won't need to be excluded?

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: 0 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[tfjs-backend-webgl/src/setup_test.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs/5984#-MsgTMGf-IIJLHTGS7LG:-MskrkuR8axgUfB2kACg:b-cq3ud) ([raw file](https://github.com/tensorflow/tfjs/blob/9719fdef17daf479ec14f724726f5d2e9cacd57b/tfjs-backend-webgl/src/setup_test.ts#L33)):*

given that your new tests are only for WebGL envs, can you make the out of bound error test for CPU only, then it won't need to be excluded?
</blockquote></details>

I'm not sure we want to restrict the out of bounds test to only run in cpu. [We also throw if out of bounds in wasm](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/kernels/GatherV2.ts#L56), which we should keep testing (although maybe there are similar performance implications for wasm and we should just fill with zeroes, but that would be a separate PR).

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5984)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

lina128:
Agree with Ping, better to use predicate to confine test conditions than using exclude, if possible. But that's fine to have it like this if it's too hard to have a predicate that specify all the envs running in cpu (cpu, wasm, node).

Thank you Matt, LGTM.

PR #5987, opened 2022-01-06
[webgpu] Resolve timestamp when read

axinging:
In the profile mode, timestamp will be resolved per op, the gpu burden is relatively heavy.  This PR introduced a lightweight way to get timestamp, resolved when read. This feature is guarded behind flag DEBUG_QUERY.

In addition to the resolve method difference,  the time returned in profile mode is gpu execution time (milliseconds).  The time returned by DEBUG_QUERY is the start time and end time of gpu execution (nanoseconds).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5987)


axinging:

PR #5994, opened 2022-01-07
[cpu] GatherV2 return zero if out of bounds when forwarding from webgl

mattsoulanille:
WebGL sometimes forwards to the CPU implementation of kernels. In #5984, the bounds check for GatherV2 was removed in favor of filling the out of bounds areas with zeroes, but the CPU implementation did not do this and filled with NaNs instead. This PR makes the CPU implementation of GatherV2 fill with zeroes. The CPU backend still checks for out of bounds before running its GatherV2 implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5994)


pyu10055 (approved):
thanks

PR #6005, opened 2022-01-12
Add 'leave_as_require' option to tfjs_bundle

mattsoulanille:
`@rollup/plugin-commonjs` rewrites 'require' calls to 'import' statements, but this is not always desirable. This commit adds the 'leave_as_require' option to 'tfjs_bundle', which allows specifying a list of modules to leave as 'require' calls.

This change does not disable `@rollup/plugin-commonjs` for cjs bundles even though it might not be required in some cases. The reason it is still enabled is because we still want to bundle commonjs modules for our node bundles. Without the plugin, they're left as `require()` calls, but with the plugin, their contents are included in the bundle.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6005)

### DIFF
--- a/tfjs-core/BUILD.bazel
+++ b/tfjs-core/BUILD.bazel
@@ -32,6 +32,9 @@ tfjs_bundle(
     entry_point = "//tfjs-core/src:index.ts",
     external = [
         "crypto",
+    ],
+    leave_as_require = [
+        "crypto",
         "node-fetch",
         "util",
     ],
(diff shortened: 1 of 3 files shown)


pyu10055 (approved):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6005)* status: :shipit: complete! 1 of 1 approvals obtained

pyu10055 (approved):

PR #6006, opened 2022-01-13
[webgpu] Resolve timestamp when read

axinging:
In the profile mode, timestamp will be resolved per op, the gpu burden is relatively heavy. This PR introduced a lightweight way to get timestamp, resolve timestamp when read. This feature is guarded behind flag TRACING.

In addition to the resolve method difference, the time returned in profile mode is gpu execution time (milliseconds). The time returned by TRACING is the start time and end time of gpu execution (nanoseconds).

For e2e, add tracing=true into the url will turn on tracing. For user application, set flag TRACING to true before model predict.


axinging:
@qjia7 @haoyunfeix @xhcao @gyagp PTAL

LwaziV (approved):

PR #6028, opened 2022-01-18
Use conv2dByMatMul when input and filter have the same width/height

qjia7:
This PR 1) moves the common part of Conv2D/FusedConv2D to conv2DImpl to reuse the code; 2) adds another situation that input width and height are equal to filter width and height, and the pad type is valid, we can also go to the `conv2dByMatMul` path;

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6028)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/Conv2D_impl.ts
+++ b/tfjs-backend-webgpu/src/kernels/Conv2D_impl.ts
@@ -183,7 +183,7 @@ function conv2dWithIm2Col({
   const matMulProgram = new MatMulPackedProgram(
       a3dShape, [1, numCols, convInfo.outChannels],
       env().get('WEBGPU_MATMUL_WORK_PER_THREAD') as number, transposeA,
-      transposeB);
+      transposeB, bias, activation, preluActivationWeights);
   const dimAOuter = a3dShape[1];
   const dimInner = a3dShape[2];
   const dimBOuter = convInfo.outChannels;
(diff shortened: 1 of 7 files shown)


qjia7:
With this change, a FusedConv2d with input0: 4D[1,2,2,32], input1: 4D[2,2,32,1] will call the matmul kernel(0.03ms) instead of conv2d kernel(0.12ms). 

@gyagp @xhcao @axinging @haoyunfeix Please take a look, thanks.

tfjs-backend-webgpu/src/kernels/Conv2D_impl.ts
```diff
@@ -183,3 +207,86 @@ export function conv2dWithIm2Col({
...
+      convInfo,
+      backend,
+      bias,
+      activation,
+      preluActivationWeights,
+      leakyreluAlpha
+    });
+  }
+
+  if (env().getBool('WEBGPU_CONV_SEPARATE_IM2COL_SHADER') && x.shape[0] === 1) {
```
xhcao, line 245: In original code, this flag is only used by Conv2D. Could fusedConv2D also use conv2dWithIm2Col here?

qjia7: Yes, it also works for `fusedConv2d`.

xhcao:
LGTM, thank you.

gyagp (approved):
LGTM

qjia7:
cc @lina128 @pyu10055

PR #6031, opened 2022-01-18
Handle IS_BROWSER and IS_NODE both being true

### LINKED ISSUE #6030: Environments where IS_BROWSER and IS_NODE are both true overwrite platform and log warnings
**System information**
- Have I written custom code: Yes
- OS Platform and Distribution: Linux Ubuntu 20.04
- TensorFlow.js installed from: npm
- TensorFlow.js version: 3.11.0
- Browser version: N/A for this example, but I've seen it happen in Electron 16
- Tensorflow.js Converter Version: 3.11.0

When `import * as tf from '@tensorflow/tfjs';` is run in either a Jest test for a Vue component, or in the Vue component being tested, the message "[Platform browser has already been set. Overwriting the platform with [object Object].](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/environment.ts#L61)" is `console.warn`'ed out. 

This is due to this Jest/Vue environment being both a browser and a node environment, thus setting both the [IS_BROWSER](https://github.com/tensorflow/tfjs/blob/1d0c63f3bedb74f553382fbcefb355da9c951c35/tfjs-core/src/flags.ts#L39) and [IS_NODE](https://github.com/tensorflow/tfjs/blob/1d0c63f3bedb74f553382fbcefb355da9c951c35/tfjs-core/src/flags.ts#L42) flags to true, and each in turn calling `setPlatform` in their respective platform's modules. This causes the console to get spammed with warnings while running tests.

This is likely not isolated to the combo of Jest and Vue. I've seen this error in Electron/Vue environments as well, but it's much harder to make a reproducible example for that.

The warning isn't logged out, since there is nothing technically wrong with this use case.

I see a couple options to fix this:
- Handling the scenario where `IS_BROWSER` and `IS_NODE` are both true such that `setPlatform` is only called once.
- A way to disable this logging. Since it occurs on import, other methods of post-import configuration won't fix this, and the current IS_TEST and PROD env checks don't seem to be working here.

I would be happy to make a contribution to fix this, but I don't know enough about the side effects of messing with platform environments well enough to do so without some conversation.

1. Install vue-cli with `npm install -g @vue/cli`
2. Create a new Vue project with `vue create tftest`
3. Add tensorflow with `npm i @tensorflow/tfjs --save`
4. Add unit tests with `vue add unit-jest` and `npm i @vue/test-utils --save-dev`
5. Add the line `import * as tf from '@tensorflow/tfjs'` to `src/components/HelloWorld.vue` at line 34
6. Run the example test with `npm run test:unit`

The result:

image

**Other info / logs**

This issue is similar #5349, but the particular warning brought up here was never solved.

austintoot40:
Fixes #6030

In certain environments, IS_BROWSER and IS_NODE can both be true, which causes `setPlatform` to be called twice along with a warning message as soon as `import * as tf from '@tensorflow/tfjs'` is called. This PR addresses this by only setting the `node` platform if IS_BROWSER is explicitly false to avoid this scenario.

Alternatively, the change here could be done in the browser platform instead of the node platform. I'm not aware of any reasons to prefer one over the other, so I did it this way.

Finally, the warning that was logged out used to say `Overwriting the platform with [object Object]`, which is not particularly helpful. It will now log out the platform name, which should make debugging a legitimate platform issue easier.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6031)

### DIFF
--- a/tfjs-core/src/environment.ts
+++ b/tfjs-core/src/environment.ts
@@ -60,7 +60,7 @@ export class Environment {
       if (!(env().getBool('IS_TEST') || env().getBool('PROD'))) {
         console.warn(
             `Platform ${this.platformName} has already been set. ` +
-            `Overwriting the platform with ${platform}.`);
+            `Overwriting the platform with ${platformName}.`);
       }
     }
     this.platformName = platformName;
--- a/tfjs-core/src/platforms/platform_node.ts
+++ b/tfjs-core/src/platforms/platform_node.ts
@@ -81,6 +81,6 @@ export class PlatformNode implements Platform {
   }
 }
 
-if (env().get('IS_NODE')) {
+if (env().get('IS_NODE') && !env().get('IS_BROWSER')) {
   env().setPlatform('node', new PlatformNode());
 }


pyu10055 (approved):
thank you

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6031#-:-MtjqKRNEMJ17Bc-RN5P:b-1cyu7w)* status: :shipit: complete! 1 of 1 approvals obtained

PR #6033, opened 2022-01-19
DRAFT: measure the esplased time of reading kernel output

xhcao:


xhcao:
Please DO NOT review it, it is just a draft and used to measure the time of reading kernel output on wasm backend, which executes between two kernels.

PR #6105, opened 2022-02-07
Fix Bazel linking of dependencies

mattsoulanille:
Create a new Bazel-only entrypoint for each Bazel package (in e.g. tfjs-core/tfjs-core_lib/index.ts) that allows the rules_nodejs linker to properly create symlinks in the root node_modules directory to Bazel's output in dist/bin/.

Before this change, packages were linked like this:
```
@tensorflow/tfjs-core/
  index.js
  foo.js
  bar.js
  dist -> . (a symlink to this directory)
```
This format confused bundlers. After this change, packages no longer have a circular symlink:
```
@tensorflow/tfjs-core/
  index.js (Bazel's index.js entrypoint. Just exports * from dist/index)
  dist/
    index.js (tfjs-core's entrypoint)
    foo.js
    bar.js
```
For more details on how this is done and why it is necessary, see `tfjs-core/tfjs-core_lib/index.ts`.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6105)

### DIFF
--- a/scripts/cloudbuild_e2e_expected.yml
+++ b/scripts/cloudbuild_e2e_expected.yml
@@ -46,22 +46,13 @@ steps:
     secretEnv:
       - BROWSERSTACK_KEY
   - name: gcr.io/learnjs-174218/release
-    dir: link-package-core
-    entrypoint: yarn
-    id: yarn-link-package-core
-    args:
-      - install
-    waitFor:
-      - bazel-tests
-  - name: gcr.io/learnjs-174218/release
     dir: link-package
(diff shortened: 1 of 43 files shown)


mattsoulanille:
This is failing because it sets the module name of `src/index.d.ts` files incorrectly to , e.g. `/// <amd-module name="@tensorflow/tfjs-core/dist" />` instead of `/// <amd-module name="@tensorflow/tfjs-core" />`. This causes an import error in Karma.

PR #6163, opened 2022-02-17
e2e: add an option to set nums for profiling kernels

xhcao:
The profiling data is not trusted if only running the model one time,
add an option to set nums of profiling kernels, and get the mean time
of each kernel.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6163)


xhcao:

qjia7:
I think the change makes benchmarks more flexible. Would like @jinjingforever @lina128 to review this change.
BTW, I can't see much difference when I adjust numProfiles on my machine.

xhcao:
Generally, when running a model in e2e, we will run 1 time for warm up, and run 50 times to compute the mean time for Subsequent average, then run 1 time for profile time and kernel time.
The below chart is the inference times for Subsequent average,
image
We could see that the beginning runs is not steady, although we had warm up the model. Run the AutoML Image on wasm backend with 1 profile times, the data is shown as below,

Subsequent average (50 runs) 8.4 ms
Best time 8.0 ms
Profile 12.8 ms
Conv2D 3.8 ms
DepthwiseConv2dNative 2.7 ms
Add 2.4 ms
Multiply 0.7 ms
Relu 0.4 ms
CropAndResize 0.2 ms
Cast 0.1 ms
Sub 0.1 ms
Mean 0.1 ms
AddN 0.1 ms
… …

We could see: 1. profile time is much larger than Subsequent average (50 runs); 2. The sum of all kernels' time is also larger than Subsequent average (50 runs).
So I think it is reasonable that run profile many times and compute the mean time of each kernel. If we also runs 50 profiling times. The data is shown as below.
Subsequent average (50 runs) 8.3 ms
Best time 7.9 ms
Profile time 8.8 ms
Conv2D 3.0 ms
DepthwiseConv2dNative 1.9 ms
Add 1.2 ms
Multiply 0.6 ms
Relu 0.3 ms
CropAndResize 0.1 ms
Cast 0.1 ms
Sub 0.1 ms
Mean 0.1 ms
AddN 0.0 ms
… …

e2e/benchmarks/local-benchmark/index.html
```diff
@@ -102,6 +103,9 @@ <h2>TensorFlow.js Model Benchmark</h2>
       if (params.has('run')) {
         runTimes = Number(params.get('run'));
       }
+      if (params.has('profile')) {
+        runTimes = Number(params.get('profile'));
```
gyagp, line 107: Should be profileTimes. I don't think we need to get parameters from url twice. Please clean up this part of code.

gyagp:
I tried this patch on my Tigerlake machine, and more profiling times did make the results much closer to the subsequent average time. My data for AutoML Image: 
Subsequent average: 6.0ms
Profile time with 1 profile: 11.9ms (url args: ?task=performance&benchmark=AutoML%20Image&backend=wasm&warmup=50&run=50&profile=1&localBuild=webgl,webgpu,core&CHECK_COMPUTATION_FOR_ERRORS=false)
Profile time with 50 profiles: 7.2ms (url args: ?task=performance&benchmark=AutoML%20Image&backend=wasm&warmup=50&run=50&profile=50&localBuild=webgl,webgpu,core&CHECK_COMPUTATION_FOR_ERRORS=false)

lina128 (approved):
Thank you!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6163#-:-MwYg0lT0NKGNhqKx8xf:b-9755fn)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @xhcao)

xhcao:
> nit: add brief comment for this new parameter above. Same for `profileInference` below

Had added the comment of the new parameter. Thank you.

xhcao:
@jinjingforever @lina128 Could you help to merge this patch? Thank you.

PR #6212, opened 2022-03-10
[tfjs-layers] Add a Transformer layer (Multihead Attention)

ierezell:
As the title says, this aim to add a transformer layer (based on multi-head attention) in the built-in tfjs-layers.

Note: This is only one layer, the full architecture with positional embeddings, Encoder-Decoder would be built on this.

- [X] Refactor code to use tensor built-in operation to be compatible (like tf.add(a,b) => a.add(b) as other layers because add is not imported)
- [X] Used the einsum mechanism as asked in https://github.com/tensorflow/tfjs/pull/4955 
- [X] Imported missing op (like matmul or sqrt from tfjs-core)
- [ ] Need the tf.layers.LayerNormalization and tf.layers.Dropout which I guess could be imported from the core (as the matMul).

Please tell me how to use the core functionalities to be fully compatible with the other layers. I tried it myself as a custom tfjs layer and it seems to be working (so I wanted to make a PR to have it built-in). However it seems a bit slow.

Have a great day.

PS : Sorry for the linting modifications, just pay attention to the new layer at the bottom of the file.
I also commented the shape for each variable, but it will be removed, it just help for reviewing and debugging.

This PR is the same as https://github.com/tensorflow/tfjs/pull/4955 just cleaned and reworked as the other one was too old...

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6212)


pyu10055 (commented):
Thank you very much for the contribution, I have a couple of high level questions:
1. there is no transformer layers in keras definition, since this is based on multi-head attention layer, which is defined in keras, would it make more sense to implement multi-head attention instead?
2. The new layer might be good to have a separate file (i.e. attention.ts), instead of be part of core.ts

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6212#-:-MzvMOdJ7u1da1f5kWGG:bz0u9ki)* status: 0 of 1 approvals obtained

Ierezell:
Hi @pyu10055, thanks a lot for the response. 

1. I can rename the whole layer to multi-head attention if needed. It's just a tiny bit more than only the multihead attention algorithm and if I remember I was inspired by https://github.com/tensorflow/models/blob/master/official/nlp/modeling/layers/transformer.py as it been a long time since I did that, I'm open to modify it to match the new standards
2. For sure I can place it in another file.

I will make the modifications.

Have a great day.

PR #6239, opened 2022-03-17
Fix nightly build.

lina128:
The number of binaryCache is different for without parallel compilation vs with parallel compilation, in Webgl1 env.

BinaryCache without parallel compilation:
1. EncodeMatrixPackedProgram_2,4_2,4_false1,1,3_2,4_false_
2. BinaryOpPackedProgram_3_2,4_false3_2,4_false3_2,4_false_
3. EncodeFloatPackedProgram_1,1,3_2,4_false1,1,3_1,3_false_

BinaryCache with parallel compilation:
In addition to above three, it also has:
4. EncodeMatrixPackedProgram_1,1_1,1_false1,1,1_2,2_false_
5. UnaryOpPackedProgram_2,2_false_1,1_false_
6. EncodeFloatPackedProgram_1,1,1_1,1_false1,1,1_1,1_false_

This may be due to how Webgl read data, so nothing is surprised. Also, there's no actual memory leak, the tensor and data buckets counts are same before and after.

The change in this PR just removes the numOfBinaryCache test because it doesn't actually indicate memory leak. This will fix nightly.


mattsoulanille (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6239#-:-MyNmlebAzod5oqadB7Z:bb6ywmx)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever)

PR #6309, opened 2022-04-08
[converter] Remove executeAsync() warning when running a sync model

### LINKED ISSUE #6249: suppress unnecessary warnings for async execution
I cannot know ahead of time if model can be executed synchronously or not  
so only safe way is to execute all models asynchronously using  

```js
model.executeAsync()
```

but then there are warnings which cannot be suppressed:

> This model execution did not contain any nodes with control flow or dynamic output shapes. You can use `model.execute()` instead.

please either provide  
1. a way to determine ahead of time how a model should be executed or  
2. a way to suppress this message when async execution is a clear choice by user  
3. alternatively, remove the message altogether as there is no downside of running model inference asynchronously in all cases

mattsoulanille:
Do not warn the user when they run a synchronous model with model.executeAsync(), which should be allowed for both types of models. Users can check if their model has async ops by trying to run it with model.execute(), which will throw an error if it contains async ops. This is a better experience because users who need synchronous execution will try this anyway, and users who don't can just use model.executeAsync() without seeing warning messages.

fixes #6249

### DIFF
--- a/tfjs-converter/src/executor/graph_executor.ts
+++ b/tfjs-converter/src/executor/graph_executor.ts
@@ -499,11 +499,6 @@ export class GraphExecutor implements FunctionExecutor {
           outputNodeNames, intermediateTensorConsumerCount, usedNodes);
       await Promise.all(promises);
     }
-    if (dynamicNode == null && !isFunctionExecution) {
-      console.warn(
-          `This model execution did not contain any nodes with control flow ` +
-          `or dynamic output shapes. You can use model.execute() instead.`);
-    }
     const missingOutputs =
         outputNodes
             .filter(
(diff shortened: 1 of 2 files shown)


pyu10055 (commented):
Thank you Matthew. But theoratically execute() should have a better performance, since the topological sorted graph is cached after the first run, which executeAsync() is not, maybe you can add that to executeAsync() if the model has no dynamic ops.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6309#-:-N-9CGMSDnvRwXUDvV8k:b-wlpd5y)* status: 0 of 1 approvals obtained (waiting on @lina128)

lina128 (commented):
Maybe only turn on the warning when DEBUG mode is on? I also agree we should add some caching to executeAsync, Ping and I discussed this awhile back, basically cache this part: https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/src/executor/graph_executor.ts#L476-L478

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6309#-:-N-9IBcq1gmoKtZXGR62:b8fsehq)* status: 0 of 1 approvals obtained

PR #6329, opened 2022-04-15
webgpu: support dataToGPU api

xhcao:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6329)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -17,7 +17,7 @@
 
 import './flags_webgpu';
 
-import {backend_util, buffer, DataStorage, DataToGPUWebGLOption, DataType, DataValues, engine, env, GPUData, KernelBackend, Rank, RecursiveArray, ShapeMap, TensorBuffer, TensorInfo, TimingInfo, TypedArray, util} from '@tensorflow/tfjs-core';
+import {backend_util, buffer, DataStorage, DataToGPUWebGPUOption, DataType, DataValues, engine, env, GPUData, KernelBackend, Rank, RecursiveArray, ShapeMap, TensorBuffer, TensorInfo, TimingInfo, TypedArray, util} from '@tensorflow/tfjs-core';
 
 import {BufferManager} from './buffer_manager';
 import {TextureManager} from './texture_manager';
@@ -478,7 +478,7 @@ export class WebGPUBackend extends KernelBackend {
    *     customBufShape: Optional. If set, will use the user defined buffer
    *     size to create the buffer.
    */
-  readToGPU(dataId: DataId, options: DataToGPUWebGLOption = {}): GPUData {
+  readToGPU(dataId: DataId, options: DataToGPUWebGPUOption = {}): GPUData {
     const srcData = this.tensorMap.get(dataId);
     const {values, dtype, bufferInfo} = srcData;
 
@@ -513,7 +513,9 @@ export class WebGPUBackend extends KernelBackend {
         [bufferSize / webgpu_util.GPUBytesPerElement(dtype)], dtype);
     // Make engine track this tensor, so that we can dispose it later.
     const tensorRef = engine().makeTensorFromDataId(
-        tensorInfo.dataId, tensorInfo.shape, tensorInfo.dtype);
+        tensorInfo.dataId,
+        [bufferInfo.byteSize / webgpu_util.GPUBytesPerElement(dtype)],
+        tensorInfo.dtype);
 
     const info = this.tensorMap.get(tensorInfo.dataId);
     info.bufferInfo.buffer = resBuffer;
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -377,19 +377,13 @@ export class WebGPUBackend extends KernelBackend {
     return this.currentComputePass;
   }
 
-  private async getBufferData(info: TensorBufferInfo):
+  public async getBufferData(buffer: GPUBuffer, size: number):
       Promise<backend_util.BackendValues> {
-    if (info.values != null) {
-      // Data is on the CPU.
-      return info.values;
-    }
     const staging = this.acquireBuffer(
-        info.bufferInfo.byteSize,
-        GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ);
+        size, GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ);
     this.ensureCommandEncoderReady();
     this.ensureComputePassEnded();
-    this.currentCommandEncoder.copyBufferToBuffer(
-        info.bufferInfo.buffer, 0, staging, 0, info.bufferInfo.byteSize);
+    this.currentCommandEncoder.copyBufferToBuffer(buffer, 0, staging, 0, size);
     this.submitQueue();
 
     await staging.mapAsync(GPUMapMode.READ);
@@ -398,8 +392,7 @@ export class WebGPUBackend extends KernelBackend {
     staging.unmap();
     if (staging != null) {
       this.bufferManager.releaseBuffer(
-          staging, info.bufferInfo.byteSize,
-          GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ);
+          staging, size, GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ);
     }
 
     // Need to get texture from swapChain to enable profiling tool
@@ -467,37 +460,14 @@ export class WebGPUBackend extends KernelBackend {
       vals = backend_util.mergeRealAndImagArrays(
           realValues as Float32Array, imagValues as Float32Array);
     } else {
-      const data = await this.getBufferData(info);
+      const data = info.values != null ?
+          info.values :
+          await this.getBufferData(
+              info.bufferInfo.buffer, info.bufferInfo.byteSize);
       vals =
           webgpu_util.ArrayBufferToTypedArray(data as ArrayBuffer, info.dtype);
     }
     this.convertAndCacheOnCPU(dataId, vals);
-    return vals;
-  }
-
-  async downloadGPUBufferData(gpuData: GPUData):
-      Promise<backend_util.BackendValues> {
-    const staging = this.acquireBuffer(
-        gpuData.bufSize, GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ);
-    this.ensureCommandEncoderReady();
-    this.ensureComputePassEnded();
-    this.currentCommandEncoder.copyBufferToBuffer(
-        gpuData.buffer, 0, staging, 0, gpuData.bufSize);
-    this.submitQueue();
-
-    await staging.mapAsync(GPUMapMode.READ);
-    const values = staging.getMappedRange().slice(0);
-
-    staging.unmap();
-    if (staging != null) {
-      this.bufferManager.releaseBuffer(
-          staging, gpuData.bufSize,
-          GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ);
-    }
-
-    const data = values as backend_util.BackendValues;
-    const vals = webgpu_util.ArrayBufferToTypedArray(
-        data as ArrayBuffer, gpuData.tensorRef.dtype);
     return vals;
   }
 
@@ -526,14 +496,17 @@ export class WebGPUBackend extends KernelBackend {
 
     if (options.customBufSize != null) {
       util.assert(
-          options.customBufSize % 4 === 0,
-          () => 'customBufSize should be a multiple of 4.');
+          options.customBufSize >= bufferInfo.byteSize,
+          () => 'customBufSize should be equal or larger than the buffer size');
     }
 
+    const bufferSize = options.customBufSize != null ?
+        Math.max(bufferInfo.byteSize, options.customBufSize) :
+        bufferInfo.byteSize;
     const copySize = options.customBufSize != null ?
         Math.min(bufferInfo.byteSize, options.customBufSize) :
         bufferInfo.byteSize;
-    const resBuffer = this.acquireBuffer(copySize);
+    const resBuffer = this.acquireBuffer(bufferSize);
     this.ensureCommandEncoderReady();
     this.ensureComputePassEnded();
     this.currentCommandEncoder.copyBufferToBuffer(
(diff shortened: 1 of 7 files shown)


xhcao:

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -471,6 +472,83 @@ export class WebGPUBackend extends KernelBackend {
     return vals;
   }
 
+  async downloadGPUBufferData(gpuData: GPUData):
```
qjia7, line 475: The function body is very similar with `getBufferData`. Can you write a common helper function to reuse the code?

tfjs-backend-webgpu/src/backend_webgpu_test.ts
```diff
@@ -246,3 +246,128 @@ describeWebGPU('backendWebGPU', () => {
...
+  it('has a valid buffer for dtype=float32.', async () => {
+    const webGPUBackend = (tf.backend() as WebGPUBackend);
+    const data = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12];
+    const size = 48;
+    const a = tf.tensor(data, [1, 3, 4]);
+    const b = tf.add(a, 0);
+    const res = b.dataToGPU();
+    expectArraysEqual(res.bufSize, size);
+    const resData = await webGPUBackend.downloadGPUBufferData(res);
+    expectArraysClose(resData, new Float32Array(data));
```
qjia7, line 271: Why not directly use `expectArraysClose(resData, data);`?

tfjs-backend-webgpu/src/backend_webgpu_test.ts
```diff
@@ -246,3 +246,128 @@ describeWebGPU('backendWebGPU', () => {
...
+  });
+
+  it('uses user defined bufSize.', async () => {
```
qjia7, line 274: For a shorter buffer, we should make it throw an error instead of making it valid.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -471,6 +472,83 @@ export class WebGPUBackend extends KernelBackend {
...
+      } else {
+        throw new Error('There is no data on GPU or CPU.');
+      }
+    }
+
+    if (options.customBufSize != null) {
+      util.assert(
+          options.customBufSize % 4 === 0,
+          () => 'customBufSize should be a multiple of 4.');
+    }
```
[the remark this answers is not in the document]
qjia7: For webgpu, I think we keep it is to conveniently copy the returned buffer to a texture and render to the canvas.  And if this point doesn't hold, we should remove `options.customBufSize` for webgpu.

tfjs-core/package.json
```diff
@@ -78,7 +78,8 @@
     "@types/webgl-ext": "0.0.30",
     "long": "4.0.0",
     "node-fetch": "~2.6.1",
-    "seedrandom": "2.4.3"
+    "seedrandom": "2.4.3",
+    "@webgpu/types": "0.1.6"
```
gyagp, line 82: We need to use "@webgpu/types" defined in tfjs/package.json instead.

xhcao:
FYI. There are issues about changing the shape when calling dataToGPU, I are discussing with LiNa in https://github.com/tensorflow/tfjs/pull/5953.

tfjs-core/package.json
xhcao: tfjs/package.json had already added this package, so I removed it here.

tfjs-backend-webgpu/src/backend_webgpu_test.ts
xhcao: I had changed `bufSize` from smaller to larger. This case is used to test that custom  buffer size is also valid for dataToGPU. And throwing an error is covered by last case when buffer size is smaller.

tfjs-backend-webgpu/src/backend_webgpu_test.ts
xhcao: Done

tfjs-backend-webgpu/src/backend_webgpu.ts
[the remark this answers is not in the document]
xhcao: Done

tfjs-backend-webgpu/src/backend_webgpu.ts
xhcao: Done

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -471,6 +472,83 @@ export class WebGPUBackend extends KernelBackend {
...
+  readToGPU(dataId: DataId, options: DataToGPUWebGLOption = {}): GPUData {
+    const srcData = this.tensorMap.get(dataId);
+    const {values, dtype, bufferInfo} = srcData;
+
+    if (dtype === 'complex64') {
+      throw new Error('Does not support reading buffer for complex64 dtype.');
+    }
+
+    if (bufferInfo.buffer == null) {
+      if (values != null) {
```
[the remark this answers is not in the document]
xhcao: Hi, Xing, Sorry that I did not fully understand your meaning. But dataToGPU is used when the data of tensor must be in GPU.

xhcao:
Please review it again, I sort of agree to add the custom size option, some details is https://github.com/tensorflow/tfjs/pull/5953#issuecomment-1109573264

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -467,14 +460,71 @@ export class WebGPUBackend extends KernelBackend {
...
+    if (options.customBufSize != null) {
+      util.assert(
+          options.customBufSize >= bufferInfo.byteSize,
+          () => 'customBufSize should be equal or larger than the buffer size');
+    }
+
+    const bufferSize = options.customBufSize != null ?
+        Math.max(bufferInfo.byteSize, options.customBufSize) :
+        bufferInfo.byteSize;
+    const copySize = options.customBufSize != null ?
```
qjia7, line 506: `const copySize = bufferInfo.byteSize;`? Or directly use `bufferInfo.byteSize` in `copyBufferToBuffer`.

qjia7 (commented):
Add @lina128 to take another look.

LGTM with one open.
For webgpu, we are exposing the buffer instead of a texture. So I am not sure whether it's still necessary to let user specify a custom buffer size.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -467,14 +460,71 @@ export class WebGPUBackend extends KernelBackend {
...
+    }
+
+    if (options.customBufSize != null) {
+      util.assert(
+          options.customBufSize >= bufferInfo.byteSize,
+          () => 'customBufSize should be equal or larger than the buffer size');
+    }
+
+    const bufferSize = options.customBufSize != null ?
+        Math.max(bufferInfo.byteSize, options.customBufSize) :
```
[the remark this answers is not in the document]
xhcao: Done

tfjs-backend-webgpu/src/backend_webgpu.ts
xhcao: Done

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -17,7 +17,7 @@
 
 import './flags_webgpu';
 
-import {backend_util, buffer, DataStorage, DataType, DataValues, engine, env, KernelBackend, Rank, RecursiveArray, ShapeMap, TensorBuffer, TensorInfo, TimingInfo, TypedArray, util} from '@tensorflow/tfjs-core';
+import {backend_util, buffer, DataStorage, DataToGPUWebGLOption, DataType, DataValues, engine, env, GPUData, KernelBackend, Rank, RecursiveArray, ShapeMap, TensorBuffer, TensorInfo, TimingInfo, TypedArray, util} from '@tensorflow/tfjs-core';
```
qjia7, line 20: Should we use `DataToGPUOptions` instead of `DataToGPUWebGLOption` for webgpu backend?

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -467,14 +460,67 @@ export class WebGPUBackend extends KernelBackend {
...
+                                                       bufferInfo.byteSize;
+    const resBuffer = this.acquireBuffer(bufferSize);
+    this.ensureCommandEncoderReady();
+    this.ensureComputePassEnded();
+    this.currentCommandEncoder.copyBufferToBuffer(
+        bufferInfo.buffer, 0, resBuffer, 0, bufferInfo.byteSize);
+    this.submitQueue();
+
+    const tensorInfo = this.makeTensorInfo(
+        [bufferSize / webgpu_util.GPUBytesPerElement(dtype)], dtype);
```
qjia7, line 513: To keep consistent with WebGL, should we use the src tensor's shape here? And maybe add a test to verify the `tensorRef`'s `shape` and `dtype`?

qjia7 (commented):
Please also fix the bots failures. Thanks.

xhcao:
Please review it again, thank you.

lina128 (approved):
Great work, thank you!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6329#-:-N0lQerDE9zYSjzFJ8lV:b-nv8j63)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @qjia7, @webgpu, and @xhcao)

___
*[`tfjs-backend-webgpu/src/backend_webgpu.ts` line 517 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6329#-N-gL_2B6hpi_CCHa9hA-r1-517:-N0lP_s77NF0ZuigSyjL:bvyho9d) ([raw file](https://github.com/tensorflow/tfjs/blob/09c371d7d6e7ddd7e49d8a8aa0454f2e7d17794c/tfjs-backend-webgpu/src/backend_webgpu.ts#L517)):*
<details><summary><i>Previously, xhcao wrote…</i></summary><blockquote>

Hi, Xing, Sorry that I did not fully understand your meaning. But dataToGPU is used when the data of tensor must be in GPU.
</blockquote></details>

Yeah, Jiajia asked me this question too for WebGL. This is the same behavior as WebGL implementation today. My thinking is that user should catch the error and read data from CPU, whether they want to upload or manipulate data directly in CPU. If data is on CPU, most likely it's not a lot of data, so the use case may actually be computation on CPU. But we don't know what the real use case is, so let's keep the design simple for now. If there's actual user request in the future, we can add the support. What do you think?

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: I still think there is a potential issue if we use the buffer size to calculate the tensor shape. Currently, it may be safe since the the buffer size is exactly equal to the shape's size. However, what if we use a larger buffer to store the data, for example, clamp all buffer to be aligned with 16 bytes for optimization in future. 
Can we directly use the source tensor's shape here? I know currently we don't keep the shape in `TensorBufferInfo`. Maybe we should add it?
Another example is webgl. WebGL is using RGBA 4 channels to store data. However, the actual result tensor shape's size may not be divisible by 16 bytes. So the original tensor shape is useful.
@xhcao @lina128 How's your opinion on this?

xhcao: There exists an issue. We cannot compute the element count from the source buffer size, because the buffer size may be larger than the elements size, for example, if dtype is 2 bytes and there are 5 elements, because the buffer size must be multiple of 4 when creating the buffer, the real buffer size is 12, if we use buffer size to compute the element count, the result is 6 and is wrong. So we should add the shape for webgpu backend.

xhcao:
Reminder: Please review it again, thank you.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -467,14 +463,69 @@ export class WebGPUBackend extends KernelBackend {
...
+    const resBuffer = this.acquireBuffer(bufferSize);
+    this.ensureCommandEncoderReady();
+    this.ensureComputePassEnded();
+    this.currentCommandEncoder.copyBufferToBuffer(
+        bufferInfo.buffer, 0, resBuffer, 0, bufferInfo.byteSize);
+    this.submitQueue();
+
+    const tensorInfo = this.makeTensorInfo(shape, dtype);
+    // Make engine track this tensor, so that we can dispose it later.
+    const tensorRef = engine().makeTensorFromDataId(
```
qjia7, line 517: nit: `makeTensorFromDataId` has been changed to `makeTensorFromTensorInfo` in latest change.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -467,14 +463,69 @@ export class WebGPUBackend extends KernelBackend {
...
+        tensorInfo.dataId, tensorInfo.shape, tensorInfo.dtype);
+
+    const info = this.tensorMap.get(tensorInfo.dataId);
+    info.bufferInfo.buffer = resBuffer;
+    // Explicitly change the buffer size that could release the buffer
+    // successfully in future.
+    info.bufferInfo.byteSize = bufferSize;
```
qjia7, line 524: Why need to explicitly change the buffer size since the buffer is acquired as the same size?

qjia7 (approved):
LGTM with two nits. Thanks.

tfjs-backend-webgpu/src/backend_webgpu.ts
xhcao: Not the same, one is computed by `shape`, and the other is real buffer size

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -467,14 +463,70 @@ export class WebGPUBackend extends KernelBackend {
...
+      } else {
+        throw new Error('There is no data on GPU or CPU.');
+      }
+    }
+
+    const size =
+        util.sizeFromShape(shape) * webgpu_util.GPUBytesPerElement(dtype);
+    if (options.customBufSize != null) {
+      util.assert(
+          options.customBufSize >= size,
```
xhcao, line 504: Here changed from buffersize to shape size for allowing to copy multiple times of original tensor content.

qjia7 (approved):
LGTM, thanks.

PR #6333, opened 2022-04-18
[webgpu] Add bazelisk back to package dev dependency

haoyunfeix:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6333)


haoyunfeix:

gyagp (changes requested):
I think a better solution is to introduce npx in root package.json, and use "npx bazel" instead in tfjs-backend-webgpu/package.json. 
I also submitted an issue https://github.com/tensorflow/tfjs/issues/6334 to ask for cleanup for all the package.json files, and our cleanup here is part of this effort.

hujiajie:
> I think a better solution is to introduce npx in root package.json, and use "npx bazel" instead in tfjs-backend-webgpu/package.json.

It sounds really bad to depend on npm/npx in a project managed by Yarn...

gyagp:
@mattsoulanille "yarn build" calls bazel, which is expected to reside in tfjs-backend-webgpu/node_modules/.bin/ only (will not check node_modules/.bin in root folder). I tried npx and it can solve this problem if we introduce it and change all the commands to call "bazel" as "npx bazel", but Jiajie raised some concern to introduce npx. Do you have a better suggestion?

gyagp:
@mattsoulanille "yarn build" calls bazel, which is expected to reside in tfjs-backend-webgpu/node_modules/.bin/ only (will not check node_modules/.bin in root folder). I tried npx  can solve this problem if we introduce it and change all the command to call "bazel" as "npx bazel", but Jiajie raised some concern to introduce npx. Do you have a better suggestion?

haoyunfeix:

gyagp (approved):
LGTM, thanks!

tfjs-backend-webgpu/package.json
```diff
@@ -11,13 +11,13 @@
...
+    "bundle": "yarn build",
     "bundle-ci": "yarn bundle",
-    "build-npm": "bazel build :tfjs-backend-webgpu_pkg",
-    "publish-npm": "bazel run :tfjs-backend-webgpu_pkg.publish",
-    "test": "yarn test-dev",
-    "test-dev": "ibazel run :tfjs-backend-webgpu_test --test-output=streamed"
+    "build-npm": "yarn build",
+    "publish-npm": "yarn --cwd .. bazel run tfjs-backend-webgpu:tfjs-backend-webgpu_pkg.publish",
+    "test": "yarn --cwd .. bazel test tfjs-backend-webgpu:tfjs-backend-webgpu_test --test_output=streamed",
+    "test-dev": "yarn --cwd .. bazel run tfjs-backend-webgpu:tfjs-backend-webgpu_test --test_output=streamed"
```
hujiajie, line 20: Why to use `bazel test` in L19, but `bazel run` in L20? @haoyunfeix @qjia7

gyagp: I had the same question yesterday, and Yunfei answered me as below (https://www.npmjs.com/package/@bazel/concatjs):
When executed under bazel test, this uses a headless browser for speed. This is also because bazel test allows multiple targets to be tested together, and we don't want to open a Chrome window on your machine for each one. Also, under bazel test the test will execute and immediately terminate.
Running under ibazel test gives you a "watch mode" for your tests. The rule is optimized for this case - the test runner server will stay running and just re-serve the up-to-date JavaScript source bundle.
To debug a single test target, run it with bazel run instead. This will open a browser window on your computer. Also you can use any other browser by opening the URL printed when the test starts up. The test will remain running until you cancel the bazel run command.

So bazel run is more interactive than bazel test, per my understanding.

hujiajie: Now I'm seeing this error on Linux during `yarn test`:

```
29 04 2022 05:33:46.922:ERROR [launcher]: No binary for Chrome browser on your platform.
  Please, set "CHROME_BIN" env variable.
```

But I've exported `CHROME_BIN`. Using `bazel run` seems to work well for me.

hujiajie: Do you know if the other backends (WASM / WebGL) will use the system browser or `CHROME_BIN` during testing?

hujiajie: The Karma Chrome launcher seems to have built-in support for Chrome Canary, so I'm going to see if that works and avoid abusing the `CHROME_BIN` environment variable.

PR #6350, opened 2022-04-26
[webgpu] Device limits maxStorageBufferBindingSize

haoyunfeix:

### DIFF
--- a/tfjs-backend-webgpu/src/index.ts
+++ b/tfjs-backend-webgpu/src/index.ts
@@ -45,6 +45,7 @@ if (isWebGPUSupported()) {
           adapterLimits.maxComputeWorkgroupStorageSize,
       'maxComputeWorkgroupsPerDimension':
           adapterLimits.maxComputeWorkgroupsPerDimension,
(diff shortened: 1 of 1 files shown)


haoyunfeix:
Use adapter limit maxStorageBufferBindingSize when request a device.
This will fix deeplab demo [ADE20K](https://github.com/tensorflow/tfjs-models/tree/master/deeplab/demo) storage buffer binding size over issue.

qjia7:
> Use adapter limit maxStorageBufferBindingSize when request a device. This will fix deeplab demo [ADE20K](https://github.com/tensorflow/tfjs-models/tree/master/deeplab/demo) storage buffer binding size over issue.

Do you know which operator is causing this issue? I just worry that this is fixed in one platform, but not in another platform if maxStorageBufferBindingSize  can't be larger. Maybe we should also consider that whether we can use a uniform buffer  or texture instead of a storage buffer.

qjia7 (approved):
LGTM, thanks.

gyagp (approved):

PR #6362, opened 2022-05-02
Fix bufferSync return type

ahmedsabie:
The existing bufferSync always returned a float template type which throws away the type, so users of the function always assume they are working with a number array which is incorrect for string types.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6362)

### DIFF
--- a/tfjs-backend-cpu/src/backend_cpu.ts
+++ b/tfjs-backend-cpu/src/backend_cpu.ts
@@ -15,7 +15,7 @@
  * =============================================================================
  */
 
-import {backend_util, BackendTimingInfo, buffer, DataStorage, DataType, DataValues, engine, env, kernel_impls, KernelBackend, Rank, ShapeMap, Tensor, Tensor2D, TensorBuffer, TensorInfo, TypedArray, util} from '@tensorflow/tfjs-core';
+import {backend_util, BackendTimingInfo, buffer, DataStorage, DataType, engine, env, kernel_impls, KernelBackend, Rank, ShapeMap, Tensor, Tensor2D, TensorBuffer, TensorInfo, TypedArray, util} from '@tensorflow/tfjs-core';
 
 const whereImpl = kernel_impls.whereImpl;
 import {assertNotComplex} from './cpu_util';
@@ -142,19 +142,21 @@ export class MathBackendCPU extends KernelBackend {
     return this.data.get(dataId).values;
   }
 
-  bufferSync<R extends Rank>(t: TensorInfo): TensorBuffer<R> {
+  bufferSync<R extends Rank, D extends DataType>(t: TensorInfo):
+      TensorBuffer<R, D> {
     const data = this.readSync(t.dataId);
-    let decodedData = data as DataValues;
     if (t.dtype === 'string') {
       try {
         // Decode the bytes into string.
-        decodedData = (data as Uint8Array[]).map(d => util.decodeString(d));
+        const strings = (data as Uint8Array[]).map(d => util.decodeString(d));
+        return buffer(t.shape as ShapeMap[R], t.dtype, strings) as
+            TensorBuffer<R, D>;
       } catch {
         throw new Error('Failed to decode encoded string bytes into utf-8');
       }
     }
-    return buffer(t.shape as ShapeMap[R], t.dtype, decodedData) as
-        TensorBuffer<R>;
+    return buffer(t.shape as ShapeMap[R], t.dtype, data as TypedArray) as
+        TensorBuffer<R, D>;
   }
 
   makeOutput<T extends Tensor>(
--- a/tfjs-backend-cpu/src/kernels/AvgPool3DGrad.ts
+++ b/tfjs-backend-cpu/src/kernels/AvgPool3DGrad.ts
@@ -15,7 +15,7 @@
  * =============================================================================
  */
 
-import {AvgPool3DGrad, AvgPool3DGradAttrs, AvgPool3DGradInputs, backend_util, buffer, KernelConfig, KernelFunc, TensorInfo} from '@tensorflow/tfjs-core';
+import {AvgPool3DGrad, AvgPool3DGradAttrs, AvgPool3DGradInputs, backend_util, buffer, KernelConfig, KernelFunc, Rank, TensorInfo} from '@tensorflow/tfjs-core';
 
 import {MathBackendCPU} from '../backend_cpu';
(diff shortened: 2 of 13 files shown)


pyu10055 (approved):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6362)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

lina128 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6362)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ahmedsabie)

___
*[`tfjs-backend-cpu/src/backend_cpu.ts` line 150 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6362#-N16GC1q3i3xBveiqPTt:-N16GC1q3i3xBveiqPTu:b-wyaz77) ([raw file](https://github.com/tensorflow/tfjs/blob/e04514eeac2689e53d6c6a91cddc23fd3c3de2d0/tfjs-backend-cpu/src/backend_cpu.ts#L150)):*
> ```XML
>   }
> 
>   bufferSync<R extends Rank, D extends DataType>(t: TensorInfo):
> ```

Can you add 'float32' as the default DataType here?



_Code quote:_
```XML
D 
```

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6362)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ahmedsabie)

___
*[`tfjs-backend-cpu/src/backend_cpu.ts` line 150 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6362#-N16GC1q3i3xBveiqPTt:-N1AvoIa51YtppboQrtf:b7pawlf) ([raw file](https://github.com/tensorflow/tfjs/blob/e04514eeac2689e53d6c6a91cddc23fd3c3de2d0/tfjs-backend-cpu/src/backend_cpu.ts#L150)):*
<details><summary><i>Previously, ahmedsabie wrote…</i></summary><blockquote>

This would reintroduce the problem of accidental cast to float since no compiler error would be generated, versus without default where you have to assert the default so it's an explicit decision. This is the same problem I ran into trying to add string support to SparseToDense
</blockquote></details>

Can you specify what error you saw when adding string support to SparseToDense? Or maybe send me that PR with the CI log? Thanks

ahmedsabie (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6362)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

___
*[`tfjs-backend-cpu/src/backend_cpu.ts` line 150 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6362#-N16GC1q3i3xBveiqPTt:-N1FPangEa9Jxo4s4p_g:b-spgvua) ([raw file](https://github.com/tensorflow/tfjs/blob/e04514eeac2689e53d6c6a91cddc23fd3c3de2d0/tfjs-backend-cpu/src/backend_cpu.ts#L150)):*

Can you specify what error you saw when adding string support to SparseToDense? Or maybe send me that PR with the CI log? Thanks
</blockquote></details>

The error involved defaultValue being type asserted into a number which compiled due to bufferSync always returning a float32 dtype implicitly. That's when I found this bug with bufferSync storing string values as a float32 dtype so that's why we need dtype template added.

I suggest we don't put a default because it forces the user to choose what type they are working with and it is explicit to the next user. The same problem above would occur if the default was float32 and the previous coder didn't explicitly specify float

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6362)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ahmedsabie and @lina128)

___
*[`tfjs-backend-cpu/src/backend_cpu.ts` line 158 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6362#-N1fBJ_j6S8EteeGOPPk:-N1fBJ_j6S8EteeGOPPl:b3kljvr) ([raw file](https://github.com/tensorflow/tfjs/blob/e04514eeac2689e53d6c6a91cddc23fd3c3de2d0/tfjs-backend-cpu/src/backend_cpu.ts#L158)):*
> ```XML
>         const strings = (data as Uint8Array[]).map(d => util.decodeString(d));
>         return buffer(t.shape as ShapeMap[R], t.dtype, strings) as
>             TensorBuffer<R, D>;
> ```

Will it throw error if we force cast type as string here and just use default everywhere else? Like this:
```
        return buffer(t.shape as ShapeMap[R], t.dtype, strings) as
            TensorBuffer<R, string>;

or

        return buffer(t.shape as ShapeMap[R], t.dtype, strings) as
            any as TensorBuffer<R, string>;


```



_Code quote:_
```XML
TensorBuffer<R, D>
```

ahmedsabie (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6362)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

___
*[`tfjs-backend-cpu/src/backend_cpu.ts` line 158 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6362#-N1fBJ_j6S8EteeGOPPk:-N1oCQYpBjA3J9Z0mZgm:b7eqph6) ([raw file](https://github.com/tensorflow/tfjs/blob/e04514eeac2689e53d6c6a91cddc23fd3c3de2d0/tfjs-backend-cpu/src/backend_cpu.ts#L158)):*

Will it throw error if we force cast type as string here and just use default everywhere else? Like this:
```
        return buffer(t.shape as ShapeMap[R], t.dtype, strings) as
            TensorBuffer<R, string>;

or

        return buffer(t.shape as ShapeMap[R], t.dtype, strings) as
            any as TensorBuffer<R, string>;


```
</blockquote></details>

It's a compile error because templates are inputs, the user could theoretically pass in a string input but ask for a float return type which is why you can't make an assertion to string

lina128 (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6362#-:-N1oING96rtCVPCmDpmR:bb6ywmx)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128)

PR #6394, opened 2022-05-11
webgpu: extend Conv2DMMVec4Program to support inChannels=3

qjia7:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6394)

### DIFF
--- a/tfjs-backend-webgpu/src/conv2d_mm_vec4_webgpu.ts
+++ b/tfjs-backend-webgpu/src/conv2d_mm_vec4_webgpu.ts
@@ -28,13 +28,13 @@ export class Conv2DMMVec4Program implements WebGPUProgram {
   dispatchLayout: {x: number[], y: number[], z: number[]};
   dispatch: [number, number, number];
   variableNames = ['x', 'W'];
+  variableTypes: string[];
   uniforms =
       `filterDims : vec2<i32>, pad : vec2<i32>, stride : vec2<i32>, dilation : vec2<i32>,
       dimAOuter : i32, dimBOuter : i32, dimInner : i32,`;
   workGroupSize: [number, number, number] = [8, 8, 1];
   elementsPerThread: [number, number, number];
   isVec4 = true;
-  isConv2dVec4 = true;
   convInfo: backend_util.Conv2DInfo;
   addBias: boolean;
   activation: backend_util.Activation;
@@ -69,15 +69,22 @@ export class Conv2DMMVec4Program implements WebGPUProgram {
     this.addBias = addBias;
     this.activation = activation;
     this.hasPreluActivationWeights = hasPreluActivationWeights;
+    this.innerElementSize = this.convInfo.inChannels % 4 === 0 ? 4 : 3;
+    if (this.innerElementSize === 3) {
+      this.variableTypes = ['f32', 'vec4<f32>'];
+    } else {
+      this.variableTypes = ['vec4<f32>', 'vec4<f32>'];
+    }
+
     if (this.addBias) {
       this.variableNames.push('bias');
+      this.variableTypes.push('vec4<f32>');
     }
 
     if (this.hasPreluActivationWeights) {
       this.variableNames.push('preluActivationWeights');
+      this.variableTypes.push('vec4<f32>');
     }
-
-    this.innerElementSize = this.convInfo.inChannels % 4 === 0 ? 4 : 3;
     this.tileAOuter = this.outputShape[1] === 1 ?
         1 :
         this.workGroupSize[1] * this.elementsPerThread[1];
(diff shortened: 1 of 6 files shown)


xhcao (approved):

xhcao:
Hi, @qjia7 , I have two questions.
1. In the Conv2DMMVec4Program, we only change the `x` variable from vec4 to vec3, but we have to explicitly define Conv2DMMVec4Program other variables' types, which are originally defined in common code. If some variables are vec4 and others are not vec4, strictly speaking, is it still a vec4 program?
2. Actually, I think you extends the condition of `inChannels=3 & padInfo.type === 'SAME' ` .  And `inChannels=3 & padInfo.type === 'VALID' ` had been already implemented, it reads data as vec4 not vec3, it also handles the issues that vec4 cross a row or channel boundary. Why you here changed to vec3, which need to change common code?

qjia7:
> 1. In the Conv2DMMVec4Program, we only change the `x` variable from vec4 to vec3, but we have to explicitly define Conv2DMMVec4Program other variables' types, which are originally defined in common code. If some variables are vec4 and others are not vec4, strictly speaking, is it still a vec4 program?

vec4 is more like to target the output. From this point, it's not a problem to keep the current name. If you think it's still confusion, we can think a more reasonable name.

> 2. Actually, I think you extends the condition of `inChannels=3 & padInfo.type === 'SAME' ` .  And `inChannels=3 & padInfo.type === 'VALID' ` had been already implemented, it reads data as vec4 not vec3, it also handles the issues that vec4 cross a row or channel boundary. Why you here changed to vec3, which need to change common code?

Currently, most of the benchmark models are not going to Conv2DMMVec4Program since they are using 'SAME' padding. In fact, in order to support `inChannels=3 & padInfo.type === 'VALID' `, the code already becomes very complicated. It's not easy to understand those extra codes. And to support `inChannels=3 & padInfo.type === 'SAME' `, the bounds checking will become more complicated. It's not easy to maintain those hard reading code. And if we use the `vec3` to support `inChannels % 3 === 0`, it's very straightforward and the code becomes much simpler and more consistent with current vec4.
As for the change to the common part, it makes our code more flexible. In the future, we can easily to support any kind of type  buffer type to do some optimizations. Currently, our shaders are limited to use either `float` or `vec4`. But for some optimization, we may need such kind of flexibility.

xhcao:
1. I think vec4 does not only mean to output as vec4 format. It could also get better performance of MAD computations for vec4. I think  there is seldom situation that use vec3 in the conv2d and matmul implementations. And all variables in the same program should be the same vecX, it will be good to design and implement the algorithm.
2. I think current implementation is OK for `inChannels=3 & padInfo.type === 'SAME'`, and float and vec4 formats are enough. You could refer other reviewers' comments for this question.

qjia7:
> 1. I think vec4 does not only mean to output as vec4 format. It could also get better performance of MAD computations for vec4. I think  there is seldom situation that use vec3 in the conv2d and matmul implementations. And all variables in the same program should be the same vecX, it will be good to design and implement the algorithm.

My change doesn't impact the original vec4 algorithm and can reuse the code as many as possible to support vec3. If you split the code to a separate program, most of the code will be same. It's not good from the point of code reusing. And for currently conv2d, some inChannels may be divisible by 2 not 4, but outChannels is divisible by 4. It can also reuse the currently code to support vec2 + vec4. I don't think it's a good idea to rewrite a new program for each type combination.

> 2. I think current implementation is OK for `inChannels=3 & padInfo.type === 'SAME'`, and float and vec4 formats are enough. You could refer other reviewers' comments for this question.

What do you mean `float and vec4 formats are enough`? So you don't think `float + vec4 -> vec4` or `vec2 + vec4 -> vec4` is necessary, yes?

gyagp (approved):
LGTM. According to the offline discussion, we need more refactoring in later PRs.

tfjs-backend-webgpu/src/conv2d_mm_vec4_webgpu.ts
```diff
@@ -100,81 +108,45 @@ export class Conv2DMMVec4Program implements WebGPUProgram {
...
-            } else if (inChCoord == 1) {
-              resData = vec4<f32>(resData.xy, temp.xy);
-            } else {
-              resData = vec4<f32>(resData.x, temp.xyz);
-            }
-          }
-          `;
+        [
+          this.elementsPerThread[0], this.elementsPerThread[1],
+          this.innerElementSize
```
xhcao, line 115: Pack these three values for `workPerThread` parameter of makeMatMulPackedVec4Source, and `workPerThread` is defined for output. Is `innerElementSize` related with output, is it reasonable to treat it as a part of `workPerThread`?

xhcao:
LGTM with a question.

tfjs-backend-webgpu/src/conv2d_mm_vec4_webgpu.ts
qjia7: Split it from workPerThread. See change https://github.com/tensorflow/tfjs/pull/6394/commits/73f4ae4db954ecc28bc6c2e3128b70b54b8b6af5 Thanks.

xhcao:
@qjia7  LGTM, thank you.

Issue #5660, opened 2021-09-24
yarn test fails on tfjs-backend-webgl on windows


qjia7:
Steps to reproduce:
1. Get latest tfjs resource code.
2. cd tfjs-backend-webgl
3.  yarn && yarn test
Below error will be reported
```
iBazel [3:14PM]: Running :tfjs-backend-webgl2_test
Loading:
Loading: 0 packages loaded
Analyzing: target //tfjs-backend-webgl:tfjs-backend-webgl2_test (0 packages loaded, 0 targets configured)
Analyzing: target //tfjs-backend-webgl:tfjs-backend-webgl2_test (1 packages loaded, 35 targets configured)
Analyzing: target //tfjs-backend-webgl:tfjs-backend-webgl2_test (1 packages loaded, 35 targets configured)
Analyzing: target //tfjs-backend-webgl:tfjs-backend-webgl2_test (1 packages loaded, 35 targets configured)
Analyzing: target //tfjs-backend-webgl:tfjs-backend-webgl2_test (1 packages loaded, 35 targets configured)
Analyzing: target //tfjs-backend-webgl:tfjs-backend-webgl2_test (1 packages loaded, 35 targets configured)
INFO: Repository python3_interpreter instantiated at:
  D:/workspace/jiajia/tfjs/tfjs/WORKSPACE:231:13: in <toplevel>
Repository rule http_archive defined at:
  C:/users/jqin7/_bazel_jqin7/6lxz3o43/external/bazel_tools/tools/build_defs/repo/http.bzl:336:31: in <toplevel>
Analyzing: target //tfjs-backend-webgl:tfjs-backend-webgl2_test (1 packages loaded, 35 targets configured)
INFO: Repository 'python3_interpreter' used the following cache hits instead of downloading the corresponding file.
 * Hash 'fb1a1114ebfe9e97199603c6083e20b236a0e007a2c51f29283ffb50c1420fb2' for https://www.python.org/ftp/python/3.8.11/
Python-3.8.11.tar.xz
If the definition of 'python3_interpreter' was updated, verify that the hashes were also updated.
ERROR: An error occurred during the fetch of repository 'python3_interpreter':
   Traceback (most recent call last):
        File "C:/users/jqin7/_bazel_jqin7/6lxz3o43/external/bazel_tools/tools/build_defs/repo/http.bzl", line 121, colum
n 10, in _http_archive_impl
                patch(ctx)
[...]

PR #6424, opened 2022-05-19
[webgpu] Make testing with Chrome Canary a little easier

hujiajie:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6424)

### DIFF
--- a/tfjs-backend-webgpu/README.md
+++ b/tfjs-backend-webgpu/README.md
@@ -81,15 +81,9 @@ yarn build
 ```
 
 ## Testing
-The `CHROME_BIN` environment variable must be set to the location of the Chrome
-Canary application.
-
-### Windows
-`set CHROME_BIN=%LOCALAPPDATA%\Google\Chrome SxS\Application\chrome.exe`
-
-### macOS
-`export CHROME_BIN="/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary"`
+Currently the Canary channel of Chrome is used for testing of the WebGPU
+backend:
 
 ```sh
-yarn test
+yarn test  # --test_env=CHROME_CANARY_BIN=/path/to/chrome
 ```
--- a/tools/karma_template.conf.js
+++ b/tools/karma_template.conf.js
@@ -115,7 +115,7 @@ module.exports = function(config) {
         flags: ['--blacklist-accelerated-compositing', '--blacklist-webgl']
       },
       chrome_webgpu: {
-        base: 'Chrome',
+        base: 'ChromeCanary',
         flags: [
           '--enable-unsafe-webgpu',
           '--disable-dawn-features=disallow_unsafe_apis'
--- a/tfjs-backend-webgpu/karma.conf.js
+++ b/tfjs-backend-webgpu/karma.conf.js
@@ -68,7 +68,7 @@ module.exports = function(config) {
     singleRun: true,
     customLaunchers: {
       chrome_webgpu: {
-        base: 'Chrome',
+        base: 'ChromeCanary',
         flags: [
           '--enable-unsafe-webgpu',
           '--disable-dawn-features=disallow_unsafe_apis'


hujiajie:
@qjia7 @haoyunfeix @gyagp @axinging @xhcao PTAL

tools/karma_template.conf.js
```diff
@@ -115,7 +115,7 @@ module.exports = function(config) {
         flags: ['--blacklist-accelerated-compositing', '--blacklist-webgl']
       },
       chrome_webgpu: {
-        base: 'Chrome',
+        base: 'ChromeCanary',
```
qjia7, line 118: Do we need to do similar change here https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/karma.conf.js#L71 ?

hujiajie: Is that still in use? I don't find a counterpart in cpu and webgl backend.

qjia7: > Is that still in use? I don't find a counterpart in cpu and webgl backend.

@haoyunfeix may have more knowlege about it. I think it's still used by the `yarn karma start --browsers=chrome_webgpu` as the workaround in windows to run the yarn test.

mattsoulanille (approved):
LGTM

tools/karma_template.conf.js
haoyunfeix: Yes, that also needs the same change.
But there maybe an important problem for WSL bazel test. The `.bazelrc.user` file needs change CHROME_BIN to CHROME_CANARY_BIN for WebGPU backend, but other backends still needs CHROME_BIN.
We may workaround in bazel test command, but I don't think it's a good idea.
https://github.com/tensorflow/tfjs/blob/master/DEVELOPMENT.md 
```
run --test_env=CHROME_BIN --test_env=WSLENV --test_env=WSL_DISTRO_NAME --define DISPLAY=:0
test --test_env=CHROME_BIN --test_env=WSLENV --test_env=WSL_DISTRO_NAME --define DISPLAY=:0"
```

tfjs-backend-webgpu/README.md
```diff
@@ -81,14 +81,8 @@ yarn build
...
 ## Testing
-The `CHROME_BIN` environment variable must be set to the location of the Chrome
-Canary application.
-
-### Windows
-`set CHROME_BIN=%LOCALAPPDATA%\Google\Chrome SxS\Application\chrome.exe`
-
-### macOS
-`export CHROME_BIN="/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary"`
+Currently the Canary channel of Chrome is used for testing of the WebGPU
```
haoyunfeix, line 84: I prefer add a reminder here to declare `CHROME_CANARY_BIN` as the ENV_CMD of chrome canary, in case some user wants to test in a specific version of canary.

tools/karma_template.conf.js
hujiajie: Then what's the problem of `yarn test` in WSL? And why do we need WSL anyway?

haoyunfeix: On my WSL env, `yarn test` will stop by a warning:
```
20 05 2022 07:11:11.204:ERROR [launcher]: No binary for ChromeCanary browser on your platform.
  Please, set "CHROME_CANARY_BIN" env variable.
```
We use WSL because bazel rules for karma [karma_web_test ](https://bazelbuild.github.io/rules_nodejs/Karma.html#karma_web_test)does not support Windows, you can refer issue https://github.com/tensorflow/tfjs/issues/5660

hujiajie: Thanks, so is `yarn karma start --browsers=chrome_webgpu` still necessary? If no, then `tfjs-backend-webgpu/karma.conf.js` is obsolete, right?

qjia7: I still use it in windows. I rarely use wsl ;(. Can we keep it for a while?

hujiajie: So `yarn karma start --browsers=chrome_webgpu` is for Windows without WSL? I think I'm on track now.

qjia7: >So yarn karma start --browsers=chrome_webgpu is for Windows without WSL?

Yes.

tfjs-backend-webgpu/README.md
hujiajie: Added a comment below. Hope that's informative enough.

qjia7 (approved):
LGTM, thanks.

tfjs-backend-webgpu/README.md
haoyunfeix: For MacOS, that's enough.
For Windows, `yarn test` depends on WSL, `--test_env=CHROME_CANARY_BIN=/path/to/chrome` is not optional but a must, unless we pass CHROME_CANARY_BIN env variable by a custom `.bazelrc.user` config (like https://github.com/tensorflow/tfjs/blob/master/DEVELOPMENT.md#developing-on-windows).
Maybe it's better to divide the comment into 2 parts:
### Windows Subsystem for Linux (WSL)
yarn test --test_env=CHROME_CANARY_BIN=/path/to/chrome
### macOS
yarn test # --test_env=CHROME_CANARY_BIN=/path/to/chrome

hujiajie: I'd argue the case of WSL is analogous to "Chrome seems unavailable in your system because it's not found in the well-known paths. It's your bad if it's installed somewhere else, without configuring your distro properly. You should create a symlink in /usr/bin, or just use the CHROME_CANARY_BIN environment variable". WSL is a Frankenstein to me, so I'm against treating it differently than other Unices.

haoyunfeix: Thanks for the explanation, jiajie.
I thought more or less we have a divergence point at "Whether Chrome locates at well-known path".
My point is, even if the Chrome is in well-known paths(default ones which SHOULD be found by karma tools) or symlink in /usr/bin, or exported CHROME_CANARY_BIN in WSL shell, the Karma(started in bazel server) still can not find the chrome start location(in WSL) unless we passed a `--test_env=CHROME_CANARY_BIN` explicitly when bazel server is starting.
e.g. 
In pure Linux, export CHROME_BIN is enough:
`export CHROME_BIN=/path/to/chrome && yarn test`
In WSL, besides CHROME_BIN, we need explicitly pass CHROME_BIN to bazel
`export CHROME_BIN=/path/to/chrome && yarn test --test_env=CHROME_BIN="$CHROME_BIN"`

I agreed that you said WSL is a Frankenstein, maybe WSL made bazel server environment lost "CHROME_BIN" or the default CHROME locations. But it's not the user's fault to make Chrome unavailable and I don't think the user should have the knowledge of "passing environment in WSL when bazel server starts karma". So I thought maybe it's good to write it down and tell user what should to do before a test.

hujiajie: I don't think `export CHROME_BIN=/path/to/chrome && yarn test` works as expected in bare metal Linux. That's exactly how the original issue was spotted, as my chrome was not installed in well-known paths and `--test_env=CHROME_BIN` is mandatory in bare metal Linux just as WSL.

I insist that it's users' responsibility to figure out the extra homework if they choose the "wrong" distro (WSL here), so I tend to avoid highlighting this in doc. I do not oppose anyone else adding that back, though.

qjia7: I will merge this one. We can always do more changes later on. Thanks.

PR #6429, opened 2022-05-19
Implement http_sync IO handler and add it to loadGraphModelSync

mattsoulanille:
Implement the httpSync IO handler, which uses a synchronous xhr to load a model. It should only be used in a webworker because it blocks the event loop.

As part of implementing the httpSync IO handler, this PR also adds a fetchSync method to `platform.ts` which is used to perform the synchronous xhr. Only the Browser platform is supported right now.

Add httpSync as the default io handler for loadGraphModelSync when it is passed a url. This is currently hard-coded, and there is no registry for sync IO handlers.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6429)

### DIFF
--- a/tfjs-converter/src/executor/graph_model.ts
+++ b/tfjs-converter/src/executor/graph_model.ts
@@ -472,24 +472,58 @@ export async function loadGraphModel(
 }
 
 /**
- * Load a graph model given a synchronous IO handler with a 'load' method.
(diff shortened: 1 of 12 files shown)


mattsoulanille:
Closing to implement the httpSync IO handler outside of TFJS instead, since a synchronous XHR implementation of sync HTTP loading would be very easy to misuse if it were part of the TFJS API.

PR #6438, opened 2022-05-23
[webgpu] update perf test

haoyunfeix:


tfjs-backend-webgpu/perf/tune.js
```diff
@@ -155,70 +130,137 @@ function initPage() {
...
+    // Collect result from here
+    for (let i = 0; i < inputs.length; i++) {
+      document.getElementById('message').innerHTML =
+        `Testing on ${backend} ...`;
+      for (let j = 0; j < numAvg; j++) {
+        const result = tf.matMul(tensors[i].tensorA, tensors[i].tensorB);
+        const m = tf.matMul(tensorsWarmUp.tensorA, tensorsWarmUp.tensorB);
+        result.dispose();
+        m.dispose();
+      }
```
xhcao, line 202: In `tf.profile`,
1. I think we mainly profile `const result = tf.matMul(tensors[i].tensorA, tensors[i].tensorB)`, would including `warmup model` and `warmup gpu` parts influence the correctness?
2. Why we still need to execute ` tf.matMul(tensorsWarmUp.tensorA, tensorsWarmUp.tensorB)` in the 3rd for-loop, I think the cost time of it may be larger than `tf.matMul(tensors[i].tensorA, tensors[i].tensorB)`, which is the real main workload.
3. How to make sure that all matMul commands are executed one by one? I think they may be executed synchronously here?

gyagp (approved):
LGTM

tfjs-backend-webgpu/perf/tune.js
```diff
@@ -155,70 +130,137 @@ function initPage() {
...
+    let n = input.split('[')[2].split(']')[0].split(',')[1];
+    inputs = [`${m},${k},${n}`];
+  }
+  else {
+    inputs = INPUTS;
+  }
+  for (let i = 0; i < inputs.length; i++) {
+    let dimAOuter = parseInt(inputs[i].split(',')[0]);
+    let dimInner = parseInt(inputs[i].split(',')[1]);
+    let dimBOuter = parseInt(inputs[i].split(',')[2]);
```
xhcao, line 165: Why we wrap the `dimAOuter | dimInner | dimBOuter` into a string here, and should encode and decode the string? Why not wrap this value into an array or a structure?

xhcao (commented):
Another comment is that why there are some many variables and all these ones are defined in functions. Could we define some of them in the global scope or wrap some variables together to reduce some definations.

tfjs-backend-webgpu/perf/tune.js
haoyunfeix: 1. Yes, `warmup model` is for the first run of each matmul shape(costing large shader generate and buffer creating time ... which leads to a significant deviation from the average value) and `warmup gpu` is to make every shape of inputs could be measured in the same frequency.
2. The idea to insert a large warmup tensor between each main workload is also to keep high frequency, the real main workload will pull down the frequency if they execute together, please see below pictures:
**with**: keep frequency at almost 1200M
image
**without**: Visible jitter, break the measurement accurate (only happens on WebGPU)
image
3. You're right, serial execution would be considered in another PR.

tfjs-backend-webgpu/perf/tune.js
haoyunfeix: Each row has a `rerun` button to help rerun a single case for double check. And this button catches shape infos(e.g. [16, 512, 16]) from HTML element, which value must be a string. So here I used the raw data what I get from the page.

qjia7 (approved):
LGTM, thanks.

As we synced offline, we can do more refactor based on it.

Issue #4227, opened 2020-11-13
[wasm] Optimization of tf.browser.fromPixels.


kylemcdonald:
I noticed that [tf.browser.fromPixels](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/browser.ts#L145-L151) can be very slow when it spends time in this loop:

```js
    const numPixels = width * height;
    values = new Int32Array(numPixels * numChannels);
    for (let i = 0; i < numPixels; i++) {
      for (let channel = 0; channel < numChannels; ++channel) {
        values[i * numChannels + channel] = vals[i * 4 + channel];
      }
    }
```

(Also reported by #4218)

I was able to get significantly better performance with my application by calling this function instead:

```js
function convertArrayBufferRGBAToUInt8RGB(buffer) {
  // create view of ArrayBuffer
  const x = new Uint8Array(buffer);
  const n = x.length;
  let i = 0;
  let j = 0;
  // loop through all pixels and shift them over to remove alpha
  while (i < n) {
    x[j] = x[i]; i++; j++;
    x[j] = x[i]; i++; j++;
    x[j] = x[i]; i++; j++;
    i++;
  }
  // return a subset of the original
  return x.slice(0, j);
}
```

And then calling it like sort of like this:

```js
const fromPixels2DContext = document.createElement('canvas').getContext('2d'); // only call this once
function fromPixelsFast(inputElt, width, height) {
  return tf.tidy(() => {
    fromPixels2DContext.canvas.width = width;
    fromPixels2DContext.canvas.height = height;
    fromPixels2DContext.drawImage(inputElt, 0, 0, width, height);
    const buffer = fromPixels2DContext.getImageData(0, 0, width, height).data.buffer;
    const UInt8RGB = convertArrayBufferRGBAToUInt8RGB(inputArray);
    return tf.tensor3d(UInt8RGB, [height, width, 3]);
  });
}
```

I think the thing that makes it faster is connected to:

1. Using Uint8Array instead of UInt32Array (small effect)
2. The unrolled loop is better optimized by the browser. (medium effect)
3. It is more efficient to shuffle data in one array than between two arrays due to cache effects. (big effect)

(But I have not done rigorous testing)

It might even be possible to get a little more speed from `.subarray` instead of `.slice` to avoid making the copy.

[...]

Issue #4218, opened 2020-11-11
Implement native tf.browser.fromPixels() for WASM backend


vladmandic:
Currently `tf.browser.fromPixels()` has native implementation for WebGL backend while WASM falls back to default `canvas.getContext('2d').getImageData(0, 0, canvas.width, canvas.height).data;`

This means it will:
a) Fail if input already has a different context  
   e.g., canvas with `webgl` context because user performed some image pre-processing
b) It is really slow

On FullHD input, WebGL native implementation executes in near 0ms while `getImageData()` takes ~20ms  

This significantly impacts overall performance of WASM backend in general.

Example, if model execution takes 33ms in both WebGL and WASM, frame rate in WebGL will be 30 FPS, but in WASM it will be 1000/(20+33) = 18 FPS - that is 36% performance difference just because of lack of native implementation to get image data.

vladmandic:
UPDATE: as an alternative, would it be possible to use WebGL implementation of `tf.browser.fromPixels()` and then continuing with WASM model. I don't mind mixing backends like this, but actually changing active backend on the fly is a no-go.

vladmandic:
Closer look, it's even worse - `browser.js`fromPixels_()` will create extra canvas if it cannot get `2d` context.

So not only there is a missing native implementation, but fallback one is really bad.
```js
    if (fromPixels2DContext == null) {
      fromPixels2DContext = document.createElement('canvas').getContext('2d');
    }
```

Plus it uses DOM (`document.createElement`), so it cannot be executed in Web Worker (if you *have to* create additional canvas, should try `OffsreenCanvas` first and use DOM as fallback).

If you can't use your accelerated webgl implemenation, why not test for canvas `webgl` context first and then rely on GL context `ctx.readPixels()` which is >2x faster than `ctx.getImagedata()`.

AmitMY:
When having a canvas with `2d` context, despite using `webgl` as the `tfjs` backend, the `fromPixels` method still takes 25ms+.
I could not find a way to perhaps copy my canvas with existing context to a new canvas in order to have a `webgl` context.

vladmandic:
@AmitMY in my tests, only first call to `fromPixels` takes ~25ms when using `WebGL` backend as GLSL needs to be initialized  
if doing 1000 calls using `HTMLVideoElement` as input (webcam in full-hd), it takes <2ms per each frame  

[...]

PR #6445, opened 2022-05-25
Add hint willReadFrequently for canvas

axinging:
Add hint willReadFrequently for canvas

This change has performance impact on cpu and wasm backend. For
getImageData involved cases, set willReadFrequently to true may
improve performance, for example, for input image 125x125 to 513x513,
the measured time on getImageData is ~ 20ms when willReadFrequently: false, 
and 6~10 ms when willReadFrequently: true.

The root cause is, getImageData runs on different hardware (CPU, GPU).
See comments here: https://bugs.chromium.org/p/chromium/issues/detail?id=1239467

Use tf.env().set('CANVAS2D_WILL_READ_FREQUENTLY', true) to turn on this feature.

Spec: https://github.com/fserb/canvas2D/blob/master/spec/will-read-frequently.md
Bug: https://github.com/tensorflow/tfjs/issues/4227, https://github.com/tensorflow/tfjs/issues/4218

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6445)

### DIFF
--- a/tfjs-core/src/ops/browser.ts
+++ b/tfjs-core/src/ops/browser.ts
@@ -157,9 +157,9 @@ function fromPixels_(
       } else {
         const willReadFrequently =
             env().getBool('CANVAS2D_WILL_READ_FREQUENTLY');
-        const options = willReadFrequently ? {willReadFrequently} : {};
-        fromPixels2DContext = document.createElement('canvas').getContext(
-                                  '2d', options) as CanvasRenderingContext2D;
+        fromPixels2DContext =
+            document.createElement('canvas').getContext(
+                '2d', {willReadFrequently}) as CanvasRenderingContext2D;
       }
     }
     fromPixels2DContext.canvas.width = width;
--- a/tfjs-core/src/ops/browser.ts
+++ b/tfjs-core/src/ops/browser.ts
@@ -157,9 +157,9 @@ function fromPixels_(
       } else {
         const willReadFrequently =
             env().getBool('CANVAS2D_WILL_READ_FREQUENTLY');
-        fromPixels2DContext =
-            document.createElement('canvas').getContext(
-                '2d', {willReadFrequently}) as CanvasRenderingContext2D;
+        const options = willReadFrequently ? {willReadFrequently} : {};
+        fromPixels2DContext = document.createElement('canvas').getContext(
+                                  '2d', options) as CanvasRenderingContext2D;
       }
     }
     fromPixels2DContext.canvas.width = width;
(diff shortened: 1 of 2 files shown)


gyagp:
Why not set willReadFrequently to true unconditionally for all the backends?

axinging:
> Why not set willReadFrequently to true unconditionally for all the backends?
From https://github.com/fserb/canvas2D/blob/master/spec/will-read-frequently.md, it says the call path of getImageData  was first "heuristics " then turned to a explicit hint willReadFrequently, because this has different performance:
"
Historically, browsers have tried heuristics to decide if a canvas is being read frequently or not, to switch code paths.
This is unreliable, complex and brittle. To prevent this, we propose a hint option that will allow developers to tell the browser that this is a canvas that rely on getImageData performance.
"
So in this PR, I tried to align it with getImageData.

axinging:
@qjia7 @gyagp @xhcao @haoyunfeix PTAL

qjia7 (approved):
Awesome!

gyagp:
@pyu10055 @lina128 This PR supposes to fix the long-standing fromPixels() perf issue, but we're not sure about its side effect. So an env variable is provided now (off by default) and we'd hope developers may have some early tests. If the side effect is minor, it can be turned on by default for CPU and WASM backends.

lina128 (approved):
Thank you! LGTM!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6445#-:-N3WWVqZBIoCOIUkx9K7:bebeqq2)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @pyu10055)

pyu10055 (approved):
Thank you for the investigation and fixing this issue!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6445#-:-N3WWzBc87zvBpp7y3um:bw102kg)* status: :shipit: complete! 3 of 1 approvals obtained

tfjs-core/src/ops/browser.ts
```diff
@@ -155,7 +155,11 @@ function fromPixels_(
               'Reason: OffscreenCanvas Context2D rendering is not supported.');
         }
       } else {
-        fromPixels2DContext = document.createElement('canvas').getContext('2d');
+        const willReadFrequently =
+            env().getBool('CANVAS2D_WILL_READ_FREQUENTLY');
+        const options = willReadFrequently ? {willReadFrequently} : {};
+        fromPixels2DContext = document.createElement('canvas').getContext(
```
qjia7, line 161: nit: It seems that you can directly write it like below and remove options. Tested [here](https://github.com/tensorflow/tfjs/pull/6486).
```
        fromPixels2DContext =
            document.createElement('canvas').getContext(
                '2d', {willReadFrequently}) as CanvasRenderingContext2D;

PR #6446, opened 2022-05-25
webgpu: Support nchw conv2d vec4

qjia7:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6446)


qjia7:
@xhcao @axinging Please take a look, thanks.

qjia7:
> 1. There is no `ColPerThreadA` in our matmul algorithm, I think it is Ok if the x and y dimensions of the workgroup size is the same. But if the our workgroup size is [4, 16, 1], could it work correctly?

For vec4, if transposeA =false, ColPerThreadA is the InnerElementSize (3 | 4). The tileInner will be workGroupSize[0] * this.InnerElementSize. So it works if the workgroup size is [4, 16, 1].
If tranposeA = true, ColPerThreadA  must be 4. It also has no problem.

> 2. I saw that you handled the `transposeA`, but did not handle `transposeB`, Is there not this situation with `transposeB = true` condition?

The target of this PR is to support nchw conv2d vec4. `transposeA` is enough to support it. No need for `transposeB`. And on my experiment on matmul, `transposeB` won't bring good performance.

xhcao:
Hi, @axinging , please take a look.

xhcao (approved):
@Jiajia, Thank you for your reply. I had noticed that `tileInner` only depended on  `tileBOuter`, which was modified by your last patch , so it was not necessary to consider `ColPerThreadA` now . 
LGTM if no performance loses on e2e models.

gyagp (approved):
LGTM

PR #6461, opened 2022-05-27
Fix kernel_registry test flakiness for async backend

hujiajie:
The kernel_registry suite fails occasionally when using the WebGPU backend. Considering Jasmine specs are executed in random order, the theory for a sample failure is:

1. `it('register same kernel on two different backends', ...)` is scheduled first. The spec calls `tf.setBackend('backend2')`, followed by `tf.removeBackend('backend2')`. Since backend2 is active while removing, `ENGINE.backendInstance` is unset to null.

2. Then in `it('register a kernel and call it', ...)`, `tf.backend()` shouts because the WebGPU backend is async and cannot be implicitly re-initialized.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6461)

### DIFF
--- a/tfjs-core/src/kernel_registry_test.ts
+++ b/tfjs-core/src/kernel_registry_test.ts
@@ -17,11 +17,16 @@
 
 import * as tf from './index';
 import {KernelBackend} from './index';
-import {ALL_ENVS, describeWithFlags} from './jasmine_util';
+import {ALL_ENVS, describeWithFlags, TestEnv} from './jasmine_util';
 import {KernelFunc, TensorInfo} from './kernel_registry';
 import {expectArraysClose} from './test_util';
 
-describeWithFlags('kernel_registry', ALL_ENVS, () => {
+describeWithFlags('kernel_registry', ALL_ENVS, (testEnv: TestEnv) => {
+  afterEach(async () => {
+    // Revert backend mutation.
+    await tf.setBackend(testEnv.backendName);
+  });
+
   it('register a kernel and call it', () => {
     // Make sure the backend is loaded. Perhaps tf.getBackend
     // should call tf.backend to make sure the backend is loaded?
@@ -94,7 +99,7 @@ describeWithFlags('kernel_registry', ALL_ENVS, () => {
     tf.removeBackend('backend1');
   });
 
-  it('register same kernel on two different backends', () => {
+  it('register same kernel on two different backends', async () => {
     interface TestBackend extends KernelBackend {
       id: number;
     }
@@ -129,12 +134,12 @@ describeWithFlags('kernel_registry', ALL_ENVS, () => {
     expect(lastStorageId).toBe(-1);
 
     // Kernel was executed on the first backend.
-    tf.setBackend('backend1');
+    await tf.setBackend('backend1');
     tf.engine().runKernel('MyKernel', {}, {});
     expect(lastStorageId).toBe(1);
 
     // Kernel was executed on the second backend.
-    tf.setBackend('backend2');
+    await tf.setBackend('backend2');
     tf.engine().runKernel('MyKernel', {}, {});
     expect(lastStorageId).toBe(2);
 
@@ -144,7 +149,7 @@ describeWithFlags('kernel_registry', ALL_ENVS, () => {
     tf.unregisterKernel('MyKernel', 'backend2');
   });
 
-  it('register kernel with setup and dispose functions', () => {
+  it('register kernel with setup and dispose functions', async () => {
     const backendName = 'custom-backend';
     const kernelName = 'MyKernel';
     interface TestBackend extends KernelBackend {}
@@ -174,7 +179,7 @@ describeWithFlags('kernel_registry', ALL_ENVS, () => {
     expect(setupCalled).toBe(false);
     expect(disposeCalled).toBe(false);
 
-    tf.setBackend(backendName);
+    await tf.setBackend(backendName);
     expect(setupCalled).toBe(true);
     expect(disposeCalled).toBe(false);
 


hujiajie:
I think this goes unnoticed before because `yarn test` is seldom used for WebGPU backend testing. Instead [this config](https://github.com/tensorflow/tfjs/blob/a7cb36a3b3bc92a7593217a17f1f0e82a2660f2d/tfjs-backend-webgpu/karma.conf.js#L78) is picked up by `yarn karma start --browsers=chrome_webgpu`, where randomness is disabled. The WASM backend does not use [karma_template.conf.js](https://github.com/tensorflow/tfjs/blob/a7cb36a3b3bc92a7593217a17f1f0e82a2660f2d/tools/karma_template.conf.js#L127) either.

tfjs-core/src/kernel_registry_test.ts
```diff
@@ -17,11 +17,11 @@
 
 import * as tf from './index';
 import {KernelBackend} from './index';
-import {ALL_ENVS, describeWithFlags} from './jasmine_util';
+import {ALL_ENVS, describeWithFlags, TestEnv} from './jasmine_util';
 import {KernelFunc, TensorInfo} from './kernel_registry';
 import {expectArraysClose} from './test_util';
 
-describeWithFlags('kernel_registry', ALL_ENVS, () => {
+describeWithFlags('kernel_registry', ALL_ENVS, (testEnv: TestEnv) => {
```
qjia7, line 24: Since this test set will call `tf.removeBackend` in most situations. How about adding below snippet here. Then we don't need to concern whether to add `await tf.setBackend(testEnv.backendName);` at the end of it.
```
  beforeAll(() => {
   // Comment the reason.
    await tf.setBackend(testEnv.backendName);
  });
```

qjia7: Ah, Just realize that it may only be called once. But we need to call it before each of iteration. Please ignore this comment.

qjia7 (approved):
LGTM, thanks.

tfjs-core/src/kernel_registry_test.ts
```diff
@@ -129,22 +129,25 @@ describeWithFlags('kernel_registry', ALL_ENVS, () => {
...
     tf.engine().runKernel('MyKernel', {}, {});
     expect(lastStorageId).toBe(2);
 
     tf.removeBackend('backend1');
     tf.removeBackend('backend2');
     tf.unregisterKernel('MyKernel', 'backend1');
     tf.unregisterKernel('MyKernel', 'backend2');
+
+    // Revert backend mutation.
+    await tf.setBackend(testEnv.backendName);
```
mattsoulanille, line 147: I think it's better to reset the backend in `afterEach`. Right now, if an error is thrown in this test, the backend will not be reset. 

Alternatively, you can wrap the entire test in `try { [test] } finally { [reset backend] }`.

tfjs-core/src/kernel_registry_test.ts
```diff
@@ -186,6 +189,9 @@ describeWithFlags('kernel_registry', ALL_ENVS, () => {
     expect(disposeCalled).toBe(true);
 
     tf.unregisterKernel(kernelName, backendName);
+
+    // Revert backend mutation.
+    await tf.setBackend(testEnv.backendName);
```
mattsoulanille, line 194: Same here. Backend may not be reset if the test throws an error.

tfjs-core/src/kernel_registry_test.ts
mattsoulanille: This is a good idea. We can use `afterEach` instead of `beforeAll`, and it will run after each test.

mattsoulanille (changes requested):

hujiajie:
FYI, testing is a little blocked by https://bugs.chromium.org/p/chromium/issues/detail?id=1332926

qjia7:
> FYI, testing is a little blocked by https://bugs.chromium.org/p/chromium/issues/detail?id=1332926

Can you use the last known good revision (64bit): https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win_x64/1007011/ to do the test?

hujiajie:
> > FYI, testing is a little blocked by https://bugs.chromium.org/p/chromium/issues/detail?id=1332926
> 
> Can you use the last known good revision (64bit): https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win_x64/1007011/ to do the test?

Then there are at least two other problems:

1. Recent breaking update: https://github.com/tensorflow/tfjs/commit/c7fdb43933de82f9940b349a4496676dce8f3a2c
2. A few tests depend on proprietary video codecs, which are only enabled in Chrome but not in Chromium (BTW I wish this could be improved by converting to royalty free formats).

hujiajie:
**The good news**

https://bugs.chromium.org/p/chromium/issues/detail?id=1332926 was fixed.

**The bad news**

`yarn test --test_env=CHROME_CANARY_BIN` in tfjs-backend-webgpu broke again:

```
...
09 06 2022 02:09:10.067:INFO [karma-server]: Karma v6.3.16 server started at http://localhost:9876/
09 06 2022 02:09:10.068:INFO [launcher]: Launching browsers ChromeHeadless with concurrency unlimited
09 06 2022 02:09:10.074:INFO [launcher]: Starting browser ChromeHeadless
09 06 2022 02:09:10.075:ERROR [launcher]: No binary for ChromeHeadless browser on your platform.
  Please, set "CHROME_BIN" env variable.
...
```

Possible culprit (@mattsoulanille ): a53aae8fd99762ab41a25ac362ece95b4bbb8cf6

mattsoulanille:
Nice catch, @hujiajie! That should be fixed by #6515.

hujiajie:
Thanks! PR has been updated, PTAL.

qjia7 (approved):
LGTM, thanks.

mattsoulanille (approved):
LGTM

PR #6525, opened 2022-06-12
[webgpu] Migrate sparseToDense to the atomic-based kernel

hujiajie:
PERF

The sparseToDense op takes an optional default value. Unlike scatterNd, the output cannot be initialized with fill(), since the default value is a scalar tensor (which could be the result of a previous op) than a scalar number. The (horrible!) workaround here is to broadcast the value with tile().

The other challenge is if the kernel should discard the original value at index or accumulate on that. The magic is performed by splitting the op into two "scatter" steps: 1) replace the default value with 0 or add 0 to 0, and 2) add the input sparse values to 0 or whatever. This avoids a bitmap for recording whether the output element at index has been updated by another invocation.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6525)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/SparseToDense.ts
+++ b/tfjs-backend-webgpu/src/kernels/SparseToDense.ts
@@ -19,9 +19,12 @@ import {backend_util, KernelConfig, KernelFunc, Rank, SparseToDense, SparseToDen
 
 import {WebGPUBackend} from '../backend_webgpu';
 import {scatterImplCPU} from '../kernel_utils/shared';
-import {ScatterProgram} from '../scatter_webgpu';
+import {ScatterOptimizedProgram} from '../scatter_optimized_webgpu';
 
+import {identity} from './Identity';
 import {reshape} from './Reshape';
+import {tile} from './Tile';
+import {zerosLike} from './ZerosLike';
 
 export function sparseToDense(args: {
   inputs: SparseToDenseInputs,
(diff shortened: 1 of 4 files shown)


tfjs-backend-webgpu/src/kernels/SparseToDense.ts
```diff
@@ -46,24 +49,91 @@ export function sparseToDense(args: {
...
+    {type: 'int32', data: [size]},
   ];
-  const program = new ScatterProgram(
-      numUpdates, sliceRank, sparseIndices.shape.length,
-      sparseValues.shape.length, strides, [outputSize, 1], sumDupeIndices);
+  const program = new ScatterOptimizedProgram(
+      flattenValues.shape, sliceRank, flattenIndices.shape.length,
+      flattenValues.shape.length, strides, flattenShape, type);
+
+  // First replace the default value with 0 at indices.
```
qjia7, line 114: Is it better to write a new program (fillZeroAtIndices?) to process this pass to fill zero at indices?
Two advantages: 1) Don't need to add extra complexity for original `ScatterOptimizedProgram`. 2) For this pass itself, it may have better performance.

hujiajie: I'm not sure. Probably the main tradeoff is

```
shader compilation * 1 + unoptimized shader execution * N
```

vs

```
shader compilation * 2 + optimized shader execution * N
```

qjia7: I vote this one.
> 
> ```
> shader compilation * 2 + optimized shader execution * N
> ```

PR #6528, opened 2022-06-12
[wasm] Add sparseToDense kernel

### LINKED ISSUE #4824: Backend WASM does not implement kernel SparseToDense
attempting to run nanodet object detection model  

converted from pytorch - conversion notes and example of model usage are at:  
<https://gist.github.com/vladmandic/d0115e3899860391382d58550bd49b3f>

```log
engine.ts:540 Uncaught (in promise) Error: Kernel 'SparseToDense' not registered for backend 'wasm'
    at Engine.runKernel (engine.ts:540)
    at sparseToDense_ (sparse_to_dense.ts:88)
    at sparseToDense__op (operation.ts:51)
    at executeOp15 (slice_join_executor.ts:183)
    at operation_executor.ts:92
    at engine.ts:467
    at Engine.scopedRun (engine.ts:478)
    at Engine.tidy (engine.ts:465)
    at tidy (globals.ts:192)
    at operation_executor.ts:91
```

same graph model works without issues in nodejs with `tensorflow` backend and browser with `webgl` backend

environment: tfjs 3.3.0 on ubuntu 20.10 with chrome 88

kon72:
Closes #4824

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6528)


kinsei0916:
Closes #4824

tfjs-backend-wasm/src/cc/scatter_impl.cc
```diff
@@ -0,0 +1,88 @@
...
+      if (sum_dupe_indices) {
+        *out_buf_ptr += *updates_ptr;
+      } else {
+        *out_buf_ptr = *updates_ptr;
+      }
+
+      out_buf_ptr++;
+      if (!update_as_scalar) {
+        updates_ptr++;
+      }
```
mattsoulanille, line 55: I don't fully understand the purpose of this. It looks like it's only used if `sparse_values_rank == 0` (SparseToDense.cc:35), in which case it assigns all the values of the output tensor to `*updates_ptr` (without incrementing).

Since this only happens when the rank of the sparse tensor is zero, wouldn't this only ever assign one value? In that case, I don't think it's necessary, since the loop will only run once.

On the other hand, it's [implemented the same way in the CPU backend](https://github.com/kinsei0916/tfjs/blob/d04e6d02360d1cf0ca4dfeab8181e92430e7d709/tfjs-backend-cpu/src/kernels/Scatter_impl.ts#L68-L71), so I think this is fine.

mattsoulanille (approved):
LGTM. Thank you!

tfjs-backend-wasm/src/cc/scatter_impl.cc
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    out_buf_ptr -= (flattened_index * slice_size + slice_size);
```
ahmedsabie, line 58: This is very confusing since you are technically resetting the pointer but using decrement. Can you leave out_buf_ptr unchanged and create a new variable that stores the start of the current iteration output (i.e. out_buf_ptr + flattened_index * slice_size)  and only increment the temporary so this decrement reset is no longer necessary?

ahmedsabie (changes requested):
See previous comment

sklum:
Would love to get this merged as it's also a pain point for my work.  Given the lack of activity, is it possible to merge despite @ahmedsabie nit?

If not, I'm happy to try to resolve in a separate PR, although I've never worked with the C++ implementation and Bazel.  Are there up to date instructions on building from source anywhere?  My (albeit brief) search came up empty.

mattsoulanille:
Sorry I didn't follow up on this earlier. Ahmed has left the TFJS team, so I'll assign this to someone else.

tfjs-backend-wasm/src/cc/kernels/SparseToDense.cc
```diff
@@ -0,0 +1,92 @@
...
+                   const size_t sparse_values_rank,
+                   const size_t default_value_id, const DType dtype,
+                   const size_t slice_rank, const size_t num_updates,
+                   const size_t slice_size, const size_t* strides_ptr,
+                   const size_t output_size, const size_t out_id) {
+  auto& sparse_indices_info = backend::get_tensor_info(sparse_indices_id);
+  auto& sparse_values_info = backend::get_tensor_info(sparse_values_id);
+  auto& default_value_info = backend::get_tensor_info(default_value_id);
+  const std::vector<size_t>& strides =
+      std::vector<size_t>(strides_ptr, strides_ptr + slice_rank);
```
chunnienc, line 62: `const std::vector<size_t>(strides_ptr, strides_ptr + slice_rank);`

It's bad to keep the lvalue ref of a temporary value.

chunnienc (changes requested):

sklum:
Thanks for the quick follow up @mattsoulanille and review @chunnienc.

@kon72 do you still have interest / availability to address the review?

sklum:
Alright @mattsoulanille, I've [opened a PR](https://github.com/kon72/tfjs/pull/1) on @kon72's fork that addresses @chunnienc's change request.  Are you able to merge that PR to the fork as you have maintainer rights?  Otherwise I'm happy to open a clean PR with my fork.

mattsoulanille:
@sklum Thanks for the fix. I can't merge the PR on the fork, but I cherry-picked the commit and pushed it here for @chunnienc to take a look at.

kon72:
Hello all,

Sorry for the long delay and thank you for reviewing and following up on this PR.
I will take a look and address these reviews quickly.

tfjs-backend-wasm/src/cc/kernels/SparseToDense.cc
kon72: I think it's valid because the lifetime of the temporary object will be extended by binding to const ref.
Anyway, the reference here was completely redundant and your suggested change is much clearer. Thank you!

tfjs-backend-wasm/src/cc/scatter_impl.cc
kon72: Done, thank you!

tfjs-backend-wasm/src/cc/scatter_impl.cc
kon72: [TensorFlow doc](https://www.tensorflow.org/api_docs/python/tf/raw_ops/SparseToDense) states that if `sparse_values` is scalar, the value is assigned to all sparce indices.

So as far as I understand, `sparse_indices` could take more than one index while `sparse_values` is a scalar, and the loop will in turn run more than once.

chunnienc (approved):

PR #6552, opened 2022-06-21
[webgpu] Migrate sparseToDense to the atomic-based kernel

hujiajie:
PERF

The sparseToDense op takes an optional default value. Unlike scatterNd, the output cannot be initialized with fill(), since the default value is a scalar tensor (which could be the result of a previous op) than a scalar number. The (horrible!) workaround here is to broadcast the value with tile().

The other challenge is if the kernel should discard the original value at index or accumulate on that. The magic is performed by splitting the op into two "scatter" steps: 1) replace the default value with 0, and 2) add the input sparse values to 0 or whatever. This avoids a bitmap for recording whether the output element at index has been updated by another invocation.

Closes #6525

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6552)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/SparseToDense.ts
+++ b/tfjs-backend-webgpu/src/kernels/SparseToDense.ts
@@ -19,9 +19,11 @@ import {backend_util, KernelConfig, KernelFunc, Rank, SparseToDense, SparseToDen
 
 import {WebGPUBackend} from '../backend_webgpu';
 import {scatterImplCPU} from '../kernel_utils/shared';
-import {ScatterProgram} from '../scatter_webgpu';
+import {ScatterOptimizedProgram} from '../scatter_optimized_webgpu';
 
+import {identity} from './Identity';
 import {reshape} from './Reshape';
+import {tile} from './Tile';
 
 export function sparseToDense(args: {
   inputs: SparseToDenseInputs,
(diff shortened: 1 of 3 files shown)


tfjs-backend-webgpu/src/scatter_optimized_webgpu.ts
```diff
@@ -64,45 +66,50 @@ export class ScatterOptimizedProgram implements WebGPUProgram {
...
-     for (; !exchanged;) {
-       let newValueF32 = bitcast<f32>(oldValue) + updateValue;
-       let newValue = bitcast<i32>(newValueF32);
-       let res = atomicCompareExchangeWeak(&(result[flatIndex]), oldValue, newValue);
-       oldValue = res.old_value;
-       exchanged = res.exchanged;
-     }
-     `;
+    const atomicRMW = (ptr: string, val: string) => {
+      const atomicAddSnippet = `
```
qjia7, line 93: Maybe keep the original `atomicAdd` path, which is better if the data is int32?
```
const atomicAddSnippet = this.type === 'int32' ?
  `atomicAdd(${ptr}, i32(val));` :
  `
        {
          var oldBits = 0;
          var newBits = bitcast<i32>(${val});
          loop {
            let info = atomicCompareExchangeWeak(${ptr}, oldBits, newBits);
            if (info.exchanged) {
              break;
            }
            oldBits = info.old_value;
            let oldValue =
                bitcast<${mapToWgslTypes(this.type, false)}>(oldBits);
            let newValue = oldValue + (${val});
            newBits = bitcast<i32>(newValue);
          }
        }
      `;

File: tfjs-backend-webgpu/src/scatter_optimized_webgpu.ts [hunk 2]
```diff
@@ -64,45 +66,50 @@ export class ScatterOptimizedProgram implements WebGPUProgram {
...
-    } else if (this.updatesRank === 2) {
-      updatesString = 'coords[0], coords[1]';
+    } else if (this.dispatchLayout.x.length === 2) {
       outCoordsString = 'vec2<i32>(flattenedIndex, coords[1])';
       getUpdatesCoordsFromFlatIndex = `
       fn getUpdatesCoordsFromFlatIndex(index : i32) -> vec2<i32> {
-        let d0 = index / uniforms.updatesShape[1];
-        let d1 = index - d0 * uniforms.updatesShape[1];
+        let d0 = index / uniforms.outShape[1];
+        let d1 = index - d0 * uniforms.outShape[1];
```
qjia7, line 83: Why changes it from `uniforms.updatesShape` to `uniforms.outShape[1]`?

tfjs-backend-webgpu/src/scatter_optimized_webgpu.ts
```diff
@@ -64,45 +66,50 @@ export class ScatterOptimizedProgram implements WebGPUProgram {
...
+            }
+            oldBits = info.old_value;
+            let oldValue =
+                bitcast<${mapToWgslTypes(this.type, false)}>(oldBits);
+            let newValue = oldValue + (${val});
+            newBits = bitcast<i32>(newValue);
+          }
+        }
+      `;
+      const atomicStoreSnippet = `atomicStore(${ptr}, bitcast<i32>(${val}));`;
```
qjia7, line 110: Why not just store `0` so that we don't need to upload a zero tensor as input?

tfjs-backend-webgpu/src/scatter_optimized_webgpu.ts
hujiajie: This is a hack. When `updates` is a scalar tensor, coords can no longer be derived from its shape as the tensor is "compressed", but `outShape[1]` happens to provide the slice size here.

tfjs-backend-webgpu/src/scatter_optimized_webgpu.ts
hujiajie: This fits my taste better: the kernel does something (accumulation or overwriting) according to indices and values given by the user.

tfjs-backend-webgpu/src/scatter_optimized_webgpu.ts
hujiajie: Let me wait for more inputs to see if it worth the complexity.

tfjs-backend-webgpu/src/scatter_optimized_webgpu.ts
qjia7: How do you ensure that `uniforms.updatesShape[1]` is equal to `uniforms.outShape[1]`?

qjia7: It's weird to use `uniforms.outShape[1]` instead of `uniforms.updatesShape[1]` for `fn getUpdatesCoordsFromFlatIndex`.
When `updates` is a scalar tensor, you probably need to do it like below. In your hack, when `updates` is a scalar tensor, getUpdatesCoordsFromFlatIndex's result is even not needed.
 ```
switch((this.updatesRank) {
case 0:
      outCoordsString = 'flattenedIndex';
      getUpdatesCoordsFromFlatIndex = '';
case 1:
      outCoordsString = 'flattenedIndex';
      getUpdatesCoordsFromFlatIndex = `
      fn getUpdatesCoordsFromFlatIndex(index : i32) -> i32 {
        return index;
      }
case 2:
      outCoordsString = 'vec2<i32>(flattenedIndex, coords[1])';
      getUpdatesCoordsFromFlatIndex = `
      fn getUpdatesCoordsFromFlatIndex(index : i32) -> vec2<i32> {
        let d0 = index / uniforms.updatesShape[1];
        let d1 = index - d0 * uniforms.updatesShape[1];
        return vec2<i32>(d0, d1);
      }
default: throw error
}

hujiajie: `flattenedIndex` also depends on `coords[0]`. Can you clarify how to handle this?

tfjs-backend-webgpu/src/kernels/SparseToDense.ts
```diff
@@ -46,24 +48,73 @@ export function sparseToDense(args: {
...
   ];
-  const program = new ScatterProgram(
-      numUpdates, sliceRank, sparseIndices.shape.length,
-      sparseValues.shape.length, strides, [outputSize, 1], sumDupeIndices);
 
-  const res = backend.runWebGPUProgram(
-      program, [sparseValues, sparseIndices, defaultValue], sparseValues.dtype,
-      uniformData);
+  {
+    // First replace the default value with 0 at indices.
```
qjia7, line 88: In API doc, it says
`If sparseIndices is scalar dense[i] = (i == sparseIndices ? sparseValues : defaultValue)`.
In this case, we don't need three trips to finish sparseToDense. Instead, we can iterate output `dense` and check whether the current index is equal to `sparseIndices`, and store the right value to output. Can you add a comment that there is an optimization opportunity for such kind of situation in case someone wants to do it in future?

tfjs-backend-webgpu/src/scatter_optimized_webgpu.ts
qjia7: Please change it as we discussed offline.

tfjs-backend-webgpu/src/kernels/SparseToDense.ts
hujiajie: I would discourage making things too fragmented, but this reminds me a change like this:

1. Fill output tensor with the default value.
2. Replace the default value with ~~`zero`~~ `sparseValues` at indices.

tfjs-backend-webgpu/src/kernels/SparseToDense.ts
```diff
@@ -46,24 +48,85 @@ export function sparseToDense(args: {
...
-      sparseValues.shape.length, strides, [outputSize, 1], sumDupeIndices);
 
-  const res = backend.runWebGPUProgram(
-      program, [sparseValues, sparseIndices, defaultValue], sparseValues.dtype,
-      uniformData);
+  switch (numUpdates) {
+    case 0:
+      break;
+    case 1:
+      if (true) {
```
qjia7, line 91: nit: Can you remove `if (true) ` and also some other similar places?

qjia7 (approved):
LGTM with one nit.

tfjs-backend-webgpu/src/kernels/SparseToDense.ts
hujiajie: Auto format looks bad to me without this.

gyagp (approved):
LGTM

PR #6346, opened 2022-04-23
Update TypeScript to 4.8.4

mattsoulanille:
This PR updates TypeScript from 3.5.3 to 4.8.4.

- [x] tfjs-converter: ~Tests need to be rewritten to avoid [spying on properties of imported objects](https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/src/operations/executors/basic_math_executor_test.ts#L52-L59)~ This is done.
- [x] tfjs-vis: ~There's a test failure with TypeScript 4.4.2~: This is now fixed
```
Chrome 92.0.4515.159 (Linux x86_64) perClassAccuracy renders histogram FAILED
	Error: Expected 0 to be 1.
	    at <Jasmine>
	    at src/show/tensor_test.ts:32:61 <- src/show/tensor_test.js:58:65
	    at Generator.next (<anonymous>)
	    at fulfilled (src/show/tensor_test.js:40:58)
```

- [x] tfjs (union package): Depends on converter
- [x] tfjs-node(-gpu), e2e: Depends on tfjs union.
- [x] tfjs-backend-webgpu: Done.
- [ ] tfjs-react-native: ~Needs to be tested.~ react-native may need other fixes before TS4 and should be updated in a different PR.

These changes will let us add the `useUnknownInCatchVariables` and `noImplicitOverride` flags to the tsconfig (in a separate PR).

Closes #5561, which has gone through review, but was not merged because of breaking changes introduced by typescript 4.4.2.

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6346)

PR #5561, opened 2021-08-31
[DRAFT] Update TypeScript to 4.4.2

mattsoulanille:
This PR updates TypeScript from 3.5.3 to 4.4.2. It is still a work in progress:

- [x] tfjs-converter: ~Tests need to be rewritten to avoid [spying on properties of imported objects](https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/src/operations/executors/basic_math_executor_test.ts#L52-L59)~ This is done.
- [x] tfjs-vis: ~There's a test failure with TypeScript 4.4.2~: This is now fixed
```
Chrome 92.0.4515.159 (Linux x86_64) perClassAccuracy renders histogram FAILED
	Error: Expected 0 to be 1.
	    at <Jasmine>
	    at src/show/tensor_test.ts:32:61 <- src/show/tensor_test.js:58:65
	    at Generator.next (<anonymous>)
	    at fulfilled (src/show/tensor_test.js:40:58)
```

- [x] tfjs (union package): Depends on converter
- [x] tfjs-node(-gpu), e2e: Depends on tfjs union.
- [x] tfjs-backend-webgpu: Done.
- [ ] tfjs-react-native: Haven't tried this package yet.

These changes will let us add the `useUnknownInCatchVariables` and `noImplicitOverride` flags to the tsconfig (in a separate PR).

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5561)

PR #6563, opened 2022-06-23
[converter] Avoid spying on properties of the 'tfOps' module

mattsoulanille:
According to the ESModule standard, properties of modules are immutable. TypeScript 4 will enforce this rule. This in particular affects tests for executors in tfjs-converter, in which we often spy on tfOps.

This PR removes all instances of `spyOn(tfOps,...)` and replaces them with a separate `spyOps` mock / fake which is passed to the `executeOp` function.

It also removes spying on the `io` module in `graph_model_test.ts`.

This PR was part of the larger TS4 upgrade PR (https://github.com/tensorflow/tfjs/pull/6346, #5561), but I'm splitting that PR into pieces that can be merged while we're still using TS3 because it's too large to keep up-to-date.

This PR also bumps `lib` to `es2019` in the root tsconfig to allow using `Object.fromEntries`. This shouldn't affect the code we ship since it's still compiled to the es2017 target.

Note that this PR does not bump TypeScript to version 4. It leaves it at 3.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6563)

### DIFF
--- a/tfjs-converter/package.json
+++ b/tfjs-converter/package.json
@@ -31,7 +31,8 @@
     "opn": "~5.1.0",
     "protobufjs": "~6.11.3",
     "ts-node": "~8.8.2",
-    "typescript": "3.5.3"
+    "typescript": "3.5.3",
+    "yalc": "~1.0.0-pre.50"
   },
   "scripts": {
     "build": "bazel build :tfjs-converter_pkg",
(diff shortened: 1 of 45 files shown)


mattsoulanille:
@pyu10055 These changes should be nearly identical to the ones in #5561, which you've already reviewed.

pyu10055 (approved):
Thanks! LGTM

> Reviewed 1 of 1 files at r1, 44 of 44 files at r2, 3 of 3 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6563#-:-N5gPZBb77LN0ual_tZg:b70hlkx)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @Linchenn)

Linchenn (approved):

PR #6591, opened 2022-06-30
Enable WebGPU testing on Linux

mattsoulanille:
Enable testing the WebGPU backend on Linux using Chrome Dev (105) instead of Chrome Canary (even though `base: 'ChromeCanary'` is listed, it launches Chrome Dev on Linux). Fix incorrect passing of flags to Chrome. Flags enabled in `chrome://flags` need to appear between `--flag-switches-begin` and `--flag-switches-end`.

Note that not all tests are passing on Linux. This could be due to Chrome Dev being behind Chrome Canary, or due to a lack of testing on Linux before this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6591)


Linchenn (approved):

tools/karma_template.conf.js
```diff
@@ -77,8 +77,11 @@ const CUSTOM_LAUNCHERS = {
   chrome_webgpu: {
     base: 'ChromeCanary',
     flags: [
+      '--disable-dawn-features=disallow_unsafe_apis',
+      '--flag-switches-begin',
       '--enable-unsafe-webgpu',
-      '--disable-dawn-features=disallow_unsafe_apis'
+      '--enable-features=Vulkan',
```
qjia7, line 83: WebGPU on Linux requires command-line flag `--enable-features=Vulkan,UseSkiaRenderer`. So `UseSkiaRenderer` is missed? And this flag is only needed for Linux. We shouldn't add it on Windows. Can we distinguish them based on platforms?

mattsoulanille: Thanks for the review! I added a [select](https://bazel.build/concepts/platforms-intro#select) to make sure the flag is added only on Linux. I also added `UseSkiaRenderer`.

tfjs-backend-webgpu/BUILD.bazel
```diff
@@ -107,6 +107,9 @@ tfjs_web_test(
         '{"WEBGPU_CPU_FORWARD": false}',
     ],
     browsers = [],
-    local_browser = "chrome_webgpu",
+    local_browser = select({
+        "@bazel_tools//src/conditions:windows": "chrome_webgpu_windows",
```
qjia7, line 111: nit: Do we need to add the macOS?

qjia7 (approved):
LGTM with a nit. Thanks.

mattsoulanille:
While testing this PR locally, I found out that "[webgpu] Replace module-scope 'let' with 'const' (#6602)" has caused most of the tests to fail on Linux (Chrome 105.0.5148.2). When it's reverted, only 8 tests fail. This could easily be related to my environment, but I thought I should mention it anyway.

tfjs-backend-webgpu/BUILD.bazel
mattsoulanille: Good point. I renamed to Windows version to just `chrome_webgpu` and set it as the default condition. macOS will use it.

qjia7 (approved):
LGTM, thanks.

> While testing this PR locally, I found out that "[webgpu] Replace module-scope 'let' with 'const' (#6602)" has caused most of the tests to fail on Linux (Chrome 105.0.5148.2). The failure mode is that all the values returned are zero. When it's reverted, only 8 tests fail. This could easily be related to my environment, but I thought I should mention it anyway.

Thanks Matt. The `wgsl` spec change of using module-scope `const` instead of `let` is a very new change. So I suppose you need to update your chrome. My chrome canary is 105.0.5164.0. And can you help to file a bug for the 8 tests failures? We will take a look, thanks

gyagp:
On Linux, the most frequently updated channel is Dev, and it doesn't include the latest WebGPU changes. Maybe you need to wait for a few days for the recent breaking changes to be rolled into Dev. 
We just enabled the tests on Linux today, and @xhcao will take care of all the failures there.

PR #6623, opened 2022-07-14
webgpu: Refactor matMul read/write method

qjia7:

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6623)


xhcao (approved):
LGTM, thank you.

tfjs-backend-webgpu/src/matmul_test.ts
```diff
@@ -1005,4 +1023,773 @@ describeWebGPU('matmul', () => {
...
+};
+
+for (let i = 0; i < MatMulProgramType.MatMULMax; i++) {
+  describeWithFlags(`matmul with program type ${i}`, MATMUL_ENVS, matmulTest);
+  // Skip MatMulSplitKProgram since it doesn't support batch > 1;
+  if (i !== MatMulProgramType.MatMulSplitKProgram) {
+    describeWithFlags(
+        `matmulBatch with program type ${i}`, MATMUL_ENVS, matmulBatchTest);
+  }
+}
```
mattsoulanille, line 1795: Another approach is to use a different environment for each MatMulProgramType. If you want to test all the WebGPU tests against the different MatMul programs, then this other approach is likely better. If you only want to test the matmul tests with different matmul programs, then your approach looks good, but skip to the end of this comment for some potential improvements.

To use a different environment for each `MatMulProgramType`, you can [register a test environment](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/jasmine_util.ts#L240-L255) for each value of the `MatMulProgramType` enum:
```ts
// backend_webgpu_test_registry.ts
for (let i = 0; i < MatMulProgramType.MatMULMax; i++) {
  registerTestEnv({
    name: `webgpu ${MatMulProgramType[i]}`,
    backendName: 'webgpu',
    flags: {
      'WEBGPU_TESTED_MATMUL': i,
    },
    isDataSync: false,
  }
}
```

Then, `describeWithFlags` will [run the specs on all of the registered environments](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/jasmine_util.ts#L218-L220) (as long as the environment satisfies the constraints).

Then, for matmul tests, you would just need to write `describeWithFlags('matmulBatch', ALL_ENVS, () => { /* tests */ })`.

However, this approach will also run all the other tests with these MatMul environments, meaning all the tests would run 5 times. If this is not what you want, you can create a constraint that requires `WEBGPU_TESTED_MATMUL` to be -1 and apply it to all the other tests:
```ts
export const DEFAULT_MATMUL_ENVS: Constraints = {
  flags: {'WEBGPU_TESTED_MATMUL': -1}
};
```
However, this may require a lot of refactoring. 

Custom environments are more suited to flags that affect all the tests instead of just a small subset of them. Another approach which is better for flags that have a small effect (as appears to be the case here) is to just set the flag in a `beforeAll` and then reset it in `afterAll` (i.e. what you've done here).

```ts
for (let i = 0; i < MatMulProgramType.MatMULMax; i++) {
  describeWithFlags(`matmul ${MatMulProgramType[i]}`, ALL_ENVS, () => {
    let savedMatmulFlag: number;
    beforeAll(() => {
      savedMatmulFlag = tf.env().get('WEBGPU_TESTED_MATMUL') as number;
      tf.env().set('WEBGPU_TESTED_MATMUL', i);
    });

    afterAll(() => {
      tf.env().set('WEBGPU_TESTED_MATMUL, savedMatmulFlag);
    });

    ...
  }
}
```
With this approach, you don't need to create a custom `Constraint` or custom `TestEnv`s. Everything is controlled by the `for` loop and `beforeAll` / `afterAll`. If you don't need to run all the WebGPU tests with different values of `MatMulProgramType` and only want to run the MatMul tests with them, then I think this is the best approach.

tfjs-backend-webgpu/src/backend_webgpu_test_registry.ts
```diff
@@ -16,7 +16,11 @@
  */
 
 // tslint:disable-next-line: no-imports-from-dist
-import {registerTestEnv} from '@tensorflow/tfjs-core/dist/jasmine_util';
+import {Constraints, registerTestEnv} from '@tensorflow/tfjs-core/dist/jasmine_util';
+
+export const MATMUL_ENVS: Constraints = {
+  flags: {'WEBGPU_TESTED_MATMUL': -1}
+};
```
mattsoulanille, line 23: You might not need this constraint. See my comment below.

tfjs-backend-webgpu/src/matmul_test.ts
```diff
@@ -17,9 +17,27 @@
...
+// tslint:disable-next-line: no-imports-from-dist
+import {describeWithFlags} from '@tensorflow/tfjs-core/dist/jasmine_util';
+
+import {MATMUL_ENVS} from './backend_webgpu_test_registry';
+import {MatMulProgramType} from './webgpu_util';
+
+const {expectArraysClose} = test_util;
+const MATMUL_SHARED_DIM_THRESHOLD = 1000;
+let type0 = MatMulProgramType.MatMulReduceProgram;
+let type1 = MatMulProgramType.MatMulReduceProgram;
```
mattsoulanille, line 29: You can avoid these variables. See my comment below for more details.

tfjs-backend-webgpu/src/matmul_test.ts
```diff
@@ -1005,4 +1023,773 @@ describeWebGPU('matmul', () => {
...
+  let savedMatmulFlag = -1;
+  beforeAll(() => {
+    savedMatmulFlag = tf.env().get('WEBGPU_TESTED_MATMUL') as number;
+    if (type1 === MatMulProgramType.MatMulSplitKProgram) {
+      // Skip MatMulSplitKProgram since it doesn't support batch > 1;
+      type1++;
+    }
+    tf.env().set('WEBGPU_TESTED_MATMUL', type1);
+    type1++;
+  });
```
mattsoulanille, line 1399: It is better to pass in which `MatMulProgramType` is used with a closure instead of storing it in `type1`. See above comment on the other matmul test.

tfjs-backend-webgpu/src/flags_webgpu.ts
```diff
@@ -33,6 +33,11 @@ ENV.registerFlag('WEBGPU_CPU_FORWARD', () => true);
  */
 ENV.registerFlag('WEBGPU_MATMUL_WORK_PER_THREAD', () => 4);
 
+/**
+ * This flag is used to test different version of matmul programs.
+ */
+ENV.registerFlag('WEBGPU_TESTED_MATMUL', () => -1);
```
mattsoulanille, line 39: Would `WEBGPU_MATMUL_PROGRAM_TYPE` be a better name for this flag?

mattsoulanille (changes requested):
Hi @qjia7. I took a look at your approach to testing different matmul programs. If you only need to run the matmul tests (and not all the tests) against the different matmul programs, then your approach looks good, and you probably don't need to use `Constraints` or `TestEnv`s. See my long comment for more details.

tfjs-backend-webgpu/src/matmul_test.ts
```diff
@@ -17,9 +17,27 @@
...
+const matmulTest = () => {
+  let savedMatmulFlag = -1;
+  beforeAll(() => {
+    savedMatmulFlag = tf.env().get('WEBGPU_TESTED_MATMUL') as number;
+    tf.env().set('WEBGPU_TESTED_MATMUL', type0);
+    type0++;
+  });
+  afterAll(() => {
+    tf.env().set('WEBGPU_TESTED_MATMUL', savedMatmulFlag);
+  });
```
mattsoulanille, line 39: This was missing from my last review:

Instead of setting WEBGPU_TESTED_MATMUL to type0 and incrementing it each time `matmulTest` is called, it would be better to pass the MatMulProgramType in a closure.

```ts
function matmulTest(programType: MatMulProgramType) {
  return () => {
    let savedMatmulFlag = -1;
    beforeAll(() => {
      savedMatmulFlag = tf.env().get('WEBGPU_TESTED_MATMUL') as number;
      tf.env().set('WEBGPU_TESTED_MATMUL', programType);
    });
    afterAll(() => {
      tf.env().set('WEBGPU_TESTED_MATMUL', savedMatmulFlag);
    });
    ...
  }
}
```

qjia7: Thanks @mattsoulanille . Great suggestions! Previously, I use `function matmulTest(programType: MatMulProgramType) {...}` but didn't have `return () => {}` to wrap the body. It will report an error to me. I thought I couldn't pass the type. So I added an ugly global variable value. Your suggestion looks better.
I have addressed all your comments. Pease take another look, thanks.

gyagp (approved):
LGTM. This brings a much better test coverage for matmul.

mattsoulanille (approved):
LGTM. The changes look great!

PR #6639, opened 2022-07-20
[WebGL] Implement packed NHWC Conv2d 

Linchenn:
Background:
Currently, WebGL's conv2d uses inefficient img2col algorithm, which needs to run 2 WebGL programs and has many duplicate reads, so this PR is trying to replace img2col algorithm for some cases. Similar to depthwise conv2d, the code is complex, so this implementation will not be used on default until the implementation is simplified. 

This PR:
1. implements packed NHWC conv2d, referring to packed NHWC depthwise's implementation. 
2. Adds 'WEBGL_EXP_CONV' flag (default as false). Only if this flag is set true, conv2d would use this implementation.
3. Adds a nightly test to test this implementation.

For reviewer, pls review this [commit](https://github.com/tensorflow/tfjs/pull/6639/commits/9889f5a0cc18a019c8dbd9a8b3a0279a9bef11b1) to see the diff from depthwise's implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6639)


pyu10055 (changes requested):
Thanks! Looking forward to see perf improvement brought by this change.

> Reviewed 1 of 1 files at r1, 4 of 4 files at r2, 2 of 2 files at r3, 2 of 2 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 2 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6639#-:-N7XFjkH02HeLiY3fE_T:b-qju7xu)* status: 0 of 1 approvals obtained (waiting on @Linchenn)

___
*[`tfjs-backend-webgl/src/conv_packed_gpu.ts` line 3 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6639#-N7XCniB8RKRrfOeK_ct:-N7XCniB8RKRrfOeK_cu:bb6ht37) ([raw file](https://github.com/tensorflow/tfjs/blob/9c1ad8aeb690bf7975edc1e823b9de171db1b6e1/tfjs-backend-webgl/src/conv_packed_gpu.ts#L3)):*
> ```XML
> /**
>  * @license
>  * Copyright 2018 Google LLC. All Rights Reserved.
> ```

2022
___
*[`tfjs-backend-webgl/src/conv_packed_gpu.ts` line 333 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6639#-N7XEFtu2iyHnNp2MZby:-N7XEFtu2iyHnNp2MZbz:bbamw6u) ([raw file](https://github.com/tensorflow/tfjs/blob/9889f5a0cc18a019c8dbd9a8b3a0279a9bef11b1/tfjs-backend-webgl/src/conv_packed_gpu.ts#L333)):*
> ```XML
>            mainLoop += `
>                wTexel = getW(r, ${colIndex + 1}, d1, d2);
>                dotProd += vec4(xC${colIndex + 1}.x, xC${colIndex + 1}.x, xC${colIndex + 1}.z, xC${colIndex + 1}.z) * vec4(wTexel.xy, wTexel.xy);
> ```

you can simplify this as xC${colIndex}.xxzz
Please apply this for other similar places as well
___
*[`tfjs-backend-webgl/src/flags_webgl.ts` line 242 at r9](https://reviewable.io/reviews/tensorflow/tfjs/6639#-N7XFeu6Dvk0AM3OExGe:-N7XFeu6Dvk0AM3OExGf:bclszs) ([raw file](https://github.com/tensorflow/tfjs/blob/f017a3ae3ec15b845958798b90fd11aad0699d82/tfjs-backend-webgl/src/flags_webgl.ts#L242)):*
> ```XML
> 
> /** Whether we will use the experimental conv op. */
> ENV.registerFlag('WEBGL_EXP_CONV', () => false);
> ```

please add the test for this flag

Linchenn (commented):
Thanks for the suggestions!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6639#-:-N7XfVAg9gItyNmBpnbe:b-yr7a7q)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

___
*[`tfjs-backend-webgl/src/conv_packed_gpu.ts` line 3 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6639#-N7XCniB8RKRrfOeK_ct:-N7XfRU6A97sp8fT4wBZ:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/9c1ad8aeb690bf7975edc1e823b9de171db1b6e1/tfjs-backend-webgl/src/conv_packed_gpu.ts#L3)):*

2022
</blockquote></details>

Done.
___
*[`tfjs-backend-webgl/src/conv_packed_gpu.ts` line 333 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6639#-N7XEFtu2iyHnNp2MZby:-N7XfSio2rgHg-sFV5Cn:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/9889f5a0cc18a019c8dbd9a8b3a0279a9bef11b1/tfjs-backend-webgl/src/conv_packed_gpu.ts#L333)):*

you can simplify this as xC${colIndex}.xxzz
Please apply this for other similar places as well
</blockquote></details>

Done.
___
*[`tfjs-backend-webgl/src/flags_webgl.ts` line 242 at r9](https://reviewable.io/reviews/tensorflow/tfjs/6639#-N7XFeu6Dvk0AM3OExGe:-N7XfTn-1maWIUITnDA_:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/f017a3ae3ec15b845958798b90fd11aad0699d82/tfjs-backend-webgl/src/flags_webgl.ts#L242)):*

please add the test for this flag
</blockquote></details>

Done.

Linchenn (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6639)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

___
*[`tfjs-backend-webgl/src/conv_packed_gpu.ts` line 72 at r10](https://reviewable.io/reviews/tensorflow/tfjs/6639#-N7XfgjmAhnfM3M43vTL:-N7XfgjmAhnfM3M43vTM:buia3ok) ([raw file](https://github.com/tensorflow/tfjs/blob/124b30622d6571fa890e6f291f91fd3f77f6823c/tfjs-backend-webgl/src/conv_packed_gpu.ts#L72)):*
> ```XML
>      mainLoop += `
>      for (int r = 0; r < ${filterHeight}; r++) {
>       for (int d1 = 0; d1 < ${convInfo.inChannels}; d1 += 2) {
> ```

Through the following experiments on Mac-chrome, find inputChannel-loop should be the inner loop, instead of the filterHeight-loop:

Cases1: input[1, 192, 192, 56] with filter[3, 3, 56, 64]
inputChannel-loop as inner loop: 29.408638880000012ms
inputChannel-loop as outer loop: 29.418943359999933ms

Cases2: input[1, 192, 192, 8] with filter[3, 3, 8, 8]
inputChannel-loop as inner loop: 0.991897599999999ms
inputChannel-loop as outer loop: 1.0030207999999998ms

Cases3: input[1, 768,  768, 8] with filter[3, 3, 8, 8]
inputChannel-loop as inner loop: 14.807828800000006ms
inputChannel-loop as outer loop: 20.644488320000043ms


Conclusion: Even though the both ways usually performance same, setting inputChannel-loop as the inner loop is always better both theoretically and practically. This deserves more experiments on more devices and parameter settings, so opened an issue https://github.com/tensorflow/tfjs/issues/6661 to follow up on this.

pyu10055 (approved):
> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6639)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @Linchenn and @pyu10055)

___
*[`tfjs-backend-webgl/src/conv_packed_gpu.ts` line 72 at r10](https://reviewable.io/reviews/tensorflow/tfjs/6639#-N7XfgjmAhnfM3M43vTL:-N7XsA1uDyWy07V_T0D5:b-mt8338) ([raw file](https://github.com/tensorflow/tfjs/blob/124b30622d6571fa890e6f291f91fd3f77f6823c/tfjs-backend-webgl/src/conv_packed_gpu.ts#L72)):*

Through the following experiments on Mac-chrome, find inputChannel-loop should be the inner loop, instead of the filterHeight-loop:

Cases1: input[1, 192, 192, 56] with filter[3, 3, 56, 64]
inputChannel-loop as inner loop: 29.408638880000012ms
inputChannel-loop as outer loop: 29.418943359999933ms

Cases2: input[1, 192, 192, 8] with filter[3, 3, 8, 8]
inputChannel-loop as inner loop: 0.991897599999999ms
inputChannel-loop as outer loop: 1.0030207999999998ms

Cases3: input[1, 768,  768, 8] with filter[3, 3, 8, 8]
inputChannel-loop as inner loop: 14.807828800000006ms
inputChannel-loop as outer loop: 20.644488320000043ms


Conclusion: Even though the both ways usually performance same, setting inputChannel-loop as the inner loop is always better both theoretically and practically. This deserves more experiments on more devices and parameter settings, so opened an issue https://github.com/tensorflow/tfjs/issues/6661 to follow up on this.
</blockquote></details>

Thank you for the quick experiment.

tfjs-backend-webgl/src/kernels/Conv2D.ts
```diff
@@ -41,6 +42,18 @@ export function conv2d(
       convInfo.strideHeight === 1 && convInfo.strideWidth === 1 &&
       (convInfo.padInfo.type === 'SAME' || convInfo.padInfo.type === 'VALID')) {
     out = conv2dByMatMul({x, filter, convInfo, backend});
+  } else if (convInfo.strideWidth <= 2 && $dataFormat === 'channelsLast'
```
qjia7, line 45: Please pay attention that we may need more strict conditions to go to this path. For example, convInfo.dilationWidth === 1, convInfo.filterWidth > 3. For such kind of conv2d, `Conv2DPackedProgram` may get more benefits. Also try the impact of large/small convInfo.inChannels. We need better to know the threshold to go to Conv2DPackedProgram path in future.

qjia7 (approved):
LGTM, Thanks.

lina128 (approved):
Thank you so much Lin!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6639#-:-N7al3rJDptCxVRoxl52:b6kz2ij)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @Linchenn and @pyu10055)

___
*[`tfjs-backend-webgl/src/conv_packed_gpu.ts` line 72 at r10](https://reviewable.io/reviews/tensorflow/tfjs/6639#-N7XfgjmAhnfM3M43vTL:-N7alPpXD8leBMXE4N24:b-87m5h0) ([raw file](https://github.com/tensorflow/tfjs/blob/124b30622d6571fa890e6f291f91fd3f77f6823c/tfjs-backend-webgl/src/conv_packed_gpu.ts#L72)):*

Thank you for the quick experiment.
</blockquote></details>

Great insight, thank you!

tfjs-backend-webgl/src/kernels/Conv2D.ts
Linchenn: Thank you Jiajia! I am not clear why we need more strict conditions here?

As we known, Conv2DPackedProgram is more efficient than the following two branches (`conv2dWithIm2Row` and `Conv2DProgram`), because it minimizes the number of 'read' instructions (such as `getX`, `getW`) in the shaders.

Currently Conv2DPackedProgram only implements NHWC format. However, NCHW format works better for larger filters, so we may implement it in future. Do you mean add more strict conditions to let NCHW-Conv2DPackedProgram to work on these cases?

(Since we are discussing the future work, I merged this PR.)

qjia7: Personally, I don't think `Conv2DPackedProgram` is always better than `conv2dWithIm2Row` even you have the condition that `convInfo.strideWidth <= 2`. But currently, almost all the models in benchmarks satisfy this condition. So you can easily test this new algorithm in benchmarks. I tested your PR locally with resnet50 model in benchmarks on Intel CFL. It becomes worse when `WEBGL_EXP_CONV` is true. So I think the conditions should be changed, at least on intel CFL. I use resnet50 because it has many non-point wise conv2d, which can better see the performance difference. That's why I suspect whether we should have more strict conditions. Have you tested some benchmarks and see obviously advantages to use `Conv2DPackedProgram` compared with `conv2dWithIm2Row`?

Linchenn: Thank you Jiajia for benchmarking this!

From my original perspective, `Conv2DPackedProgram ` aims to optimize all `non-point wise conv2d` against `conv2dWithIm2Row` (from your comments, this point of view deserves more experiments to support). The models (whose inference time is dominated by non-pointwise conv2d) are supposed to perform better if  set WEBGL_EXP_CONV as true.

`Conv2DPackedProgram` performs great for one of our internal models (whose inference time is dominated by non-pointwise conv2d) (the model is speed up from 170ms to 50ms). A comprehensive benchmarking is a great idea and I will do it and share with you later. 

I just tested bodypix with ResNet50 base on my Mac, and it just performance similar under `WEBGL_EXP_CONV` as true and as false. (The aggregated Conv2d time drops from 73.09ms to 52.11ms, when set it as true. It's wired that the final inference time keeps similar. I will take a look over this)
image
Since I do not have Intel CFL model, I am curious how worse it is for Intel CFL?

qjia7: @Linchenn Thanks for your data. I rechecked it on Intel CFL, TGL and NV RTX3090 on Windows. All of them have obvious regressions on `bodypix-ResNet50-image-0.5`.  Since `Conv2DPackedProgram` is behind a flag, it's ok now as it is. But once we plan to enable it by default. We need really to be careful about the conditions when to use it.

Take Intel TGL as an example:

- **WEBGL_EXP_CONV as true**

webgl_exp_conv_true
webgl_exp_conv_true_avg_time
webgl_exp_env_true_individual

- WEBGL_EXP_CONV as false

webgl_exp_conv_false
webgl_exp_conv_false_avg_time
webgl_exp_conv_false_individual

Linchenn: Thank you Jiajia for the detailed checking on various devices! I will check it before removing the flag and share with yoiu latter.

PR #6654, opened 2022-07-21
webgpu: dataToGPU shares same buffer with source tensor

xhcao:


tfjs-core/src/tensor.ts
```diff
@@ -378,8 +378,10 @@ export class Tensor<R extends Rank = Rank> {
    * For WebGL backend, the data will be stored on a densely packed texture.
    * This means that the texture will use the RGBA channels to store value.
    *
-   * For WebGPU backend, the data will be stored on a buffer. There is no
-   * parameter, so can not use an user defined size to create the buffer.
+   * For WebGPU backend, the source tensor and the `tensorRef` tensor share the
+   * same underlying GPUBuffer, it means that modifying one tensor data will
+   * affects the other one. There is no parameter, so can not use an user
```
qjia7, line 383: nit: I remember @pyu10055 ever said that `tensor` is immutable. So users shouldn't modify the buffer we exposed. But if it happens, we just tell user the result is unknown in doc here?

qjia7 (approved):
@pyu10055 @Linchenn This is another case that we want to support zero copy for webgpu for performance. 
But it's not that exact same with the one to create tensor from external buffer. For `dataToGPU`, the returned buffer is controlled by tfjs. We can decide when to really destroy it.

Please let me know if you have further concerns for it. Thanks.

Linchenn (commented):
It looks like WebGL backend creates a new tensor and exposes it's texture, instead of exposing input tensor's texture:

https://github.com/tensorflow/tfjs/blob/8fae9c50ae1a6503718961ba0ba7b77f38285fba/tfjs-backend-webgl/src/backend_webgl.ts#L464

WebGL backend has to do this, because it has to convert data layout from square packing to dense packing for ease of use.

For WebGPU backend, this PR would result in a different behaviors for the two backends. So, I am curious, why you want to avoid copying? Is the overhead significant for some use cases?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6654#-:-NH5lxO1C5d4VJ25Brsp:bjaf7p9)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @gyagp, @pyu10055, and @xhcao)

pyu10055 (commented):
Given the buffer copy of dataToGPU is one time cost, I am wondering how expensive it could be?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6654#-:-NH5uhqB7J8dXDns6ZOy:bf6kv0f)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @gyagp and @qjia7)

___
*[`tfjs-core/src/tensor.ts` line 383 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6654#-N7ThIo4r3ZahUsBL7ED-r1-383:-NH5uGtu0SAhwu4hMnlp:b-uoiqyn) ([raw file](https://github.com/tensorflow/tfjs/blob/415476d2bb78b1d4be1ee35cf175f21b79824c4d/tfjs-core/src/tensor.ts#L383)):*

nit: I remember @pyu10055 ever said that `tensor` is immutable. So users shouldn't modify the buffer we exposed. But if it happens, we just tell user the result is unknown in doc here?
</blockquote></details>

is it possible to mark the GPUBuffer readonly?

Issue #6664, opened 2022-07-22
fromPixels has different results on CPU and GPU backends


gyagp:
Using image, imageBitamp or video as input, fromPixels op follows below steps:
1) A 2D canvas is created
To get the drawing context of the canvas, willReadFrequently is set to true (canvas is backed by CPU) for CPU backends and false (canvas is backed by GPU) for GPU backends, so that the following data upload can be more efficient. 

2) drawImage is called to hold the content
Unfortunately, different canvas back resource leads to different Chrome rendering paths. Chrome chooses skia CPU pipeline for CPU back resource and skia GPU pipeline for GPU back resource. In general, image decoding and rendering to canvas requires color space conversion which applies matrix multiplication with floats. The difference between CPU and GPU doing float operations introduces a slight difference in the final rendering result (e.g. +/- 1 in a single channel).

3) Canvas is fed to different backends
getImageData (CPU), texImage2D(WebGL1), texSubImage2D (WebGL 2) and copyExternalImageToTexture (WebGPU) are used to serve the data from canvas to different backends.

with “tennis_standing.jpg” as input, we get below results from fromPixels op. We may notice CPU and GPU backends have some difference.

[CPU data]
Int32Array(789507) [85, 96, 61, 61, 78, 27, 43, 66, 0, 35, 62, 0, 27, 58, 0, 24, 54, 0, 21, 52, 0, 18, 50, 0, 21, 52, 0, 27, 51, 0, 28, 52, 0, 31, 55, 0, 40, 60, 1, 39, 57, 4, 27, 44, 0, 41, 57, 14, 49, 62, 11, 42, 55, 6, 50, 60, 12, 96, 106, 57, 118, 131, 72, 113, 129, 62, 120, 136, 69, 121, 140, 78, 85, 105, 43, 44, 65, 0, 31, 56, 0, 34, 57, 0, 34, 56, 0, 40, 62, 7, 32, 54, 0, 37, 61, 0, 42, 65, 0, 37, …]

[GPU data] (WebGL and WebGPU are the same)
[...]

PR #6665, opened 2022-07-22
Set willReadFrequently to true for GPU correctness test

gyagp:
If willReadFrequently is set to true, 2D canvas will always be backed by
CPU. Otherwise, it will be backed by GPU in first call, and backed by
CPU in the rest calls.
For CPU backends, including CPU and WASM, we set this to true for
perfomance. So to compare the results of GPU with CPU, we need to ensure
it's always backed by CPU.

BUG: #6664

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6665)


gyagp:
@axinging PTAL

PR #6667, opened 2022-07-23
webgpu: dont use mod as it is a reserved keyword

crowlKats:
as per https://gpuweb.github.io/gpuweb/wgsl/#reserved-words, `mod` is a reserved keyword, which is causing the webgpu implementation of deno to error. seems chrome doesnt check for these keywords yet.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6667)


qjia7 (approved):
LGTM, thanks!

gyagp (approved):
LGTM

PR #6680, opened 2022-07-26
[local-benchmark tool] Enable populate modelUrl and input shapes via URL

Linchenn:
This PR enables users add the following parameters via URL (would be helpful for custom models):
1. Model URL, `modelUrl`. For example, `modelUrl=https://storage.googleapis.com/tfjs-models/savedmodel/bodypix/mobilenet/float/075/model-stride16.json`.
2. Input shapes, `${InputName}Shape`. For example, bodypix has an input named `sub_2`, then users could add `sub_2Shape=1,1,1,3` in the URL to populate its shape.

After this PR, you could try:
https://tfjs-benchmarks.web.app/local-benchmark/index.html?benchmark=custom&backend=webgl&run=5&modelUrl=https://storage.googleapis.com/tfjs-models/savedmodel/bodypix/mobilenet/float/075/model-stride16.json&sub_2Shape=1,1,1,3&task=performance

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6680)


e2e/benchmarks/local-benchmark/index.html
```diff
@@ -496,6 +504,8 @@ <h2>TensorFlow.js Model Benchmark</h2>
         state.numProfiles = Number(urlState.get('profile'));
       if (urlState.has('benchmark'))
         state.benchmark = urlState.get('benchmark');
+      if (urlState.has('modelUrl'))
+        state.modelUrl = urlState.get('modelUrl');
```
mattsoulanille, line 508: This is probably okay for testing and benchmarking, but keep in mind [it may fail if the model url has characters like `?` or `&`](https://stackoverflow.com/questions/1737935/what-is-the-recommended-way-to-pass-urls-as-url-parameters). It would be better to encode the URL with `encodeURIComponent`, but the format makes it hard for users to create new URLs.

It looks like you're hooking into the UI element that allows setting the custom model URL. If we can make this bi-directional (i.e. if you set the url in the UI, it appears as a URL parameter), then we can use the correct encoding, and users can input new URLs in the UI.

pyu10055 (commented):
thank you, do we have a section in the doc that explains the available keys for url configuration?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6680#-:-N7wmE9Z3BpBa1MMCC0w:b87mv9z)* status: 0 of 1 approvals obtained (waiting on @Linchenn and @mattsoulanille)

___
*[`e2e/benchmarks/local-benchmark/index.html` line 508 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6680#-N7wh56WBgkxYxKjJtvG-r2-508:-N7wml-x-qFkHDFzFJj3:bolqryu) ([raw file](https://github.com/tensorflow/tfjs/blob/0bc9857cd5870e0ccb402d4689409f2da6b238e3/e2e/benchmarks/local-benchmark/index.html#L508)):*

This is probably okay for testing and benchmarking, but keep in mind [it may fail if the model url has characters like `?` or `&`](https://stackoverflow.com/questions/1737935/what-is-the-recommended-way-to-pass-urls-as-url-parameters). It would be better to encode the URL with `encodeURIComponent`, but the format makes it hard for users to create new URLs.

It looks like you're hooking into the UI element that allows setting the custom model URL. If we can make this bi-directional (i.e. if you set the url in the UI, it appears as a URL parameter), then we can use the correct encoding, and users can input new URLs in the UI.
</blockquote></details>

good point, we should encode the url string and other parameter values.

mattsoulanille (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6680)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @Linchenn and @pyu10055)

___
*[`e2e/benchmarks/local-benchmark/index.html` line 508 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6680#-N7wh56WBgkxYxKjJtvG-r2-508:-N7woFzz-3cKlc0bTjvy:b-m5fias) ([raw file](https://github.com/tensorflow/tfjs/blob/0bc9857cd5870e0ccb402d4689409f2da6b238e3/e2e/benchmarks/local-benchmark/index.html#L508)):*

good point, we should encode the url string and other parameter values.
</blockquote></details>

Good idea. In a different PR, we should make all the settings appear as URL parameters when you change them. We should also add new settings for any URL parameters that aren't currently listed in the UI. These changes should make it easier for people to share benchmark configs, and no one will have to look up what URL parameters are supported since all the config will be done in the UI.

We can probably make a generic function that adds a new setting and hooks it up to a url parameter. It looks like these have to be set all at once, so the URL update function will probably just read the current state and transform it to a string of URL params to push with `window.history.replaceState()`.

I'm happy with merging this PR as-is if it makes it easier to make these changes.

Linchenn (commented):
Yes, we have a section here, https://github.com/tensorflow/tfjs/tree/master/e2e/benchmarks/local-benchmark#benchmark-test. Updated. Thanks!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6680#-:-N7x5wet5edYgMf_kzY1:brhm3sh)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @mattsoulanille and @pyu10055)

___
*[`e2e/benchmarks/local-benchmark/index.html` line 451 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6680#-N7wh56WBgkxYxKjJtvG-r2-451:-N7wr8f15T0-Ob6b61E4:b-896fix) ([raw file](https://github.com/tensorflow/tfjs/blob/0bc9857cd5870e0ccb402d4689409f2da6b238e3/e2e/benchmarks/local-benchmark/index.html#L451)):*

Nit: Use the `Number` function.
```suggestion
                  shapeInput.value.split(',').map(Number);
```
</blockquote></details>

Done.
___
*[`e2e/benchmarks/local-benchmark/index.html` line 508 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6680#-N7wh56WBgkxYxKjJtvG-r2-508:-N7wsbatBadhTVGhdzbd:b-7htm05) ([raw file](https://github.com/tensorflow/tfjs/blob/0bc9857cd5870e0ccb402d4689409f2da6b238e3/e2e/benchmarks/local-benchmark/index.html#L508)):*

Good idea. In a different PR, we should make all the settings appear as URL parameters when you change them. We should also add new settings for any URL parameters that aren't currently listed in the UI. These changes should make it easier for people to share benchmark configs, and no one will have to look up what URL parameters are supported since all the config will be done in the UI.

We can probably make a generic function that adds a new setting and hooks it up to a url parameter. It looks like these have to be set all at once, so the URL update function will probably just read the current state and transform it to a string of URL params to push with `window.history.replaceState()`.

I'm happy with merging this PR as-is if it makes it easier to make these changes.
</blockquote></details>

Thank you for thinking about it with details!

From my perspective, for most model urls, they are supposed to be a `.json` file (also ends up with `.json`) and they do not have `?` or `&`. For other parameters, they also typically do not have special characters. Also, without encoding url and other parameters, the url itself is more informational (users could see the model url or other parameter values directly from the url).

However, I think encoding/decoding still have non-trivial numbers of use cases and generating URL functions would be vary useful. I opened an ISSUE https://github.com/tensorflow/tfjs/issues/6682 to add this feature. Great catch!

pyu10055 (approved):
> Reviewed 1 of 1 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6680)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @mattsoulanille)

PR #6689, opened 2022-07-28
[webgpu] Remove extra device stuff passing to the constructor

haoyunfeix:
BUG

While passing both device and device feature to the constructor,
there is a potential mismatch between them and cause error, so only keep
unique device in backend constructor's parameter.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6689)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -487,14 +487,14 @@ export class WebGPUBackend extends KernelBackend {
   }
 
   async time(f: () => void): Promise<WebGPUTimingInfo> {
-    if(!this.supportTimeQuery) {
+    if (!this.supportTimeQuery) {
       console.warn(
           `This device doesn't support timestamp-query extension. ` +
           `Start Chrome browser with flag ` +
           `--disable-dawn-features=disallow_unsafe_apis then try again. ` +
-          `Or zero will shown for the kernel time when profiling mode is` +
-          `enabled. Using performance.now is not workable for webgpu since` +
-          `it doesn't support synchronously to read data from GPU.`);
+          `Otherwise, zero will be shown for the kernel time when profiling ` +
+          `mode is enabled. Using performance.now is not workable for webgpu ` +
+          `since it doesn't support synchronous data read from GPU.`);
     }
     const oldActiveTimers = this.activeTimers;
     const newActiveTimers: TimerNode[] = [];
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -155,14 +155,6 @@ export class WebGPUBackend extends KernelBackend {
         type: 'timestamp',
         count: 2,
       });
-    } else {
-      console.warn(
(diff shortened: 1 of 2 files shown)


haoyunfeix:
@qjia7 PTAL

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -155,6 +155,14 @@ export class WebGPUBackend extends KernelBackend {
         type: 'timestamp',
         count: 2,
       });
+    } else {
+      console.warn(
+          `This device doesn't support timestamp-query extension. ` +
```
qjia7, line 160: nit: remove the intermediate `+`. Only one pair of `` is enough.

qjia7 (approved):
LGTM with one nit. Thanks.

tfjs-backend-webgpu/src/backend_webgpu.ts
haoyunfeix: Only one pair of `` will separate long warnings to several lines and keep extra spaces at the beginning of each line. image
Below is what is looks like now (in one line):
image

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -155,6 +155,14 @@ export class WebGPUBackend extends KernelBackend {
...
+          `Start Chrome browser with flag ` +
+          `--disable-dawn-features=disallow_unsafe_apis then try again. ` +
+          `Or zero will shown for the kernel time when profiling mode is` +
+          `enabled. Using performance.now is not workable for webgpu since` +
+          `it doesn't support synchronously to read data from GPU.`);
```
gyagp, line 165: I don't quite understand why we always need this warning. This is a development only extension.

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: I think you can adjust your output layout.
For example:
```
      console.warn(`
This device doesn't support timestamp-query extension.
Start Chrome browser with flag
--disable-dawn-features=disallow_unsafe_apis then try again.
Or zero will shown for the kernel time when profiling mode is
enabled. Using performance.now is not workable for webgpu since
it doesn't support synchronously to read data from GPU.`);

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: Just noticed a bug https://github.com/tensorflow/tfjs/issues/6691 related with it. Agree with Yang's opinion. We should only print this warning when the users really use it.

tfjs-backend-webgpu/src/backend_webgpu.ts
haoyunfeix: I thought this also break one sentence into 6 lines, sometimes may cause the typography issue.
For example when console window is less than 70 width(debugging in a small corner), 6 lines will became 6 paragraphs(12+ lines), instead 1 line message will stay in 1 paragraph, easy to read.

gyagp (approved):
LGTM

Issue #6232, opened 2022-03-15
Invoke tfjs operations functions with GPU resource


BruceDai:
Hi. I have this scenario of invoking tfjs operations functions for processing GPU resource, for example, invoking `tf.add(a, b)` with `gpuBufferA` and `gpuBufferB` by WebGPU backend.

Current my workaround likes:
```ts
// Get TypedArray f32ArrayA
const readbackBufferA = gpuDevice.createBuffer({size: sizeInBytes, usage: GPUBufferUsage.MAP_READ | GPUBufferUsage.COPY_DST});
const readbackEncoder = gpuDevice.createCommandEncoder();
readbackEncoder.copyBufferToBuffer(gpuBufferA, 0, readbackBufferA, 0, sizeInBytes);
device.queue.submit([readbackEncoder.finish()]);
readbackBufferA.mapAsync(GPUMapMode.READ);
const f32ArrayA = new Float32Array(readbackBufferA.getMappedRange());

// Same to get f32ArrayB
...
const f32ArrayB = new Float32Array(readbackBufferB.getMappedRange());

// Invoke tf.add()
const result = tf.add(f32ArrayA, f32ArrayB);
```
I don't think it's good on memory copy / performance.

Is there any efficient solution for this scenario? Any suggestion? Thanks.

huningxin:
This would readback the data from GPU to CPU. If tf.js uses webgpu backend, it would ideal to let the data stay on GPU.

@pyu10055 @qjia7 , is there a method to create a tf.Tensor from an existing WebGPU buffer? tf.tensor only supports CPU  buffer (TypedArrayBuffer).

qjia7:
@huningxin Currently, there is no such kind of interface to directly import an external GPU resource to TFJS. I just offline talked with @lina128. We think it's helpful to have a such kind of API, like `tf.tensor (GPUData)` to take GPU resource as the input to create a tensor. Thanks.

PR #6690, opened 2022-07-28
Create tensor from GPUBuffer (GPUReadData)

axinging:
This only works on WebGPU.
TODO: refine API docs; Add more dtype.
Use case: 
```
const shape: Shape = [16];
const gpuReadData = new tf.GPUReadData(aBuffer, shape, dtype);
const a = tf.tensor1d(gpuReadData);
```

 BUG: https://github.com/tensorflow/tfjs/issues/6232

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6690)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -17,7 +17,7 @@
 
 import './flags_webgpu';
 
-import {backend_util, buffer, DataStorage, DataType, engine, env, GPUData, KernelBackend, Rank, RecursiveArray, ShapeMap, TensorBuffer, TensorInfo, TimingInfo, TypedArray, util} from '@tensorflow/tfjs-core';
+import {backend_util, buffer, DataStorage, DataType, engine, env, GPUData, GPUReadData, KernelBackend, Rank, RecursiveArray, ShapeMap, TensorBuffer, TensorInfo, TimingInfo, TypedArray, util} from '@tensorflow/tfjs-core';
 
 import {BufferManager} from './buffer_manager';
 import {TextureManager} from './texture_manager';
@@ -50,6 +50,7 @@ type TensorData = {
   shape: number[],
   refCount: number,
   resourceInfo?: BufferInfo|TextureInfo,
+  external?: boolean,
   // For complex numbers, the real and imaginary parts are stored as their own
   // individual tensors, with a parent joining the two with the
   // complexTensorInfos field.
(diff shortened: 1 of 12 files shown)


axinging:
@qjia7 @xhcao @haoyunfeix @gyagp , This is for API discussion.

The use case for this PR:
```
const shape: Shape = [16];
const gpuReadData = new tf.GPUReadData(aBuffer, shape, dtype);
const a = tf.tensor1d(gpuReadData);
```

In another PR https://github.com/tensorflow/tfjs/pull/6684, the use case is as below:
```
const a = tf.tensor1d(gpuReadBuffer, 'float32', shape);
``

**Comment by qjia7** [comment]
Some initial opinions for this PR.
Currently, the `GPUData`  returned from `dataToGPU` is as below:
```
export interface GPUData {
  tensorRef: Tensor;
  texture?: WebGLTexture;
  buffer?: GPUBuffer;
  texShape?: [number, number];
  bufSize?: number;
}
```
I am wondering how about refactoring it like below (however, it breaks the usage of webgl for `dataToGPU` ):
```
export interface GPUTextureData {
  texture: WebGLTexture;
  texShape: [number, number];
}
export interface GPUBufferData {
  buffer: GPUBuffer;
  size: number;
  type: string;
}
export interface GPUData {
  tensorRef: Tensor;
  textureData?: GPUTextureData;
  bufferData?: GPUBufferData;
}
```

Then the interface of create tensor from gpu data is like below:

- Option 1: Reuse the current interface `export function tensor<R extends Rank>(values: TensorLike|GPUBufferData|GPUTextureData, shape?: ShapeMap[R], dtype?: DataType): Tensor<R>`

In this interface, it just extends the current tensor creation function to also support GPU data (`GPUBufferData|GPUTextureData`) as input.

For implementation, maybe we can reuse the current function `makeTensor` `write`. 

A small issue is that if the input is `GPUBufferData|GPUTextureData`, we can't infer the shape from values and dtype with the current rules. Maybe we can provide a default rule for GPU resource to get a shape or it's mandatory to provide the shape when values' type are `GPUBufferData|GPUTextureData`.

- Option 2: Add a new interface `export function tensorGPU<R extends Rank>(values: GPUBufferData|GPUTextureData, shape: ShapeMap[R], dtype?: DataType): Tensor<R>`

@pyu10055 @lina128 How do you think? (Personally, I think option 1 is ok since we don't need to add a new interface.)
This change also involves `dataToGPU`. But that API has been released for WebGL. So I don't know whether we still can change the returned value's structure or keep webgl as it is. But I think it will be great if we can keep consistent data structure for dataToGPU and create tensor from GPU. 

BTW, @gyagp and I have some offline discussions about `dataToGPU` and find some issues. @gyagp will provide more details description about them.

pyu10055 (commented):
I do like option 1, @lina128 what is your thoughts on the breaking change for dataToGPU, maybe we can release this as part of 4.0 release.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6690#-:-N8dkVY475Pg70D2sZF1:b1ph0zd)* status: 0 of 1 approvals obtained

axinging:
Now that the API has been defined in https://github.com/tensorflow/tfjs/pull/6853,  I will close this and create a new PR for WebGPU.

Issue #6687, opened 2022-07-28
major difference in performance between sync and async methods to download tensor data


vladmandic:
when using `webgl` backend, `tensor.dataSync()` is about 35% faster than `await tensor.data()` when returning larger amounts of data  
(in my case, model takes 720x720x3 image as input and produces 720x720x3 image as output)

actual timings are ~100ms for `tensor.dataSync()` and ~140ms for `await tensor.data()`  
note that there is nothing else executing, this is a single active loop

i wouldn't mind using `dataSync()`, but then again any sync calls are not compatible with `webgpu` backend and maintaining two separate codepaths for `webgl` and `webgpu` is a no-op

btw, model in question can be found at <https://github.com/vladmandic/anime>

environment: tfjs 3.19..0 on chrome 103

gyagp:
We once did some investigation on why there is a big perf diff between WebGL data() and dataSync(). The basic finding is below:
WebGLRenderingContext.readPixels() is a blocking operation. It waits for all previous rendering operations to finish, then reads data from GPU to CPU, which may block the main thread for quite a long time. To mitigate the situation, in async path (that's TFJS implementation of tensor.data()), we may insert a fence right after rendering operations, then use some timer (For example, setTimeout) to poll fence’s status. If fence is not satisfied, we can do something in parallel. Otherwise, we then readPixels() without a long blocking as rendering operations have been finished.
With an ideal timer, fence may unblock the UI, hide the latency using parallel tasks and timely read data back. 
However, according to html spec (https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#:~:text=If%20nesting%20level%20is%20greater%20than%205%2C%20and%20timeout%20is%20less%20than%204%2C%20then%20set%20timeout%20to%204), If the nesting level is greater than 5 and timeout is less than 4ms, timeout will be changed to 4ms, which causes a delay of at least 4ms. You have about 40ms diff, so I'm not sure if there are other reasons. 
To mitigate this, there are 3 possibilities:
1. We already found a tricky way to introduce a new setTimeout(), which will be implemented in TFJS (Sorry that this work has been delayed for a while). 
2. In Chrome Canary, it also changed the nesting threshold from 5 to 100 (https://groups.google.com/a/chromium.org/g/blink-dev/c/-TjeYs7shTQ/m/FhJq0mQyDAAJ?pli=1). You may have a quick try with Chrome Canary (not Stable) to see if your problem can be alleviated.
[...]

PR #6694, opened 2022-07-29
Customize setTimeout

gyagp:
If the setTimeout nesting level is greater than 5 and timeout is less
than 4ms, timeout will be clamped to 4ms, which hurts the perf. A custom
setTimeout is provided to mitigate the perf impact.

BUG: 6687


gyagp:
This PR is still WIP. @qjia7 @xhcao  PTAL to see if you have any high-level comments. Do you have a good idea to write some test cases for this?

qjia7:
> This PR is still WIP. @qjia7 @xhcao PTAL to see if you have any high-level comments. Do you have a good idea to write some test cases for this?

The implementation looks good for me. How about enabling setTimeoutWPM behind a flag for a while? For the test case, maybe add a new file util_base_test.js. Test that repeat call `setTimeoutWPM ` 20(or a larger number) times. And expect the avg time of the last 15 is less than 4 ms or less?

xhcao:
The function could be called by tf.profile, if you want to write unit cases. You could also view the total of tfjs-models' demos, the time could be reduced.

gyagp:
With above changes, I get below from unit test:
LOG: 'averageTime of setTimeoutWPM is 0.06666664282480876 ms'
LOG: 'averageTime of setTimeout is 4 ms'

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6694)* status: 0 of 1 approvals obtained (waiting on @gyagp, @lina128, @Linchenn, @mattsoulanille, @qjia7, and @xhcao)

___
*[`tfjs-core/src/util_base.ts` line 765 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8_i2tBEkEmf2wKaGfa:-N8_i2tC1MBovy0TyuOl:b9x8ddk) ([raw file](https://github.com/tensorflow/tfjs/blob/d4a2ab1e3907e87838211f575c82e269399c2a82/tfjs-core/src/util_base.ts#L765)):*
> ```XML
>   // If the setTimeout nesting level is greater than 5 and timeout is less than
>   // 4ms, timeout will be clamped to 4ms, which hurts the perf. Interleaving
>   // window.postMessage and setTimeout will trick the browser and avoid the
> ```

when the postMessage call will reset the timeout counting?
___
*[`tfjs-core/src/util_base.ts` line 770 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8_i_HC9YUc9aQ8-59q:-N8_i_HC9YUc9aQ8-59r:b-1lqfsh) ([raw file](https://github.com/tensorflow/tfjs/blob/d4a2ab1e3907e87838211f575c82e269399c2a82/tfjs-core/src/util_base.ts#L770)):*
> ```XML
>     fns.push(fn);
>     setTimeout(() => {
>       window.postMessage({name: messageName, index: fns.length - 1}, '*');
> ```

I am curious why you use the index of the functions, instead of simply popping from the function array?
___
*[`tfjs-core/src/util_base.ts` line 780 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8_h63m0lx9F831BNk4:-N8_h63m0lx9F831BNk5:b-wb7tx7) ([raw file](https://github.com/tensorflow/tfjs/blob/d4a2ab1e3907e87838211f575c82e269399c2a82/tfjs-core/src/util_base.ts#L780)):*
> ```XML
>       fn();
>       handledMessageCount++;
>       // console.log(`handledMessageCount=${handledMessageCount},
> ```

remove the comments?

tfjs-core/src/util_base.ts
```diff
@@ -749,3 +754,38 @@ export function isPromise(object: any): object is Promise<unknown> {
...
+  };
+
+  const handleMessage = function(event: MessageEvent) {
+    if (event.source == window && event.data.name == messageName) {
+      event.stopPropagation();
+      const fn = fns[event.data.index];
+      fn();
+      handledMessageCount++;
+      // console.log(`handledMessageCount=${handledMessageCount},
+      // fns.length=${fns.length}`);
```
mattsoulanille, line 781:
```suggestion
```

tfjs-core/src/util_base.ts
```diff
@@ -749,3 +754,38 @@ export function isPromise(object: any): object is Promise<unknown> {
...
+      if (handledMessageCount === fns.length) {
+        fns = [];
+        handledMessageCount = 0;
+      }
+    }
+  };
+
+  window.addEventListener('message', handleMessage, true);
+  (window as any).setTimeoutWPM = setTimeoutWPM;
```
mattsoulanille, line 790: It may be better to store this in `_tfGlobals` (instead of `window`), which you can get by calling `getGlobalNamespace()` from [global_util.ts](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/global_util.ts#L24).

tfjs-core/src/util_test.ts
```diff
@@ -665,3 +665,58 @@ describe('util.decodeString', () => {
     expect(util.isPromise(promise3)).toBeFalsy();
   });
 });
+
+describe('setTimeout', () => {
+  // If we set a larger number here, an error will be reported as "'expect'
+  // was used when there was no current spec, this could be because an
+  // asynchronous test timed out".
```
mattsoulanille, line 672: This is because Jasmine does not know you're still testing `setTimeout`, so the `expect` called in the callback is actually called during another test (or after all tests are complete, which causes the error you're seeing). Marking the test as `async` or using a `done` callback should fix this. [Docs](https://jasmine.github.io/tutorials/async)

tfjs-core/src/util_test.ts
```diff
@@ -665,3 +665,58 @@ describe('util.decodeString', () => {
...
+      let endTime = performance.now();
+      count++;
+      if (count > skipCount) {
+        totalTime += endTime - startTime;
+      }
+      if (count === totalCount) {
+        let averageTime = totalTime / (totalCount - skipCount);
+        console.log(`averageTime of setTimeout is ${averageTime} ms`);
+        // We don't have expect here as in some browsers, like Chrome Canary,
+        // nesting level threshold is set to 100 instead of 5.
```
mattsoulanille, line 692: With the fix for totalCount above, you might be able to increase the count above 100 and add an `expect`. Otherwise, it may be best to remove this test or mark it with `xit`, since it doesn't appear to be testing anything (unless you're testing that no errors are thrown. If so, you should call `expect().nothing()`).

tfjs-core/src/util_test.ts
```diff
@@ -665,3 +665,58 @@ describe('util.decodeString', () => {
...
 });
+
+describe('setTimeout', () => {
+  // If we set a larger number here, an error will be reported as "'expect'
+  // was used when there was no current spec, this could be because an
+  // asynchronous test timed out".
+  const totalCount = 8;
+  const skipCount = 5;
+
+  it('setTimeout', () => {
```
mattsoulanille, line 676: Make this test `async` or add a `done` callback.

tfjs-core/src/util_test.ts
```diff
@@ -665,3 +665,58 @@ describe('util.decodeString', () => {
...
+        // We don't have expect here as in some browsers, like Chrome Canary,
+        // nesting level threshold is set to 100 instead of 5.
+        return;
+      }
+      startTime = performance.now();
+      setTimeout(_testSetTimeout, 0);
+    }
+  });
+
+  it('setTimeoutWPM', () => {
```
mattsoulanille, line 700: Make this test `async` or add a `done` callback.

tfjs-core/src/util_test.ts
```diff
@@ -665,3 +665,58 @@ describe('util.decodeString', () => {
     expect(util.isPromise(promise3)).toBeFalsy();
   });
 });
+
+describe('setTimeout', () => {
+  // If we set a larger number here, an error will be reported as "'expect'
+  // was used when there was no current spec, this could be because an
+  // asynchronous test timed out".
+  const totalCount = 8;
+  const skipCount = 5;
```
mattsoulanille, line 674: Nit: Document what this does and why it exists.
```suggestion
  // Skip the first few samples because the browser does not throttle them.
  const skipCount = 5;
```

tfjs-core/src/util_base.ts
```diff
@@ -749,3 +754,38 @@ export function isPromise(object: any): object is Promise<unknown> {
...
+      // console.log(`handledMessageCount=${handledMessageCount},
+      // fns.length=${fns.length}`);
+      if (handledMessageCount === fns.length) {
+        fns = [];
+        handledMessageCount = 0;
+      }
+    }
+  };
+
+  window.addEventListener('message', handleMessage, true);
```
mattsoulanille, line 789: It might be better to add this event listener when the user first calls `setTimeoutWPM`. That way, we don't pollute the event listener list if `setTimeoutWPM` is never used.

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6694)* status: 0 of 1 approvals obtained (waiting on @gyagp, @lina128, @Linchenn, @mattsoulanille, @qjia7, and @xhcao)

___
*[`tfjs-core/src/util_base.ts` line 758 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8_sDKP-A6Qw5g00wrB:-N8_sDKP-A6Qw5g00wrC:b-ta9e2) ([raw file](https://github.com/tensorflow/tfjs/blob/d4a2ab1e3907e87838211f575c82e269399c2a82/tfjs-core/src/util_base.ts#L758)):*
> ```XML
> }
> 
> if (!('setTimeoutWPM' in window)) {
> ```

is this only targeting browser? might be better to register this on the browser platform file.

mattsoulanille (changes requested):

tfjs-core/src/util_base.ts
mattsoulanille: Actually, since this relies on `window`, _tfGlobals may be the wrong place for it. Maybe we should add an optional `setTimeoutWPM` function to [platform.ts](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/platforms/platform.ts) and add the function itself to [platform_browser.ts](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/platforms/platform_browser.ts) (like Ping suggested in his review).

Linchenn (commented):
Agree with Ping. If this targets to browser, we probably have to move it to browser platform. 

LGTM! Thank you Yang so much for investigating and fulfilling it!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6694#-:-N8a8wS_-L_e5c-P1P2P:b-b4ls3v)* status: 0 of 1 approvals obtained (waiting on @gyagp, @lina128, @qjia7, and @xhcao)

tfjs-core/src/util_base.ts
```diff
@@ -303,7 +303,7 @@ export function rightPad(a: string, size: number): string {
 
 export function repeatedTry(
     checkFn: () => boolean, delayFn = (counter: number) => 0,
-    maxCounter?: number): Promise<void> {
+    maxCounter?: number, scheduleFn?: Function): Promise<void> {
```
gyagp, line 306: environment.ts imports util_base, so we couldn't import environment in util_base, which will result in circular import. Thus I introduce scheduleFn here. If we need a custom scheduling function other than setTimeout, we can pass the function in. In our case, it's setTimeoutCustom() from WebGL side. In this way, repeatedTry() keeps to be platform independent.

tfjs-core/src/flags.ts
```diff
@@ -82,3 +82,6 @@ ENV.registerFlag('ENGINE_COMPILE_ONLY', () => false);
 
 /** Whether to enable canvas2d willReadFrequently for GPU backends */
 ENV.registerFlag('CANVAS2D_WILL_READ_FREQUENTLY_FOR_GPU', () => false);
+
+/** Whether to use setTimeoutCustom */
+ENV.registerFlag('USE_SETTIMEOUTCUSTOM', () => false);
```
gyagp, line 87: I changed to this name as even if we set it, we can't guarantee postMessage version setTimeout is called, as window might be not available in browser.

tfjs-core/src/platforms/platform.ts
```diff
@@ -46,4 +46,6 @@ export interface Platform {
   encode(text: string, encoding: string): Uint8Array;
   /** Decode the provided bytes into a string using the provided encoding. */
   decode(bytes: Uint8Array, encoding: string): string;
+
+  setTimeoutCustom?(functionRef: Function, delay: number): void;
```
gyagp, line 50: I changed the parameter names to be aligned with setTimeout definition in browser (https://developer.mozilla.org/en-US/docs/Web/API/setTimeout).

tfjs-core/src/util_base.ts
gyagp: done

tfjs-core/src/util_base.ts
gyagp: Moved to platform_browser.ts

tfjs-core/src/util_test.ts
gyagp: Thanks for the suggestion. I used done callback to solve this.

tfjs-core/src/util_test.ts
gyagp: Increased to 100

tfjs-core/src/util_test.ts
gyagp: done

tfjs-core/src/util_test.ts
gyagp: done

tfjs-core/src/util_test.ts
gyagp: comments added

tfjs-core/src/util_base.ts
gyagp: Moved to first call to setTimeoutCustom.

gyagp (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6694)* status: 0 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, @pyu10055, @qjia7, and @xhcao)

___
*[`tfjs-core/src/util_base.ts` line 758 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8_sDKP-A6Qw5g00wrB:-N8hYUVa5IU29XN8jXT0:bvk0utr) ([raw file](https://github.com/tensorflow/tfjs/blob/d4a2ab1e3907e87838211f575c82e269399c2a82/tfjs-core/src/util_base.ts#L758)):*

is this only targeting browser? might be better to register this on the browser platform file.
</blockquote></details>

moved to platform_browser
___
*[`tfjs-core/src/util_base.ts` line 765 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8_i2tBEkEmf2wKaGfa:-N8hXAe7CYgbTSowNQWz:beu6db1) ([raw file](https://github.com/tensorflow/tfjs/blob/d4a2ab1e3907e87838211f575c82e269399c2a82/tfjs-core/src/util_base.ts#L765)):*

when the postMessage call will reset the timeout counting?
</blockquote></details>

The trick here is if we interleave the setTimeout and window.postMessage, it will break the nesting rule of setTimeout (calling setTimeout inside callback).
___
*[`tfjs-core/src/util_base.ts` line 770 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8_i_HC9YUc9aQ8-59q:-N8hXewLBr5iJS_jU4Pk:b-g3cmxc) ([raw file](https://github.com/tensorflow/tfjs/blob/d4a2ab1e3907e87838211f575c82e269399c2a82/tfjs-core/src/util_base.ts#L770)):*

I am curious why you use the index of the functions, instead of simply popping from the function array?
</blockquote></details>

We may call setTimeout with different delays, which means the sequence of call may not be the sequence of callbacks. For example, setTimeout(cb0, 2); setTimeout(cb1, 1); cb1 should be called first, then cb0. 
___
*[`tfjs-core/src/util_base.ts` line 780 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8_h63m0lx9F831BNk4:-N8hX98073Y-Gft-tEpt:bb7f2mf) ([raw file](https://github.com/tensorflow/tfjs/blob/d4a2ab1e3907e87838211f575c82e269399c2a82/tfjs-core/src/util_base.ts#L780)):*

remove the comments?
</blockquote></details>

done

gyagp:
Thank you all for the comments! I think I have addressed all of them. Could you please take another look?
I also tested bodypix-MobileNetV1-image-0.25 and got below data (best time, average time):
[setTimeout]
16.5, 22.5
15.3, 20.7
15.8, 21.4
17.6, 23.2
17.4, 22.5

[setTimeoutCustom]
13.1, 16.0
13.2, 18.3
13.6, 15.8
12.7, 17.0
11.7, 15.2

The data shows we can get 2-4ms back. 

It seems the bots are still not satisfied, but I couldn't figure out the reason after checking the detailed log. Could you please point me to the reason? Thanks!

Linchenn (commented):
Thank you Yang! LGTM.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6694#-:-N8k_rBX0rvaHnKWAbQP:brj75ku)* status: 0 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, @pyu10055, @qjia7, and @xhcao)

mattsoulanille (changes requested):

gyagp (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6694)* status: 0 of 1 approvals obtained (waiting on @gyagp, @lina128, @mattsoulanille, @pyu10055, @qjia7, and @xhcao)

___
*[`tfjs-core/src/util_base.ts` line 328 at r4](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N89SdVlWYNwNjMyLCGI-r4-328:-N8nxlP11H3KAVnpoacZ:b-122fdm) ([raw file](https://github.com/tensorflow/tfjs/blob/cc75f040b645e55715cb9330dbd1f74db95dcdc8/tfjs-core/src/util_base.ts#L328)):*

Set the default value of `scheduleFunction` to `setTimeout` to avoid this if statement.
```suggestion
      scheduleFn(tryFn, nextBackoff);
```
</blockquote></details>

This looks more clean. done!
___
*[`tfjs-core/src/platforms/platform_browser_test.ts` line 102 at r4](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8fIBHXNtLpzQrumdwE-r4-102:-N8nwopv7jVrJZpnQ-Ws:b92cdst) ([raw file](https://github.com/tensorflow/tfjs/blob/cc75f040b645e55715cb9330dbd1f74db95dcdc8/tfjs-core/src/platforms/platform_browser_test.ts#L102)):*

Actually, if all the tests in this `describe` block need to run in the browser, you can replace it with 
```ts
describeWithFlags('setTimeout', BROWSER_ENVS, () => {
  ...
});
```
Then, you won't have to check the platform on each test.

[describeWithFlags](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/jasmine_util.ts#L209)
[BROWSER_ENVS](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/jasmine_util.ts#L45)
</blockquote></details>

BROWSER_ENVS is much better. Thanks for the suggestion!

tfjs-core/src/util_base.ts
```diff
@@ -321,7 +323,7 @@ export function repeatedTry(
         reject();
         return;
       }
-      setTimeout(tryFn, nextBackoff);
+      scheduleFn(tryFn, nextBackoff);
```
qjia7, line 326: How do you think to always call `setTimeoutCustom(tryFn, nextBackoff)` here? And move `setTimeoutCustom(tryFn, nextBackoff)` to a common place since your current implementation `setTimeoutCustom` is already a common way and not only for `webgl` backend? In this case, you don't need to change `gpgpu_context.ts`?

mattsoulanille (approved):

tfjs-core/src/util_base.ts
gyagp: We should replace setTimeout with a custom one only in browser and window is ready. So platform browser may be a good place to host it. 
util_base should not be dependent on any env, but setTimeoutCustom relies on env (for both platform check and flag check). So I'm afraid even with your suggestion, we still need to pass setTimeoutCustom as a parameter (Actually I introduced scheduleFn to avoid the dependency with env).

gyagp:
@pyu10055 Can you review again the latest changes? Thanks!

lina128 (approved):
Thank you Yang! LGTM

vladmandic:
I love this improvement as it results in quite significant performance improvements,  
but at the moment it causes conflict with `@tensorflow/tfjs-backend-wasm` as soon as that module is loaded   
regardless if its used or not (IMO its due to bad behavior by `EMScripten`, but its a conflict non-the-less),  

see <https://github.com/tensorflow/tfjs/issues/6687#issuecomment-1209286270> for details

gyagp:
Thanks @vladmandic , I will investigate this.

pyu10055 (approved):
Thanks @gyagp looks good to me, just couple minor comments.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6694#-:-N99GB_n9tL6C0OSXROv:b-c32iml)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @gyagp, @mattsoulanille, @pyu10055, @qjia7, and @xhcao)

___
*[`tfjs-core/src/platforms/platform.ts` line 50 at r3](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8fIBHXNtLpzQrumdwC-r3-50:-N99AbZS1Ep426lemctw:b-99zdgf) ([raw file](https://github.com/tensorflow/tfjs/blob/ae45f6a587028a3a5f2d31f481b614951e6a487e/tfjs-core/src/platforms/platform.ts#L50)):*
<details><summary><i>Previously, gyagp (Yang Gu) wrote…</i></summary><blockquote>

I changed the parameter names to be aligned with setTimeout definition in browser (https://developer.mozilla.org/en-US/docs/Web/API/setTimeout).
</blockquote></details>

To align with setTimeout, should it return the id for allowing clearTimeout API?
___
*[`tfjs-core/src/platforms/platform_browser_test.ts` line 112 at r5](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N99BD-43aBdpfeuR6iY:-N99BD-43aBdpfeuR6iZ:b-6gghhn) ([raw file](https://github.com/tensorflow/tfjs/blob/f9af3579826697efc3682b1144a7a60b2b90ec99/tfjs-core/src/platforms/platform_browser_test.ts#L112)):*
> ```XML
>       if (count === totalCount) {
>         const averageTime = totalTime / (totalCount - skipCount);
>         console.log(`averageTime of setTimeout is ${averageTime} ms`);
> ```

remove the console log
___
*[`tfjs-core/src/platforms/platform_browser_test.ts` line 140 at r5](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N99BFBh6mJe9q1ch6jK:-N99BFBh6mJe9q1ch6jL:b-6gghhn) ([raw file](https://github.com/tensorflow/tfjs/blob/f9af3579826697efc3682b1144a7a60b2b90ec99/tfjs-core/src/platforms/platform_browser_test.ts#L140)):*
> ```XML
>       if (count === totalCount) {
>         const averageTime = totalTime / (totalCount - skipCount);
>         console.log(`averageTime of setTimeoutCustom is ${averageTime} ms`);
> ```

remove the console log
___
*[`tfjs-core/src/platforms/platform_browser_test.ts` line 141 at r5](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N99BSCfBzz1qPzjwaf2:-N99BSCfBzz1qPzjwaf3:b6vslog) ([raw file](https://github.com/tensorflow/tfjs/blob/f9af3579826697efc3682b1144a7a60b2b90ec99/tfjs-core/src/platforms/platform_browser_test.ts#L141)):*
> ```XML
>         const averageTime = totalTime / (totalCount - skipCount);
>         console.log(`averageTime of setTimeoutCustom is ${averageTime} ms`);
>         if (window) {
> ```

since you are specifying browser_env, there is no need to test window variable, and the false branch should not be triggered

gyagp (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6694)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @gyagp, @mattsoulanille, @pyu10055, @qjia7, and @xhcao)

___
*[`tfjs-core/src/platforms/platform.ts` line 50 at r3](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N8fIBHXNtLpzQrumdwC-r3-50:-N9BRqVe1amL1o6imMLV:bk4cevk) ([raw file](https://github.com/tensorflow/tfjs/blob/ae45f6a587028a3a5f2d31f481b614951e6a487e/tfjs-core/src/platforms/platform.ts#L50)):*

To align with setTimeout, should it return the id for allowing clearTimeout API?
</blockquote></details>

I don't want to introduce the complexity to implement clearTimeoutCustom. Without it, it's useless to return id, so I intentionally set it as void. 
___
*[`tfjs-core/src/platforms/platform_browser_test.ts` line 112 at r5](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N99BD-43aBdpfeuR6iY:-N9BPjsH6uKo7hM1eF2e:bwrhyot) ([raw file](https://github.com/tensorflow/tfjs/blob/f9af3579826697efc3682b1144a7a60b2b90ec99/tfjs-core/src/platforms/platform_browser_test.ts#L112)):*

remove the console log
</blockquote></details>

This is in the test and I intentionally added this to understand the setTimeout delay on various configurations, like node, safari, etc. The results may change as time goes by. For example, Chrome is increasing the throttle nesting level from 5 to 100, which may cause failure of this test in future. With this info, it should be easier to understand the reason.
___
*[`tfjs-core/src/platforms/platform_browser_test.ts` line 140 at r5](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N99BFBh6mJe9q1ch6jK:-N9BQs93AvazE0auZFYl:bnj7hl0) ([raw file](https://github.com/tensorflow/tfjs/blob/f9af3579826697efc3682b1144a7a60b2b90ec99/tfjs-core/src/platforms/platform_browser_test.ts#L140)):*

remove the console log
</blockquote></details>

same as above
___
*[`tfjs-core/src/platforms/platform_browser_test.ts` line 141 at r5](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N99BSCfBzz1qPzjwaf2:-N9BRZUD-rpqCjyDeCRm:b-5m8w5p) ([raw file](https://github.com/tensorflow/tfjs/blob/f9af3579826697efc3682b1144a7a60b2b90ec99/tfjs-core/src/platforms/platform_browser_test.ts#L141)):*

since you are specifying browser_env, there is no need to test window variable, and the false branch should not be triggered
</blockquote></details>

I was thinking about running this in a worker. It should be unnecessary. Will remove in a follow-up PR.

pyu10055 (approved):
> Reviewed 3 of 6 files at r2, 2 of 4 files at r3, 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6694)* status: :shipit: complete! 3 of 1 approvals obtained (waiting on @gyagp, @mattsoulanille, @pyu10055, @qjia7, and @xhcao)

___
*[`tfjs-core/src/platforms/platform_browser_test.ts` line 141 at r5](https://reviewable.io/reviews/tensorflow/tfjs/6694#-N99BSCfBzz1qPzjwaf2:-N9Xm6ecDsLPpU4c6xMr:b-rbc6rj) ([raw file](https://github.com/tensorflow/tfjs/blob/f9af3579826697efc3682b1144a7a60b2b90ec99/tfjs-core/src/platforms/platform_browser_test.ts#L141)):*
<details><summary><i>Previously, gyagp (Yang Gu) wrote…</i></summary><blockquote>

I was thinking about running this in a worker. It should be unnecessary. Will remove in a follow-up PR.
</blockquote></details>

we run worker tests separately, and those tests are typically a separate file.

gyagp:
@pyu10055 @mattsoulanille, given related Emscripten issues were fixed (Thanks @vladmandic !) and @mattsoulanille upgraded Emscripten to include the fixes, can we merge this change now?

tfjs-core/src/platforms/platform_browser_test.ts
```diff
@@ -88,3 +89,62 @@ describeWithFlags('PlatformBrowser', BROWSER_ENVS, async () => {
     expect(s).toEqual('Здраво');
   });
 });
+
+describeWithFlags('setTimeout', BROWSER_ENVS, () => {
+  const totalCount = 100;
+  // Skip the first few samples because the browser does not clamp the timeout
+  const skipCount = 5;
+
+  it('setTimeout', (done) => {
```
hujiajie, line 98: @gyagp I began to see this test fail on Linux, maybe related with https://chromestatus.com/feature/5710690097561600

gyagp: @hujiajie What's your channel and version of Chrome? I tried both  Stable (108.0.5359.124) and Dev (110.0.5478.4) on Linux,  but the nesting levels are still 5. I also double checked Windows, with Stable (108.0.5359.125) and Canary (111.0.5491.0), and the nesting levels are also 5.
My testing program is timer.html at https://github.com/webatintel/webatintel.github.io. You need to download it to local (otherwise the result is impacted by the network condition) and type "<pathToScript>/timer.html?test=setTimeout&maxCount=30" in the address bar.

hujiajie: Ah, it seems chromestatus is lying here, and the feature is still in field trial. What if you add `--enable-features=MaxUnthrottledTimeoutNestingLevel`?

PR #6712, opened 2022-08-03
Make link-package's build-deps-for script check transitive deps

mattsoulanille:
Make build-deps-for check for Bazel dependencies transitively instead of just on the packages passed to it. Additionally, change the behavior of `build-deps-for` so that when it is passed a Bazel package, it will directly build the package instead of building just its dependencies.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6712)

### DIFF
--- a/BAZEL_MIGRATION.md
+++ b/BAZEL_MIGRATION.md
@@ -384,12 +384,12 @@ Update all downstream dependencies that depend on the package to point to its lo
 
 To find downstream packages, run `grep -r --exclude=yarn.lock --exclude-dir=node_modules "link:.*tfjs-foo" .` in the root of the repository.
 
-#### Update Downstream Packages' build-link-package Scripts
-Make sure to list the new package in the call to `yarn build-deps-for ...` script for downstream packages. This includes packages that have `tfjs-foo` as a transitive dependency.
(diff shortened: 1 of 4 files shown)


lina128 (approved):
I think the original reason we resolve transitive dependency ourselves is because of circular dependency of cpu and core. If this is not longer the case, then the refactor is the right way to go. Thank you Matt!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6712#-:-N8WOkj44t1FC4AejQTx:b-7xmoy3)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @Linchenn)

mattsoulanille (commented):
Thanks for the review! Like you said, we no longer have circular devDependencies, but the solution here should still work if we somehow add one. It can only ever visit each package once, so it should not get stuck in a loop (see the `visited` set in `findTransitiveBazelDeps`).

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6712#-:-N8WYsMP2nKAWQDCp3i_:b8gswtl)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @Linchenn)

Linchenn (approved):
LGTM! Thank you Matt!

PR #6626, opened 2022-07-15
De-flake fromPixels test failure

hujiajie:
`HTMLVideoElement.requestVideoFrameCallback()` provides a realiable way to guarantee the readiness of a video frame, whereas the browser may not update the ready state until a few frames later. In such cases, false alarms will be raised by the validation in `fromPixels()`.

Fixes #6577

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6626)

PR #6735, opened 2022-08-09
Strengthen frame readiness while testing fromPixels with video element

hujiajie:
This fixes the error in WebGPU-backed unit tests:

OperationError: Failed to execute 'importExternalTexture' on 'GPUDevice': Failed to import texture from video element that doesn't have back resource.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6735)


hujiajie:
The original readyState check is not deleted because of [concerns on old browsers](https://github.com/tensorflow/tfjs/pull/6626#pullrequestreview-1048308808).

The original readyState check is left as compulsory even when rvfc is present because [having a unified codepath that optionally awaits for the rvfc is nicer than divergent behavior](https://github.com/KhronosGroup/WebGL/pull/3319#discussion_r692576529).

hujiajie:
> Another place that has readState check is tfjs-data/src/util/test_utils.ts. I'm not sure if we need to also fix it.

I'm not sure either. Let me wait for more feedback.

> Given waiting for video is useful in multiple places, can we implement it as a utility function?

Maybe, but why it was not a helper before? And is there anything changed so you are proposing this now?

gyagp:
> > Another place that has readState check is tfjs-data/src/util/test_utils.ts. I'm not sure if we need to also fix it.
> 
> I'm not sure either. Let me wait for more feedback.
I think it's better to fix all the places.

> > Given waiting for video is useful in multiple places, can we implement it as a utility function?
> 
> Maybe, but why it was not a helper before? And is there anything changed so you are proposing this now?

To guarantee video is loaded is not straightforward and simple as we expected. That's the change from our previous understanding, so a helper function can be handy and sync all the related code on the same page.

PR #6751, opened 2022-08-13
Strengthen frame readiness while testing fromPixels with video element

hujiajie:
This fixes the error in WebGPU-backed unit tests:

  OperationError: Failed to execute 'importExternalTexture' on 'GPUDevice': Failed to import texture from video element that doesn't have back resource.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6751)

### DIFF
--- a/tfjs-backend-webgpu/src/from_pixels_webgpu_test.ts
+++ b/tfjs-backend-webgpu/src/from_pixels_webgpu_test.ts
@@ -16,6 +16,7 @@
  */
 
 import * as tf from '@tensorflow/tfjs-core';
+import {test_util} from '@tensorflow/tfjs-core';
 import {WebGPUBackend} from './backend_webgpu';
 import {describeWebGPU} from './test_util';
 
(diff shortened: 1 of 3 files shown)


hujiajie:
`replaceHTMLVideoElementSource()` in tfjs-data is exempted at the moment because it cannot be built on top of the proposed utils here. It might deserve a separate PR to deal with that.

gyagp:
Can we further simplify the usage, like "await createAndWaitForVideo(src)"? Then you may hide all the details behind this function. Also for simplicity, we may not need to support multiple sources.

hujiajie:
The unit test removes the video element when it's finished, so for symmetry `document.body.appendChild(video)` also lives at the top level. Meanwhile, it reads the autoplay policy may cause `video.play()` to fail if the element is not yet on screen. I feel it inevitably breaks `createAndWaitForVideo()` apart.

gyagp (approved):
According to browserstack tests, I think it's safe to use a single function. But anyway, it's already a good improvement comparing with exsiting code, so LGTM.

pyu10055 (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6751)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128)

pyu10055 (approved):
thanks!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6751#-:-N9dEPG_DW9MkKngbaGf:bd5l6mr)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128)

PR #6760, opened 2022-08-17
Add cpu forward for Cast

qjia7:
It's requred by cityscapes architecture for DeepLabV3. Otherwise, it has
two issues: 1. it will call dataSync, which is not supported for webgpu.
2. Even it's supported. It's bad to happen cpu/gpu data sync during the
model execution.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6760)


pyu10055 (approved):
@linchenn Will this be needed for WebGL?

> Reviewed 1 of 1 files at r1, 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6760#-:-N9gUHymEQXujnTM7_f3:blou2c5)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @gyagp and @Linchenn)

Linchenn (commented):
Hi @qjia7 , have you seen the performance gain from this change? 

I am not clear if we could get performance gain for WebGL. I just checked the graph of this model and found all three `cast` op appear at the beginning of the graph and are followed by `pack` op, `ExpandDims` op and `ResizeBilinear` op, which could be also executed on WebGL without forwarding to CPU. So I am considering, if `cast` op happens at WebGL, the following ops will still work on WebGL without forwarding to CPU.



> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6760#-:-N9iFcmm-JRzQQj6olHp:b-wk5yaw)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @gyagp and @qjia7)

___
*[`tfjs-backend-cpu/src/kernels/Cast.ts` line 82 at r3](https://reviewable.io/reviews/tensorflow/tfjs/6760#-N9iFEGI4oDdtyTjfYhI:-N9iFEGI4oDdtyTjfYhJ:bfwt8id) ([raw file](https://github.com/tensorflow/tfjs/blob/e98b2f206b0dcb265d1e6a0fdbd6d827af52a602/tfjs-backend-cpu/src/kernels/Cast.ts#L82)):*
> ```XML
> 
>     return result;
>   }
> ```

Do we need to move the casting logics related to complex64 into castImpl?



_Code quote:_
```XML
  // Casting to complex64.
  if (dtype === 'complex64') {
    if (x.dtype === 'complex64') {
      return identity({inputs: {x}, backend});
    }

    const zerosTensorInfo = zeros(backend, x.shape, x.dtype);
    const floatX = cast({inputs: {x}, backend, attrs: {dtype: 'float32'}});

    const result =
        complex({inputs: {real: floatX, imag: zerosTensorInfo}, backend});

    backend.disposeIntermediateTensorInfo(zerosTensorInfo);
    backend.disposeIntermediateTensorInfo(floatX);

    return result;
  }

  // Casting from complex64
  if (x.dtype === 'complex64') {
    const realPart = real({inputs: {input: x}, backend});
    const result = cast({inputs: {x: realPart}, backend, attrs: {dtype}});

    backend.disposeIntermediateTensorInfo(realPart);

    return result;
  }
```

qjia7 (commented):
Yes, it's also needed by webgl. For debugging, you can set a breakpoint in https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/kernels/Cast.ts#L71. It will be caught and goes to `int` which is a pure gpu implementation. However, the result of this `Cast` is expected to be as the input of parameter `size` for op `resizeBilinear`.  However, the `size`'s type is `[number, number]` which is expected to call dataSync to get the parameter value. Then it will happen a gpu-cpu copy in webgl.

>ExpandDims op and ResizeBilinear op, which could be also executed on WebGL without forwarding to CPU.

`ExpandDims` is as the first parameter of `resizeBilinear`. There are no problem for them. But the issue is the second parameter `size` of `resizeBilinear`, which is expected in cpu. However, the Cast makes it in gpu. That's the place where happens cpu-> gpu- > cpu.

I test the performance. It may have 3~5 ms impact using NV3080 (160ms -> 155ms).

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6760#-:-N9nqah16l2iEGwBwb3P:b-90qdz4)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @gyagp and @Linchenn)

___
*[`tfjs-backend-cpu/src/kernels/Cast.ts` line 82 at r3](https://reviewable.io/reviews/tensorflow/tfjs/6760#-N9iFEGI4oDdtyTjfYhI:-N9nuEQy3zCNzpydTsiJ:b-hf46zg) ([raw file](https://github.com/tensorflow/tfjs/blob/c24be2695e96722b83c298aa916a2c08963bf0de/tfjs-backend-cpu/src/kernels/Cast.ts#L82)):*

Do we need to move the casting logics related to complex64 into castImpl?
</blockquote></details>

I roughly looked at all the xxxImp. It seems that it won't take a `backend` parameter and only process the `value`. So I am not sure how to process `complex64`. 
For simplicity, I did the current change.

Linchenn (approved):
Sorry for the late replay. Thank you for the detailed investigation! LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6760#-:-NA2vkaV3TpqasIdGiUd:b-ym6eop)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @gyagp)

___
*[`tfjs-backend-cpu/src/kernels/Cast.ts` line 82 at r3](https://reviewable.io/reviews/tensorflow/tfjs/6760#-N9iFEGI4oDdtyTjfYhI:-NA2xbY71WKSDx9gPAvq:b2x5vij) ([raw file](https://github.com/tensorflow/tfjs/blob/c24be2695e96722b83c298aa916a2c08963bf0de/tfjs-backend-cpu/src/kernels/Cast.ts#L82)):*

I roughly looked at all the xxxImp. It seems that it won't take a `backend` parameter and only process the `value`. So I am not sure how to process `complex64`. 
For simplicity, I did the current change.
</blockquote></details>

LGTM. Let me check this, https://github.com/tensorflow/tfjs/issues/6777. If it is needed, we could do it in a seperate PR. (I just consider if we should put all casting logics into one function, instead of scattering.)

qjia7 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6760)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @gyagp)

___
*[`tfjs-backend-cpu/src/kernels/Cast.ts` line 82 at r3](https://reviewable.io/reviews/tensorflow/tfjs/6760#-N9iFEGI4oDdtyTjfYhI:-NA3Dg4BAQLX-aSgQHsY:bwqi2ri) ([raw file](https://github.com/tensorflow/tfjs/blob/c24be2695e96722b83c298aa916a2c08963bf0de/tfjs-backend-cpu/src/kernels/Cast.ts#L82)):*

LGTM. Let me check this, https://github.com/tensorflow/tfjs/issues/6777. If it is needed, we could do it in a seperate PR. (I just consider if we should put all casting logics into one function, instead of scattering.)
</blockquote></details>

Sounds good. Thank you!

PR #6764, opened 2022-08-18
[webgpu] Tweak WGSL entry point emission so '{' isn't visually swallowed

hujiajie:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6764)

### DIFF
--- a/tfjs-backend-webgpu/src/webgpu_program.ts
+++ b/tfjs-backend-webgpu/src/webgpu_program.ts
@@ -59,7 +59,7 @@ export const compileProgram =
       const module = device.createShaderModule(
           {code: source, label: program.constructor.name});
       const pipeline = device.createComputePipeline({
-        compute: {module, entryPoint: 'main'},
+        compute: {module, entryPoint: '_start'},
         label: program.constructor.name,
         layout: 'auto'
       });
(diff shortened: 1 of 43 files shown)


gyagp (approved):
LGTM. Thanks for the unification!

tellink1 (approved):

tfjs-backend-webgpu/src/webgpu_program.ts
```diff
@@ -103,23 +103,45 @@ export function getCoordsXYZ(index: number): string {
...
+  switch (params.length) {
+    case 0:
+      snippet = `
+        ${getWorkGroupSizeString()}
+        fn _start(@builtin(local_invocation_id) LocalId : vec3<u32>,
+                  @builtin(global_invocation_id) GlobalId : vec3<u32>,
+                  @builtin(num_workgroups) NumWorkgroups : vec3<u32>) {
+          localId = LocalId;
+          globalId = GlobalId;
+          numWorkgroups = NumWorkgroups;
```
qjia7, line 119: Can you check if `numWorkgroups = NumWorkgroups;` is still needed for this case? I expect if we use non-linear work group size, we can totally remove the dependency on `num_workgroups` and `getGlobalIndex()`. I am ok to do it in a follow-up PR. Thanks.

qjia7 (approved):
LGTM with a comment. But I am ok to merge this PR first and see if we can do further cleanup to address the comment in follow-up PRs.

tfjs-backend-webgpu/src/webgpu_program.ts
hujiajie: I don't feel it's safe to assume `getGlobalIndex()` is unused elsewhere. E.g. we currently have this dependency chain in depth-wise conv2d: `main() -> getOutputCoords() -> getGlobalIndex()`.

qjia7: `getGlobalIndex()` is only called when `x.length === outRank` in `getOutputCoords`, which means a flat dispatch layout. Let's keep it for now. Maybe it's not necessary to worry about the very small overhead of  `NumWorkgroups `.

hujiajie: I do find the `x.length === outRank` branch is taken while running depthwise_conv2d_test, is this expected?

Issue #6769, opened 2022-08-18
[browserstack benchmark tool] Windows instances could not run WebGL backend


Linchenn:
Follow https://github.com/tensorflow/tfjs/pull/6768 and run `yarn test --browserstack --browsers=Windows_11_1`, which will invoke a browserstack's win11 chrome103 instance to benchmark MobileNetV3 on WebGL backend.

However, I received the following message, which means BrowserStack's Windows instances do not support WebGL backend:
![image](https://user-images.githubusercontent.com/40653845/185507926-8067a07b-5953-48f2-a308-948343dd4c01.png)

I also tried BrowserStack Live to run on WebGL backend, but it also does not support:
![image](https://user-images.githubusercontent.com/40653845/185508053-6c847a0c-8ccb-4311-b67e-3f768481e243.png)

Linchenn:
This may have the following effects:
1. We may consider other solutions to benchmark on desktop devices for optimization. @pyu10055 
2. We may want to remove all windows instances from nightly benchmark, until we want to add `wasm` and `cpu` backends to nightly benchmark. @lina128

PR #6772, opened 2022-08-19
[WebGL backend] Add a flag to enable Software WebGL

Linchenn:
Previously, if the device does not have hardware GPU (even though it provides software webgl), it will disable webgl backend.

Example: for this case, WebGL backend will be disabled.
image

This works for our users, but we want to run tests and benchmarks on WebGL backend, even if the devices only have software WebGL only (BrowserStack's windows devices). Then this PR adds a flag to disable/enable software WebGL to make it flexible.

Reference: https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext (search for `failIfMajorPerformanceCaveat `)

Fix https://github.com/tensorflow/tfjs/issues/6769.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6772)

### DIFF
--- a/e2e/benchmarks/browserstack-benchmark/benchmark_models.js
+++ b/e2e/benchmarks/browserstack-benchmark/benchmark_models.js
@@ -118,6 +118,7 @@ describe('BrowserStack benchmark', () => {
     jasmine.DEFAULT_TIMEOUT_INTERVAL = 1000000;
     const response = await fetch(`${KARMA_SERVER}/benchmark_parameters.json`);
     benchmarkParameters = await response.json();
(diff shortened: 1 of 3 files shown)


lina128 (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6772#-:-N9rxmw93V59iBQegkZ1:bb6ywmx)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

pyu10055 (approved):

PR #6797, opened 2022-08-26
webgpu: tighten the condition of Conv2DDerInputProgram

qjia7:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6797)

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/Conv2DBackpropInput.ts
+++ b/tfjs-backend-webgpu/src/kernels/Conv2DBackpropInput.ts
@@ -54,17 +54,17 @@ export function conv2DBackpropInput(args: {
     },
   ];
   let program: Conv2DDerInputProgram|Conv2DDerInputMMProgram;
-  const dimBOuter = convInfo.inChannels;
   const dimInner =
       convInfo.filterHeight * convInfo.filterWidth * convInfo.outChannels;
   if (env().getBool('WEBGPU_USE_NAIVE_CONV2D_TRANSPOSE') ||
-      dimInner <= 64 && dimBOuter < 32) {
+      dimInner <= 64 && convInfo.inChannels <= 16) {
     // When dimInner and dimBOuter are small,
     // Conv2DDerInputProgram is much faster than Conv2DDerInputMMProgram.
     program = new Conv2DDerInputProgram(convInfo);
   } else {
     program = new Conv2DDerInputMMProgram(convInfo);
     const dimAOuter = convInfo.inHeight * convInfo.inWidth;
+    const dimBOuter = convInfo.inChannels;
     dimensions.push(
         {type: 'uint32', data: [dimAOuter]},
         {type: 'uint32', data: [dimBOuter]},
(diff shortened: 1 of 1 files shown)


qjia7:
Just find the original conditions are too loose. `Conv2DBackpropInput` with shape `[1,128,96,16]`, `[4,4,16,16]` in ArPortraitDepth falls into `Conv2DDerInputProgram` with worse performance. So this PR tightens the conditions. 

cc @Linchenn @gyagp

qjia7:
Have verified all benchmark models with `Conv2DBackpropInput` op. No regressions found with this PR.

tfjs-backend-webgpu/src/kernels/Conv2DBackpropInput.ts
```diff
@@ -54,17 +54,17 @@ export function conv2DBackpropInput(args: {
     },
   ];
   let program: Conv2DDerInputProgram|Conv2DDerInputMMProgram;
+  const dimBOuter = convInfo.inChannels;
+  const dimInner =
+      convInfo.filterHeight * convInfo.filterWidth * convInfo.outChannels;
   if (env().getBool('WEBGPU_USE_NAIVE_CONV2D_TRANSPOSE') ||
-      convInfo.inChannels < 32 && convInfo.outChannels < 32) {
-    // When inChannels and outChannels are both less than 32,
+      dimInner <= 64 && dimBOuter < 32) {
```
xhcao, line 61: 1. Why is `dimBOuter < 32`, not `dimBOuter <= 32`? Assume to define the work group size is 32, `31` and `32` use only 1 work group, and `33` uses 2 groups, `31` and `32` are the same behavior, but `32` and `33` are not the same behavior.
2. Do we need to consider the value of `dimAOuter` for the condition here?

tfjs-backend-webgpu/src/kernels/Conv2DBackpropInput.ts
```diff
@@ -54,17 +54,17 @@ export function conv2DBackpropInput(args: {
     },
   ];
   let program: Conv2DDerInputProgram|Conv2DDerInputMMProgram;
+  const dimBOuter = convInfo.inChannels;
```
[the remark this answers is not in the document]
qjia7: `dimBOuter` and `dimInner` are also used by the condition checking. So I moved them out.

tfjs-backend-webgpu/src/kernels/Conv2DBackpropInput.ts
qjia7: I don't have so many cases to cover all shapes. I just roughly look at the shapes we used in benchmark models. And find the shapes of `Conv2DBackpropInput` in SelfieSegmentation-General/LandScape are very poor. So I added the conditions to allow some shapes to fall into `Conv2DDerInputProgram`. Obviously, the original conditions are too loose. That's why I tighten these conditions. As for the threshold, I don't have a good solution to decide them. I choose them because they are close to the shapes, which has the bad perf. And per my observation, the `dimBOuter`  is usually divisible by 2/4 or equal to 1.

gyagp (approved):
LGTM

xhcao (approved):

PR #6803, opened 2022-08-29
[webgpu] Remove 'yarn --cwd' from publish script

mattsoulanille:
Some scripts call `yarn` multiple times. Each additional time `yarn` is nested, an additional `--` needs to be passed before arguments can be passed (since we're using yarn 1). Other packages' publishing scripts require two instances of `--`, and this is [what the publish-npm script assumes](https://github.com/tensorflow/tfjs/blob/master/scripts/publish-npm.ts#L132).

tfjs-backend-webgpu uses an additional call to yarn to change directory to the root of the monorepo, so it requires three instances of `--` before `--otp=######` can be passed during publishing.

This PR removes the extra call to yarn and instead uses the full name of the Bazel target, `//tfjs-backend-webgpu:tfjs-backend-webgpu_pkg.publish`. This should not affect the published artifacts since Bazel always creates a separate output tree for them anyway.

### DIFF
--- a/tfjs-backend-webgpu/package.json
+++ b/tfjs-backend-webgpu/package.json
@@ -15,7 +15,7 @@
     "bundle": "yarn build",
     "bundle-ci": "yarn bundle",
     "build-npm": "yarn build",
-    "publish-npm": "yarn --cwd .. bazel run tfjs-backend-webgpu:tfjs-backend-webgpu_pkg.publish",
+    "publish-npm": "bazel run //tfjs-backend-webgpu:tfjs-backend-webgpu_pkg.publish",
     "test": "yarn --cwd .. bazel test tfjs-backend-webgpu:tfjs-backend-webgpu_test --test_output=streamed",
     "test-dev": "yarn --cwd .. bazel run tfjs-backend-webgpu:tfjs-backend-webgpu_test --test_output=streamed"
   },


mattsoulanille:
Tested locally by running `yarn publish-npm -- -- --otp=123` in tfjs-backend-webgpu, which should fail with a bad OTP instead of `ERROR: --otp=123 :: Unrecognized option: --otp=123`.

lina128 (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6803#-:-NAfEvUm3KATPwO0EPuG:bb6ywmx)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

hujiajie (commented):
IIRC `yarn --cwd ..` was added because bazelisk has been removed from `tfjs-backend-webgpu/package.json` and the bazel command was not found in `tfjs-backend-webgpu/node_modules/.bin` in a fresh build. How can this change work then?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6803#-:-NAhl2O76cLl5sUA3MAu:bqaryen)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

mattsoulanille:
You're right, @hujiajie. I tested in a docker container, and got "command not found" with this change. I've reverted the change to `webgpu` and added a special case to `publish-npm` instead. I published webgpu@0.0.1-alpha.13 using this change, so it appears to work.

scripts/publish-npm.ts
```diff
@@ -129,7 +129,13 @@ async function main() {
...
 
     if (BAZEL_PACKAGES.has(pkg)) {
-      $(`YARN_REGISTRY="https://registry.npmjs.org/" yarn publish-npm -- -- --otp=${
+      let dashes = '-- --';
+      if (pkg === 'tfjs-backend-webgpu') {
+        // Special case for webgpu, which has an additional call to `yarn`
+        // in publish-npm.
+        dashes = '-- -- --';
+      }
+      $(`YARN_REGISTRY="https://registry.npmjs.org/" yarn publish-npm ${dashes} --otp=${
```
hujiajie, line 138: The line exceeds the 80 column limit a bit, is that fine?

hujiajie (approved):

pyu10055 (approved):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6803)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @hujiajie)

PR #6806, opened 2022-08-30
[webgpu] Handling validation error

axinging:


axinging:
@qjia7  @xhcao  @haoyunfeix @gyagp PTAL

qjia7 (commented):
I think it will impact the performance if you unconditionally add `pushErrorScope` and `popErrorScope`. Do you observe any perf loss?

And in another side, I don't see we can catch the error in time and have to delay it when executing `getBufferData`. In this case, I can also get the similar message in console as the warning without this PR. So what's the advantage of the PR?

axinging:
> I think it will impact the performance if you unconditionally add `pushErrorScope` and `popErrorScope`. Do you observe any perf loss?
> 
> And in another side, I don't see we can catch the error in time and have to delay it when executing `getBufferData`. In this case, I can also get the similar message in console as the warning without this PR. So what's the advantage of the PR?

Yes, I am also thinking about if we should add a validation flag on this.

I think warning here is misleading. This kind of validation error should be fixed immediately, and abort running. And the validation error in dawn is generated async, do you have any better place (instead of getBufferData) to get these errors?

Issue #6775, opened 2022-08-22
Mismatch in packed depthwise conv 2d results on Mali GPU


shanumantesc:
<em>Please make sure that this is a bug. As per our
[GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md),
we only address code/doc bugs, performance issues, feature requests and
build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): Yes, test case shared below
- Mobile device : Pixel 6 Pro (Reproduces on any android device with Mali GPU that I tried)
- TensorFlow.js installed from (npm or script link): 3.19.0
- Browser version: Chrome 103.0.5060.53

**Describe the current behavior**
Packed depthwise conv2d produces incorrect result on Mali GPUs when `WEBGL_MAX_TEXTURE_SIZE` is left at the default value (which is 4096 on most modern android devices). In one of our networks, we end up creating a 3672x1 sized texture for weights, which produces incorrect outputs (presumably some error in sampling the texture, but that is just a guess). Setting max texture size lower than 3672 fixes the issue.

I have attached sample code below to produce the issue (uses the same filter dims as the layer which caused inaccuracy in our original network). The code does the following:
- We first run packed depthwise conv with 4096 as the texture size (the default value of max texture size on all browsers I tried, the value is hardcoded for consistent results)
- Next, we re-run the same node with a max size of 2048.
- Finally, we set backend to cpu to get the reference output.
- With size 2048, the outputs match the reference, but with the default size, the outputs do not match. 

Note: The mismatch occurs only on Mali GPUs with Android based on my tests. iOS, MacOS chrome, Androids with Adreno GPU all produce the correct result with default texture size (of 4096)

**Standalone code to reproduce the issue**
```
tf.ENV.set('WEBGL_PACK_DEPTHWISECONV', true)

let w = Array.from({length: 3 * 3 * 816}, () => Math.random())
let x = Array.from({length: 12 * 10 * 816}, () => Math.random())

let inputs = {
    filter: tf.tensor(w, [3, 3, 816, 1]),
    x: tf.tensor(x, [1, 12, 10, 816]),
    strides: 1,
    pad: [[0, 0], [1, 1], [1, 1], [0, 0]],
    dataFormat: "channelsLast",
    dilations: 1,
    activation: 'relu'
};

tf.setBackend('webgl')
[...]

PR #6808, opened 2022-08-31
[WebGL backend] Add max 1D texture dimension flag

Linchenn:
Fix: https://github.com/tensorflow/tfjs/issues/6775
### Problem
Mali GPU could not provide interpolation values with enough precision for long skinny triangles, proved by the [demo](https://github.com/Linchenn/Toolkit/blob/main/GPGPU/gpgpu.js). The varying from the interpolation values will be used to calculate coords (in type) for sampling, finally causing sampling mismatch.

```js
tf.ENV.set('WEBGL_PACK_DEPTHWISECONV', true)

let w = Array.from({length: 3 * 3 * 816}, () => Math.random())
let x = Array.from({length: 12 * 10 * 816}, () => Math.random())

let inputs = {
    filter: tf.tensor(w, [3, 3, 816, 1]),
    x: tf.tensor(x, [1, 12, 10, 816]),
    strides: 1,
    pad: [[0, 0], [1, 1], [1, 1], [0, 0]],
    dataFormat: "channelsLast",
    dilations: 1,
    activation: 'relu'
};

tf.setBackend('webgl')
tf.ENV.set('WEBGL_MAX_TEXTURE_SIZE', 4096)
let out_4096 = tf.fused.depthwiseConv2d(inputs);

tf.ENV.set('WEBGL_MAX_TEXTURE_SIZE', 2048)
inputs.x = tf.tensor(x, [1, 12, 10, 816])
inputs.filter = tf.tensor(w, [3, 3, 816, 1])
let out_2048 = tf.fused.depthwiseConv2d(inputs);

tf.setBackend('cpu')
inputs.x = tf.tensor(x, [1, 12, 10, 816])
inputs.filter = tf.tensor(w, [3, 3, 816, 1])
let out_reference = tf.fused.depthwiseConv2d(inputs);

const doTensorsDiffer = function(t0, t1) {
	return tf.any(tf.greater(tf.abs(tf.sub(t0, t1)), tf.scalar(1e-2))).dataSync()[0];
}

console.log("Default and 2048 differ? " + doTensorsDiffer(out_4096, out_2048));
console.log("Reference and 2048 differ? " + doTensorsDiffer(out_reference, out_2048));
console.log("Reference and 4096 differ? " + doTensorsDiffer(out_reference, out_4096));
```
The three differs are expected to be 0, while Mali GPU generates 1 for the first and the third.

### Findings
Take the texture with physical size of [3672, 1] for example. As the second dimension increase, the precision becomes better. Actually, when the second dimension is increased to 2, the precision for calculating coords is enough. ([reference data](https://docs.google.com/spreadsheets/d/1uGfdksY1icQJnWEeV9nVX63DHB6bj-xc2IMTjk0sN-Y/edit?usp=sharing): for each table: the second column is the computed index from varying values; the third and forth columns are ground-truth; the fifth column is the difference between indices. You could find all difference between indices are 0.)

### Solution
This PR expose a flag, `WEBGL_MAX_TEXTURE_DIMENSION_FOR_1D_TEXTURE`. For 1D texture (physical height or physical width is 1), if the physical length of any texture edges exceed the threshold, the other edge will be increased to 2.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6808)


Linchenn:
cc @shanumantesc

qjia7 (approved):
LGTM, thanks.

Can you also add the `depthwiseConv2d` case described in the bug for webgl?

Linchenn (commented):
Done. Thanks!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6808#-:-NAm_C153QGmUz7CtF6l:boyexb3)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

qjia7 (commented):
Forget to submit the test file?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6808#-:-NAmb9YW9PM5u3HpKFk9:byueq9h)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

pyu10055 (changes requested):
Thanks Lin for investigating and fixing this deep hidden bug. I am curious why we take the approach to double the dimension of shape 1, instead of using the squarish function?
I assume the squarish function to produce smaller texture size and potentially reduce the computation if the texture is output.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6808#-:-NAobeVM0KXxSzn9spx6:bpyv5f9)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @Linchenn)

___
*[`tfjs-backend-webgl/src/webgl_util.ts` line 446 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6808#-NAocgGr1ZB3k7CumQW5:-NAocgGsDJ795qR_b_0l:b5r3wzb) ([raw file](https://github.com/tensorflow/tfjs/blob/52dc64443001bbca65252b9311744ad887f310dd/tfjs-backend-webgl/src/webgl_util.ts#L446)):*
> ```XML
>       env().getNumber('WEBGL_MAX_TEXTURE_DIMENSION_FOR_1D_TEXTURE') *
>       (isPacked ? 2 : 1);
>   if (textureShape[0] > maxDimFor1DTex || textureShape[1] > maxDimFor1DTex) {
> ```

It would be good to separate these two conditions to avoid unnecessary computation later.
___
*[`tfjs-backend-webgl/src/webgl_util.ts` line 449 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6808#-NAod10q-FkQ9MbZ50sP:-NAod10q-FkQ9MbZ50sQ:b-k76crw) ([raw file](https://github.com/tensorflow/tfjs/blob/52dc64443001bbca65252b9311744ad887f310dd/tfjs-backend-webgl/src/webgl_util.ts#L449)):*
> ```XML
>     // For 1D texture, if the length exceeds maxDimFor1DTex, the texture will be
>     // upgraded to 2D with a physical shape as [length, 2] or [2, length].
>     textureShape[0] = Math.max(isPacked ? 4 : 2, textureShape[0]);
> ```

should the packed texture be 2?

Linchenn (commented):
@qjia7 Since this PR only updates the `getTextureShapeFromLogicalShape` function, I just added some tests for it. 

For the depthwise's mismatch bug here, we do not guarantee this PR could fix it. Instead, this PR expose the flags to users and let users to tune the flag and correct the mismatch.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6808#-:-NApBbrB6pll0BD0KTXi:bthlgid)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6808)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @Linchenn and @pyu10055)

___
*[`tfjs-backend-webgl/src/webgl_util.ts` line 374 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6808#-NApon02DHH0R7qvamD9:-NApon02DHH0R7qvamDA:ba0xdpj) ([raw file](https://github.com/tensorflow/tfjs/blob/16770325837b02b62374f6eea87ec9dddfa9b71d/tfjs-backend-webgl/src/webgl_util.ts#L374)):*
> ```XML
>       env().getNumber('WEBGL_MAX_SIZE_FOR_NARROW_TEXTURE');
>   if (maxSizeForNarrorTex === Infinity &&
>       env().getBool('WEBGL_AUTO_SQUARIFY_NARROW_TEXTURE_SHAPE_FOR_MALI_GPU')) {
> ```

we need to detect the GPU type if it is MALI, to ensure it only applies to MALI GPU.
or you can change the name to WEBGL_AUTO_SQUARIFY_NARROW_TEXTURE_SHAPE
so it is user's responsibility to detect MALI GPU.
___
*[`tfjs-backend-webgl/src/webgl_util.ts` line 404 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6808#-NApo1et81sxOEECJqRe:-NApo1et81sxOEECJqRf:b-eqnnuw) ([raw file](https://github.com/tensorflow/tfjs/blob/16770325837b02b62374f6eea87ec9dddfa9b71d/tfjs-backend-webgl/src/webgl_util.ts#L404)):*
> ```XML
>       (textureShape: [number, number]) => {
>         return Math.max(...textureShape) > maxSizeForNarrorTex &&
>             Math.min(...textureShape) <= (isPacked ? 2 : 1);
> ```

if the texture is a 0 shaped tensor, we should not change it.
Which means `Math.min(...textureShape) === (isPacked ? 2 : 1)`
___
*[`tfjs-backend-webgl/src/webgl_util.ts` line 420 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6808#-NApoLTIBoZciQW0a4k9:-NApoLTIBoZciQW0a4kA:bh75ehr) ([raw file](https://github.com/tensorflow/tfjs/blob/16770325837b02b62374f6eea87ec9dddfa9b71d/tfjs-backend-webgl/src/webgl_util.ts#L420)):*
> ```XML
>       logShape.length === 2 && logShape[0] <= maxTexSize &&
>       logShape[1] <= maxTexSize &&
>       !isLongNarrowTex(logShape as [number, number])) {
> ```

cache this boolean value, since it is checked in multiple places.

Linchenn (commented):
Thank you Ping!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6808#-:-NAptZGzCYU0BP3uUFnI:b-t1trzh)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

___
*[`tfjs-backend-webgl/src/webgl_util.ts` line 374 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6808#-NApon02DHH0R7qvamD9:-NApr5Pa43oeEAsvj_2d:bpw4nqf) ([raw file](https://github.com/tensorflow/tfjs/blob/16770325837b02b62374f6eea87ec9dddfa9b71d/tfjs-backend-webgl/src/webgl_util.ts#L374)):*

we need to detect the GPU type if it is MALI, to ensure it only applies to MALI GPU.
or you can change the name to WEBGL_AUTO_SQUARIFY_NARROW_TEXTURE_SHAPE
so it is user's responsibility to detect MALI GPU.
</blockquote></details>

Renamed to `WEBGL_AUTO_SQUARIFY_NARROW_TEXTURE_SHAPE`, as we are not sure about when Mali GPU's varying becomes correct. We could add it when we want to maintain this.
___
*[`tfjs-backend-webgl/src/webgl_util.ts` line 404 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6808#-NApo1et81sxOEECJqRe:-NApsDpl5K3oV4O-JNt_:bm5ihck) ([raw file](https://github.com/tensorflow/tfjs/blob/16770325837b02b62374f6eea87ec9dddfa9b71d/tfjs-backend-webgl/src/webgl_util.ts#L404)):*

if the texture is a 0 shaped tensor, we should not change it.
Which means `Math.min(...textureShape) === (isPacked ? 2 : 1)`
</blockquote></details>

Since we need to include both `Math.min(...textureShape) === 1` and `Math.min(...textureShape) ===2` for `isPacked === true`, I instead added one more condition here.
___
*[`tfjs-backend-webgl/src/webgl_util.ts` line 420 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6808#-NApoLTIBoZciQW0a4k9:-NApsmWL4XcKvtCdM90h:b-8ufmin) ([raw file](https://github.com/tensorflow/tfjs/blob/16770325837b02b62374f6eea87ec9dddfa9b71d/tfjs-backend-webgl/src/webgl_util.ts#L420)):*

cache this boolean value, since it is checked in multiple places.
</blockquote></details>

Since the input for `isLongNarrowTex` is different for each branch, maybe we could not cache it

Linchenn (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6808)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

___
*[`tfjs-backend-webgl/src/webgl_util.ts` line 420 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6808#-NApoLTIBoZciQW0a4k9:-NAqnRkt-z2GcSdEsCIA:b-pulve2) ([raw file](https://github.com/tensorflow/tfjs/blob/16770325837b02b62374f6eea87ec9dddfa9b71d/tfjs-backend-webgl/src/webgl_util.ts#L420)):*

Since the input for `isLongNarrowTex` is different for each branch, maybe we could not cache it
</blockquote></details>

Moves checking `isLongNarrowTex` to the end of the function. Then we do not need to add `isLongNarrowTex` to multiple places.

tfjs-backend-webgl/src/webgl_util.ts
```diff
@@ -395,30 +403,42 @@ export function getTextureShapeFromLogicalShape(
   }
 
   let size = util.sizeFromShape(logShape);
-  if (logShape.length <= 1 && size <= maxTexSize) {
-    return [1, size];
+  let textureShape: [number, number];
+  if (logShape.length <= 1 && size <= maxTexSize &&
+      size <= maxSizeForNarrorTex) {
```
qjia7, line 408: nit: `size <= maxSizeForNarrorTex` is not needed?

tfjs-backend-webgl/src/webgl_util.ts
```diff
@@ -395,30 +403,42 @@ export function getTextureShapeFromLogicalShape(
   }
 
   let size = util.sizeFromShape(logShape);
-  if (logShape.length <= 1 && size <= maxTexSize) {
-    return [1, size];
+  let textureShape: [number, number];
```
qjia7, line 406: nit: `let textureShape = null`. Then the last `else {}` can be removed?

qjia7 (approved):
LGTM with nits. Thanks.

tfjs-backend-webgl/src/webgl_util.ts
Linchenn: Done.

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6808)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @pyu10055)

tfjs-backend-webgl/src/webgl_util.ts
```diff
@@ -368,8 +368,16 @@ export function getShapeAs3D(shape: number[]): [number, number, number] {
 export function getTextureShapeFromLogicalShape(
     logShape: number[], isPacked = false): [number, number] {
   let maxTexSize = env().getNumber('WEBGL_MAX_TEXTURE_SIZE');
+  let maxSizeForNarrorTex =
```
shanumantesc, line 371:
```suggestion
  let maxSizeForNarrowTex =
```
(same everywhere else)

shanumantesc (approved):

tfjs-backend-webgl/src/webgl_util.ts
Linchenn: Thanks! Done.

PR #6814, opened 2022-09-02
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6814)


jinjingforever (approved):

Neilblaze (approved):

sashashura:
Ready to merge?

rthadur:
We need one more approval , @mattsoulanille could you please help review this PR.Thank you!

sashashura:
> LGTM. Thanks! We should also update [deploy-benchmark-preview.yml](https://github.com/tensorflow/tfjs/blob/master/.github/workflows/deploy-benchmark-preview.yml), which runs on PRs and is therefore more dangerous.

It is triggered on `pull_request` only, which has restricted permissions.

PR #6820, opened 2022-09-07
webgpu: Expand DepthwiseConv2DVec4Program to support any stride

qjia7:
This PR expands DepthwiseConv2DVec4Program's capabilities to support any stride size. Meanwhile, 
Constraint stride <= 2 to go to this path.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6820)


tfjs-backend-webgpu/src/depthwise_conv2d_vec4_webgpu.ts
```diff
@@ -58,57 +60,66 @@ export class DepthwiseConv2DVec4Program implements WebGPUProgram {
...
+        for (var i = 0; i < ${this.workPerThread}; i++)
+        {
+          dotProd[i] = vec4<f32>(0.0);
+        }
 
         // Use constant instead of uniform can give better performance.
         for (var wR = 0; wR < ${this.convInfo.filterHeight}; wR = wR + 1) {
           let xR = xRCorner + wR;
-          for (var i = 0; i < ${xNumber}; i++)
+          if (xR >=0 && xR < uniforms.inDims[0])
```
Linchenn, line 106: Since you moves `xR`'s boundary checks out of `readX` function, should we also move `col`'s boundary check out of `readX` function?

Linchenn (approved):
LGTM, thank you!

tfjs-backend-webgpu/src/depthwise_conv2d_vec4_webgpu.ts
qjia7: We can move `xR`'s boundary checks out of `readX` is because we are processing the values in the same row. However, we need to check whether each value is out of boundary for `col`. So `col`'s boundary check is necessary for each value.

gyagp (approved):
LGTM

gyagp (commented):
In WGSL, we have some format issue. For example, "{" should be in the same line with for and if.

tfjs-backend-webgpu/src/depthwise_conv2d_vec4_webgpu.ts
Linchenn: Yes. I mean,  in `readX`, it may be confusing that we only do boundary check for `col` while do not for `row`. Since `readX` is only used once and it's very straight forward, this LGTM and I was kindly reminding.

Approved.

PR #6828, opened 2022-09-09
webgl: Fix NaN issue

### LINKED ISSUE #6822: [webgl] enabling WEBGL_USE_SHAPES_UNIFORMS results in model execution returning null values
user reported issue with my library and after a while it was traced down to `WEBGL_USE_SHAPES_UNIFORMS`  
which i have enabled by default for over half a year now due to massive performance advantages  

issue itself is that model execution shows no issues and it returns tensors in expected shape,  
but tensor data is an array with all `null` values instead of expected `float32` values

what is strange that on majority of systems there are no issues (plus i cannot reproduce locally)  
and there is only single (quite small) affected model out of 10+ used  

affected model  
- <https://vladmandic.github.io/human/models/faceres.json>

very simple code reproduction (again, i cannot reproduce myself)  
(it runs a model using predefined image input and checks output validity)  
- <https://vladmandic.github.io/human/test/issue-faceres.html?default>: returns arrays with correct float values
- <https://vladmandic.github.io/human/test/issue-faceres.html?uniforms>: returns array with all null values

test has tfjs debug enabled and you can see all ops in the browser console

environment: 
- tfjs: 3.20.0 (older versions are also affected)
- browser: chrome 104 and 105
- os: windows 10 pro

it's reported on two different systems using different graphics adapters (AMD Radeon and nVidia RTX)  
both affected systems report no issues looking at <chrome://gpu> and webgl v2 is working fine

only other thing worth noting is that user is using non-latin os locale

link to original issue: <https://github.com/vladmandic/human/issues/291>

qjia7:
Fix #6822

Problem
1： On some GPUs, even if `a` and `b` are both non-NaN, the value of  `isNaN` in `vec4 isNaN = min(vec4(isnan(a)) + vec4(isnan(b)), vec4(1.0));` are still larger than `0.`, which misleads all values become `NAN`.
2： After resolving `NAN` issue, the result is still incorrect. It seems that the `isnan_custom` is not well supported on the problem GPU. After switching back to builtin `isnan`, everything works well.

Solution:
Use the bool type `bvec4` instead of float type `vec4` to calculate `isNaN` to avoid the the float precision issue when comparing with zero.
Meanwhile, add an env flag `WEBGL_ISNAN_CUSTOM` to allow user to specify which `isnan` to use.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6828)


qjia7:
@vladmandic @shurshilov 
Please have the last test with this PR. I also make it ready for review.
The verified solution https://github.com/tensorflow/tfjs/issues/6822#issuecomment-1247460709:
```
vec4 nanValue = a;
bvec4 isNaN = isnan(a);
result.r = isNaN.r ? nanValue.r : result.r;
result.g = isNaN.g ? nanValue.g : result.g;
result.b = isNaN.b ? nanValue.b : result.b;
result.a = isNaN.a ? nanValue.a : result.a;

nanValue = b;
isNaN = isnan(b);
result.r = isNaN.r ? nanValue.r : result.r;
result.g = isNaN.g ? nanValue.g : result.g;
result.b = isNaN.b ? nanValue.b : result.b;
result.a = isNaN.a ? nanValue.a : result.a;

return result;
```
The solution in this PR:
```
  bvec4 isNaNA = isnan(a);
  bvec4 isNaNB = isnan(b);
  bvec4 isNaN = bvec4(isNaNA.x || isNaNB.x, isNaNA.y || isNaNB.y, isNaNA.z || isNaNB.z, isNaNA.w || isNaNB.w);
  result.r = isNaN.r ? NAN : result.r;
  result.g = isNaN.g ? NAN : result.g;
  result.b = isNaN.b ? NAN : result.b;
  result.a = isNaN.a ? NAN : result.a;

  return result;
```
I think the second one should also be workable. But if not, I need to revert this change [Use NAN instead of nanValue.xxx](https://github.com/tensorflow/tfjs/pull/6828/commits/bbfa53bb67747c1cde303083ad876f79b9c649a9)

pyu10055 (commented):
@qjia7 thank you for fixing the NaN check, seems there are still some discussion on the validity of the fix? can you add some comment on the snippets to explain why the check needs to be breaking up. thanks!

> Reviewed 9 of 9 files at r1, 1 of 1 files at r2, 5 of 5 files at r3, 7 of 7 files at r4, 6 of 6 files at r5, 9 of 9 files at r6.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6828#-:-NC7J6A_D8sxhsBUWpd0:bxgiqrq)* status: 0 of 1 approvals obtained (waiting on @lina128)

qjia7 (commented):
@pyu10055 Breaking up the checks won't resolve the problem. And our previous verified solution is a misleading, which used the same test url (which totally removing NAN checking) by accident. So it's reasonable that whether to break up the checks, the result is similar.
After further debugging, it shows that `isnan_custom` doesn't work well on the problem gpu. But if we switch back to the shader builtin `isnan`, everything works well. So the current fixing is to allow the user to specify which `isnan` to use. Please take another look. Thanks.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6828#-:-NCOZ0fbDxVM-Vk7uN1g:bmwbljl)* status: 0 of 1 approvals obtained (waiting on @lina128)

qjia7 (commented):
Sorry, please hold on the review. I find another issue. Will ping you when it's ready.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6828#-:-NCOkrUbDyB8fq7MQ4xE:b-x2i4z4)* status: 0 of 1 approvals obtained (waiting on @lina128)

qjia7:
Kindly ping you guys. Also add @Linchenn in case you are in vacation.

pyu10055 (changes requested):
thanks

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6828#-:-NCv3ZuE1NNYGcqcY_IH:b-k73z76)* status: 0 of 1 approvals obtained (waiting on @lina128, @Linchenn, and @qjia7)

___
*[`tfjs-backend-webgl/src/flags_webgl.ts` line 276 at r8](https://reviewable.io/reviews/tensorflow/tfjs/6828#-NCv3Uji0izM7qMFC-y6:-NCv3Uji0izM7qMFC-y7:b-jys2ib) ([raw file](https://github.com/tensorflow/tfjs/blob/51d13727a2a1406a8a2b7dce5d3a45d5e6412954/tfjs-backend-webgl/src/flags_webgl.ts#L276)):*
> ```XML
>  * doesn't have the builtin isnan.
>  */
> ENV.registerFlag('WEBGL_ISNAN_CUSTOM', () => false);
> ```

should this default to true to match the previous behavior?
___
*[`tfjs-backend-webgl/src/glsl_version.ts` line 78 at r8](https://reviewable.io/reviews/tensorflow/tfjs/6828#-NCv37oM-9qjpM8mPZXS:-NCv37oN6qqSDS2eFs4M:bh7mzgn) ([raw file](https://github.com/tensorflow/tfjs/blob/51d13727a2a1406a8a2b7dce5d3a45d5e6412954/tfjs-backend-webgl/src/glsl_version.ts#L78)):*
> ```XML
>       #define isnan(value) isnan_custom(value)
>     ` :
>                                                              '';
> ```

The formatting seems to be weird, is it auto-formatted?

vladmandic:
> should this default to true to match the previous behavior?

if previous behavior is proven to be broken on some gpus, why?  

afaik, only reason for `isnan_custom` is because **webgl1** does not have native `isnan`, so why not enable it for **webgl1** and disable otherwise. we should avoid creating new flags that nobody will remember few months from now.

lina128 (commented):
Thank you Jiajia!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6828#-:-NCw0fO-6e599KJvMEZI:b-juxzzv)* status: 0 of 1 approvals obtained (waiting on @Linchenn, @pyu10055, and @qjia7)

___
*[`tfjs-backend-webgl/src/flags_webgl.ts` line 276 at r8](https://reviewable.io/reviews/tensorflow/tfjs/6828#-NCv3Uji0izM7qMFC-y6:-NCw-whP6ohAYlij197T:bkcyfe6) ([raw file](https://github.com/tensorflow/tfjs/blob/51d13727a2a1406a8a2b7dce5d3a45d5e6412954/tfjs-backend-webgl/src/flags_webgl.ts#L276)):*

should this default to true to match the previous behavior?
</blockquote></details>

If it's only useful for webgl2, should this flag be set to true if webgl1 is true?

qjia7 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6828)* status: 0 of 1 approvals obtained (waiting on @Linchenn, @pyu10055, and @vladmandic)

___
*[`tfjs-backend-webgl/src/flags_webgl.ts` line 276 at r8](https://reviewable.io/reviews/tensorflow/tfjs/6828#-NCv3Uji0izM7qMFC-y6:-NCwK8iv332pZW0wDYJi:bog9hz9) ([raw file](https://github.com/tensorflow/tfjs/blob/51d13727a2a1406a8a2b7dce5d3a45d5e6412954/tfjs-backend-webgl/src/flags_webgl.ts#L276)):*

If it's only useful for webgl2, should this flag be set to true if webgl1 is true?
</blockquote></details>

webgl1 is not controlled by this flag. Since webgl1 doesn't have a builtin `isnan`, we always use a custom `isnan`. For webgl2, previously, we use a customized `isnan`, which is proved to be problem on some GPUs. So I add this flag to use builtin `isnan` for webgl2 by default. Meanwhile, I keep the original customized `isnan` for testing if necessary. I agree with @vladmandic's opinion. Maybe I can rename it to `WEBGL2_ISNAN_CUSTOM` to more reflect the meaning.  
___
*[`tfjs-backend-webgl/src/glsl_version.ts` line 78 at r8](https://reviewable.io/reviews/tensorflow/tfjs/6828#-NCv37oM-9qjpM8mPZXS:-NCwJyKS4A0rRtAXI_qY:b-fmrkmb) ([raw file](https://github.com/tensorflow/tfjs/blob/51d13727a2a1406a8a2b7dce5d3a45d5e6412954/tfjs-backend-webgl/src/glsl_version.ts#L78)):*

The formatting seems to be weird, is it auto-formatted?
</blockquote></details>

Yes, it's auto-formatted.

qjia7 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6828)* status: 0 of 1 approvals obtained (waiting on @Linchenn, @pyu10055, and @vladmandic)

___
*[`tfjs-backend-webgl/src/flags_webgl.ts` line 276 at r8](https://reviewable.io/reviews/tensorflow/tfjs/6828#-NCv3Uji0izM7qMFC-y6:-NCwMN58Ele-aQ9UFo0p:begox09) ([raw file](https://github.com/tensorflow/tfjs/blob/51d13727a2a1406a8a2b7dce5d3a45d5e6412954/tfjs-backend-webgl/src/flags_webgl.ts#L276)):*

webgl1 is not controlled by this flag. Since webgl1 doesn't have a builtin `isnan`, we always use a custom `isnan`. For webgl2, previously, we use a customized `isnan`, which is proved to be problem on some GPUs. So I add this flag to use builtin `isnan` for webgl2 by default. Meanwhile, I keep the original customized `isnan` for testing if necessary. I agree with @vladmandic's opinion. Maybe I can rename it to `WEBGL2_ISNAN_CUSTOM` to more reflect the meaning.  
</blockquote></details>

Or I can totally remove the customized isnan for webgl2 since we have the builtin isnan ?

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6828)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @Linchenn, @pyu10055, and @vladmandic)

___
*[`tfjs-backend-webgl/src/flags_webgl.ts` line 276 at r8](https://reviewable.io/reviews/tensorflow/tfjs/6828#-NCv3Uji0izM7qMFC-y6:-ND0V4f89XLoU9AkUVRj:b-3ftjzd) ([raw file](https://github.com/tensorflow/tfjs/blob/51d13727a2a1406a8a2b7dce5d3a45d5e6412954/tfjs-backend-webgl/src/flags_webgl.ts#L276)):*

Or I can totally remove the customized isnan for webgl2 since we have the builtin isnan ?
</blockquote></details>

got it, if we believe using build-in isnan will fix the error and not creating any new issues, we can remove this flag.

qjia7 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6828)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @Linchenn, @pyu10055, and @vladmandic)

___
*[`tfjs-backend-webgl/src/flags_webgl.ts` line 276 at r8](https://reviewable.io/reviews/tensorflow/tfjs/6828#-NCv3Uji0izM7qMFC-y6:-ND0WA3J4UBze0rp0Uzz:b6euo8c) ([raw file](https://github.com/tensorflow/tfjs/blob/51d13727a2a1406a8a2b7dce5d3a45d5e6412954/tfjs-backend-webgl/src/flags_webgl.ts#L276)):*

got it, if we believe using build-in isnan will fix the error and not creating any new issues, we can remove this flag. 
</blockquote></details>

Thanks Ping. Let's keep it for a while to see if any bug is reported by the builtin `isnan`. If everything works well, I will submit another PR to totally remove the customized one.

hujiajie:
@qjia7 It sounds the correctness of the builtin/polyfilled `isnan` can be programmatically probed with a small snippet and then the right one could be picked up _automatically_ without "creating new flags that nobody will remember few months from now", or no?

qjia7:
> @qjia7 It sounds the correctness of the builtin/polyfilled `isnan` can be programmatically probed with a small snippet and then the right one could be picked up _automatically_ without "creating new flags that nobody will remember few months from now", or no?

In fact, It's hard to construct a small case to reflect the `isnan` issue. The current solution is to always use the builtin `isnan`.  I know you may have concern about the comment [here](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/glsl_version.ts#L53) `Use custom isnan definition to work across differences between implementations on various platforms.`. But I think it's compare to the old custom `isnan` [`return (val > 0. || val < 1. || val == 0.) ? false : true;` ]. So I prefer we directly use the builtin isnan for simplification.

lina128 (approved):

PR #6913, opened 2022-10-07
Use VAOs for save+restore of vertexAttribPointer state between different webgl programs.

kdashg:
gpgpucontext.createProgram() calls vertexAttribPointer, but if you call this twice, the second createProgram() might overwrite the vertexAttribPointer data with different values. We believe this is the case more commonly in Firefox, and WebGL's spec has no guarantees per se here.

vertexAttribPointer should either be refreshed in gpgpucontext.useProgram, or ideally we can use a VAO and just (re)bind that in gpgpucontext.useProgram.

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6913)

### DIFF
--- a/tfjs-backend-webgl/src/gpgpu_context.ts
+++ b/tfjs-backend-webgl/src/gpgpu_context.ts
@@ -29,6 +29,12 @@ export interface FenceContext {
   isFencePassed(): boolean;
 }
 
+type WebGLVao = WebGLVertexArrayObject | WebGLVertexArrayObjectOES;
+
+export interface GPGPUContextProgram extends WebGLProgram {
+  vao: WebGLVao;
+}
+
 export class GPGPUContext {
   gl: WebGLRenderingContext;
   textureFloatExtension: {};
@@ -42,11 +48,16 @@ export class GPGPUContext {
   indexBuffer: WebGLBuffer;
   framebuffer: WebGLFramebuffer;
   outputTexture: WebGLTexture|null = null;
-  program: WebGLProgram|null = null;
+  program: GPGPUContextProgram|null = null;
   private disposed = false;
   private disjoint: boolean;
   private vertexShader: WebGLShader;
   textureConfig: TextureConfig;
+
+  createVertexArray: () => WebGLVao | null;
+  bindVertexArray: (vao: WebGLVao | null) => void;
+  deleteVertexArray: (vao: WebGLVao | null) => void;
+  getVertexArray: () => WebGLVao | null;
 
   constructor(gl?: WebGLRenderingContext) {
     const glVersion = env().getNumber('WEBGL_VERSION');
@@ -56,6 +67,50 @@ export class GPGPUContext {
     } else {
       this.gl = getWebGLContext(glVersion);
     }
+    gl = this.gl;
+
+    if (gl instanceof WebGL2RenderingContext) {
+      this.createVertexArray = () => {
+        return webgl_util.callAndCheck(gl,
+          () => gl.createVertexArray());
+      };
+      this.bindVertexArray = (vao: WebGLVao|null) => {
+        return webgl_util.callAndCheck(gl,
+          () => gl.bindVertexArray(vao as WebGLVertexArrayObject));
+      };
+      this.deleteVertexArray = (vao: WebGLVao|null) => {
+        return webgl_util.callAndCheck(gl,
+          () => gl.deleteVertexArray(vao as WebGLVertexArrayObject));
+      };
+      this.getVertexArray = () => {
+        return webgl_util.callAndCheck(gl,
+          () => gl.getParameter(gl.VERTEX_ARRAY_BINDING));
+      };
+    } else if (gl != null) {
+      const ext = gl.getExtension('OES_vertex_array_object');
+      if (ext == null) {
+        throw new Error(
+            'All WebGL1 implementations are expected to offer' +
+            ' OES_vertex_array_object.');
+      }
+      this.createVertexArray = () => {
+        return webgl_util.callAndCheck(gl,
+          () => ext.createVertexArrayOES());
+      };
+      this.bindVertexArray = (vao: WebGLVao|null) => {
+        return webgl_util.callAndCheck(gl,
+          () => ext.bindVertexArrayOES(vao as WebGLVertexArrayObjectOES));
+      };
+      this.deleteVertexArray = (vao: WebGLVao|null) => {
+        return webgl_util.callAndCheck(gl,
+          () => ext.deleteVertexArrayOES(vao as WebGLVertexArrayObjectOES));
+      };
+      this.getVertexArray = () => {
+        return webgl_util.callAndCheck(gl,
+          () => gl.getParameter(ext.VERTEX_ARRAY_BINDING_OES));
+      };
+    }
+
     // WebGL 2.0 enables texture floats without an extension.
     let COLOR_BUFFER_FLOAT = 'WEBGL_color_buffer_float';
     const COLOR_BUFFER_HALF_FLOAT = 'EXT_color_buffer_half_float';
@@ -273,7 +328,7 @@ export class GPGPUContext {
             this.gl, physicalRows, physicalCols));
   }
 
-  public createProgram(fragmentShader: WebGLShader): WebGLProgram {
+  public createProgram(fragmentShader: WebGLShader): GPGPUContextProgram {
     this.throwIfDisposed();
     const gl = this.gl;
     if (this.vertexShader == null) {
@@ -285,40 +340,46 @@ export class GPGPUContext {
     webgl_util.callAndCheck(gl, () => gl.attachShader(program, fragmentShader));
     webgl_util.linkProgram(gl, program);
 
-    const boundVao = webgl_util.callAndCheck(gl, () => gl.getParameter(gl.VERTEX_ARRAY_BINDING));
+    let program2: GPGPUContextProgram;
+    const vaoWas = this.getVertexArray();
     {
-      program.vao = webgl_util.callAndCheck(gl, () => gl.createVertexArray());
-      webgl_util.callAndCheck(gl, () => gl.bindVertexArray(program.vao));
+      program2 = Object.assign(program, {
+        vao: this.createVertexArray(),
+      });
+      this.bindVertexArray(program2.vao);
       console.assert(
-        gpgpu_util.bindVertexProgramAttributeStreams(gl, program, this.vertexBuffer),
+        gpgpu_util.bindVertexProgramAttributeStreams(gl, program2,
+                                                     this.vertexBuffer),
         'gpgpu_util.bindVertexProgramAttributeStreams not fully successful.');
 
       if (this.debug) {
-        webgl_util.validateProgram(gl, program);
+        webgl_util.validateProgram(gl, program2);
       }
     }
-    webgl_util.callAndCheck(gl, () => gl.bindVertexArray(boundVao));
+    this.bindVertexArray(vaoWas);
 
-    return program;
+    this.setProgram(program2);
+
+    return program2;
   }
 
-  public deleteProgram(program: WebGLProgram) {
+  public deleteProgram(program: GPGPUContextProgram) {
     this.throwIfDisposed();
     if (program === this.program) {
       this.program = null;
     }
     if (program != null) {
       webgl_util.callAndCheck(this.gl, () => this.gl.deleteProgram(program));
-      webgl_util.callAndCheck(this.gl, () => this.gl.deleteVertexArray(program.vao));
+      this.deleteVertexArray(program.vao);
     }
   }
 
-  public setProgram(program: WebGLProgram|null) {
+  public setProgram(program: GPGPUContextProgram|null) {
     this.throwIfDisposed();
     this.program = program;
 
     if (this.program != null) {
-      webgl_util.callAndCheck(gl, () => gl.bindVertexArray(this.program.vao));
+      this.bindVertexArray(this.program.vao);
 
       if (this.debug) {
         webgl_util.validateProgram(this.gl, this.program);
@@ -400,8 +461,9 @@ export class GPGPUContext {
     this.throwIfNoProgram();
     const gl = this.gl;
     if (this.debug) {
-      const boundVao = webgl_util.callAndCheck(gl, () => gl.getParameter(gl.VERTEX_ARRAY_BINDING));
-      console.assert(boundVao == this.program.vao, "VAO changed between setProgram and executeProgram!");
+      const boundVao = this.getVertexArray();
+      console.assert(boundVao === this.program.vao,
+                     'VAO changed between setProgram and executeProgram!');
 
       this.debugValidate();
     }
--- a/tfjs-backend-webgl/src/gpgpu_context.ts
+++ b/tfjs-backend-webgl/src/gpgpu_context.ts
@@ -273,8 +273,6 @@ export class GPGPUContext {
             this.gl, physicalRows, physicalCols));
   }
 
-  private vertexAttrsAreBound = false;
-
   public createProgram(fragmentShader: WebGLShader): WebGLProgram {
     this.throwIfDisposed();
     const gl = this.gl;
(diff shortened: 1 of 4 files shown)


kdashg:
I believe this PR will address this bug filed against Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1793986

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6913)* status: 0 of 1 approvals obtained (waiting on @kdashg and @lina128)

___
*[`tfjs-backend-webgl/src/gpgpu_context.ts` line 288 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6913#-NE2d0Eq-bKd0SBL359W:-NE2d0Eq-bKd0SBL359X:bsyf7ge) ([raw file](https://github.com/tensorflow/tfjs/blob/2d3a1d71233fbc33ab8f0aa195ade7f54aa7a2b6/tfjs-backend-webgl/src/gpgpu_context.ts#L288)):*
> ```XML
>     webgl_util.linkProgram(gl, program);
> 
>     const boundVao = webgl_util.callAndCheck(gl, () => gl.getParameter(gl.VERTEX_ARRAY_BINDING));
> ```

Is this a blocking call?
@lina128 This might break the parallel compilation.
___
*[`tfjs-backend-webgl/src/gpgpu_context.ts` line 293 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6913#-NE2dOrEAIZ5KcDaLr7V:-NE2dOrEAIZ5KcDaLr7W:b21xy3w) ([raw file](https://github.com/tensorflow/tfjs/blob/91679aa9c69b3815ced76072b09eeb3634f145d5/tfjs-backend-webgl/src/gpgpu_context.ts#L293)):*
> ```XML
>     }
>     if (!this.vertexAttrsAreBound) {
>       this.setProgram(program);
> ```

The setProgram call is missing in the change.

lina128 (approved):
Thank you for digging the root cause and fix this issue!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6913#-:-NE2nZpF30gamRoA7GK8:b-3xh0hm)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @kdashg and @pyu10055)

___
*[`tfjs-backend-webgl/src/gpgpu_context.ts` line 288 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6913#-NE2d0Eq-bKd0SBL359W:-NE2lHKcAkB87jJNN12a:bnld162) ([raw file](https://github.com/tensorflow/tfjs/blob/2d3a1d71233fbc33ab8f0aa195ade7f54aa7a2b6/tfjs-backend-webgl/src/gpgpu_context.ts#L288)):*

Is this a blocking call?
@lina128 This might break the parallel compilation.
</blockquote></details>

Yeah, I think all get calls are blocking calls. Let's allow this PR to get merged. And then I can have a follow up PR that separates the setProgram part out to be controlled by the parallel compilation flag in the compileProgram.

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6913)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @kdashg)

___
*[`tfjs-backend-webgl/src/gpgpu_context.ts` line 75 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6913#-NE2qyiBAhDONQ8lYoyY:-NE2qyiBAhDONQ8lYoyZ:b-bdlcpo) ([raw file](https://github.com/tensorflow/tfjs/blob/347094a7fde88bb62ecae634da18134d085a350e/tfjs-backend-webgl/src/gpgpu_context.ts#L75)):*
> ```XML
>       this.createVertexArray = () => {
>         return webgl_util.callAndCheck(gl,
>           () => gl.createVertexArray());
> ```

gl needs to be casted to WebGL2RenderingContext type, otherwise it is failing ts check.

75           () => gl.createVertexArray());
                      ~~~~~~~~~~~~~~~~~
tfjs-backend-webgl/src/gpgpu_context.ts:79:20 - error TS2339: Property 'bindVertexArray' does not exist on type 'WebGLRenderingContext'.

79           () => gl.bindVertexArray(vao as WebGLVertexArrayObject));
                      ~~~~~~~~~~~~~~~
tfjs-backend-webgl/src/gpgpu_context.ts:83:20 - error TS2339: Property 'deleteVertexArray' does not exist on type 'WebGLRenderingContext'.

83           () => gl.deleteVertexArray(vao as WebGLVertexArrayObject));
                      ~~~~~~~~~~~~~~~~~
tfjs-backend-webgl/src/gpgpu_context.ts:87:36 - error TS2339: Property 'VERTEX_ARRAY_BINDING' does not exist on type 'WebGLRenderingContext'.

87           () => gl.getParameter(gl.VERTEX_ARRAY_BINDING));
                                      ~~~~~~~~~~~~~~~~~~~~
___
*[`tfjs-backend-webgl/src/gpgpu_context.ts` line 344 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6913#-NE2k0im40U4DyakxdYQ:-NE2k0im40U4DyakxdYR:bnkv10i) ([raw file](https://github.com/tensorflow/tfjs/blob/347094a7fde88bb62ecae634da18134d085a350e/tfjs-backend-webgl/src/gpgpu_context.ts#L344)):*
> ```XML
> 
>     let program2: GPGPUContextProgram;
>     const vaoWas = this.getVertexArray();
> ```

this is a blocking call, which would likely break our parallel compilation code. @lina128 
___
*[`tfjs-backend-webgl/src/gpgpu_context.ts` line 359 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6913#-NE2kMMr28OOLGSrLlYn:-NE2kMMr28OOLGSrLlYo:btnonjr) ([raw file](https://github.com/tensorflow/tfjs/blob/347094a7fde88bb62ecae634da18134d085a350e/tfjs-backend-webgl/src/gpgpu_context.ts#L359)):*
> ```XML
>       }
>     }
>     this.bindVertexArray(vaoWas);
> ```

why you need to rebind the old vertex array? The setProgram seems to bind the new vertex immediately?


<!-- Sent from Reviewable.io -->

kdashg:
I'm really sorry, I don't really feel comfortable giving Reviewable access to "read and write all public repository data" for my account.

I'll include my responses here:

> gl needs to be casted to WebGL2RenderingContext type, otherwise it is failing ts check.
Weird, typechecking worked for me locally! Generally it should be inferring that gl is constrained to WebGL2RenderingContext from the instanceof branch!

> this is a blocking call, which would likely break our parallel compilation code. @lina128
This isn't a blocking call normally in Firefox, at least! (it doesn't interact with program stuff)

> The setProgram call is missing in the change.
It's still there, it's just further down now.

> why you need to rebind the old vertex array? The setProgram seems to bind the new vertex immediately?
Yeah, it looks like that's probably true in the final form of the patch here.

I did it because it's the safest thing to do correctness-wise. If anything else in the library or in an app expects the bound VertexArray to remain unchanged, neglecting to reset this to its previous value would cause breakage.
(I believe this is exactly the class of errors we're running into in the case I'm investigating: https://bugzilla.mozilla.org/show_bug.cgi?id=1793986 )

It's kinda guessable that createProgram might change GL's current program, but it's not at all expected that it would change the VAO.

An alternative would be to lazily create the VAO in executeProgram.

I'm not familiar with this codebase, and I can't figure out how to build and test it locally. (I can get it to do lint checks, but not integrate with the testcase I need to investigate)
As such, I'm being very cautious and paranoid about correctness. :)

pyu10055 (commented):
@kdashg Thank you for the explanation. I ran tests your forked branch, I am seeing this errors for most of the shaders (on Chrome)
[.WebGL-0x7e000325400] GL_INVALID_OPERATION: Must have element array buffer bound.
Do you know what might be the cause? thanks

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6913#-:-NEC9WYs3PAzJS2qTw3M:bufr2r5)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @kdashg)

kdashg:
I can't figure out how to run linting, which is weird, because I had it at least partially working on a different computer.
What invocations do you use?

kdashg:
But regardless, PTAL!

pyu10055 (approved):
> Reviewed 1 of 2 files at r1, 2 of 3 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6913)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kdashg)

pyu10055 (commented):
thanks, all tests are passing now.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6913#-:-NEmWYvm-vhDQMtgcasu:b-g7dix9)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kdashg)

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6913)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @kdashg)

kdashg (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6913)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[`tfjs-backend-webgl/src/gpgpu_context.ts` line 293 at r1](https://reviewable.io/reviews/tensorflow/tfjs/6913#-NE2dOrEAIZ5KcDaLr7V:-NE3AJdG7UH2b6gxGnVq:bfh5l5f) ([raw file](https://github.com/tensorflow/tfjs/blob/91679aa9c69b3815ced76072b09eeb3634f145d5/tfjs-backend-webgl/src/gpgpu_context.ts#L293)):*

The setProgram call is missing in the change.
</blockquote></details>

It's still there, it's just further down now.
___
*[`tfjs-backend-webgl/src/gpgpu_context.ts` line 75 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6913#-NE2qyiBAhDONQ8lYoyY:-NE3Cw5D8fnNMPqn9e5y:bbbg1op) ([raw file](https://github.com/tensorflow/tfjs/blob/347094a7fde88bb62ecae634da18134d085a350e/tfjs-backend-webgl/src/gpgpu_context.ts#L75)):*

gl needs to be casted to WebGL2RenderingContext type, otherwise it is failing ts check.

75           () => gl.createVertexArray());
                      ~~~~~~~~~~~~~~~~~
tfjs-backend-webgl/src/gpgpu_context.ts:79:20 - error TS2339: Property 'bindVertexArray' does not exist on type 'WebGLRenderingContext'.

79           () => gl.bindVertexArray(vao as WebGLVertexArrayObject));
                      ~~~~~~~~~~~~~~~
tfjs-backend-webgl/src/gpgpu_context.ts:83:20 - error TS2339: Property 'deleteVertexArray' does not exist on type 'WebGLRenderingContext'.

83           () => gl.deleteVertexArray(vao as WebGLVertexArrayObject));
                      ~~~~~~~~~~~~~~~~~
tfjs-backend-webgl/src/gpgpu_context.ts:87:36 - error TS2339: Property 'VERTEX_ARRAY_BINDING' does not exist on type 'WebGLRenderingContext'.

87           () => gl.getParameter(gl.VERTEX_ARRAY_BINDING));
                                      ~~~~~~~~~~~~~~~~~~~~
</blockquote></details>

Weird, typechecking worked for me locally! Generally it should be inferring that gl is constrained to WebGL2RenderingContext from the instanceof branch!
___
*[`tfjs-backend-webgl/src/gpgpu_context.ts` line 344 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6913#-NE2k0im40U4DyakxdYQ:-NE3CU0t1Kfq9rnPFLBb:b-hjbhfp) ([raw file](https://github.com/tensorflow/tfjs/blob/347094a7fde88bb62ecae634da18134d085a350e/tfjs-backend-webgl/src/gpgpu_context.ts#L344)):*

this is a blocking call, which would likely break our parallel compilation code. @lina128 
</blockquote></details>

This isn't a blocking call for Firefox, at least! (it doesn't interact with program stuff)
___
*[`tfjs-backend-webgl/src/gpgpu_context.ts` line 359 at r2](https://reviewable.io/reviews/tensorflow/tfjs/6913#-NE2kMMr28OOLGSrLlYn:-NE3AUES8JAC13oNB6Jw:bee1wjn) ([raw file](https://github.com/tensorflow/tfjs/blob/347094a7fde88bb62ecae634da18134d085a350e/tfjs-backend-webgl/src/gpgpu_context.ts#L359)):*

why you need to rebind the old vertex array? The setProgram seems to bind the new vertex immediately?
</blockquote></details>

Yeah, it looks like that's probably true in the final form of the patch here.

I did it because it's the safest thing to do correctness-wise. If anything else in the library or in an app expects the bound VertexArray to remain unchanged, neglecting to reset this to its previous value would cause breakage.
(I believe this is exactly the class of errors we're running into in the case I'm investigating: https://bugzilla.mozilla.org/show_bug.cgi?id=1793986 )

It's kinda guessable that createProgram might change GL's current program, but it's not at all expected that it would change the VAO.

An alternative would be to lazily create the VAO in executeProgram.

I'm not familiar with this codebase, and I can't figure out how to build and test it locally. (I can get it to do lint checks, but not integrate with the testcase I need to investigate)
As such, I'm being very cautious and paranoid about correctness. :)


<!-- Sent from Reviewable.io -->

kdashg:
Sorry, that comment was stuck in the queue somehow.

kdashg:
@pyu10055 Heya, do you need anything else from me here?

pyu10055 (commented):
I think the PR is ready, thank you for the fixes.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6913#-:-NFKsAxn4FqfZXE21Bst:b1oscb2)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128)

shanumantesc:
This change is generating an `Assertion failed: gpgpu_util.bindVertexProgramAttributeStreams not fully successful.` when running fill. This is because the uv attribute is being compiled out by the shader (reproduces on Mali GPU on Pixel 7 Pro) and so `bindVertexProgramAttributeStreams` returns `false`.

The corresponding vertex and fragment sources are

```
VertexShaderSource
#version 300 es
    precision highp float;
    in vec3 clipSpacePos;
    in vec2 uv;
    out vec2 resultUV;

    void main() {
      gl_Position = vec4(clipSpacePos, 1);
      resultUV = uv;
    }
```

```
FragmentSource
#version 300 es
    precision highp float;
    precision highp int;
    precision highp sampler2D;
    in vec2 resultUV;
    out vec4 outputColor;
    const vec2 halfCR = vec2(0.5, 0.5);

    struct ivec5
    {
      int x;
      int y;
      int z;
      int w;
      int u;
    };

    struct ivec6
    {
      int x;
      int y;
      int z;
      int w;
      int u;
      int v;
    };

    uniform float NAN;
    
    
    
      #define round(value) newRound(value)
      int newRound(float value) {
        return int(floor(value + 0.5));
      }

      ivec4 newRound(vec4 value) {
        return ivec4(floor(value + vec4(0.5)));
      }
    

    int imod(int x, int y) {
      return x - y * (x / y);
    }

    int idiv(int a, int b, float sign) {
      int res = a / b;
      int mod = imod(a, b);
      if (sign < 0. && mod != 0) {
        res -= 1;
      }
      return res;
    }

    //Based on the work of Dave Hoskins
    //https://www.shadertoy.com/view/4djSRW
    #define HASHSCALE1 443.8975
    float random(float seed){
      vec2 p = resultUV * seed;
      vec3 p3  = fract(vec3(p.xyx) * HASHSCALE1);
      p3 += dot(p3, p3.yzx + 19.19);
      return fract((p3.x + p3.y) * p3.z);
    }

    
vec2 uvFromFlat(int texNumR, int texNumC, int index) {
  int texR = index / texNumC;
  int texC = index - texR * texNumC;
  return (vec2(texC, texR) + halfCR) / vec2(texNumC, texNumR);
}
vec2 packedUVfrom1D(int texNumR, int texNumC, int index) {
  int texelIndex = index / 2;
  int texR = texelIndex / texNumC;
  int texC = texelIndex - texR * texNumC;
  return (vec2(texC, texR) + halfCR) / vec2(texNumC, texNumR);
}

    
vec2 packedUVfrom2D(int texelsInLogicalRow, int texNumR,
  int texNumC, int row, int col) {
  int texelIndex = (row / 2) * texelsInLogicalRow + (col / 2);
  int texR = texelIndex / texNumC;
  int texC = texelIndex - texR * texNumC;
  return (vec2(texC, texR) + halfCR) / vec2(texNumC, texNumR);
}

    
vec2 packedUVfrom3D(int texNumR, int texNumC,
    int texelsInBatch, int texelsInLogicalRow, int b,
    int row, int col) {
  int index = b * texelsInBatch + (row / 2) * texelsInLogicalRow + (col / 2);
  int texR = index / texNumC;
  int texC = index - texR * texNumC;
  return (vec2(texC, texR) + halfCR) / vec2(texNumC, texNumR);
}

  

    float sampleTexture(sampler2D textureSampler, vec2 uv) {
      return texture(textureSampler, uv).r;
    }
  

    void setOutput(float val) {
      outputColor = vec4(val, 0, 0, 0);
    }
  
uniform float value;

    ivec4 getOutputCoords() {
      ivec2 resTexRC = ivec2(resultUV.yx *
        vec2(1, 64));
      int index = resTexRC.x * 64 + resTexRC.y;
      int r = index / 64; index -= r * 64;int c = index / 64; index -= c * 64;int d = index / 64; int d2 = index - d * 64;
      return ivec4(r, c, d, d2);
    }
  


      void main() {
        // Input can be obtained from uniform value.
        setOutput(value);
      }
```

Note that fragment shader is just setting `outputColor = vec4(val, 0, 0, 0);` without trying to sample any texture.

PR #6947, opened 2022-10-15
cleanup conv class definitions

vladmandic:
with `tfjs` **4.0** its almost there to a clean build using standard `tsc`

few minor cleanups:
- `Conv2DTranspose` and `Conv3DTranspose` unnecessarily redeclare `inputSpec` when its already declared in parent class
- `ConvRNN2D` redeclares `cell` with a different type than parent class, but without explicit declaration

cc @mattsoulanille - can you take a look since you've done most of the work for `typescript` upgrade?

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6947)


tfjs-layers/src/layers/convolutional.ts
```diff
@@ -706,7 +706,6 @@ serialization.registerClass(Conv3D);
 export class Conv2DTranspose extends Conv2D {
   /** @nocollapse */
   static className = 'Conv2DTranspose';
-  inputSpec: InputSpec[];
```
mattsoulanille, line 709: Nice catch. Note to other reviewer: This is declared in the `Layer` interface in `src/engine/topology.ts`.

tfjs-layers/src/layers/convolutional_recurrent.ts
```diff
@@ -116,7 +116,7 @@ class ConvRNN2D extends RNN {
   /** @nocollapse */
   static className = 'ConvRNN2D';
 
-  readonly cell: ConvRNN2DCell;
+  declare readonly cell: ConvRNN2DCell;
```
mattsoulanille, line 119: Nice catch here as well. `cell` is already a property of `RNN`, so `declare` is required here to assert its new type. I'm not completely sure why this didn't throw an error when compiling with Bazel.

mattsoulanille (approved):
LGTM. Thanks!

tfjs-layers/src/layers/convolutional.ts
Linchenn: Thank you Matt!

Linchenn (approved):
LGTM, thanks!

PR #5659, opened 2021-09-24
Add model debug mode support

axinging:
The model debug mode will print the values of all intermediate tensors during model inference.
This can be used for model debugging when the inference result is incorrect.

Current model debug is enabled for below models: 
mobilenet_v2, mesh_128, face_detector, hand_detector, hand_skeleton, blazeface.

For above models,  add KEEP_INTERMEDIATE_TENSORS=true in url will turn on model debug mode.
For other graph model, below changes are required:
1. Use executeAsync instead of predict.
2. When predict is done, get intermediate tensors, then print and dispose:
      await printTensors(model.getIntermediateTensors());
      model.disposeIntermediateTensors();

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/5659)

PR #6953, opened 2022-10-19
Support dump with graphmodel.execute

axinging:
Currently only graph model predicting with executeAsync supports dump. This has two drawbackes:
1. Models predict with execute don't support dump.
2. Some models have a wrapping layer over graph model (such as bodypix, pose-detection), to support dumping by graphModel.executeAsync, non trival changes are required, example change: https://github.com/tensorflow/tfjs-models/pull/841.

This change removes these limitations and enables dumping on more models.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6953)

### DIFF
--- a/tfjs-converter/src/executor/graph_executor.ts
+++ b/tfjs-converter/src/executor/graph_executor.ts
@@ -15,7 +15,7 @@
  * =============================================================================
  */
 
-import {DataType, env, NamedTensorMap, Tensor, tidy, util} from '@tensorflow/tfjs-core';
+import {DataType, env, keep, NamedTensorMap, Tensor, tidy, util} from '@tensorflow/tfjs-core';
 
 import {ISignatureDef} from '../data/compiled_api';
 import {NamedTensorsMap, TensorArrayMap, TensorInfo, TensorListMap} from '../data/types';
@@ -33,6 +33,13 @@ interface NodeWithContexts {
   node: Node;
 }
 
+// Default, not dump. Sync, dump by execute. Async, dump by executeAsync.
+enum DumpMode {
+  Default = -1,
+  Sync,
+  Async,
+}
+
 export class GraphExecutor implements FunctionExecutor {
   private compiledMap: Map<string, Node[]> = new Map();
   private _weightMap: NamedTensorsMap = {};
@@ -46,9 +53,12 @@ export class GraphExecutor implements FunctionExecutor {
   private _functionExecutorMap: {[key: string]: FunctionExecutor} = {};
   private _resourceManager: ResourceManager;
   private intermediateTensors: NamedTensorsMap = {};
-  private keepIds: Set<number>;
+  private keepIdsForExecuteAsync: Set<number>;
   private tensorsMap: NamedTensorsMap;
+  private keepInputTensorsForExecute: Tensor[];
+  private keepTensorsForExecute: Tensor[];
   private keepTensorForDebug = false;
+  private dumpMode = DumpMode.Default;
 
   get weightIds(): number[] {
     return this.parent ? this.parent.weightIds : this._weightIds;
@@ -182,6 +192,18 @@ export class GraphExecutor implements FunctionExecutor {
         this.graph, this.weightMap, executionInfo);
   }
 
+  private keepTensors(keepTensors: Tensor[], tensors: Tensor[]) {
+    if (tensors == null || this.dumpMode !== DumpMode.Sync) {
+      return;
+    }
+    tensors.forEach(tensor => {
+      if (tensor && !tensor.kept) {
+        keepTensors.push(tensor);
+        keep(tensor);
+      }
+    });
+  }
+
   /**
    * Executes the inference for given input tensors.
    * @param inputs Tensor map for the model inputs, keyed by the input node
@@ -217,20 +239,34 @@ export class GraphExecutor implements FunctionExecutor {
       this.compiledMap.set(compilationKey, orderedNodes);
     }
 
+    try {
+      this.keepTensorForDebug = env().getBool('KEEP_INTERMEDIATE_TENSORS');
+      if (this.keepTensorForDebug) {
+        this.dumpMode = DumpMode.Sync;
+      }
+    } catch (e) {
+      console.warn(e.message);
+    }
     const tensorArrayMap: TensorArrayMap = {};
     const tensorListMap: TensorListMap = {};
+    const tensorsMap: NamedTensorsMap = {...this.weightMap};
 
-    return tidy(() => {
+    const result = tidy(() => {
       const context = new ExecutionContext(
           this.weightMap, tensorArrayMap, tensorListMap,
           this.functionExecutorMap);
-      const tensorsMap: NamedTensorsMap = {...this.weightMap};
+      if (this.keepTensorForDebug) {
+        this.keepTensorsForExecute = [];
+        this.keepInputTensorsForExecute = [];
+      }
 
       Object.keys(inputs).forEach(name => {
         const [nodeName, index] = parseNodeName(name);
         const tensors: Tensor[] = [];
         tensors[index] = inputs[name];
         tensorsMap[nodeName] = tensors;
+        // Input tensors should be disposed by user.
+        this.keepTensors(this.keepInputTensorsForExecute, tensors);
       });
 
       const tensorsToKeep = this.getFrozenTensorIds(tensorsMap);
@@ -247,6 +283,7 @@ export class GraphExecutor implements FunctionExecutor {
                 `Please use model.executeAsync() instead.`);
           }
           tensorsMap[node.name] = tensors;
+          this.keepTensors(this.keepTensorsForExecute, tensors);
           this.checkTensorForDisposal(
               node.name, node, tensorsMap, context, tensorsToKeep,
               outputNodeNames, intermediateTensorConsumerCount);
@@ -258,6 +295,13 @@ export class GraphExecutor implements FunctionExecutor {
       }
       return outputs.map(name => getTensor(name, tensorsMap, context));
     });
+
+    if (this.keepTensorForDebug) {
+      this.tensorsMap = tensorsMap;
+    } else {
+      this.tensorsMap = {};
+    }
+    return result;
   }
 
   private getFrozenTensorIds(tensorMap: NamedTensorsMap): Set<number> {
@@ -268,6 +312,7 @@ export class GraphExecutor implements FunctionExecutor {
             .map(tensors => tensors.map(tensor => tensor.id)));
     return new Set(ids);
   }
+
   private checkTensorForDisposal(
       nodeName: string, node: Node, tensorMap: NamedTensorsMap,
       context: ExecutionContext, tensorsToKeep: Set<number>,
@@ -337,25 +382,35 @@ export class GraphExecutor implements FunctionExecutor {
   }
 
   disposeIntermediateTensors() {
-    if (!this.intermediateTensors) {
-      return;
+    if (this.dumpMode === DumpMode.Sync) {
+      if (this.keepTensorsForExecute) {
+        this.keepTensorsForExecute.forEach(tensor => {
+          tensor.dispose();
+        });
+        this.keepTensorsForExecute = null;
+      }
+    } else {
+      if (!this.intermediateTensors) {
+        return;
+      }
+      Object.keys(this.intermediateTensors)
+          .forEach(
+              key => this.intermediateTensors[key].forEach(
+                  tensor => tensor.dispose()));
+      this.disposeTensorsMap();
     }
-    Object.keys(this.intermediateTensors)
-        .forEach(
-            key => this.intermediateTensors[key].forEach(
-                tensor => tensor.dispose()));
-    this.disposeTensorsMap();
   }
 
   private disposeTensorsMap() {
     if (!this.tensorsMap) {
       return;
     }
+
     Object.keys(this.tensorsMap).forEach(key => {
       const tensorArray = this.tensorsMap[key];
       tensorArray.forEach(tensor => {
         if (tensor && !tensor.kept && !tensor.isDisposed &&
-            !this.keepIds.has(tensor.id)) {
+            !this.keepIdsForExecuteAsync.has(tensor.id)) {
           tensor.dispose();
         }
       });
@@ -367,9 +422,11 @@ export class GraphExecutor implements FunctionExecutor {
   }
 
   private resetIntermediateTensors() {
-    for (const key in this.intermediateTensors) {
-      this.intermediateTensors[key].forEach(tensor => tensor.dispose());
-      delete this.intermediateTensors[key];
+    if (this.dumpMode === DumpMode.Async) {
+      for (const key in this.intermediateTensors) {
+        this.intermediateTensors[key].forEach(tensor => tensor.dispose());
+        delete this.intermediateTensors[key];
+      }
     }
   }
 
@@ -402,6 +459,9 @@ export class GraphExecutor implements FunctionExecutor {
     // For model debug.
     try {
       this.keepTensorForDebug = env().getBool('KEEP_INTERMEDIATE_TENSORS');
+      if (this.keepTensorForDebug) {
+        this.dumpMode = DumpMode.Async;
+      }
     } catch (e) {
       console.warn(e.message);
     }
@@ -422,7 +482,7 @@ export class GraphExecutor implements FunctionExecutor {
     // dispose all the intermediate tensors
     const outputIds = results.map(t => t.id);
     const inputIds = Object.keys(inputs).map(name => inputs[name].id);
-    this.keepIds =
+    this.keepIdsForExecuteAsync =
         new Set<number>([...outputIds, ...inputIds, ...this.weightIds]);
     if (!this.keepTensorForDebug) {
       this.disposeTensorsMap();
@@ -430,7 +490,7 @@ export class GraphExecutor implements FunctionExecutor {
 
     // dispose the context for the root executor
     if (this.parent == null) {
-      context.dispose(this.keepIds);
+      context.dispose(this.keepIdsForExecuteAsync);
     }
 
     return results;
--- a/tfjs-converter/src/executor/graph_executor.ts
+++ b/tfjs-converter/src/executor/graph_executor.ts
@@ -45,8 +45,7 @@ export class GraphExecutor implements FunctionExecutor {
   private _functions: {[key: string]: Graph} = {};
   private _functionExecutorMap: {[key: string]: FunctionExecutor} = {};
   private _resourceManager: ResourceManager;
-  private tensorsPendingDisposal: Set<string>;
-  private tensorsMap: NamedTensorsMap;
+  private clonedTensorsMap: NamedTensorsMap;
   private keepIntermediateTensors = false;
 
   get weightIds(): number[] {
@@ -181,20 +180,17 @@ export class GraphExecutor implements FunctionExecutor {
         this.graph, this.weightMap, executionInfo);
   }
 
-  private keepTensors(
-      tensorsToKeep: Tensor[], nodeName: string = null,
-      tensorsPendingDisposal: Set<string> = null) {
-    if (!this.keepIntermediateTensors || tensorsToKeep == null) {
-      return;
-    }
-    if (nodeName && tensorsPendingDisposal) {
-      tensorsPendingDisposal.add(nodeName);
-    }
-    tensorsToKeep.forEach(tensor => {
-      if (tensor && !tensor.kept) {
-        keep(tensor);
-      }
-    });
+  private cloneTensorMap(tensorsMap: NamedTensorsMap) {
+    return Object.fromEntries(
+        Object.entries(tensorsMap).map(([name, tensorsList]) => {
+          return [
+            name, tensorsList.map(tensor => {
+              const clone = tensor.clone();
+              keep(clone);
+              return clone;
+            })
+          ];
+        }));
   }
 
   /**
@@ -242,24 +238,16 @@ export class GraphExecutor implements FunctionExecutor {
     const tensorListMap: TensorListMap = {};
     const tensorsMap: NamedTensorsMap = {...this.weightMap};
 
-    const result = tidy(() => {
+    return tidy(() => {
       const context = new ExecutionContext(
           this.weightMap, tensorArrayMap, tensorListMap,
           this.functionExecutorMap);
-      if (this.keepIntermediateTensors) {
-        this.tensorsPendingDisposal = new Set<string>();
-      }
 
       Object.keys(inputs).forEach(name => {
         const [nodeName, index] = parseNodeName(name);
         const tensors: Tensor[] = [];
         tensors[index] = inputs[name];
         tensorsMap[nodeName] = tensors;
-        // TODO(xing.xu@intel.com): for some models, such as bodypix, it will
-        // dispose the input tensors in its top level tidy. In keep intermediate
-        // tensors mode, these tensors are required, so we should call keep to
-        // ensure they are preserved. However, this comes with tensor leak in
-        // keep intermediate tensors mode.
       });
 
       const tensorsToKeep = this.getFrozenTensorIds(tensorsMap);
@@ -276,7 +264,6 @@ export class GraphExecutor implements FunctionExecutor {
                 `Please use model.executeAsync() instead.`);
           }
           tensorsMap[node.name] = tensors;
-          this.keepTensors(tensors, node.name, this.tensorsPendingDisposal);
           this.checkTensorForDisposal(
               node.name, node, tensorsMap, context, tensorsToKeep,
               outputNodeNames, intermediateTensorConsumerCount);
@@ -286,15 +273,12 @@ export class GraphExecutor implements FunctionExecutor {
       if (this.parent == null) {
         context.dispose(tensorsToKeep);
       }
+      if (this.keepIntermediateTensors) {
+        this.clonedTensorsMap = this.cloneTensorMap(tensorsMap);
+      }
+
       return outputs.map(name => getTensor(name, tensorsMap, context));
     });
-
-    if (this.keepIntermediateTensors) {
-      this.tensorsMap = tensorsMap;
-    } else {
-      this.tensorsMap = null;
-    }
-    return result;
   }
 
   private getFrozenTensorIds(tensorMap: NamedTensorsMap): Set<number> {
@@ -366,25 +350,22 @@ export class GraphExecutor implements FunctionExecutor {
   }
 
   disposeIntermediateTensors() {
-    if (!this.keepIntermediateTensors || this.tensorsPendingDisposal == null ||
-        this.tensorsMap == null) {
+    if (!this.clonedTensorsMap) {
       return;
     }
-    for (const nodeName of this.tensorsPendingDisposal) {
-      const tensorArray = this.tensorsMap[nodeName];
-      tensorArray.forEach(tensor => {
+    Object.entries(this.clonedTensorsMap).forEach(([, tensorsList]) => {
+      tensorsList.forEach(tensor => {
         if (tensor && !tensor.isDisposed) {
           tensor.dispose();
         }
       });
-    }
+    });
 
-    this.tensorsPendingDisposal = null;
-    this.keepIntermediateTensors = false;
+    this.clonedTensorsMap = null;
   }
 
   getIntermediateTensors(): NamedTensorsMap {
-    return this.tensorsMap;
+    return this.clonedTensorsMap;
   }
 
   /**
@@ -416,9 +397,6 @@ export class GraphExecutor implements FunctionExecutor {
     // Keep tensors if KEEP_INTERMEDIATE_TENSORS is on.
     try {
       this.keepIntermediateTensors = env().getBool('KEEP_INTERMEDIATE_TENSORS');
-      if (this.keepIntermediateTensors) {
-        this.tensorsPendingDisposal = new Set<string>();
-      }
     } catch (e) {
       this.keepIntermediateTensors = false;
       console.warn(e.message);
@@ -431,28 +409,27 @@ export class GraphExecutor implements FunctionExecutor {
     // Graph with control flow op requires runtime evaluation of the execution
     // order, while without control flow the execution order is pre-determined
     // in the compile method.
-    this.tensorsMap = await this.executeWithControlFlow(
+    const tensorsMap = await this.executeWithControlFlow(
         inputs, context, outputs, isFunctionExecution);
-    const results =
-        outputs.map(name => getTensor(name, this.tensorsMap, context));
+    const results = outputs.map(name => getTensor(name, tensorsMap, context));
 
     // dispose all the intermediate tensors
     const outputIds = results.map(t => t.id);
     const inputIds = Object.keys(inputs).map(name => inputs[name].id);
     const keepIds =
         new Set<number>([...outputIds, ...inputIds, ...this.weightIds]);
-    if (!this.keepIntermediateTensors) {
-      Object.keys(this.tensorsMap).forEach(key => {
-        const tensorArray = this.tensorsMap[key];
-        tensorArray.forEach(tensor => {
-          if (tensor && !tensor.kept && !tensor.isDisposed &&
-              !keepIds.has(tensor.id)) {
-            tensor.dispose();
-          }
-        });
-      });
-      this.tensorsMap = null;
+    if (this.keepIntermediateTensors) {
+      this.clonedTensorsMap = this.cloneTensorMap(tensorsMap);
     }
+
+    Object.entries(tensorsMap).forEach(([, tensorsList]) => {
+      tensorsList.forEach(tensor => {
+        if (tensor && !tensor.kept && !tensor.isDisposed &&
+            !keepIds.has(tensor.id)) {
+          tensor.dispose();
+        }
+      });
+    });
 
     // dispose the context for the root executor
     if (this.parent == null) {
@@ -596,9 +573,6 @@ export class GraphExecutor implements FunctionExecutor {
           }));
         } else {
           tensorMap[nodeName] = tensors;
-          if (this.keepIntermediateTensors) {
-            this.tensorsPendingDisposal.add(nodeName);
-          }
           this.checkTensorForDisposal(
               nodeName, item.node, tensorMap, context, tensorsToKeep,
               outputNames, intermediateTensorConsumerCount);
--- a/tfjs-converter/src/executor/graph_executor.ts
+++ b/tfjs-converter/src/executor/graph_executor.ts
@@ -192,25 +192,21 @@ export class GraphExecutor implements FunctionExecutor {
     return clone;
   }
 
+  private cloneTensorList(tensors: Tensor[]) {
+    if (!tensors) {
+      return null;
+    }
+    const clonedTensor = tensors.map(tensor => {
+      return this.cloneAndKeepTensor(tensor);
+    });
+    return clonedTensor;
+  }
+
   private cloneTensorMap(tensorsMap: NamedTensorsMap): NamedTensorsMap {
     return Object.fromEntries(
         Object.entries(tensorsMap).map(([name, tensorsList]) => {
-          return [
-            name, tensorsList.map(tensor => {
-              return this.cloneAndKeepTensor(tensor);
-            })
-          ];
+          return [name, this.cloneTensorList(tensorsList)];
         }));
-  }
-
-  private cloneTensorList(
-      tensors: Tensor[], nodeName: string, tensorsMap: NamedTensorsMap) {
-    if (!tensorsMap || !tensors || !nodeName) {
-      return;
-    }
-    tensorsMap[nodeName] = tensors.map(tensor => {
-      return this.cloneAndKeepTensor(tensor);
-    });
   }
 
   /**
@@ -273,7 +269,9 @@ export class GraphExecutor implements FunctionExecutor {
         const tensors: Tensor[] = [];
         tensors[index] = inputs[name];
         tensorsMap[nodeName] = tensors;
-        this.cloneTensorList(tensors, nodeName, this.clonedTensorsMap);
+        if (this.keepIntermediateTensors) {
+          this.clonedTensorsMap[nodeName] = this.cloneTensorList(tensors);
+        }
       });
 
       const tensorsToKeep = this.getFrozenTensorIds(tensorsMap);
@@ -290,7 +288,9 @@ export class GraphExecutor implements FunctionExecutor {
                 `Please use model.executeAsync() instead.`);
           }
           tensorsMap[node.name] = tensors;
-          this.cloneTensorList(tensors, node.name, this.clonedTensorsMap);
+          if (this.keepIntermediateTensors) {
+            this.clonedTensorsMap[node.name] = this.cloneTensorList(tensors);
+          }
           this.checkTensorForDisposal(
               node.name, node, tensorsMap, context, tensorsToKeep,
               outputNodeNames, intermediateTensorConsumerCount);
@@ -377,11 +377,11 @@ export class GraphExecutor implements FunctionExecutor {
       return;
     }
     Object.values(this.clonedTensorsMap).forEach(tensorsList => {
-      tensorsList.forEach(tensor => {
+      for (const tensor of tensorsList) {
         if (tensor && !tensor.isDisposed) {
           tensor.dispose();
         }
-      });
+      }
     });
 
     this.clonedTensorsMap = null;
@@ -450,8 +450,7 @@ export class GraphExecutor implements FunctionExecutor {
 
     Object.values(tensorsMap).forEach(tensorsList => {
       tensorsList.forEach(tensor => {
-        if (tensor && !tensor.kept && !tensor.isDisposed &&
-            !keepIds.has(tensor.id)) {
+        if (tensor && !tensor.isDisposed && !keepIds.has(tensor.id)) {
           tensor.dispose();
         }
       });
@@ -585,7 +584,9 @@ export class GraphExecutor implements FunctionExecutor {
         if (util.isPromise(tensors)) {
           promises.push(tensors.then(t => {
             tensorMap[nodeName] = t;
-            this.cloneTensorList(t, nodeName, this.clonedTensorsMap);
+            if (this.keepIntermediateTensors) {
+              this.clonedTensorsMap[nodeName] = this.cloneTensorList(t);
+            }
             context.currentContext = currentContext;
             this.checkTensorForDisposal(
                 nodeName, item.node, tensorMap, context, tensorsToKeep,
@@ -596,7 +597,9 @@ export class GraphExecutor implements FunctionExecutor {
           }));
         } else {
           tensorMap[nodeName] = tensors;
-          this.cloneTensorList(tensors, nodeName, this.clonedTensorsMap);
+          if (this.keepIntermediateTensors) {
+            this.clonedTensorsMap[nodeName] = this.cloneTensorList(tensors);
+          }
           this.checkTensorForDisposal(
               nodeName, item.node, tensorMap, context, tensorsToKeep,
               outputNames, intermediateTensorConsumerCount);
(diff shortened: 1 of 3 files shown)


axinging:

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -45,10 +52,14 @@ export class GraphExecutor implements FunctionExecutor {
   private _functions: {[key: string]: Graph} = {};
   private _functionExecutorMap: {[key: string]: FunctionExecutor} = {};
   private _resourceManager: ResourceManager;
-  private intermediateTensors: NamedTensorsMap = {};
-  private keepIds: Set<number>;
+  // Variables with Async suffix is used for dumping by executeAsync.
+  private tensorsPendingDisposalForAsync: NamedTensorsMap;
+  private idsKeepForAsync: Set<number>;
+  // Variables with Sync suffix is used for dumping by execute.
+  private tensorsPendingDisposalForSync: Tensor[];
```
qjia7, line 59: Can you unify to use `tensorsPendingDisposal` or `intermediateTensorsPendingDisposal` for both async and sync for simplicity?

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -45,10 +52,14 @@ export class GraphExecutor implements FunctionExecutor {
...
+  // tensorsMap is used by both execute and executeAsync.
   private tensorsMap: NamedTensorsMap;
-  private keepTensorForDebug = false;
+  private dumpMode = DumpMode.Default;
```
qjia7, line 62: A boolean variable `keepTensorForDebug` like before is enough? I think the `tidy` won't be used for async execute.

tfjs-converter/src/executor/graph_executor.ts
axinging: Done

tfjs-converter/src/executor/graph_executor.ts
axinging: This is for: 
https://github.com/axinging/tfjs/blob/ca5dbc94cbae0848af0a38d61f58aefe81240c4b/tfjs-converter/src/executor/graph_executor.ts#L346

In Default mode,  it does dispose; Sync mode,  does nothing; Async mode: pushes tensor to disposal queue.
```
                if (this.dumpMode === DumpMode.Default) {
                  tensor.dispose();
                } else if (this.dumpMode === DumpMode.Async) {
                  this.tensorsPendingDisposal.push(tensor);
                }
```

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -217,20 +239,33 @@ export class GraphExecutor implements FunctionExecutor {
...
+        this.tensorsPendingDisposal = [];
+      }
 
       Object.keys(inputs).forEach(name => {
         const [nodeName, index] = parseNodeName(name);
         const tensors: Tensor[] = [];
         tensors[index] = inputs[name];
         tensorsMap[nodeName] = tensors;
+        // Input tensors should be disposed by user.
+        this.keepTensors(tensors);
```
qjia7, line 268: Why we need to add `keepTensors` for inputs compared with the original code?

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -258,6 +294,13 @@ export class GraphExecutor implements FunctionExecutor {
       }
       return outputs.map(name => getTensor(name, tensorsMap, context));
     });
+
+    if (this.dumpMode === DumpMode.Sync) {
+      this.tensorsMap = tensorsMap;
+    } else {
+      this.tensorsMap = null;
```
qjia7, line 301: It's safe to say `this.tensorsMap = tensorsMap` since L280 make sure that async won't go to this path?

tfjs-converter/src/executor/graph_executor.ts
qjia7: In sync mode, it won't go to this code since `tensor.kept = true` since you add `keepTensors` before it?

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -337,25 +374,32 @@ export class GraphExecutor implements FunctionExecutor {
...
-                tensor => tensor.dispose()));
-    this.disposeTensorsMap();
+    if (this.tensorsPendingDisposal) {
+      this.tensorsPendingDisposal.forEach(tensor => {
+        tensor.dispose();
+      });
+      this.tensorsPendingDisposal = null;
+    }
+    if (this.dumpMode === DumpMode.Async) {
+      this.disposeTensorsMap();
```
qjia7, line 387: why we need to call `this.disposeTensorsMap()` only for async mode?

axinging: This is intended behaviour: 
In sync mode,  intermediat tensors are kept in tensorsPendingDisposal.
In async mode, intermediat tensors = = tensormap - tensorsToKeep.

This is why we have two different dispose. 
Please note in the updated version, I tried to align these two path, which means, for both mode, all intermediat tensors are kept in tensorsPendingDisposal.

tfjs-converter/src/executor/graph_executor.ts
axinging: I add some comments above this line:
```
        // For some models, such as bodypix, it will dispose the input tensors
        // in its top level tidy. In dump mode, these tensors are required, so
        // call keep to eusure they are preserved. However, this comes with a
        // side effect in dump mode, tensor leak.
```

tfjs-converter/src/executor/graph_executor.ts
axinging: You are right.  Put code under "(this.dumpMode === DumpMode.Async)" help to remind this is dump specific.
BTW, I move this logic outside this checkTensorForDisposal in the latest change.

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -45,10 +52,11 @@ export class GraphExecutor implements FunctionExecutor {
   private _functions: {[key: string]: Graph} = {};
   private _functionExecutorMap: {[key: string]: FunctionExecutor} = {};
   private _resourceManager: ResourceManager;
-  private intermediateTensors: NamedTensorsMap = {};
-  private keepIds: Set<number>;
+  // Variables with Async suffix is used for dumping by executeAsync.
```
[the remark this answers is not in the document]
axinging: done

tfjs-converter/src/executor/graph_executor.ts
axinging: Yes. This 'if' help to remind this is dump specific. Under non-dump mode, this.tensorsMap will never be used.

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -247,6 +276,7 @@ export class GraphExecutor implements FunctionExecutor {
                 `Please use model.executeAsync() instead.`);
           }
           tensorsMap[node.name] = tensors;
+          this.keepTensors(tensors, this.tensorsPendingDisposal);
```
qjia7, line 279:
```
if (this.keepIntermediateTensors)
{
  this.keepTensors(tensors, this.tensorsPendingDisposal);
}

File: tfjs-converter/src/executor/graph_executor.ts (hunk 3 above)
Comment (axinging) [reply tfjs-converter/src/executor/graph_executor.ts:262]: This leak only happens in dump mode.  And 'tensors' includes some tensors will be used in later inference, especially in e2e, which will run inference on two different backends.

File: tfjs-converter/src/executor/graph_executor.ts (hunk 7 above)
Comment (axinging) [reply tfjs-converter/src/executor/graph_executor.ts:279]: keepTensors is being call at two place, so I move this conditional inside keepTensors,

File: tfjs-converter/src/executor/graph_executor.ts [hunk 8]
```diff
@@ -45,10 +45,10 @@ export class GraphExecutor implements FunctionExecutor {
   private _functions: {[key: string]: Graph} = {};
   private _functionExecutorMap: {[key: string]: FunctionExecutor} = {};
   private _resourceManager: ResourceManager;
-  private intermediateTensors: NamedTensorsMap = {};
-  private keepIds: Set<number>;
+  private idsKeepForAsync: Set<number>;
+  private tensorsPendingDisposal: Tensor[];
   private tensorsMap: NamedTensorsMap;
-  private keepTensorForDebug = false;
+  private keepTensorsForDump = false;
```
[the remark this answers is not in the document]
axinging: done

tfjs-converter/src/executor/graph_executor.ts
qjia7: Anyway, I think we need to fix it. We need to find which place releases it and debug why it's released and then reused again in debug mode. Or give reason that why tensor leak is unavoidable.

qjia7 (commented):
Thanks Xing. 
The code looks good in my side. But the tensor leak issue still needs you to do more investigation to figure out the reason.
Will add more reviewers. Thanks.

axinging:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6953)_ status: 0 of 1 approvals obtained (waiting on @axinging, @jinjingforever, and @qjia7)
> 
> _[`tfjs-converter/src/executor/graph_executor.ts` line 193 at r6](https://reviewable.io/reviews/tensorflow/tfjs/6953#-NFPLIDJ8GQxBg0KSfZk:-NFPLIDJ8GQxBg0KSfZl:bfgygr6) ([raw file](https://github.com/tensorflow/tfjs/blob/ca5dbc94cbae0848af0a38d61f58aefe81240c4b/tfjs-converter/src/executor/graph_executor.ts#L193)):_
> 
> > ```
> >   }
> > 
> >   private keepTensors(
> > ```
> 
> this assumes there will be no duplication of tensors from the tensorsToKeep array, might be better to change tensorsPendingDisposal array to a set to enable auto-dedupe.

=> Done

> 
> _[`tfjs-converter/src/executor/graph_executor.ts` line 268 at r6](https://reviewable.io/reviews/tensorflow/tfjs/6953#-NEi3uiI4S1nH6oTL-PG-r6-268:-NFPKHCVCFRmSqvxLbGh:b-spkb8s) ([raw file](https://github.com/tensorflow/tfjs/blob/ca5dbc94cbae0848af0a38d61f58aefe81240c4b/tfjs-converter/src/executor/graph_executor.ts#L268)):_
> 
> Previously, qjia7 (Jiajia Qin) wrote…
> I am against keeping the input tensors as intermediate tensor of graph model, since they are not generated inside the model. This is beyond the tfjs model dump, it is part of the model API pipeline dump.

=> Done. With this change, for bodypix, we can not get the whole intermediate tensors because the inputs are disposed.

> 
> _[`tfjs-converter/src/executor/graph_executor.ts` line 308 at r6](https://reviewable.io/reviews/tensorflow/tfjs/6953#-NFPNHKl7fab3-eE_hAA:-NFPNHKl7fab3-eE_hAB:b-4hjg8o) ([raw file](https://github.com/tensorflow/tfjs/blob/8f0ece516143e9956b3eab53901a04842d765af9/tfjs-converter/src/executor/graph_executor.ts#L308)):_
> 
> > ```
> >                     this.intermediateTensors[nodeName][index] = tensor;
> >                   } else {
> >                     this.intermediateTensors[nodeName] = [];
> > ```
> 
> the intermediateTensors contains the name for the tensors, how is the current approach tracking the name?
=> The original intermediateTensors is a misnomer, it should be tensorsPendingDisposal.  The API getIntermediateTensors returns this.tensorMap. So the name in original intermediateTensors is of no use.

@pyu10055 ,updated, PTAL

axinging:
@pyu10055 @jinjingforever 
I will use bodypix to explain the background if or not keep input Tensors.
```
// https://github.com/tensorflow/tfjs-models/blob/master/body-pix/src/base_model.ts#L74
return tf.tidy(() => {
  const asFloat = this.preprocessInput(tf.cast(input, 'float32'));
  const asBatch = tf.expandDims(asFloat, 0);
  const results = this.model.predict(asBatch) as tf.Tensor4D[];
```

From above code, we can see asBatch will be disposed after tidy returns. 
However, in the later dump, the input(asBatch) is required when running predictOp.
We will use the input(asBatch) together with other inputs to construct a tensorMap, and run predictOp.
So if the value of asBatch is being disposed, predictOp will fail at this op.

We have two options about this:
1. do not keep input tensors,  some models, such as bodypix, will be un-dumpable.
2. keep input tensors,  bodypix will have tensor leaks when dump is enabled.

Our goal is to support dumping on more models. And this tensor leak happens only when dump is on. So I personally prefer keeping these input tensors. WDYT?

pyu10055 (commented):
thank @axinging for explaining, if I understand correctly, you are using the dump mode for the graph model to do an tensor audit for tfjs-models API that does not share a specific model file?
That sounds like quite different from the intermediate tensor dumping use case, still not clear in this case, why the inputs cannot be disposed?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6953#-:-NFiw7sr-A66w1dwSjcl:b-y4nzww)* status: 0 of 1 approvals obtained (waiting on @jinjingforever and @qjia7)

axinging:
> thank @axinging for explaining, if I understand correctly, you are using the dump mode for the graph model to do an tensor audit for tfjs-models API that does not share a specific model file? That sounds like quite different from the intermediate tensor dumping use case, still not clear in this case, why the inputs cannot be disposed?
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6953#-:-NFiw7sr-A66w1dwSjcl:b-y4nzww)_ status: 0 of 1 approvals obtained (waiting on @jinjingforever and @qjia7)

**About: "why the inputs cannot be disposed?" ?**
I think we have two considerations regarding to this question: 
1. graphModel.getIntermediateTensors is expected to get all tensors: input tensors, weights, intermediate output tensors ,outputs. If the inputs are disposed, graphModel.getIntermediateTensors is not full functional.
2.  As we know, the previouse debug mode has a problem: Errors may accumulate, resulting in fake errors.  For example, subgraph {opA => opB => opC}. If opA is wrong, both opB and opC will be wrong.
  In order to get rid of fake errors, the new Dump works in two steps:
  First,Dump tensors into files according to dumpLevel.  This is similar to normal predict, but all the intermediate tensors(input tensors, weights, intermediate output tensors ,outputs) are reserved. Code: https://github.com/tensorflow/tfjs/blob/master/e2e/benchmarks/local-benchmark/index.html#L239
  Second, When tensor diffs spotted, apply below to each tensor related op: 
  use the reference as input, run the op(by predictOp) again under actual backend. Then dump all the results into files.
  code: https://github.com/tensorflow/tfjs/blob/master/e2e/benchmarks/local-benchmark/dump.js#L204
  The predictOp in 2nd step ensures all fake errors are removed. And it also requires all the input tensors, otherwise this predictOp will fail, which means the model dump is incomplete.

**Conclusion**: for both graphModel.getIntermediateTensors and remove fake errors, all input tensors are required.

**Input tensors are used differently on e2e**

In current e2e, there are two kinds of input tensor use scenarios, examples are MobileNetV3 and bodypix.
1. MobileNetV3
```
    predictFunc: () => {
      return predictFunction(input);
    },
```
In MobileNetV3, the input is never disposed at user side(This is a tensor leak).  So the dump of MobileNetV3 is of full functionality. And keep this tensor in graph model is meanless for dump.

2. bodypix
```
// https://github.com/tensorflow/tfjs-models/blob/master/body-pix/src/base_model.ts#L74
return tf.tidy(() => {
  const asFloat = this.preprocessInput(tf.cast(input, 'float32'));
  const asBatch = tf.expandDims(asFloat, 0);
  const results = this.model.predict(asBatch) as tf.Tensor4D[];
```
As mentioned before, input tensor asBatch will be disposed in tidy. If we do not keep it(call keep) in graph model, graphModel.getIntermediateTensors can not get all the required tensors, and remove fake errors is incomplete. (This means dump on bodypix is incomplete)

Based on the above MobileNetV3&bodypix, to have better dump dupport, we'd better keep the input tensors.

pyu10055 (commented):
@axinging Thanks, this makes a lot more sense now, thanks. One nitpick is the name - intermediate tensors is not very intuitive, probably better be called 'inferenceTensorAudit', which could contains inputs + intermediateTensors+outputs.
I think for both MobileNetV3 and bodypix, the input tensor should be disposed as the benchmark is completed.
So, given that the dump has tracked for all tensors, when caller request dispose those tensors, it should be able to remove them all?  
Why there will be memory leak?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6953#-:-NFlDixf2hU6traJz9Us:b7kmjzz)* status: 0 of 1 approvals obtained (waiting on @jinjingforever and @qjia7)

axinging:
> @axinging Thanks, this makes a lot more sense now, thanks. One nitpick is the name - intermediate tensors is not very intuitive, probably better be called 'inferenceTensorAudit', which could contains inputs + intermediateTensors+outputs. I think for both MobileNetV3 and bodypix, the input tensor should be disposed as the benchmark is completed. So, given that the dump has tracked for all tensors, when caller request dispose those tensors, it should be able to remove them all? Why there will be memory leak?
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6953#-:-NFlDixf2hU6traJz9Us:b7kmjzz)_ status: 0 of 1 approvals obtained (waiting on @jinjingforever and @qjia7)

## about: Why there will be memory leak?
We can understand this by models: e2e/MobileNetv3, bodypix.

**e2e/MobileNetv3 : tensor leak can be fixed.**
```
    predictFunc: () => {
      return predictFunction(input);
    },
```
In above code, 'input' will never be disposed, this is a **tensor leak** (This exists long before, un-relating to dump feature). I think this a potential bug and this can be fixed by call input.dispose after predict is done (I will try to work out a fix for this later).

**bodypix: no tensor leak or un-dumpable.**
The original bodypix has no tensor leak, but this will fail dump because asBatch is disposed in tidy.
```
// https://github.com/tensorflow/tfjs-models/blob/master/body-pix/src/base_model.ts#L74
return tf.tidy(() => {
  const asFloat = this.preprocessInput(tf.cast(input, 'float32'));
  const asBatch = tf.expandDims(asFloat, 0);
  const results = this.model.predict(asBatch) as tf.Tensor4D[];
```

To support bodypix dump or avoid tensor leak at dump mode, we have three options at below line:
https://github.com/tensorflow/tfjs/pull/6953/files#diff-44e0a825cd7c6f31c03d9333db5dc21a8937d66a5b28d719c699863eef96ad8dR258

**Option 1: do not keep input tensors.** 
Pros: no tensor leak at bodypix
Cons: bodypix can not be dumped
```
      Object.keys(inputs).forEach(name => {
        const [nodeName, index] = parseNodeName(name);
        const tensors: Tensor[] = [];
        tensors[index] = inputs[name];
        tensorsMap[nodeName] = tensors;
      });
```
**Option 2:  Keep input tensors, not add it to the this.tensorsPendingDisposal.**
Pros: bodypix can be dumped. 
Cons: User need to dispose this input tensor(require changes in tfjs-model/bodypix), otherwise tensor leak.
```
      Object.keys(inputs).forEach(name => {
        const [nodeName, index] = parseNodeName(name);
        const tensors: Tensor[] = [];
        tensors[index] = inputs[name];
        tensorsMap[nodeName] = tensors;
        this.keepTensors(tensors);
      });
```
**Option 3:  Keep input tensors, add it to the this.tensorsPendingDisposal.**
Pros: bodypix can be dumped.  
Cons: User need to clone the input in e2e/MobileNetV3(And other model calls predictFunction in e2e) so that it can be used for the second inference.
```
      Object.keys(inputs).forEach(name => {
        const [nodeName, index] = parseNodeName(name);
        const tensors: Tensor[] = [];
        tensors[index] = inputs[name];
        tensorsMap[nodeName] = tensors;
        this.keepTensors(tensors, nodeName, this.tensorsPendingDisposal);
      });
```

We need consider e2e/MobileNetv3 together with e2e/bodypix here. Because they have different useage of input.
In e2e conformance, all models run two inferences. For e2e/MobileNetv3 the input tensors are shared by two inferences. For e2e/bodypix, input tensors are not shared.
This means, for e2e/MobileNetv3, input tensors can not be disposed after first inference(Option 3 will dispose this). For bodypix, its ok to dispose input. The fix is to clone the input tensors for e2e/MobileNetv3.

The basic workflow of current e2e conformance test:
```
// First inference
predictAndGetData(ExpectedBackend);
if(enableDump) {
 graphModel.getIntermediateTensors();
 graphModel.disposeIntermediateTensors();
}

// Second inference
// Input is shared between two backends. But it will be disposed at first disposeIntermediateTensors if "Input is added to the this.tensorsPendingDisposal".
predictAndGetData(ActualBackend);
if(enableDump) {
 graphModel.getIntermediateTensors();
 graphModel.disposeIntermediateTensors();
}
```
## About name inferenceTensorAudit
I prefer a name like "GraphModel.getNamedTensorsMap", because getIntermediateTensors returns NamedTensorsMap which contains all tensors:
```
  getIntermediateTensors(): NamedTensorsMap {
    return this.tensorsMap;
  }
```
BTW,  if change name from getIntermediateTensors to getNamedTensorsMap, it seems not related to flag KEEP_INTERMEDIATE_TENSORS namely. But from logic, getNamedTensorsMap do return a NamedTensorsMap, and KEEP_INTERMEDIATE_TENSORS do keep all "INTERMEDIATE TENSORS".

## So summarize my proposal:
1. name: GraphModel.getIntermediateTensors => GraphModel.getNamedTensorsMap
2. Keep tensor policy:  prefer "Option 2:  Keep input tensors, not add it to the this.tensorsPendingDisposal"
https://github.com/tensorflow/tfjs/pull/6953/files#diff-44e0a825cd7c6f31c03d9333db5dc21a8937d66a5b28d719c699863eef96ad8dR258
(Tensor leak at dump mode, but enable dump on bodypix.)
3.  Fix e2e/MobileNetv3 tensor leak in a followup PR, and try "Option 3:  Keep input tensors, add it to the this.tensorsPendingDisposal." in a followup PR.

WDYT?
@pyu10055 @jinjingforever

pyu10055 (commented):
If I understand it correct, the key obstacle is the input tensors, can we separate those from the intermediate tensors? 
Give the dump mode another explicit option to dump the intermediate tensor plus the input tensors
On disposal call, we can also separate disposal of input tensors from intermediate tensors, can be an parameter to the disposal function or a separate function.
In this way, both scenarios can control if it needs to dump input tensors and when to dispose the input tensors.
Does this make sense? thanks!
@axinging  

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6953#-:-NFny8IS19iKBERtFSG_:b-keli1v)* status: 0 of 1 approvals obtained (waiting on @jinjingforever and @qjia7)

mattsoulanille:
To add to Ping's discussion, I think we can use `tf.clone` and references to solve this problem without needing special logic to determine when we can `dispose` each tensor. 

Instead of using `keep` to determine when to dispose each tensor, we can `tf.clone` it to get another reference. To do this, instead of saving the `tensorsMap` by setting [`this.tensorsMap = tensorsMap;`](https://github.com/tensorflow/tfjs/pull/6953/files#diff-44e0a825cd7c6f31c03d9333db5dc21a8937d66a5b28d719c699863eef96ad8dR293), we create a copy of it, where we clone each tensor in the map.
```typescript
this.tensorsMap = Object.fromEntries(Object.entries(tensorsMap).map(([name, tensorsList]) => {
  const cloned = tensor.clone();
  // This clone needs to be 'kept' because model.execute may be called within a tidy(). We don't want
  // tidy() to dispose these cloned tensors because we need to look at them after the model has finished
  // executing (after the 'tidy()'). However, we don't check whether the tensor is 'kept' when we free it. 
  keep(clone);
  return [name, tensorsList.map(tensor => tensor.clone())];
}));
```
When a user calls [`getIntermediateTensors`](https://github.com/tensorflow/tfjs/pull/6953/files#diff-44e0a825cd7c6f31c03d9333db5dc21a8937d66a5b28d719c699863eef96ad8dL365), they get this map of cloned tensors, which are guaranteed to not be disposed since nothing else interacts with this map (if `tensorsMap` _is_ used elsewhere, we can just create an extra variable `clonedTensorsMap` and use it instead).

When we need to clean up the intermediate tensors, we no longer need to [iterate `tensorsPendingDisposal`](https://github.com/tensorflow/tfjs/pull/6953/files#diff-44e0a825cd7c6f31c03d9333db5dc21a8937d66a5b28d719c699863eef96ad8dR373-R374) to make sure we only dispose tensors that we created. We can just dispose all the cloned tensors in `tensorsMap`, since they're just clones / extra references to the original tensors:
```typescript
for (const tensorsList in Object.keys(this.tensorsMap)) {
  for (const tensor of tensorsList) {
    // This is a clone of the real tensor (i.e. another reference), so it's okay to dispose it.
    // We're not disposing the user's input tensor here. Just the clone.
    tensor.dispose();
  }
}
```

If the user calls `model.execute` again before calling `disposeIntermediateTensors`, we just call it for them before we clear `tensorsMap` and run the model again.

## Works when a tensor is reused

Suppose the user reuses an input tensor like this:

```typescript

function runModel() {
  return model.execute(input);
}
``` 
We can count the number of references to the underlying data of the `input` tensor as this runs. 
* When `input` is declared, there is 1 reference to the data. `RC = 1`
* When we run `model.execute` with dump mode, we get another reference in `tensorsData`. `RC = 2`.
* We can look at this cloned tensor (and the other intermediate tensors) now.
* When we run `model.disposeIntermediateTensors`, the cloned input tensor is disposed, so `RC = 1` again.
* When we eventually dispose `input`, then `RC = 0` and the data gets deleted.

## Works when a new tensor is created inside a `tidy()`

Suppose the user creates a new tensor each time like this:

```typescript
function runModel() {
  tf.tidy(() => {
    return model.execute(input);
  });
}
``` 
We can count the number of references to the underlying data of the `input` tensor as this runs. 
* When `input` is declared, there is 1 reference to the data. `RC = 1`
* When we run `model.execute` with dump mode, we get another reference in `tensorsData`. `RC = 2`
* When we exit the `tidy`, two things happen:
  * The original `input` tensor is disposed, so `RC = 1` now.
  * The cloned `input` tensor, is _not_ disposed because it was marked as `kept`, so `RC = 1`.
* We can now inspect the intermediate tensors, including the input tensor, because they were all `kept` through the `tidy`.
* When we run `model.disposeIntermediateTensors`, the cloned input tensor is disposed, so `RC = 0` and the data gets deleted.

axinging:
Thanks @pyu10055  @mattsoulanille.  The updated change is with keep+clone, PTAL.

pyu10055 (changes requested):
Thank you @axinging, the PR LGTM with some minor questions. 

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6953#-:-NFt7UF-23jZvik5V4zu:bxp986u)* status: 0 of 1 approvals obtained (waiting on @axinging, @jinjingforever, and @qjia7)

___
*[`tfjs-converter/src/executor/graph_executor.ts` line 193 at r6](https://reviewable.io/reviews/tensorflow/tfjs/6953#-NFPLIDJ8GQxBg0KSfZk:-NFt7KhZ17fkNkibRb13:b-gyedkd) ([raw file](https://github.com/tensorflow/tfjs/blob/ca5dbc94cbae0848af0a38d61f58aefe81240c4b/tfjs-converter/src/executor/graph_executor.ts#L193)):*

=> Done
</blockquote></details>

just want to confirm that inputs are included in the tensorsMap, thanks
___
*[`tfjs-converter/src/executor/graph_executor.ts` line 560 at r13](https://reviewable.io/reviews/tensorflow/tfjs/6953#-NFt6dB553DmxcGNqZL7:-NFt6dB553DmxcGNqZL8:bhy8l61) ([raw file](https://github.com/tensorflow/tfjs/blob/926c46e571f17a4e073a0824e48d1453f7e59472/tfjs-converter/src/executor/graph_executor.ts#L560)):*
> ```XML
>         const currentContext = context.currentContext;
>         if (util.isPromise(tensors)) {
>           if (this.keepIntermediateTensors) {
> ```

why keepIntermediateTensors would not work with promise?
These promise should be resolved to tensors and put into the tensorMap.

axinging (commented):
@pyu10055  @jinjingforever @mattsoulanille 
PTAL.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6953#-:-NFv0CP85hM0b4y_x6Zo:b5a56ud)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @pyu10055, and @qjia7)

___
*[`tfjs-converter/src/executor/graph_executor.ts` line 193 at r6](https://reviewable.io/reviews/tensorflow/tfjs/6953#-NFPLIDJ8GQxBg0KSfZk:-NFut5T3DIEFljGrGbvE:b-s5yvhz) ([raw file](https://github.com/tensorflow/tfjs/blob/ca5dbc94cbae0848af0a38d61f58aefe81240c4b/tfjs-converter/src/executor/graph_executor.ts#L193)):*

just want to confirm that inputs are included in the tensorsMap, thanks
</blockquote></details>

For both async and sync, inputs are in the tensorMap: 
```
    Object.keys(inputs).forEach(name => {
      const [nodeName, index] = parseNodeName(name);
      const tensors: Tensor[] = [];
      tensors[index] = inputs[name];
      tensorsMap[nodeName] = tensors;
    });
```
___
*[`tfjs-converter/src/executor/graph_executor.ts` line 560 at r13](https://reviewable.io/reviews/tensorflow/tfjs/6953#-NFt6dB553DmxcGNqZL7:-NFutGYm9TNWMSnMfmxW:b-8zwfqj) ([raw file](https://github.com/tensorflow/tfjs/blob/926c46e571f17a4e073a0824e48d1453f7e59472/tfjs-converter/src/executor/graph_executor.ts#L560)):*

why keepIntermediateTensors would not work with promise?
These promise should be resolved to tensors and put into the tensorMap.
</blockquote></details>

Yes, this limitation can be removed. And remove it in the latest update.

pyu10055 (approved):
> Reviewed 1 of 3 files at r1, 1 of 2 files at r14, 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6953)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever and @qjia7)

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -256,6 +276,10 @@ export class GraphExecutor implements FunctionExecutor {
       if (this.parent == null) {
         context.dispose(tensorsToKeep);
       }
+      if (this.keepIntermediateTensors) {
+        this.clonedTensorsMap = this.cloneTensorMap(tensorsMap);
+      }
```
qjia7, line 281: nit: Should this `if` happen before L276?

qjia7 (approved):
LGTM with one nit.

tfjs-converter/src/executor/graph_executor.ts
axinging: done

qjia7 (approved):
LGTM, thanks.

@pyu10055 @mattsoulanille @jinjingforever Currently, `getIntermediateTensors()` will return all cloned tensors in the graph, which including inputs, weights, intermediate tensors, outputs. Maybe we need a new meaningful name for it in future. Just list here to get your attention!

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -182,6 +180,22 @@ export class GraphExecutor implements FunctionExecutor {
...
+  private cloneTensorMap(tensorsMap: NamedTensorsMap) {
+    return Object.fromEntries(
+        Object.entries(tensorsMap).map(([name, tensorsList]) => {
+          return [
+            name, tensorsList.map(tensor => {
+              if (tensor == null) {
+                return null;
+              }
+              const clone = tensor.clone();
+              keep(clone);
```
mattsoulanille, line 192: Nit: Adding some docs on why we `keep` the cloned tensor.
```suggestion
               // Keep the clone because `model.execute()` may be called within
               // a `tidy()`, but the user may inspect these tensors after the tidy.
              keep(clone);
```

mattsoulanille (changes requested):
I'm going to take a look at this PR in a few minutes, but I'm sending this review now to prevent it from being merged before I can take a look (since It already has two approvals).

mattsoulanille (changes requested):
Thanks for waiting for my review!

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -202,7 +216,6 @@ export class GraphExecutor implements FunctionExecutor {
         names.map(name => this.graph.nodes[parseNodeName(name)[0]]);
     const outputNodeNames = outputs.map(name => parseNodeName(name)[0]);
     let outputNodes = outputNodeNames.map(name => this.graph.nodes[name]);
```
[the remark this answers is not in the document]
mattsoulanille: Are there any cases where a user would want to collect intermediate tensors from multiple runs of the model? My fix above would make that very difficult because the user would have to clone the tensors again to avoid them being disposed on the next call to `model.execute`.

If this is an expected use case, maybe the model should have a public list of all the intermediate tensors for its past runs, and when `model.disposeIntermediateTensors` is called, it disposes all the tensors in the list.

To be clear, the list would have type `NamedTensorsMap[]`, not `Tensor[]`, and each entry would correspond to a run of the model.

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -252,10 +272,14 @@ export class GraphExecutor implements FunctionExecutor {
               outputNodeNames, intermediateTensorConsumerCount);
         }
       }
+      if (this.keepIntermediateTensors) {
+        this.clonedTensorsMap = this.cloneTensorMap(tensorsMap);
+      }
```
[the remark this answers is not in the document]
axinging: This is possible. But 1), this.cloneTensorMap can not be reused, we need some new logic to do this.
2), Your above mentioned tensorMapForEach refactor is also impacted.

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -414,23 +412,31 @@ export class GraphExecutor implements FunctionExecutor {
...
         new Set<number>([...outputIds, ...inputIds, ...this.weightIds]);
-    if (!this.keepTensorForDebug) {
-      this.disposeTensorsMap();
+    if (this.keepIntermediateTensors) {
+      this.clonedTensorsMap = this.cloneTensorMap(tensorsMap);
     }
 
+    Object.entries(tensorsMap).forEach(([, tensorsList]) => {
+      tensorsList.forEach(tensor => {
+        if (tensor && !tensor.kept && !tensor.isDisposed &&
```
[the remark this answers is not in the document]
axinging: In fact I am not clear why kept is checked here, this is from the original version: https://github.com/tensorflow/tfjs/pull/5659/files#diff-44e0a825cd7c6f31c03d9333db5dc21a8937d66a5b28d719c699863eef96ad8dL369.

I just try to not change the code I am not clear.

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -337,40 +353,22 @@ export class GraphExecutor implements FunctionExecutor {
...
-      tensorArray.forEach(tensor => {
-        if (tensor && !tensor.kept && !tensor.isDisposed &&
-            !this.keepIds.has(tensor.id)) {
+    Object.entries(this.clonedTensorsMap).forEach(([, tensorsList]) => {
+      tensorsList.forEach(tensor => {
+        if (tensor && !tensor.isDisposed) {
           tensor.dispose();
         }
       });
     });
```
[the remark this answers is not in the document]
axinging: Maybe we can do this in a follow up change?

tfjs-converter/src/executor/graph_executor.ts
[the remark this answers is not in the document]
axinging: collect intermediate tensors from multiple runs is not supported by dump yet.

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -399,13 +397,13 @@ export class GraphExecutor implements FunctionExecutor {
...
 
-    // For model debug.
+    // Keep tensors if KEEP_INTERMEDIATE_TENSORS is on.
     try {
-      this.keepTensorForDebug = env().getBool('KEEP_INTERMEDIATE_TENSORS');
+      this.keepIntermediateTensors = env().getBool('KEEP_INTERMEDIATE_TENSORS');
     } catch (e) {
+      this.keepIntermediateTensors = false;
       console.warn(e.message);
     }
```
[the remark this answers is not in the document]
axinging: This is great, thanks. Done

tfjs-converter/src/executor/graph_executor.ts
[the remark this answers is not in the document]
axinging: collect intermediate tensors from multiple runs is not supported by dump yet.

tfjs-converter/src/executor/graph_executor.ts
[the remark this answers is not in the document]
qjia7: I think Matt is right. For the original intermediate tensors, we should clone each of them when we get them in L274. Otherwise, we are leaking them due to the changes (L330) you add in `this.checkTensorForDisposal`. Since you already use cloned tensor, theoretically, `this.checkTensorForDisposal` should be unchanged.  And the code logic you mentioned can be changed based on our requirement.
Similar for the async execute.

tfjs-converter/src/executor/graph_executor.ts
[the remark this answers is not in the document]
mattsoulanille: True, the last call's tensors will still be leaked. However, this change will limit the amount of leaking that's possible, so a model can only ever leak however many intermediate tensors it has (and not more).

I wouldn't be opposed to removing this, though, especially since it makes it hard to collect intermediate tensors from multiple runs. I'll leave it up to you.

tfjs-converter/src/executor/graph_executor.ts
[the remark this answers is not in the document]
mattsoulanille: I should have made this a Nit suggestion. We can keep the current implementation as-is.

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -225,12 +264,16 @@ export class GraphExecutor implements FunctionExecutor {
...
+      if (this.keepIntermediateTensors) {
+        this.clonedTensorsMap = this.cloneTensorMap(this.weightMap);
+      }
 
       Object.keys(inputs).forEach(name => {
         const [nodeName, index] = parseNodeName(name);
         const tensors: Tensor[] = [];
         tensors[index] = inputs[name];
         tensorsMap[nodeName] = tensors;
+        this.cloneTensorList(tensors, nodeName, this.clonedTensorsMap);
```
mattsoulanille, line 276: Nit: Make `this.cloneTensorsList` return a cloned list of tensors instead of storing them in a map. I think this is a bit easier to read, and it matches `this.cloneTensorsMap`. Also, then `this.cloneTensorsMap` can use `this.cloneTensorsList` in its implementation.

```suggestion
        this.clonedTensorsMap[nodeName] = this.cloneTensorList(tensors);
```

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -247,15 +290,18 @@ export class GraphExecutor implements FunctionExecutor {
                 `Please use model.executeAsync() instead.`);
           }
           tensorsMap[node.name] = tensors;
+          this.cloneTensorList(tensors, node.name, this.clonedTensorsMap);
```
mattsoulanille, line 293: Does this need a check for `this.keepIntermediateTensors`?

tfjs-converter/src/executor/graph_executor.ts
mattsoulanille, line 276: Tensors should only be cloned if `this.keepIntermediateTensors` is set, right?

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -337,40 +373,22 @@ export class GraphExecutor implements FunctionExecutor {
...
-      return;
-    }
-    Object.keys(this.tensorsMap).forEach(key => {
-      const tensorArray = this.tensorsMap[key];
-      tensorArray.forEach(tensor => {
-        if (tensor && !tensor.kept && !tensor.isDisposed &&
-            !this.keepIds.has(tensor.id)) {
+    Object.values(this.clonedTensorsMap).forEach(tensorsList => {
+      tensorsList.forEach(tensor => {
+        if (tensor && !tensor.isDisposed) {
```
mattsoulanille, line 381: Nit: Google's TS style guide [prefers `for (... of ...)` instead of `.forEach(...)`](https://google.github.io/styleguide/tsguide.html#iterating-containers) (`.map(...)` is still okay when using the resulting value).

tfjs-converter/src/executor/graph_executor.ts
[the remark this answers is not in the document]
mattsoulanille: My guess is this was from the previous logic that used `kept` to save the intermediate tensors, although I could be wrong.

It looks like the `keepIds` set is what prevents this from disposing the input, output, and weight tensors, so we're not using `tensor.kept` to prevent them from being disposed. If all the other tensors in `tensorsMap` are intermediate tensors, then I think it's safe to remove this check.

tfjs-converter/src/executor/graph_executor.ts
mattsoulanille: Oh, I see, `cloneTensorsList` checks if `this.clonedTensorsMap` is null before running. I find that a bit confusing. Can we do the check here instead?

Also, I think it might be better for `this.cloneTensorsList` to return a list of cloned tensors instead of mutating the tensors map. See my nit.

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -182,6 +180,39 @@ export class GraphExecutor implements FunctionExecutor {
...
+          return [
+            name, tensorsList.map(tensor => {
+              return this.cloneAndKeepTensor(tensor);
+            })
+          ];
+        }));
+  }
+
+  private cloneTensorList(
+      tensors: Tensor[], nodeName: string, tensorsMap: NamedTensorsMap) {
```
mattsoulanille, line 207:
```suggestion
      tensors?: Tensor[], nodeName?: string, tensorsMap?: NamedTensorsMap) {
```
We'd like to enable strictNullChecks in the future, so if you don't accept the nit that changes this function to return a list of tensors instead of mutating `tensorsMap`, please mark all these arguments as optional.

tfjs-converter/src/executor/graph_executor.ts
mattsoulanille: No, it doesn't, because `this.clonedTensorsMap` is null, but IMO that's a bit confusing. See my other comment on this.

tfjs-converter/src/executor/graph_executor.ts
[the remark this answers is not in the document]
mattsoulanille: If you do end up removing it, make sure to remove it from `execute()` as well.

mattsoulanille (changes requested):
Thanks for making the changes I requested! Here are a few more, and then I think this will be ready to merge.

tfjs-converter/src/executor/graph_executor.ts
axinging: return a list of tensors

tfjs-converter/src/executor/graph_executor.ts
axinging: done

tfjs-converter/src/executor/graph_executor.ts
axinging: done

tfjs-converter/src/executor/graph_executor.ts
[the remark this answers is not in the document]
axinging: removed

axinging (commented):
@mattsoulanille PTAL

mattsoulanille (approved):
LGTM. Thanks!

PR #6957, opened 2022-10-19
Add path mapping to the root tsconfig

mattsoulanille:
Use TypeScript [path mapping](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) to specify paths between monorepo packages. This allows editors to correctly "go to definition" across packages, and ensures they reference the `.ts` file instead of `.d.ts`.

This could also work for `@tensorflow/tfjs-converter` generated `op_list` files, but that would require changing the relative imports to absolute imports from dist, e.g. `@tensorflow/tfjs-converter/dist/operations/op_list/basic_math`.

Before:
image

After:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6957)


tfjs-converter/src/operations/executors/basic_math_executor_test.ts
```diff
@@ -19,7 +19,7 @@ import {scalar} from '@tensorflow/tfjs-core';
 import * as tfOps from '@tensorflow/tfjs-core/dist/ops/ops_for_converter';
 
 import {ExecutionContext} from '../../executor/execution_context';
-import * as basic_math from '../op_list/basic_math';
+import * as basic_math from '@tensorflow/tfjs-converter/dist/operations/op_list/basic_math';
```
mattsoulanille, line 22: I'm using full paths here because path mapping does not appear to work on relative paths.

Linchenn (approved):
This is really helpful, thank you Matt!

pyu10055 (approved):

PR #6969, opened 2022-10-22
[wasm] Fix AvgPool and MaxPool for 1x1 kernels

### LINKED ISSUE #6867: [benchmarks][wasm] Test correctness of MobileNetV3 is false
Steps to reproduce:
1. Open https://tensorflow.github.io/tfjs/e2e/benchmarks/local-benchmark/index.html 
2. Choose `MobileNetV3` and `wasm` backend, click `Test correctness`.

Expect `Prediction matches CPU` is `true`. Actually got `false`.
OS: windows
browser: 105.0.5195.127

mattsoulanille:
XNNPack does not support 1x1 kernels for AvgPool or MaxPool. Implement these cases manually, including support for strides.

Support for padding is still TODO.

#6867 could be fixed by this issue, but I've yet to confirm that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6969)


tfjs-backend-wasm/src/cc/util.h
```diff
@@ -152,6 +152,15 @@ const std::vector<size_t> assert_and_get_broadcast_shape(
 const std::vector<size_t> get_broadcast_dims(
     const std::vector<size_t> in_shape, const std::vector<size_t> out_shape);
 
+// Generates the output for AvgPool, MaxPool, etc where xnnpack does not support
+// a 1x1 filter. Applies batching, channels, and strides.
+// TODO(mattsoulanille): Support padding.
```
Linchenn, line 157: Could we either add an error in `identity_pool` or add a condition in the references for cases `padding != 0`? Otherwise, this will silently produce a wrong result.

Linchenn (approved):
LGTM, just some nits. Thank you!

tfjs-backend-wasm/src/cc/util.h
mattsoulanille: I did some more reading on AvgPool padding, and it seems like it's meaningless for the 1x1 kernel case.

Unlike for convolution ops, padding for AvgPool and MaxPool does not take into account the padded values when computing the result at a given index. It [only uses the values that are part of the original tensor](https://www.tensorflow.org/api_docs/python/tf/nn#difference_between_convolution_and_pooling_layers_2) (and padding just allows it to compute at a given index instead of returning a tensor of smaller dimension than the input). Pooling with a 1x1 kernel will always output the same dimension as the input, so padding should not affect the result. 

I also think I discovered a bug in the CPU backend; Padding a 1x1 kernel can result in NaNs in the output tensor.

As far as I can tell right now, the wasm version is actually correct to completely ignore the padding (for 1x1), so I'll remove the TODO.

tfjs-backend-wasm/src/cc/util.cc
```diff
@@ -87,5 +88,53 @@ const std::vector<size_t> get_broadcast_dims(
...
+                         const size_t input_height, const size_t input_width,
+                         const size_t stride_height, const size_t stride_width,
+                         const size_t channels) {
+  // Early bailout for the identity case to use memcpy for efficiency.
+  if (stride_width == 1 && stride_height == 1) {
+    std::memcpy(out_buf, x_buf, out_size * sizeof(*out_buf));
+    return;
+  }
+
+  // Values per row and column as determined by the stride size.
```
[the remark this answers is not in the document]
mattsoulanille: Done. Thanks!

tfjs-backend-wasm/src/cc/util.h
Linchenn: Makes sense. Thank you!

jinjingforever (approved):
Thank you fixing the bug and adding tests!

PR #6994, opened 2022-11-01
webgpu: support bincount and denseBincount operators

xhcao:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6994)

### DIFF
--- a/tfjs-backend-webgpu/src/bincount_webgpu.ts
+++ b/tfjs-backend-webgpu/src/bincount_webgpu.ts
@@ -47,7 +47,6 @@ export class BincountProgram implements WebGPUProgram {
   variableNames = ['x'];
   uniforms = 'binCountSize : i32,';
   workgroupSize: [number, number, number] = [64, 1, 1];
-  size = true;
   atomic = true;
   hasWeights = true;
   binaryOutput = false;
@@ -98,7 +97,7 @@ export class BincountProgram implements WebGPUProgram {
                 this.binaryOutput ?
                     1. :
                     (this.hasWeights ? 'f32(getW(coord[0], coord[1]))' : '1.')};
-        bincount_write(coord.x * uniforms.outShape[1] + indexVal, value);
+        bincount_write(coord.x * uniforms.binCountSize + indexVal, value);
       }
     }`}
   }
--- a/tfjs-backend-webgpu/src/bincount_webgpu.ts
+++ b/tfjs-backend-webgpu/src/bincount_webgpu.ts
@@ -39,7 +39,7 @@ const binaryWriteSnippet = `
   }
 `;
 
-export class BincountRankOneProgram implements WebGPUProgram {
+export class BincountProgram implements WebGPUProgram {
   outputShape: number[] = [];
   shaderKey: string;
   dispatchLayout: {x: number[]};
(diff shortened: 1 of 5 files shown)


tfjs-backend-webgpu/src/bincount_webgpu.ts
```diff
@@ -0,0 +1,134 @@
...
+            1. :
+            (this.hasWeights ? 'f32(getW(coord[0], coord[1]))' : '1.')};
+        bincount_write(coord.x * uniforms.outShape[1] + indexVal, value);
+      }
+    }
+  }
+  `;
+    return userCode;
+  }
+}
```
gyagp, line 134: The code of these 2 programs are very similar, can we merge them?

xhcao:
webgl backend uses cpu implementation, which will read the gpu data back to cpu side, and it will throw an error when input's value is negative. There is different for webgpu backend, which throw a shader compiling error for access out of array bound. Is it reasonable here?

tfjs-backend-webgpu/src/kernels/Bincount.ts
```diff
@@ -0,0 +1,63 @@
...
+        {inputs: BincountInputs, backend: WebGPUBackend, attrs: BincountAttrs}):
+    TensorInfo {
+  const {inputs, backend, attrs} = args;
+  const {x, weights} = inputs;
+  const {size} = attrs;
+
+  const xSize = util.sizeFromShape(x.shape);
+  const weightsSize = util.sizeFromShape(weights.shape);
+  const hasWeights = weightsSize > 0;
+  const outputSize: [number] = [Math.max(xSize, size)];
```
qjia7, line 37: Please don't use the same `outputSize` for `BincountProgram` and `fill`. So you don't need to do extra `slice`.

tfjs-backend-webgpu/src/bincount_webgpu.ts
```diff
@@ -0,0 +1,108 @@
...
+        let value = ${
+                this.binaryOutput ?
+                    1. :
+                    (this.hasWeights ? 'f32(getW(index))' : '1.')};
+        bincount_write(indexVal, value);
+      }
+    }` :
+            `let coord = getCoordsFromIndex(index);
+    if (coordsInBounds2D(coord, uniforms.xShape)) {
+      let indexVal = i32(getX(coord[0], coord[1]));
```
qjia7, line 95: Can we reshape x to a 1D tensor so that we can unify the code in shader?

qjia7:
> webgl backend uses cpu implementation, which will read the gpu data back to cpu side, and it will throw an error when input's value is negative. There is different for webgpu backend, which throw a shader compiling error for access out of array bound. Is it reasonable here?

Please clamp the index value for webgpu.

tfjs-backend-webgpu/src/kernels/Bincount.ts
xhcao: Using `slice` aims to dispatch threads according output shapes, which is the normal case in almost all operators.
If dispatching threads according input shapes, it can indeed get better performance. @haoyunfeix , had you mentioned that there was an error for `matmul_splitK` when using input shapes to dispatch threads?
I personally prefer to using output shapes here. @gyagp @hujiajie any comments here?

tfjs-backend-webgpu/src/bincount_webgpu.ts
xhcao: It is more difficult to compute the 2D coords if unifying the code.

tfjs-backend-webgpu/src/kernels/Bincount.ts
xhcao: Had changed dispatch size based on input size.

xhcao (commented):
Please take a look again. The other comment is about negative input value, I think it is better to throw a runtime error than clamping the value, which tell the user a incorrect input, instead of hiding the error.

tfjs-backend-webgpu/src/bincount_webgpu.ts
```diff
@@ -0,0 +1,108 @@
...
+    var oldValue = atomicLoad(& (result[index]));
+    var exchanged = false;
+    for (; !exchanged;) {
+      let newValueF32 = bitcast<f32>(oldValue) + value;
+      let newValue = bitcast<i32>(newValueF32);
+      let res = atomicCompareExchangeWeak(
+          &(result[index]), oldValue, newValue);
+      oldValue = res.old_value;
+      exchanged = res.exchanged;
+    }
```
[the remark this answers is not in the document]
xhcao: @hujiajie Thanks, good comment, I will refactor this code in the next patch, let us keep them here in this patch.

gyagp (approved):
LGTM, thanks!

tfjs-backend-webgpu/src/bincount_webgpu.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export class BincountProgram implements WebGPUProgram {
+  outputShape: number[] = [];
+  shaderKey: string;
+  dispatchLayout: {x: number[]};
+  dispatch: [number, number, number];
+  variableNames = ['x'];
+  uniforms = 'binCountSize : i32,';
+  workgroupSize: [number, number, number] = [64, 1, 1];
+  size = true;
```
qjia7, line 50: nit: It seems that `uniforms.size` is never used. So it can be removed?

qjia7 (approved):
LGTM with one nit.

tfjs-backend-webgpu/src/bincount_webgpu.ts
xhcao: Done

qjia7:
@pyu10055 FYI  For these two ops, webgl is using [cpu implementation](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/kernels/Bincount.ts#L36). In this PR, webgpu uses compute shader to support them.

Issue #7008, opened 2022-11-02
[all] guard against decimal shape value across TF ops


pyu10055:
<em>Please make sure that this is a bug. As per our
[GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md),
we only address code/doc bugs, performance issues, feature requests and
build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js):
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04):
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device:
- TensorFlow.js installed from (npm or script link): 4.0.0
- TensorFlow.js version (use command below): 4.0.0
- Browser version:
- Tensorflow.js Converter Version:

**Describe the current behavior**
Although all TFJS op APIs require integer shapes, users can still use shape values in decimal when passing the shape parameter.
This would cause backend errors (i.e. webgl shader error when interpolating the shape values in the shader)
For ops or backends that does not throw errors, the tensor created will contain invalid state for shape and strides (i.e. [1, 2.5, 3.3333]).
For ops on backends that do throw errors, the error message is very [cryptic](https://github.com/tensorflow/tfjs/issues/6987). 

This invalid state should be guarded and provide a meaningful error message.

**Describe the expected behavior**

**Standalone code to reproduce the issue**
Provide a reproducible test case that is the bare minimum necessary to generate
the problem. If possible, please share a link to Colab/CodePen/any notebook.

**Other info / logs** Include any logs or source code that would be helpful to
diagnose the problem. If including tracebacks, please include the full
traceback. Large logs and files should be attached.

Issue #6987, opened 2022-10-29
Unclear error message with fill invalid shape using WebGL


yolisses:
**System information**
- Linux Ubuntu 22.04
- laptop Acer Aspire A515-52G
- TensorFlow.js version 4.0.0 
- installed using yarn
- Google Chrome Version 106.0.5249.119 (Official Build) (64-bit)

**Describe the current behavior**
Using `fill` with a invalid not integer shape throws a unclear WebGl error.

**Describe the expected behavior**
Preferably, `fill` should raise a clear error, like `Invalid shape received to fill: [2.32, 2]`

**Standalone code to reproduce the issue**
`tf.fill([2.32, 2], 3).print();` throws
```
1    #version 300 es
2        precision highp float;
3        precision highp int;
4        precision highp sampler2D;
5        in vec2 resultUV;
6        out vec4 outputColor;
7        const vec2 halfCR = vec2(0.5, 0.5);
8    
9        struct ivec5
10       {
11         int x;
12         int y;
13         int z;
14         int w;
15         int u;
16       };
17   
18       struct ivec6
19       {
20         int x;
21         int y;
22         int z;
23         int w;
24         int u;
25         int v;
26       };
27   
28       uniform float NAN;
29       
30       
31       
32         #define round(value) newRound(value)
33         int newRound(float value) {
34           return int(floor(value + 0.5));
35         }
36   
37         ivec4 newRound(vec4 value) {
38           return ivec4(floor(value + vec4(0.5)));
39         }
40       
41   
42       int imod(int x, int y) {
43         return x - y * (x / y);
44       }
45   
46       int idiv(int a, int b, float sign) {
47         int res = a / b;
48         int mod = imod(a, b);
49         if (sign < 0. && mod != 0) {
50           res -= 1;
51         }
52         return res;
53       }
54   
55       //Based on the work of Dave Hoskins
56       //https://www.shadertoy.com/view/4djSRW
57       #define HASHSCALE1 443.8975
58       float random(float seed){
59         vec2 p = resultUV * seed;
60         vec3 p3  = fract(vec3(p.xyx) * HASHSCALE1);
61         p3 += dot(p3, p3.yzx + 19.19);
62         return fract((p3.x + p3.y) * p3.z);
63       }
64   
65       
66   vec2 uvFromFlat(int texNumR, int texNumC, int index) {
67     int texR = index / texNumC;
68     int texC = index - texR * texNumC;
[...]

PR #7014, opened 2022-11-04
[ops ] guard shape for decimal and negative values

pyu10055:
fixed https://github.com/tensorflow/tfjs/issues/7008 and https://github.com/tensorflow/tfjs/issues/6987

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7014)

### DIFF
--- a/tfjs-core/src/ops/broadcast_to.ts
+++ b/tfjs-core/src/ops/broadcast_to.ts
@@ -22,6 +22,7 @@ import {Tensor} from '../tensor';
 import {NamedTensorMap} from '../tensor_types';
 import {convertToTensor} from '../tensor_util_env';
 import {Rank, ShapeMap, TensorLike} from '../types';
+import {assertNonNegativeIntegerDimensions} from '../util_base';
 
 import {clone} from './clone';
 import {op} from './operation';
@@ -49,6 +50,8 @@ function broadcastTo_<R extends Rank>(
   if (shape.some(d => !(d > 0) || d % 1 !== 0)) {
     throw new Error(`broadcastTo(): Invalid broadcast shape [${shape}].`);
   }
+
+  assertNonNegativeIntegerDimensions(shape);
 
   if (shape.length < input.rank) {
     throw new Error(`broadcastTo(): shape.length=${shape.length} < input.rank=${
--- a/tfjs-core/src/ops/broadcast_to.ts
+++ b/tfjs-core/src/ops/broadcast_to.ts
@@ -49,7 +49,6 @@ function broadcastTo_<R extends Rank>(
 
   assertNonNegativeIntegerDimensions(shape);
 
-
   if (shape.length < input.rank) {
     throw new Error(`broadcastTo(): shape.length=${shape.length} < input.rank=${
         input.rank}.`);
--- a/tfjs-core/src/ops/broadcast_to.ts
+++ b/tfjs-core/src/ops/broadcast_to.ts
@@ -47,11 +47,8 @@ function broadcastTo_<R extends Rank>(
   let input = convertToTensor(x, 'broadcastTo', 'x');
   const xShape = input.shape;
 
-  if (shape.some(d => !(d > 0) || d % 1 !== 0)) {
-    throw new Error(`broadcastTo(): Invalid broadcast shape [${shape}].`);
-  }
-
   assertNonNegativeIntegerDimensions(shape);
+
 
   if (shape.length < input.rank) {
     throw new Error(`broadcastTo(): shape.length=${shape.length} < input.rank=${
--- a/tfjs-core/src/ops/broadcast_to_test.ts
+++ b/tfjs-core/src/ops/broadcast_to_test.ts
@@ -73,4 +73,9 @@ describeWithFlags('broadcastTo', ALL_ENVS, () => {
 
     expectArraysClose(await df(a).array(), await dh(a).array());
   });
+
+  it('should throw error when shape is not integer', () => {
+    const a = tf.scalar(4.2);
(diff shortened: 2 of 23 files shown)


Linchenn (approved):
LGTM, thank you!

tfjs-core/src/ops/broadcast_to.ts
```diff
@@ -50,6 +51,8 @@ function broadcastTo_<R extends Rank>(
     throw new Error(`broadcastTo(): Invalid broadcast shape [${shape}].`);
   }
 
+  assertNonNegativeIntegerDimensions(shape);
```
mattsoulanille, line 54: We can avoid having to check this for values we get from `convertToTensor` if we add this check to every `convertToTensor` call. [We already check dtype](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/tensor_util_env.ts#L97-L100), so checking shape as well seems fine to me.

mattsoulanille: Unfortunately, I don't think there's a similar fix for ops that make new tensors unless we make changes to the [`op` function](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/operation.ts#L27-L66), and even then, it's probably not really possible since you don't know the order of the args.

mattsoulanille (approved):
LGTM with a suggestion.

pyu10055 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7014)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @mattsoulanille)

___
*[`tfjs-core/src/ops/broadcast_to.ts` line 54 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7014#-NG3B_wGN3Ryy3BCviJq-r1-54:-NG3Xi8s2SecZKFXIX2D:bz9pex3) ([raw file](https://github.com/tensorflow/tfjs/blob/ac2db1c041b2df8907eeb2c5776019b84cc44bd6/tfjs-core/src/ops/broadcast_to.ts#L54)):*

Unfortunately, I don't think there's a similar fix for ops that make new tensors unless we make changes to the [`op` function](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/operation.ts#L27-L66), and even then, it's probably not really possible since you don't know the order of the args.
</blockquote></details>

are you suggesting to add shape check to `convertToTensor` method? That method does not have shape value issue, since it is either from a existing tensor or from an array of array, the shape is derived.

pyu10055 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7014)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @mattsoulanille)

___
*[`tfjs-core/src/ops/broadcast_to.ts` line 54 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7014#-NG3B_wGN3Ryy3BCviJq-r1-54:-NG3_V3bEKwqWPrWD-Jy:bxjl89b) ([raw file](https://github.com/tensorflow/tfjs/blob/ac2db1c041b2df8907eeb2c5776019b84cc44bd6/tfjs-core/src/ops/broadcast_to.ts#L54)):*

are you suggesting to add shape check to `convertToTensor` method? That method does not have shape value issue, since it is either from a existing tensor or from an array of array, the shape is derived. 
</blockquote></details>

but looks like the previous line is doing the shape check for negative and integer. I will consolidate.

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7014)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @pyu10055)

___
*[`tfjs-core/src/ops/broadcast_to.ts` line 54 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7014#-NG3B_wGN3Ryy3BCviJq-r1-54:-NG3eLKDBI6jvAavWkPm:b88m6j6) ([raw file](https://github.com/tensorflow/tfjs/blob/ac2db1c041b2df8907eeb2c5776019b84cc44bd6/tfjs-core/src/ops/broadcast_to.ts#L54)):*

but looks like the previous line is doing the shape check for negative and integer. I will consolidate.
</blockquote></details>

Oh. oops. That was the wrong line.

Here's where we check for dtype in convertToTensor when the input is a TensorLike and not a Tensor. I think we can add a shape check there as well

https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/tensor_util_env.ts#L108

mattsoulanille (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7014)* status: :shipit: complete! 2 of 1 approvals obtained

___
*[`tfjs-core/src/ops/broadcast_to.ts` line 54 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7014#-NG3B_wGN3Ryy3BCviJq-r1-54:-NG3gE1s4W24hOve9I90:b-85srne) ([raw file](https://github.com/tensorflow/tfjs/blob/ac2db1c041b2df8907eeb2c5776019b84cc44bd6/tfjs-core/src/ops/broadcast_to.ts#L54)):*

Oh. oops. That was the wrong line.

Here's where we check for dtype in convertToTensor when the input is a TensorLike and not a Tensor. I think we can add a shape check there as well

https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/tensor_util_env.ts#L108
</blockquote></details>

Wait, actually, I misread the type of `convertToTensor` and thought it was using a TensorInfo. TensorLike should not need its shape checked for decimals or negatives.

PR #7025, opened 2022-11-09
webgpu: refactor atomicAdd code

xhcao:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7025)

### DIFF
--- a/tfjs-backend-webgpu/src/bincount_webgpu.ts
+++ b/tfjs-backend-webgpu/src/bincount_webgpu.ts
@@ -20,8 +20,8 @@ import {getMainHeaderString as main, WebGPUProgram} from './webgpu_program';
 import {computeDispatch, flatDispatchLayout} from './webgpu_util';
 
 const writeSnippet = `
-  fn bincount_write(flatIndex: i32, value: f32) {
-    ${atomicAddSnippet(1)}
+  fn bincount_write(index: i32, value: f32) {
+    ${atomicAddSnippet('result', 'index', 'value', 1)}
   }
 `;
 
--- a/tfjs-backend-webgpu/src/bincount_webgpu.ts
+++ b/tfjs-backend-webgpu/src/bincount_webgpu.ts
@@ -15,21 +15,13 @@
  * =============================================================================
  */
 
+import {atomicAddSnippet} from './shader_util';
 import {getMainHeaderString as main, WebGPUProgram} from './webgpu_program';
 import {computeDispatch, flatDispatchLayout} from './webgpu_util';
 
 const writeSnippet = `
-  fn bincount_write(index: i32, value: f32) {
-    var oldValue = atomicLoad(& (result[index]));
-    var exchanged = false;
-    for (; !exchanged;) {
-      let newValueF32 = bitcast<f32>(oldValue) + value;
-      let newValue = bitcast<i32>(newValueF32);
-      let res = atomicCompareExchangeWeak(
-          &(result[index]), oldValue, newValue);
-      oldValue = res.old_value;
-      exchanged = res.exchanged;
-    }
+  fn bincount_write(flatIndex: i32, value: f32) {
+    ${atomicAddSnippet(1)}
   }
 `;
 
--- a/tfjs-backend-webgpu/src/bincount_webgpu.ts
+++ b/tfjs-backend-webgpu/src/bincount_webgpu.ts
@@ -21,13 +21,13 @@ import {computeDispatch, flatDispatchLayout} from './webgpu_util';
 
 const writeSnippet = `
   fn bincount_write(index: i32, value: f32) {
-    ${atomicAddSnippet('result', 'index', 'value', 1)}
+    ${atomicAddSnippet('&result[index]', 'value', 'float32')}
   }
 `;
 
 const binaryWriteSnippet = `
   fn bincount_write(index: i32, value: f32) {
-    result[index] = value;
+    atomicStore(&result[index], bitcast<i32>(value));
   }
 `;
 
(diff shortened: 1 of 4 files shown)


xhcao:
@hujiajie I had refactored the atomicAdd code, PTAL, thanks.

tfjs-backend-webgpu/src/shader_util.ts
```diff
@@ -32,3 +32,23 @@ export function symbolicallyComputeStrides(
...
+        let newValueF32 = bitcast<f32>(oldValue) + ${
+      component > 1 ? 'value[i]' : 'value'};
+        let newValue = bitcast<i32>(newValueF32);
+        let res = atomicCompareExchangeWeak(&(result[flatIndex + i]), oldValue, newValue);
+        oldValue = res.old_value;
+        exchanged = res.exchanged;
+      }
+    }
+  `;
+};
```
hujiajie, line 54: For symmetry, shall we add other atomic utils for `atomicSub`/`atomicStore`/...?

xhcao:
PTAL again, thanks.

tfjs-backend-webgpu/src/shader_util.ts
```diff
@@ -32,3 +32,24 @@ export function symbolicallyComputeStrides(
 
   return strides;
 }
+
+// atomicAdd only supports uint/int type. For float, we use
+// atomicCompareExchangeWeak to simulate.
+export const atomicAddSnippet =
+    (dst: string, index: string, val: string, component: number) => {
+      return `
+    for (var i = 0; i < ${component}; i = i + 1) {
```
hujiajie, line 41: Is this really necessary to have a vectorized variant here? Can we achieve the same with `for (var i = 0; i < 4; i++) { atomicAdd(...); }` in `main()`?

tfjs-backend-webgpu/src/shader_util.ts
```diff
@@ -32,3 +32,24 @@ export function symbolicallyComputeStrides(
...
+      var oldValue = atomicLoad(&(${dst}[${index} + i]));
+      var exchanged = false;
+      for (; !exchanged;) {
```
hujiajie, line 44: I think the initialization of `exchanged` is a waste of cycles. Can we rewrite the loop to

```rust
loop {
  if exchanged {
    break;
  }
}
```

xhcao:
@hujiajie PTAL, thanks

tfjs-backend-webgpu/src/shader_util.ts
```diff
@@ -32,3 +32,23 @@ export function symbolicallyComputeStrides(
...
+// atomicAdd only supports uint/int type. For float, we use
+// atomicCompareExchangeWeak to simulate.
+export const atomicAddSnippet = (dst: string, val: string) => {
+  return `
+      var oldValue = atomicLoad(&(${dst}));
+      var exchanged = false;
+      loop {
+        if exchanged {
+          break;
+        }
```
hujiajie, line 45: Thanks for iterating on this, but I think the point of my previous comment was lost. The if statement should be moved to the end, otherwise we are still initializing `exchanged` here and making an unnecessary comparison before hitting `atomicCompareExchangeWeak()` for the first time.

xhcao: Done

tfjs-backend-webgpu/src/bincount_webgpu.ts
```diff
@@ -15,21 +15,13 @@
  * =============================================================================
  */
 
+import {atomicAddSnippet} from './shader_util';
 import {getMainHeaderString as main, WebGPUProgram} from './webgpu_program';
 import {computeDispatch, flatDispatchLayout} from './webgpu_util';
 
 const writeSnippet = `
   fn bincount_write(index: i32, value: f32) {
```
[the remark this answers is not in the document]
hujiajie: Multiple threads may write to the same memory location in L30, right? I'm worried that's undefined behavior from the spec perspective, but I'm not familiar with the WebGPU memory model. I suspect atomic store is required there, even if intuitively non-atomic store seems correct.

tfjs-backend-webgpu/src/shader_util.ts
```diff
@@ -32,3 +32,20 @@ export function symbolicallyComputeStrides(
 
   return strides;
 }
+
+// atomicAdd only supports uint/int type. For float, we use
+// atomicCompareExchangeWeak to simulate.
+export const atomicAddSnippet = (dst: string, val: string) => {
```
gyagp, line 38: Can we also pass in type so that i32 and f32 can be both covered?
It's better to follow wgsl spec, and name the parameters as ptr and v, then we need to pass in address. 
With above, we can remove atomicRMW.

xhcao:
Hope already address all comments, please take a look again, thank you.

tfjs-backend-webgpu/src/scatter_webgpu.ts
```diff
@@ -95,33 +96,17 @@ export class ScatterProgram implements WebGPUProgram {
...
-              newBits = bitcast<i32>(newValue);
-            }
-          }
-        `;
-      }
-      const atomicStoreSnippet = `atomicStore(${ptr}, bitcast<i32>(${val}));`;
-      return this.sumDupeIndices ? atomicAddSnippet : atomicStoreSnippet;
-    };
+    let $atomicAddSnippet;
+    if (this.type === 'float32') {
```
gyagp, line 100: atomicAddSnippet() already supports the type, so we can pass in this.type directly.

xhcao: this.type is one of `DataType`, which includes float32, int32, bool, complex64 and string. I think it cannot be passed directly.

gyagp (approved):
LGTM

tfjs-backend-webgpu/src/bincount_webgpu.ts
```diff
@@ -96,7 +88,8 @@ export class BincountProgram implements WebGPUProgram {
         let value = ${
                 this.binaryOutput ?
                     1. :
-                    (this.hasWeights ? 'f32(getW(coord[0], coord[1]))' : '1.')};
+                    (this.hasWeights ? 'f32(getW(coord[0], coord[1]))' :
```
qjia7, line 91: `f32(getW)` -> `getW`?

qjia7 (approved):
LGTM with one nit.

PR #7027, opened 2022-11-09
Add TFDF setLocateFile function

ahmedsabie:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7027)


tfjs-tfdf/src/tfdf_web_api_client.d.ts
```diff
@@ -19,4 +19,8 @@ import {TFDFWebAPIClient} from './types/tfdf_web_api';
 /**
  * The main export for tfdf_web_api_client.js types.
  */
-export declare let tfdfWeb: Promise<TFDFWebAPIClient>;
+export declare let tfdfWeb: () => Promise<TFDFWebAPIClient>;
+
+type LocateFileFunction = (path: string, prefix?: string) => string;
+/** The global function to set WASM path. */
+export declare let setLocateFile: (locateFile: LocateFileFunction) => void;
```
mattsoulanille, line 26: Tiny nit: I think this can be const, although I'm not actually that knowledgeable about .d.ts files.
```suggestion
export declare const setLocateFile: (locateFile: LocateFileFunction) => void;
```

tfjs-tfdf/src/tfdf_web_api_client.js
```diff
@@ -22,4 +22,17 @@
 // imported correctly by typescript code.
 const YggdrasilDecisionForests = require('../wasm/inference.js');
 
-exports.tfdfWeb = YggdrasilDecisionForests();
+let tfdfWeb = null;
+
+// Function is used here instead of variable to make loading lazy in case user
+// uses setLocateFile before module is used.
+exports.tfdfWeb = () => {
```
mattsoulanille, line 29: Nit: If the user has to call this regardless of whether they set the wasm path, could we add an optional argument that just sets the wasm path for them? Might make it a bit easier to call.
```js
exports.tfdfWeb = (locateFile?) => {
  if (locateFile !== null) {
    exports.setLocateFile(locateFile);
  }
  ...
}
```

This probably also requires changing some types and reordering declarations in this file, but I think it's a more convenient interface.

tfjs-tfdf/src/tfdf_web_api_client.js
```diff
@@ -22,4 +22,17 @@
...
+  if (tfdfWeb == null) {
+    tfdfWeb = YggdrasilDecisionForests();
+  }
+  return tfdfWeb;
+};
+
+exports.setLocateFile = (locateFile) => {
+  tfdfWeb = YggdrasilDecisionForests({locateFile});
+};
```
mattsoulanille, line 38: Should we throw an error if the user calls this twice?

mattsoulanille (approved):
LGTM with some nits, mostly on the interface that the user interacts with.

tfjs-tfdf/src/tfdf_web_api_client.js
```diff
@@ -22,4 +22,17 @@
...
 
-exports.tfdfWeb = YggdrasilDecisionForests();
+let tfdfWeb = null;
+
+// Function is used here instead of variable to make loading lazy in case user
+// uses setLocateFile before module is used.
+exports.tfdfWeb = () => {
+  if (tfdfWeb == null) {
+    tfdfWeb = YggdrasilDecisionForests();
+  }
```
mattsoulanille, line 32: Does TFDF have state? If so, could a user ever want multiple instances of it?

tfjs-tfdf/src/tfdf_web_api_client.js
ahmedsabie: The user can't create tfdfWeb, as for calling setLocateFile twice it would be similar behavior to tflite's setWasmPath which reassigns until the path is used. I do agree with storing the function though so that later accidental sets don't recreate needlessly, once library is loaded without error we are done.

tfjs-tfdf/src/tfdf_web_api_client.js
ahmedsabie: No

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7027)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @ahmedsabie and @mattsoulanille)

___
*[`tfjs-tfdf/README.md` line 59 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7027#-NGrJYjTAJQzMEs5wv8o:-NGrJYjTAJQzMEs5wv8p:b-8cqagh) ([raw file](https://github.com/tensorflow/tfjs/blob/f781f3d552eb8e57ca96ba7edb936aa907b92f1b/tfjs-tfdf/README.md#L59)):*
> ```Markdown
> 
> ```js
> tfdf.setLocateFile((path, base) => {
> ```

is the second parameter `base` ever used? If not might be better to remove or add comments.

PR #7033, opened 2022-11-10
webgpu: support fft, ifft, rfft and irfft operators

xhcao:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7033)

### DIFF
--- a/tfjs-backend-webgpu/src/fft_webgpu.ts
+++ b/tfjs-backend-webgpu/src/fft_webgpu.ts
@@ -1,0 +1,82 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


qjia7 (approved):
LGTM, thanks!

tfjs-backend-webgpu/src/fft_webgpu.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      var result = 0.0;
+
+      for (var i = 0; i < uniforms.realShape[1]; i = i + 1) {
+        // x = (-2|2 * PI / N) * index * i;
+        let x = exponentMultiplierTimesIndexRatio * f32(i);
+        let expR = cos(x);
+        let expI = sin(x);
+        let real = getReal(batch, i);
+        let imag = getImag(batch, i);
```
hujiajie, line 64: Is this a direct mapping from the definition of discrete Fourier transform:
$$X_k = \sum_{n=0}^{N-1} x_n e^{-\frac{2kn\pi{}i}{N}}$$

It has $\mathcal{O}(N^2)$ complexity, so the first F in FFT ( $\mathcal{O}(N\log{}N)$ ) is lost. FFT should be super suitable for parallelism using the butterfly diagram, and could be a good example where WebGPU beats WebGL, though I have never remembered the details of the algorithm :sweat:

xhcao: Yes, It directly uses DFT to implement FFT here. FFT implementation is more difficult, I plan to implementation it in the future as an optimization task.

gyagp (approved):
LGTM

PR #6853, opened 2022-09-24
[WebGL] Texture to tensor API

Linchenn:
This fixes https://github.com/tensorflow/tfjs/issues/3937

#### Changes
- Add a new method to the API: `tf.tensor(values: WebGLData, shape, dtype)` that allows users to create tensors from textures and `WebGLData` is an object of `WebGLTexture`, `height`, `width` and `channels` properties and the `channels` could be a non-empty subset of 'RGBA' in any order.

Reference:
https://github.com/tensorflow/tfjs/pull/6848 (The current PR starts from this PR of Na.)
https://github.com/tensorflow/tfjs/pull/4376

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/6853)

PR #7034, opened 2022-11-10
[webgpu] Create tensor from GPUBuffer

axinging:
BUG: https://github.com/tensorflow/tfjs/issues/6232

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7034)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -463,6 +463,10 @@ export class WebGPUBackend extends KernelBackend {
     if (values.buffer.size < size) {
       throw new Error(`GPUBuffer size(${
           values.buffer.size}) is smaller than tensor size(${size})!`);
+    } else if (
+        (values.buffer.usage & GPUBufferUsage.STORAGE) !==
+        GPUBufferUsage.STORAGE) {
+      throw new Error('GPUBuffer.usage should include GPUBufferUsage.STORAGE!');
     }
 
     tensorData.resourceInfo = {size: buffer.size, usage: buffer.usage, buffer};
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -461,9 +461,9 @@ export class WebGPUBackend extends KernelBackend {
     const sizeFromShape = util.sizeFromShape(tensorData.shape);
     const size =
         webgpu_util.GPUBytesPerElement(tensorData.dtype) * sizeFromShape;
-    if (values.size < sizeFromShape) {
+    if (values.buffer.size < sizeFromShape) {
       throw new Error(`GPUBuffer size(${
-          values.size}) is smaller than tensor size(${sizeFromShape})!`);
+          values.buffer.size}) is smaller than tensor size(${sizeFromShape})!`);
     }
 
     tensorData
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -458,16 +458,14 @@ export class WebGPUBackend extends KernelBackend {
     this.tensorMap.set(
         dataId, {dtype, shape, values: null, refCount: 1, external: true});
     const tensorData = this.tensorMap.get(dataId);
-    const sizeFromShape = util.sizeFromShape(tensorData.shape);
-    const size =
-        webgpu_util.GPUBytesPerElement(tensorData.dtype) * sizeFromShape;
-    if (values.buffer.size < sizeFromShape) {
+    const size = webgpu_util.GPUBytesPerElement(tensorData.dtype) *
+        util.sizeFromShape(tensorData.shape);
+    if (values.buffer.size < size) {
       throw new Error(`GPUBuffer size(${
-          values.buffer.size}) is smaller than tensor size(${sizeFromShape})!`);
+          values.buffer.size}) is smaller than tensor size(${size})!`);
     }
 
-    tensorData
-        .resourceInfo = {size, usage: this.defaultGpuBufferUsage(), buffer};
+    tensorData.resourceInfo = {size: buffer.size, usage: buffer.usage, buffer};
     return engine().makeTensorFromDataId(dataId, shape, dtype, this);
   }
 
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -51,8 +51,12 @@ type TensorData = {
   shape: number[],
   refCount: number,
   resourceInfo?: BufferInfo|TextureInfo,
-  // Indicate the tensor is created from an external GPU resource.
-  external?: boolean,
+  // zeroCopy is used for creating tensor from GPUBuffer. When zeroCopy is false
+  // or undefined (default), this GPUBuffer will be copied to the tensor's
+  // resource buffer. When zeroCopy is true, tensor will use this GPUBUffer as
+  // tensor's resource buffer, user should not destroy this GPUBuffer until all
+  // access are done.
+  zeroCopy?: boolean,
   // For complex numbers, the real and imaginary parts are stored as their own
   // individual tensors, with a parent joining the two with the
   // complexTensorInfos field.
@@ -244,8 +248,9 @@ export class WebGPUBackend extends KernelBackend {
     if (!tensorData || !tensorData.resourceInfo) {
       return;
     }
-    // If tensor data is from external resource, do not release.
-    if (tensorData.external) {
+    // If tensor's resource buffer is from a zero copy GPUBuffer, do not
+    // release.
+    if (tensorData.zeroCopy) {
       tensorData.resourceInfo = null;
       return;
     }
@@ -445,18 +450,31 @@ export class WebGPUBackend extends KernelBackend {
     return vals;
   }
 
+  // The source GPUBuffer and destination GPUBuffer have the same size and
+  // usage.
+  private copyBuffer(srcBuffer: GPUBuffer, size: number, usage: number) {
+    const dstBuffer = this.bufferManager.acquireBuffer(size, usage);
+    this.ensureCommandEncoderReady();
+    this.ensureComputePassEnded();
+    this.currentCommandEncoder.copyBufferToBuffer(
+        srcBuffer, 0, dstBuffer, 0, size);
+    this.submitQueue();
+    return dstBuffer;
+  }
+
   /**
    * Create a TF.js tensor out of an existing WebGPU buffer.
    */
   override createTensorFromGPUData(
       values: WebGPUData, shape: number[], dtype: DataType): Tensor {
-    const buffer = values.buffer;
+    let buffer = values.buffer;
     if (dtype === 'complex64') {
       throw new Error(`Cannot write to a complex64 dtype. `);
     }
     const dataId = {id: this.nextDataId()};
+    const zeroCopy = env().getBool('WEBGPU_TENSOR_FROM_BUFFER_WITH_ZERO_COPY');
     this.tensorMap.set(
-        dataId, {dtype, shape, values: null, refCount: 1, external: true});
+        dataId, {dtype, shape, values: null, refCount: 1, zeroCopy});
     const tensorData = this.tensorMap.get(dataId);
     const size = webgpu_util.GPUBytesPerElement(tensorData.dtype) *
         util.sizeFromShape(tensorData.shape);
@@ -469,6 +487,10 @@ export class WebGPUBackend extends KernelBackend {
       throw new Error('GPUBuffer.usage should include GPUBufferUsage.STORAGE!');
     }
 
+    // Do buffer copy by default.
+    if (zeroCopy === false) {
+      buffer = this.copyBuffer(buffer, size, buffer.usage);
+    }
     tensorData.resourceInfo = {size: buffer.size, usage: buffer.usage, buffer};
     return engine().makeTensorFromDataId(dataId, shape, dtype, this);
   }
@@ -659,9 +681,8 @@ export class WebGPUBackend extends KernelBackend {
       // TODO: WebGPU doesn't support read data synchronously from GPU to CPU.
       // So it will report error when switching backend from WebGPU to others.
       // There are two situations: 1) swithcing the backend after running a
-      // model; 2) swithcing the backend within the model. Temporarilly keep the
-      // values on CPU to solve the first issue.
-      // tensorData.values = null;
+      // model; 2) swithcing the backend within the model. Temporarilly keep
+      // the values on CPU to solve the first issue. tensorData.values = null;
     }
   }
 
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -17,7 +17,7 @@
 
 import './flags_webgpu';
 
-import {backend_util, buffer, DataStorage, DataType, engine, env, GPUData, KernelBackend, Rank, RecursiveArray, ShapeMap, TensorBuffer, TensorInfo, TimingInfo, TypedArray, util} from '@tensorflow/tfjs-core';
+import {backend_util, buffer, DataStorage, DataType, engine, env, GPUData, KernelBackend, Rank, RecursiveArray, ShapeMap, Tensor, TensorBuffer, TensorInfo, TimingInfo, TypedArray, util, WebGPUData} from '@tensorflow/tfjs-core';
 
 import {AdapterInfo} from './adapter_info';
 import {BufferManager} from './buffer_manager';
@@ -51,6 +51,8 @@ type TensorData = {
   shape: number[],
   refCount: number,
   resourceInfo?: BufferInfo|TextureInfo,
+  // Indicate the tensor is created from an external GPU resource.
+  external?: boolean,
   // For complex numbers, the real and imaginary parts are stored as their own
   // individual tensors, with a parent joining the two with the
   // complexTensorInfos field.
@@ -242,6 +244,11 @@ export class WebGPUBackend extends KernelBackend {
     if (!tensorData || !tensorData.resourceInfo) {
       return;
     }
+    // If tensor data is from external resource, do not release.
+    if (tensorData.external) {
+      tensorData.resourceInfo = null;
+      return;
+    }
     if ('texture' in tensorData.resourceInfo) {
       const textureInfo = tensorData.resourceInfo;
       if (textureInfo.texture instanceof GPUTexture) {
@@ -282,7 +289,8 @@ export class WebGPUBackend extends KernelBackend {
     }
   }
 
-  override write(values: backend_util.BackendValues, shape: number[],
+  override write(
+      values: backend_util.BackendValues, shape: number[],
       dtype: DataType): DataId {
     if (dtype === 'complex64' && values != null) {
       throw new Error(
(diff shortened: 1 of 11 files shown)


axinging:
@qjia7 @xhcao @haoyunfeix @gyagp , PTAL

axinging:
@qjia7 @xhcao @haoyunfeix @gyagp  PTAL

tfjs-core/src/ops/tensor.ts
```diff
@@ -92,6 +92,67 @@ import {makeTensor} from './tensor_ops_util';
  *
  * const tex = a.dataToGPU();
  * ```
+ *
+ * ```js
+ * // Pass a `WebGPUData` object and specify a shape yourself.
```
qjia7, line 97: Please also emphasize that we will directly bind this buffer to the new created tensor to support zero copy. So please DONOT destroy this buffer until all calculations are finished in TFJS.

tfjs-backend-webgpu/src/backend_webgpu_test.ts
```diff
@@ -366,3 +366,105 @@ describeWebGPU('keeping data on gpu ', () => {
...
+    tf.test_util.expectArraysClose(await result.data(), expected);
+    a.dispose();
+    b.dispose();
+    result.dispose();
+    const endNumBytes = tf.memory().numBytes;
+    const endNumTensors = tf.memory().numTensors;
+    expect(endNumBytes - startNumBytes).toEqual(0);
+    expect(endNumTensors - startNumTensors).toEqual(0);
+    aBuffer.destroy();
+  });
```
qjia7, line 469: Please also add tests for buffer.size > shape and buffer.size < shape.

Linchenn (changes requested):
Great work and thank you Xu! Overall LGTM except a few nits and a question: 

If I understand correctly, this implementation will create a new tensor using the GPUBuffer passed from users. However, WebGL's implementation always creates a new texture and copies values from the texture passed from users. I am curious, is this intended? should we keep the same approach for the two backends, even though this implementation is good for performance (also directly resolves https://github.com/tensorflow/tfjs/issues/6232)? @qjia7 @axinging 

> Reviewed 4 of 10 files at r1, 6 of 7 files at r4.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7034#-:-NGuAkhb5LYovo4RvNF_:b-pz1xaa)* status: 0 of 1 approvals obtained (waiting on @axinging and @xhcao)

___
*[`tfjs-core/src/tensor_util_env.ts` line 33 at r4](https://reviewable.io/reviews/tensorflow/tfjs/7034#-NGu7H5s67ERY8p17PCW:-NGu7H5s67ERY8p17PCX:bqxagly) ([raw file](https://github.com/tensorflow/tfjs/blob/153471be7e48d7f63c5598e9734342846db883c5/tfjs-core/src/tensor_util_env.ts#L33)):*
> ```XML
>   const isObject = typeof val === 'object';
>   if (isObject) {
>     if ('texture' in val && val.texture instanceof WebGLTexture) {
> ```

Great catch! Thanks!



_Code quote:_
```XML
val.texture instanceof WebGLTexture
```
___
*[`tfjs-core/src/ops/tensor.ts` line 105 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7034#-NGu4pnZ0-6or37maAqZ:-NGu4pnZ0-6or37maAq_:bohz2v9) ([raw file](https://github.com/tensorflow/tfjs/blob/7d560d7b897c1388458f493855c736c382bd0f78/tfjs-core/src/ops/tensor.ts#L105)):*
> ```XML
>  *
>  * // Example for WebGPU:
>  * async function createReadonlyGPUBufferFromData(device, data, dtype) {
> ```

Could this be a non-async function?
___
*[`tfjs-core/src/ops/tensor.ts` line 157 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7034#-NGu5Rkk67BP1bhD-hcR:-NGu5Rkk67BP1bhD-hcS:b-pwlxq0) ([raw file](https://github.com/tensorflow/tfjs/blob/7d560d7b897c1388458f493855c736c382bd0f78/tfjs-core/src/ops/tensor.ts#L157)):*
> ```XML
>  * ```
>  * @param values The values of the tensor. Can be nested array of numbers,
>  *     or a flat array, or a `TypedArray`, or a `WebGLData` object. If the
> ```

Could you also add some description about `WebGPUData` in `value` parameter's description?

axinging:
> Great work and thank you Xu! Overall LGTM except a few nits and a question:
> 
> If I understand correctly, this implementation will create a new tensor using the GPUBuffer passed from users. However, WebGL's implementation always creates a new texture and copies values from the texture passed from users. I am curious, is this intended? should we keep the same approach for the two backends, even though this implementation is good for performance (also directly resolves #6232)? @qjia7 @axinging
> 
> > Reviewed 4 of 10 files at r1, 6 of 7 files at r4.
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7034#-:-NGuAkhb5LYovo4RvNF_:b-pz1xaa)_ status: 0 of 1 approvals obtained (waiting on @axinging and @xhcao)
> 
> _[`tfjs-core/src/tensor_util_env.ts` line 33 at r4](https://reviewable.io/reviews/tensorflow/tfjs/7034#-NGu7H5s67ERY8p17PCW:-NGu7H5s67ERY8p17PCX:bqxagly) ([raw file](https://github.com/tensorflow/tfjs/blob/153471be7e48d7f63c5598e9734342846db883c5/tfjs-core/src/tensor_util_env.ts#L33)):_
> 
> > ```
> >   const isObject = typeof val === 'object';
> >   if (isObject) {
> >     if ('texture' in val && val.texture instanceof WebGLTexture) {
> > ```
> 
> Great catch! Thanks!
> 
> _Code quote:_
> 
> ```
> val.texture instanceof WebGLTexture
> ```
> 
> _[`tfjs-core/src/ops/tensor.ts` line 105 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7034#-NGu4pnZ0-6or37maAqZ:-NGu4pnZ0-6or37maAq_:bohz2v9) ([raw file](https://github.com/tensorflow/tfjs/blob/7d560d7b897c1388458f493855c736c382bd0f78/tfjs-core/src/ops/tensor.ts#L105)):_
> 
> > ```
> >  *
> >  * // Example for WebGPU:
> >  * async function createReadonlyGPUBufferFromData(device, data, dtype) {
> > ```
> 
> Could this be a non-async function?
> 
> _[`tfjs-core/src/ops/tensor.ts` line 157 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7034#-NGu5Rkk67BP1bhD-hcR:-NGu5Rkk67BP1bhD-hcS:b-pwlxq0) ([raw file](https://github.com/tensorflow/tfjs/blob/7d560d7b897c1388458f493855c736c382bd0f78/tfjs-core/src/ops/tensor.ts#L157)):_
> 
> > ```
> >  * ```
> >  * @param values The values of the tensor. Can be nested array of numbers,
> >  *     or a flat array, or a `TypedArray`, or a `WebGLData` object. If the
> > ```
> 
> Could you also add some description about `WebGPUData` in `value` parameter's description?

𝑇hanks @Linchenn 
All yopur comments are updated, except the WebGL has copy and WebGPU not. BTW, I prefer the not copy one.

qjia7:
>If I understand correctly, this implementation will create a new tensor using the GPUBuffer passed from users. However, WebGL's implementation always creates a new texture and copies values from the texture passed from users. I am curious, is this intended?

Yes, it's intended for performance. But like you said, it will result different behavior between webgl and webgpu. For webgl, once the user creates the tensor from the external texture. The user can directly destroy that texture. However, for webgpu, the user can't destroy the external buffer until tfjs finished the calculation. To keep the user have consistent usage experience, can we tell user that once users use this API to create tensor from GPU data no matter it's webgl or webgpu, to support zero copy for performance, we require that they can't destroy this external gpu resource until tfjs finished all the calculations? For webgl, it's not true, but it leaves room for optimization for webgl in future. How do you think? @Linchenn @pyu10055 @lina128

pyu10055 (commented):
Thank @qjia7 for the clarification. I am look at the API itself, the parameter `values` typically refer passing by value mechanism, which means the values is copied during the call comparing to passing by reference.
I think we might want to have two APIs to differentiate the mechanisms ( 1 for values and 1 for reference).

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7034#-:-NGx8qqT5b_SYq5yvnxb:bfge9qo)* status: 0 of 1 approvals obtained (waiting on @axinging, @qjia7, and @xhcao)

axinging:
> Thank @qjia7 for the clarification. I am look at the API itself, the parameter `values` typically refer passing by value mechanism, which means the values is copied during the call comparing to passing by reference. I think we might want to have two APIs to differentiate the mechanisms ( 1 for values and 1 for reference).
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7034#-:-NGx8qqT5b_SYq5yvnxb:bfge9qo)_ status: 0 of 1 approvals obtained (waiting on @axinging, @qjia7, and @xhcao)

Thanks @pyu10055.  I am trying to understand if Javascript is "passing by value" or "passing by reference" here.
In the API side, tensor accepts TensorLike:
```
type TensorLike = string | number | boolean | TypedArray | RecursiveArray<number | TypedArray | number[]> | RecursiveArray<boolean> | RecursiveArray<...> | Uint8Array[]
export function tensor<R extends Rank>(
    values: TensorLike|WebGLData|WebGPUData, shape?: ShapeMap[R],
    dtype?: DataType): Tensor<R> 
```
Regarding to above definition, number|boolean should be look like  "passing by value". But How about TypedArray?
Based on this thread https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language, I wrote a simple TypeScript case. From below case, you can see TypedArray behaves more like "pass by reference."
```
class Animal {
  name: string;
  constructor(theName: string) {
    this.name = theName;
    console.log("Constructor:" + theName);
  }
}

function passByValueOrReference_TypedArray(array: Float32Array) {
   array[0] = 100;
}

function passByValueOrReference_NumberArray(array: number[]) {
   array[0] = 200;
}

function passByValueOrReference_Class(animal: Animal) {
   console.log(animal);
}

{
    const array = new Float32Array(2);
    array[0] = 42;
    passByValueOrReference_TypedArray(array);
    console.log(array);
}
{
    const array: number[] = [1,2];
    array[0] = 42;
    passByValueOrReference_NumberArray(array);
    console.log(array);
}
{
    // I am not confidient about this case.
    const animal = new Animal("Dog");
    passByValueOrReference_Class(animal);
    console.log(animal);
}

```
```
[LOG]: Float32Array: {
  "0": 100,
  "1": 0
} 
[LOG]: [200, 2] 
[LOG]: "Constructor:Dog" 
[LOG]: Animal: {
  "name": "Dog"
} 
[LOG]: Animal: {
  "name": "Dog"
} 
```

axinging:
Hi， @pyu10055 @qjia7, I think @huningxin @BruceDai is the user of this API, maybe they have some inputs about this?

The WebGL version which requires an extra copy: https://github.com/tensorflow/tfjs/pull/6853
This WebGPU version is zero copy, but user should not destroy the GPUBuffer until all access is done.

Linchenn:
> From below case, you can see TypedArray behaves more like "pass by reference."

You are right, but I think Ping's point is that the current `function tensor(values, shape, dtype)` is designed to be 'passing by value'. Then, whatever the values' type is, we will make a copy from values (if `values` is `TypedArray`, we also make a copy, instead of using users' directly).

axinging:
> > From below case, you can see TypedArray behaves more like "pass by reference."
> 
> You are right, but I think Ping's point is that the current `function tensor(values, shape, dtype)` is designed to be 'passing by value'. Then, whatever the values' type is, we will make a copy from values (if `values` is `TypedArray`, we also make a copy, instead of using users' directly).

Oh, Thanks @Linchenn @pyu10055 , sorry I misunderstand ping's idea.  I think there are four situations here based your and ping:
1.  primitive type, such as number[], string, boolean;
2.  TypedArray: copy or not, or both?
3.  WebGL: required copy.
4.  WebGPU: zero copy

So does you mean we should support something like below?
```
// Create tensor requires copy
export function tensor<R extends Rank>(
    values: TensorLike|WebGLData, shape?: ShapeMap[R],
    dtype?: DataType): Tensor<R> 

// Create tensor without copy
export function tensorWithoutCopy<R extends Rank>(
    values: TensorLike|WebGPUData, shape?: ShapeMap[R],
    dtype?: DataType): Tensor<R> 
```

Linchenn:
> * primitive type, such as number[], string, boolean;
> * TypedArray: copy or not, or both?

I think for the first two cases, we are copying values here:
https://github.com/tensorflow/tfjs/blob/48b2009e15b36927c7663a19ef8503eb83e2619c/tfjs-core/src/ops/tensor_ops_util.ts#L83-L85

> So do you mean we should support something like below?
> // Create tensor requires copy
> export function tensor<R extends Rank>(
>     values: TensorLike|WebGLData, shape?: ShapeMap[R],
>     dtype?: DataType): Tensor<R>
>
> // Create tensor without copy
> export function tensorWithoutCopy<R extends Rank>(
>     values: TensorLike|WebGPUData, shape?: ShapeMap[R],
>     dtype?: DataType): Tensor<R>

Could we implement the 'passing by value' one for WebGPU at first? like:
```js
export function tensor<R extends Rank>(
    values: TensorLike|WebGLData|WebGPUData, shape?: ShapeMap[R],
    dtype?: DataType): Tensor<R> 
```

From my perspective, this has some overheads (I am not sure if the overhead is significant), but it complies with our current design and avoids using data managed by users.

axinging:
For TypedArray, it has a "noConversionNeeded" case, this requires no copy.
I summarized the deep copy or shallow copy of all possible tensor inputs, please corrct me if anything wrong:

| Tensor input type             | Copy         |
| ----------------------------- | ------------ |
| string - number - boolean     | Deep copy    |
| TypedArray noConversionNeeded | Shallow copy |
| TypedArray ConversionNeeded   | Deep copy    |
| Array                         | Deep copy    |
| RecursiveArray                | Deep copy    |
| WebGL                         | Deep copy    |
| WebGPU(Draft)                 | Shalow copy  |

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -51,6 +51,12 @@ type TensorData = {
   shape: number[],
   refCount: number,
   resourceInfo?: BufferInfo|TextureInfo,
+  // zeroCopy is used for creating tensor from GPUBuffer. When zeroCopy is false
+  // or undefined (default), this GPUBuffer will be copied to the tensor's
+  // resource buffer. When zeroCopy is true, tensor will use this GPUBUffer as
+  // tensor's resource buffer, user should not destroy this GPUBuffer until all
+  // access are done.
+  zeroCopy?: boolean,
```
qjia7, line 59: I still like `external` as the name to indicate the resource comes from external and can't be destroyed in tfjs.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -437,6 +450,51 @@ export class WebGPUBackend extends KernelBackend {
...
+    this.tensorMap.set(
+        dataId, {dtype, shape, values: null, refCount: 1, zeroCopy});
+    const tensorData = this.tensorMap.get(dataId);
+    const size = webgpu_util.GPUBytesPerElement(tensorData.dtype) *
+        util.sizeFromShape(tensorData.shape);
+    if (values.buffer.size < size) {
+      throw new Error(`GPUBuffer size(${
+          values.buffer.size}) is smaller than tensor size(${size})!`);
+    } else if (
+        (values.buffer.usage & GPUBufferUsage.STORAGE) !==
```
qjia7, line 485: Let's unify to use `GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC` as the minimum requirements. I don't want users meet errors that they can't call `tensor.data`.

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: And change `WEBGPU_TENSOR_FROM_BUFFER_WITH_ZERO_COPY` flag to `WEBGPU_TENSOR_USE_EXTERNAL_BUFFER` to keep consistency?

qjia7 (commented):
@pyu10055 @Linchenn Thanks for your comments and suggestions. We internally synced this issue and reached an agreement to support create tensor from buffer with one copy by default to keep consistent with other backends. Meanwhile, we add a webgpu flag to enable zero copy path so that we can collect some perf data conveniently in future. Are you ok with this solution?

pyu10055 (changes requested):
Thanks @qjia7 @axinging the approach looks good to me, one suggestion is to add the zeroCopy flag (or call something else) to the WebGPUData interface instead of being a global flag on the backend.
I think this way, it is easier to document the meaning of the flag and the related API, and in the future if WebGL would support the same functionality it can add this flag as well.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7034#-:-NHQg4N45yOvFUCIUkhB:bz8hfjh)* status: 0 of 1 approvals obtained (waiting on @axinging, @qjia7, and @xhcao)

___
*[`tfjs-backend-webgpu/src/backend_webgpu.ts` line 59 at r10](https://reviewable.io/reviews/tensorflow/tfjs/7034#-NGUfISHUE6y5EJdhTs7-r10-59:-NHQeQ3l8HNMcwisVCEZ:b32v57) ([raw file](https://github.com/tensorflow/tfjs/blob/6e0937cd8e893f28370eccc976899a16601b97a3/tfjs-backend-webgpu/src/backend_webgpu.ts#L59)):*

And change `WEBGPU_TENSOR_FROM_BUFFER_WITH_ZERO_COPY` flag to `WEBGPU_TENSOR_USE_EXTERNAL_BUFFER` to keep consistency?
</blockquote></details>

+1
___
*[`tfjs-core/src/types.ts` line 191 at r10](https://reviewable.io/reviews/tensorflow/tfjs/7034#-NHQf3ma2ktXQXszoCGt:-NHQf3ma2ktXQXszoCGu:bj3y00x) ([raw file](https://github.com/tensorflow/tfjs/blob/6e0937cd8e893f28370eccc976899a16601b97a3/tfjs-core/src/types.ts#L191)):*
> ```XML
>  * creating a tensor, tensor type is float32.
>  */
> export interface WebGPUData {
> ```

can we have the zeroCopy flag here, instead of a global flag?

axinging:
@pyu10055 @Linchenn  @qjia7 , zeroCopy is added into WebGPUData,  PTAL.

The current status of copy is summarized as below:

| Tensor input type             | Copy         |  
| ----------------------------- | ------------ | 
| string - number - boolean     | Deep copy    | 
| TypedArray noConversionNeeded | Shallow copy | 
| TypedArray ConversionNeeded   | Deep Copy| 
| Array                         | Deep copy    | 
| RecursiveArray                | Deep copy    |           
| WebGL                         | Deep copy    |           
| WebGPU(zeroCopy = false)                 | Shallow copy  |         
| WebGPU(zeroCopy = true)                 | Deep copy  |

qjia7 (approved):
LGTM, thanks!

Linchenn (approved):
Thank you Xu so much for the amazing work! LGTM except some small nits.

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7034#-:-NHVneCj4XBtTVR2CO3L:boqqj89)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @axinging, @qjia7, and @xhcao)

___
*[`tfjs-core/src/ops/tensor.ts` line 105 at r13](https://reviewable.io/reviews/tensorflow/tfjs/7034#-NHVn-4M3cAQDMwNvJp2:-NHVn-4M3cAQDMwNvJp3:b-rp4vel) ([raw file](https://github.com/tensorflow/tfjs/blob/81b08d3fc435b6ca93f41b762279d725383a7033/tfjs-core/src/ops/tensor.ts#L105)):*
> ```XML
>  * // by WebGPUData.zeroCopy. When zeroCopy is false or undefined(default), this
>  * // passing GPUBuffer can be destroyed after tensor is created. When zeroCopy
>  * // is true, this GPUBuffer is bound directly by the tensor, so donot destroy
> ```

do not
___
*[`tfjs-core/src/ops/tensor.ts` line 184 at r13](https://reviewable.io/reviews/tensorflow/tfjs/7034#-NHVnXAM-el3h-xUXUV1:-NHVnXAM-el3h-xUXUV2:b-rp4vel) ([raw file](https://github.com/tensorflow/tfjs/blob/81b08d3fc435b6ca93f41b762279d725383a7033/tfjs-core/src/ops/tensor.ts#L184)):*
> ```XML
>  * zero copy by flag zeroCopy. When zeroCopy is false or undefined(default),
>  * this passing GPUBuffer can be destroyed after tensor is created. When
>  * zeroCopy is true, this GPUBuffer is bound directly by the tensor, so donot
> ```

do not

pyu10055 (approved):

PR #7035, opened 2022-11-10
webgpu: Correct depthwise conv2d vec4 condition

qjia7:
We should use the output width/height instead of input width/height as the condition of whether going to vec4 version since the task is splitted as output shape.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7035)


gyagp (approved):

xhcao (approved):
LGTM

PR #7048, opened 2022-11-15
[e2e] Add type check for TypedArray

axinging:


e2e/benchmarks/local-benchmark/util.js
```diff
@@ -102,6 +102,18 @@ function expectObjectsPredicate(actual, expected, epsilon, predicate) {
     throw new Error(`Actual length ${
         actualKeys.length} not equal Expected length ${expectedKeys.length}`);
   }
+
+  // For TypedArray, check type first.
+  if (tf.util.isTypedArray(actual) && tf.util.isTypedArray(expected)) {
```
qjia7, line 107: It seems that you missed the situation to check whether `actual` and `expected` are directly an TypedArray before. So will it be better to change it like below?
```
if (tf.util.isTypedArray(actual) && tf.util.isTypedArray(expected))
{
  expectArraysClose(actual, expected, epsilon, key);
  return true;
} 

for (let i = 0; i < actualKeys.length; i++) {
  ...
}

axinging: This case is specially for our version of "areClose(a, b, epsilon)". So we should avoid call expectArraysClose(actual, expected, epsilon, key) here.

PR #7057, opened 2022-11-17
Remove 'sideEffects' field from package.jsons

mattsoulanille:
Some of tfjs's package.json files have a `sideEffects` field to mark certain files as side-effectful for bundlers. However, bundlers can do static analysis to find side effects automatically, and this is usually more accurate and less error-prone than relying on a manually maintained list.

This PR removes the `sideEffects` field from the package.json files.

### DIFF
--- a/tfjs-backend-cpu/package.json
+++ b/tfjs-backend-cpu/package.json
@@ -42,14 +42,6 @@
     "util": false,
     "crypto": false
   },
-  "sideEffects": [
-    "./dist/register_all_kernels.js",
-    "./dist/base.js",
-    "./dist/index.js",
-    "./src/register_all_kernels.mjs",
-    "./src/base.mjs",
-    "./src/index.mjs"
-  ],
   "resolutions": {
     "minimist": "1.2.6"
   }
--- a/tfjs-backend-webgl/package.json
+++ b/tfjs-backend-webgl/package.json
@@ -42,19 +42,5 @@
   "browser": {
     "util": false,
     "crypto": false
-  },
-  "sideEffects": [
-    "./dist/register_all_kernels.js",
-    "./dist/flags_webgl.js",
-    "./dist/base.js",
-    "./dist/index.js",
-    "./dist/register_all_kernels.mjs",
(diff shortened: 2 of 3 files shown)


mattsoulanille:
I've run [nightly tests](https://pantheon.corp.google.com/cloud-build/builds/e69a73ce-3957-4549-a4fa-aee25bffef0e;step=38?project=learnjs-174218) on this PR and found no failing tests. I still need to check if this change affects bundle sizes.

pyu10055 (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7057)* status: :shipit: complete! 1 of 1 approvals obtained

mattsoulanille:
I'm seeing some size differences in custom bundles:
Without this change:
```
1000K	dist/full/app_rollup.js
432K	dist/custom/app_rollup.js
```

With this change:
```
1004K	dist/full/app_rollup.js
524K	dist/custom/app_rollup.js
```

Without `sideEffects`, rollup seems to be ignoring the custom bundle's list of ops and including everything.

mattsoulanille:
This may require a more extensive refactor than I initially anticipated. tfjs-core [exports all](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/base.ts#L60) the ops from [`ops.ts`](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/ops.ts) (these are eventually [exported in `index.ts`](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/index.ts#L21)), which I think means they can not be tree-shaken. Custom bundles replace the [`ops_for_converter.ts`](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/ops_for_converter.ts) file (which re-exports all the ops) with a new file that exports only a subset of the ops. However, they still export all of tfjs-core's `base.ts`, which re-exports all the ops, so I'm not really sure how this works when `sideEffects` is set in the package.json but not when it's not set.

Issue #7061, opened 2022-11-18
Incorrect result by matmul 4D x 3D 


BruceDai:
<em>Please make sure that this is a bug. As per our
[GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md),
we only address code/doc bugs, performance issues, feature requests and
build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): Yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): all platforms
- TensorFlow.js version (use command below): 3.19.0
- Browser version: 107.0.5304.107

**Describe the current behavior**
Run below simple JavaScript code, get incorrect result comparing with the result by Python.
```js
// matmul 4D x 3D
const a = tf.tensor4d([1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16], [2, 2, 2, 2]);
const b = tf.tensor3d([1,  2,  3,  4], [2, 2, 1]);
a.matMul(b);
```
```
Output result:

Tensor
    [[[[5  ],
       [11 ]],

      [[39 ],
       [53 ]]],


     [[[67 ], // incorrect, expected 29
       [81 ]], // incorrect, expected 35

      [[95 ],
       [109]]]]
```
**Describe the expected behavior**
```python
# Run Python code using numpy module
import numpy as np
a = np.array(np.arange(1, 17))
b = np.array(np.arange(1, 5))
c = a.reshape([2, 2, 2, 2])
d = b.reshape([2, 2, 1])
np.matmul(c, d)
```

```
Output result:

array([[[[  5],
         [ 11]],

        [[ 39],
         [ 53]]],


       [[[ 29],
         [ 35]],

        [[ 95],
         [109]]]])
```

**Standalone code to reproduce the issue**
Provide a reproducible test case that is the bare minimum necessary to generate
the problem. If possible, please share a link to Colab/CodePen/any notebook.
Please see above code.

@lina128  PTAL, thanks.

PR #7062, opened 2022-11-18
Fix batch matching for batch mat mul

Linchenn:
Fix https://github.com/tensorflow/tfjs/issues/7061 for CPU, WASM (native implementation) and WebGL backends.

This error is because of batch mismatch when A has more dimensions than B, and vice versa. 

For example: A's shape is `[2,4,3,3]` and B's shape is `[4,3,3]`. Then A's batch is the first two dimensions `[2, 4]` while B's batch is the first dimension `[4]`, so B's batch is supposed to be broadcasted to be `[2, 4]`. Then, to compute `output[1][0][0][0]`, we have to do dot product of `A[1][0]   [0][...]` with `B[0]   [...][0]`, but the current algorithm is doing dot product of `A[1][0][0][...]` with `B[3][...][0]`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7062)

### DIFF
--- a/tfjs-backend-cpu/src/kernels/BatchMatMul.ts
+++ b/tfjs-backend-cpu/src/kernels/BatchMatMul.ts
@@ -114,8 +114,10 @@ export function batchMatMul(args: {
                   batchIndexB = bi % batchDimB;
                 }
                 const aVal =
+                    // tslint:disable-next-line: max-line-length
                     a3dValues[batchIndexA * aBatch + i * aOuterStep + k * aInnerStep];
                 const bVal =
+                    // tslint:disable-next-line: max-line-length
                     b3dValues[k * bInnerStep + j * bOuterStep + batchIndexB * bBatch];
                 sum += aVal * bVal;
               }
--- a/tfjs-backend-cpu/src/kernels/BatchMatMul.ts
+++ b/tfjs-backend-cpu/src/kernels/BatchMatMul.ts
@@ -106,12 +106,17 @@ export function batchMatMul(args: {
               let sum = 0.0;
 
               for (let k = k0; k < kBlock; k++) {
-                const batchOffsetA = Math.min(bi, batchDimA - 1) * aBatch;
-                const batchOffsetB = Math.min(bi, batchDimB - 1) * bBatch;
+                let batchIndexA = bi;
+                let batchIndexB = bi;
+                if (bi >= batchDimA) {
+                  batchIndexA = bi % batchDimA;
+                } else if (bi >= batchDimB) {
+                  batchIndexB = bi % batchDimB;
+                }
                 const aVal =
-                    a3dValues[batchOffsetA + i * aOuterStep + k * aInnerStep];
+                    a3dValues[batchIndexA * aBatch + i * aOuterStep + k * aInnerStep];
                 const bVal =
-                    b3dValues[k * bInnerStep + j * bOuterStep + batchOffsetB];
+                    b3dValues[k * bInnerStep + j * bOuterStep + batchIndexB * bBatch];
                 sum += aVal * bVal;
               }
               resVals[bi * size + (i * rightDim + j)] += sum;
--- a/tfjs-backend-cpu/src/kernels/BatchMatMul.ts
+++ b/tfjs-backend-cpu/src/kernels/BatchMatMul.ts
@@ -106,13 +106,8 @@ export function batchMatMul(args: {
               let sum = 0.0;
 
               for (let k = k0; k < kBlock; k++) {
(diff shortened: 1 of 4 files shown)


tfjs-backend-cpu/src/kernels/BatchMatMul.ts
```diff
@@ -106,12 +106,19 @@ export function batchMatMul(args: {
...
               for (let k = k0; k < kBlock; k++) {
-                const batchOffsetA = Math.min(bi, batchDimA - 1) * aBatch;
-                const batchOffsetB = Math.min(bi, batchDimB - 1) * bBatch;
+                let batchIndexA = bi;
+                let batchIndexB = bi;
+                if (bi >= batchDimA) {
+                  batchIndexA = bi % batchDimA;
+                } else if (bi >= batchDimB) {
+                  batchIndexB = bi % batchDimB;
+                }
```
mattsoulanille, line 115: Is the `if` statement here necessary? `batchDim = Math.max(batchDimA, batchDimB)`, and `bi < batchDim`, so it's never the case that `bi >= batchDimA && bi >= batchDimB` (`bi` can only be greater than or equal to one of them (or none of them)), meaning the `else if` case will always run when its condition is true (since the `if` condition will have been false). That lets us rewrite this code as two if statements:
```ts
if (bi >= batchDimA) {
  batchIndexA = bi % batchDimA;
}
if (bi >= batchDimB) {
  batchIndexB = bi % batchDimB;
}
```
Then, we can remove the `if` statements and just use modulus:

```suggestion
                const batchIndexA = bi % batchDimA;
                const batchIndexB = bi % batchDimB;
```

tfjs-backend-wasm/src/cc/batch_mat_mul_impl.cc
```diff
@@ -235,14 +235,17 @@ void slow_batch_matmul(const size_t a_id, const size_t* a_shape_ptr,
...
-                sum +=
-                    a_buf[batch_offset_a + i * a_outer_step +
-                          k * a_inner_step] *
-                    b_buf[k * b_inner_step + j * b_outer_step + batch_offset_b];
+                size_t batch_index_a = b;
+                size_t batch_index_b = b;
+                if (b >= a_shape[0]) {
+                  batch_index_a = b % a_shape[0];
+                } else if (b >= b_shape[0]) {
+                  batch_index_b = b % b_shape[0];
```
mattsoulanille, line 243: See my comment in the CPU implementation for why this should work.
```suggestion
                size_t batch_index_a = b % a_shape[0];
                size_t batch_index_b = b % b_shape[0];
```

tfjs-backend-cpu/src/kernels/BatchMatMul.ts
```diff
@@ -106,12 +106,19 @@ export function batchMatMul(args: {
...
                 const aVal =
-                    a3dValues[batchOffsetA + i * aOuterStep + k * aInnerStep];
+                    // tslint:disable-next-line: max-line-length
+                    a3dValues[batchIndexA * aBatch + i * aOuterStep + k * aInnerStep];
                 const bVal =
-                    b3dValues[k * bInnerStep + j * bOuterStep + batchOffsetB];
+                    // tslint:disable-next-line: max-line-length
+                    b3dValues[k * bInnerStep + j * bOuterStep + batchIndexB * bBatch];
```
mattsoulanille, line 121: I was wondering why `k` matched with `aInnerStep` but did not then match with `bOuterStep`, but I see now that `bOuterStep` and `bInnerStep` are opposite to `aOuterStep` and `aInnerStep` (lines 82 - 87). They don't refer to 'rows' and 'columns' of the matrices being multiplied, where you would step `a`'s row index with `b`'s column index for the dot product.

No action necessary.

mattsoulanille (approved):
LGTM with a nit related to modulus. Thanks!

tfjs-backend-cpu/src/kernels/BatchMatMul.ts
Linchenn: Thank you Matt! Done.

tfjs-backend-wasm/src/cc/batch_mat_mul_impl.cc
Linchenn: Done.

Linchenn:
cc @qjia7 @xhcao In case WebGPU has similar problem. This problem happens in BatchMatMul when A's batch size does not match B's batch size. For details, you could take a look at my descriptions.

PR #7145, opened 2022-12-07
Increase timeout and add Gpu info to nightly benchmark

Linchenn:
This PR:
1. Set a timeout for child processes spawned from `app.js` program. Even though we have set a couple of timeout in karma, we also want to restrict the time of the child processes running the karma test (in a higher level). Otherwise, some benchmark item takes too much time. (for example, the following test on A11 machine takes approximately 21 minutes.)
image
2. Append GPU info to device info. (Alternative solution: add a new filed for GPU info. However, we have to change the schema of the firestore collection.)
3. Changes `maxBenchmarks` from 9 to 12. Because currently we have 23 target devices, this change possibly reduces time each model (if no re-try happens, approximately reduce from `ceil(23 / 9) * average benchmark time` to `ceil(23 / 12) * average benchmark time`).

### DIFF
--- a/e2e/benchmarks/browserstack-benchmark/app.js
+++ b/e2e/benchmarks/browserstack-benchmark/app.js
@@ -281,7 +281,7 @@ function runBrowserStackBenchmark(tabId) {
     const command = `yarn ${args.join(' ')}`;
     console.log(`Running: ${command}`);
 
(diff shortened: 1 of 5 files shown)


mattsoulanille (approved):
I agree that we should modify the schema to add additional fields like GPU. For now, though, this looks good to me.

PR #7152, opened 2022-12-08
webgpu: support dilation2d operator

xhcao:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7152)

### DIFF
--- a/tfjs-backend-webgpu/src/dilation_webgpu.ts
+++ b/tfjs-backend-webgpu/src/dilation_webgpu.ts
@@ -18,21 +18,22 @@
 import {backend_util} from '@tensorflow/tfjs-core';
 
 import {getMainHeaderString as main, WebGPUProgram} from './webgpu_program';
-import {computeDispatch} from './webgpu_util';
+import {computeDispatch, flatDispatchLayout} from './webgpu_util';
 
 export class Dilation2DProgram implements WebGPUProgram {
   outputShape: number[];
   shaderKey: string;
-  dispatchLayout: {x: number[], y: number[], z: number[]};
+  dispatchLayout: {x: number[]};
   dispatch: [number, number, number];
   variableNames = ['x', 'w'];
   uniforms =
       'filterDims: vec2<i32>, pad: vec2<i32>, stride: vec2<i32>, dilation: vec2<i32>';
-  workgroupSize: [number, number, number] = [8, 16, 1];
+  workgroupSize: [number, number, number] = [64, 1, 1];
+  size = true;
 
   constructor(convInfo: backend_util.Conv2DInfo) {
     this.outputShape = convInfo.outShape;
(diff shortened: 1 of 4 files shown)


xhcao:
I had tried to implement this operator using shared memory feature, which is similar as conv2d operator. But when defining the shared memory for the `x` and `filter`, this dimensions are different, `x` dimension is 3d, `filter` dimension is 2d, each thread must load `x` data multiple times when there is only one time for `filter`, it is difficult to fix the situation for high performance.
So I referred the webgl implementation here firstly, there are 2-loos for each thread, and the loop size is filter size and small, so I think there is no huge performance loss as naive conv2d, which is 3-loops, including large channel loop. But we could optimize it in the future.

tfjs-backend-webgpu/src/dilation_webgpu.ts
```diff
@@ -0,0 +1,78 @@
...
+  dispatchLayout: {x: number[], y: number[], z: number[]};
+  dispatch: [number, number, number];
+  variableNames = ['x', 'w'];
+  uniforms =
+      'filterDims: vec2<i32>, pad: vec2<i32>, stride: vec2<i32>, dilation: vec2<i32>';
+  workgroupSize: [number, number, number] = [8, 16, 1];
+
+  constructor(convInfo: backend_util.Conv2DInfo) {
+    this.outputShape = convInfo.outShape;
+    this.dispatchLayout = {x: [3], y: [1, 2], z: [0]};
```
qjia7, line 35: Please use the flat dispatch layout if your algorithm is irrelevant with compute work group layout.

qjia7 (approved):

tfjs-backend-webgpu/src/kernels/Dilation2D.ts
```diff
@@ -0,0 +1,60 @@
...
+    {type: 'int32', data: [convInfo.strideHeight, convInfo.strideWidth]},
+    {type: 'int32', data: [convInfo.dilationHeight, convInfo.dilationWidth]}
+  ];
+
+  const program = new Dilation2DProgram(convInfo);
+  const out =
+      backend.runWebGPUProgram(program, [x, filter], 'float32', uniformData);
+
+  const outReshaped =
+      reshape({inputs: {x: out}, backend, attrs: {shape: convInfo.outShape}});
```
qjia7, line 50: Sorry, forgot one. Why you still need the additional `reshape` since the `out` shape is already `convInfo.outShape`?

gyagp (approved):
LGTM % Jiajia's comment

tfjs-backend-webgpu/src/kernels/Dilation2D.ts
xhcao: Done, I had remove this reshape code. And changed the type of output tensor to rely on `x` dtype.

PR #7153, opened 2022-12-09
[webgpu] Add a huge matmul test

hujiajie:
Currently the matMul splitK kernel has a threshold of K=128 for dispatching multiple workgroups along the Z axis, but the existing 'matMul A x B multiple tiles' spec is not large enough to cover the case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7153)

### DIFF
--- a/tfjs-backend-webgpu/src/matmul_test.ts
+++ b/tfjs-backend-webgpu/src/matmul_test.ts
@@ -1219,6 +1219,17 @@ function matmulTest(programType: MatMulProgramType) {
       ]);
     });
 
+    it('A x B with large K. [16,2048]x[2048,16]', async () => {
+      const a = tf.tensor2d(new Array(16 * 2048).fill(1), [16, 2048]);
+      const b = tf.tensor2d(new Array(2048 * 16).fill(1), [2048, 16]);
+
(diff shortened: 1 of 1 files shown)


qjia7 (approved):
LGTM, thanks!

tfjs-backend-webgpu/src/matmul_test.ts
```diff
@@ -1219,6 +1219,17 @@ function matmulTest(programType: MatMulProgramType) {
       ]);
     });
 
+    it('[16,2048]x[2048,16]', async () => {
```
xhcao, line 1222: Is it necessary to describe this test name in more detail here?

hujiajie: Do you have any recommendation for this and `[8,4]x[4,8]` right above?

qjia7: How about changing it like below?
```
...
it('A x B with M/N/K divisible by 4. [8,4]x[4,8]',
...
it('A x B with large K. [16,2048]x[2048,16]',
```

xhcao (approved):

PR #7154, opened 2022-12-09
webgpu: support multinomial operator

xhcao:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7154)

### DIFF
--- a/tfjs-backend-webgpu/src/multinomial_webgpu.ts
+++ b/tfjs-backend-webgpu/src/multinomial_webgpu.ts
@@ -1,0 +1,77 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import {getMainHeaderString as main, WebGPUProgram} from './webgpu_program';
+import {computeDispatch, flatDispatchLayout} from './webgpu_util';
+
+export class MultinomialProgram implements WebGPUProgram {
+  variableNames: string[] = ['probs'];
+  outputShape: number[] = [];
+  shaderKey: string;
+  dispatchLayout: {x: number[]};
+  dispatch: [number, number, number];
+  uniforms = 'seed : f32, numOutcomes: i32,';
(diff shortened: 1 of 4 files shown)


qjia7 (approved):

tfjs-backend-webgpu/src/multinomial_webgpu.ts
```diff
@@ -0,0 +1,77 @@
...
+    const userCode = `
+    //Based on the work of Dave Hoskins
+    //https://www.shadertoy.com/view/4djSRW
+    fn random (seed : f32, resultUV : vec2<f32>) -> f32 {
+      let HASHSCALE1 = 443.8975;
+      let p = resultUV * seed;
+      var p3  = fract(vec3<f32>(p.xyx) * HASHSCALE1);
+      p3 = p3 + dot(p3, p3.yzx + 19.19);
+      return fract((p3.x + p3.y) * p3.z);
+    }
```
gyagp, line 50: We can put this in shader_util.ts, so that other code may share.

tfjs-backend-webgpu/src/multinomial_webgpu.ts
```diff
@@ -0,0 +1,77 @@
...
+    ${main('index')} {
+      if (index < uniforms.size) {
+        let coords = getOutputCoords();
+        let batch = coords[0];
+
+        let resUV = vec2<f32>(f32(coords[1]) / f32(uniforms.outShape[1]),
+            f32(coords[0]) / f32(uniforms.outShape[0]));
+        let r = random(uniforms.seed, resUV);
+        var cdf = 0.0;
+        for (var i = 0; i < uniforms.numOutcomes; i = i + 1) {
```
gyagp, line 61: In WebGL impl, it's (uniforms.numOutcomes -1) here.

tfjs-backend-webgpu/src/multinomial_webgpu.ts
xhcao: The function could be only used for 2D coordinate axes. If we want to move it to shader_util.ts, we should design a  more common function for 1~4D  coordinate axes on webgpu backend. Is it OK to do it in the future if necessary?

tfjs-backend-webgpu/src/multinomial_webgpu.ts
xhcao: Thanks for finding the error.

qjia7: Is it easy to add a case to catch such kind of error?

tfjs-backend-webgpu/src/multinomial_webgpu.ts
gyagp: Sure, we can make it a util function when necessary in the future.

tfjs-backend-webgpu/src/multinomial_webgpu.ts
xhcao: `r` is random value, I think it is difficult to design the case.

xhcao:
@gyagp  PTAL

gyagp (approved):

PR #7155, opened 2022-12-09
Mark all calls to 'op()' as pure

mattsoulanille:
Mark calls to the `op()` function that creates the exported op as pure by using [`/* @__PURE__ */` annotations](https://esbuild.github.io/api/#ignore-annotations) (this also works for Rollup, but I can't find the docs). This comment instructs bundlers that the function call has no side-effects, so it can be removed if the result is not used.

This is okay for the `op` function because, although it references ENGINE, it does so [in a closure](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/operation.ts#L48-L61) that it never calls, so while its return value may cause side effects when called, it itself does not.

This has no immediate effect because we still maintain a list of `sideEffects` in the package.json, but it is a step towards removing that list.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7155)

### DIFF
--- a/tfjs-core/src/ops/abs.ts
+++ b/tfjs-core/src/ops/abs.ts
@@ -48,4 +48,4 @@ function abs_<T extends Tensor>(x: T|TensorLike): T {
   }
 }
 
-export const abs = op({abs_});
+export const abs = /* @__PURE__ */ op({abs_});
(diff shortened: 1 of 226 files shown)


pyu10055 (approved):
> Reviewed 226 of 226 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7155)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @Linchenn)

Linchenn (approved):

PR #7168, opened 2022-12-12
Use static getters to get optimizer class names

mattsoulanille:
Each `Optimizer` lists its class name as a static property of the class so it can be serialized and deserialized. This prevents the class from being tree-shaken because bundlers will compile it like this:

```ts
class SomeOptimizer {
  ...
}

// The bundler can not remove this assignment because
// SomeOptimizer.className could be a setter with a side effect.
SomeOptimizer.className = 'SomeOptimizer';
```

This PR uses a static getter for the class name instead, which bundlers can tree-shake properly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7168)


pyu10055 (approved):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7168)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @Linchenn)

PR #7169, opened 2022-12-12
Enable esModuleInterop and update Long to 5.x

mattsoulanille:
Enable the TypeScript [`esModuleInterop` flag](https://www.typescriptlang.org/tsconfig#esModuleInterop), which was released in TS2.7 and is a recommended config option. This removes the need for some workarounds we use for rollup plugins and Long.

Upgrade Long from 4.0.0 to 5.2.1.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7169)


mattsoulanille:
Actually, this is a breaking change. I'm not sure we want to do it.

See `Module interop` in https://rfcs.emberjs.com/id/0730-semver-for-ts/

PR #7185, opened 2022-12-17
[webgpu] Further tweak vectorized NaN handling in binary ops

hujiajie:
The big picture will be:

```
  fn binaryOperation(a : vec4<f32>, b : vec4<f32>) -> vec4<f32> {
    let isNaN = false;
    {
      // op-specific code begins
      let result = ...;
      let isNaN = ...;  // optional, declare this only if the op defines
                        // its own extra rules about NaN, say mod(0, 0)
      // op-specific code ends

      return select(result, vec4<f32>(valueForNaN),
                    vec4<bool>(isNaN) | isnanVec4(a) | isnanVec4(b));
    }
  }
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7185)

### DIFF
--- a/tfjs-backend-webgpu/src/binary_op_util.ts
+++ b/tfjs-backend-webgpu/src/binary_op_util.ts
@@ -46,8 +46,9 @@ const CHECK_NAN_SNIPPET = `
   `;
 
 const CHECK_NAN_SNIPPET_VEC4 = `
-  let isNaN = isnanVec4(a) | isnanVec4(b);
-  resultTemp = select(resultTemp, vec4<f32>(valueForNaN), isNaN);
+  resultTemp = select(
+      resultTemp, vec4<f32>(valueForNaN),
+      vec4<bool>(isNaN) | isnanVec4(a) | isnanVec4(b));
   `;
 
 const ADD = 'return a + b;';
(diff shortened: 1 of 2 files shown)


tfjs-backend-webgpu/src/binary_op_util.ts
```diff
@@ -197,7 +186,9 @@ const POW_VEC4 = `
     resultTemp.a = 1.0;
   }
   let isNaN = (a < vec4<f32>(0.0)) & (floor(b) < b);
-  return select(resultTemp, vec4<f32>(uniforms.NAN), isNaN);
+  let valueForNaN = uniforms.NAN;
+  ${CHECK_NAN_SNIPPET_VEC4}
+  return resultTemp;
```
qjia7, line 191: The original code only checks `isNan`. Here you also add `isnanVec4(a) | isnanVec4(b)`. Is it expected?

hujiajie: It's almost certain `binOp(NaN, b)` or `binOp(a, NaN)` will return `valueForNaN`. I guess previously this case was just quietly covered by the `pow()` builtin.

tfjs-backend-webgpu/src/binary_op_webgpu.ts
```diff
@@ -87,7 +87,10 @@ export class BinaryOpProgram implements WebGPUProgram {
     const dType = this.isVec4 ? 'vec4<f32>' : 'f32';
     const opFnStr = `
     fn binaryOperation(a : ${dType}, b : ${dType}) -> ${dType} {
-      ${getBinaryOpString(this.op, this.isVec4)}
+      let isNaN = false;
```
qjia7, line 90: Can you add a comment to `isNaN` in the code? Like you did in the commit message. And in `MOD_VEC4` and `POW_VEC4`. You can directly use `isNaN = ...`.
```
     // Define op own extra rules about NaN, say mod(0, 0).
      var isNaN = ${this.isVec4 ?  'vec4<bool>(false)' : 'false'};

qjia7 (approved):

tfjs-backend-webgpu/src/binary_op_webgpu.ts
hujiajie: I feel it's better to have a combined comment after `valueForNaN` is restructured similarly, or even later after everything calms down.

The `let` before `isNaN` in op-specific code is an intentional design. IMO, the snippet looks more self-contained that way, so I won't wonder where is the declaration of this `isNaN` identifier while reading `MOD_VEC4`.

qjia7: Sounds good.

qjia7 (approved):
Still LGTM, thanks.

xhcao (approved):
LGTM, thanks.

gyagp (approved):
LGTM

PR #7191, opened 2022-12-20
[webgpu] Enable parallel compilation

axinging:
Example:
```
  // Parallel compile.
  tf.env().set('WEBGPU_ENGINE_COMPILE_ONLY', true);
  const result1 = predict(model);
  await tf.backend().checkCompileCompletionAsync();
  tf.dispose(result1);
  // Actual inference.
  tf.env().set('WEBGPU_ENGINE_COMPILE_ONLY', false);
  const result2 = predict(model);
  await result2.data();
  tf.dispose(result2);
```

This change is based on: https://github.com/tensorflow/tfjs/pull/5815

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7191)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -373,9 +373,9 @@ export class WebGPUBackend extends KernelBackend {
 
   public async getBufferData(buffer: GPUBuffer, size: number):
       Promise<ArrayBuffer> {
-    if (env().getBool('ENGINE_COMPILE_ONLY')) {
+    if (env().getBool('WEBGPU_ENGINE_COMPILE_ONLY')) {
       console.warn(
-          'The data may be invalid since ENGINE_COMPILE_ONLY is true, this can only be called when ENGINE_COMPILE_ONLY is false');
+          'The data may be invalid since WEBGPU_ENGINE_COMPILE_ONLY is true, this can only be called when WEBGPU_ENGINE_COMPILE_ONLY is false');
       return null;
     }
     const staging = this.bufferManager.acquireBuffer(
@@ -929,7 +929,7 @@ export class WebGPUBackend extends KernelBackend {
     program.shaderKey =
         webgpu_program.makeShaderKey(program, inputsData, output);
 
-    const parallelCompilation = env().getBool('ENGINE_COMPILE_ONLY');
+    const parallelCompilation = env().getBool('WEBGPU_ENGINE_COMPILE_ONLY');
     if (!(program.shaderKey in this.pipelineCache)) {
       this.pipelineCache[program.shaderKey] = webgpu_program.compileProgram(
           this.device, program, inputsData, output, parallelCompilation);
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -129,7 +129,8 @@ export class WebGPUBackend extends KernelBackend {
   private dummyContext: GPUCanvasContext;
   private tensorDataPendingDisposal: DataId[] = [];
   private static nextDataId = 0;
-  private pipelineCache: {[key: string]: GPUComputePipeline};
+  private pipelineCache:
+      {[key: string]: GPUComputePipeline|Promise<GPUComputePipeline>};
   private programTimersStack: TimerNode[];
   private querySet: GPUQuerySet;
   private stagingPendingDisposal: BufferInfo[] = [];
@@ -356,8 +357,27 @@ export class WebGPUBackend extends KernelBackend {
     return this.currentComputePass;
   }
 
+  // Check if parallel compilation is done.
+  async checkCompileCompletionAsync() {
+    let pipelines: GPUComputePipeline[];
+    try {
+      pipelines = await Promise.all(Object.values(this.pipelineCache));
+    } catch (e) {
+      // TODO: Add test case to catch this exception.
+      throw new Error(e.message);
+    }
+    Object.keys(this.pipelineCache).map((key, i) => {
+      this.pipelineCache[key] = pipelines[i];
+    });
+  }
+
   public async getBufferData(buffer: GPUBuffer, size: number):
       Promise<ArrayBuffer> {
+    if (env().getBool('ENGINE_COMPILE_ONLY')) {
+      console.warn(
+          'The data may be invalid since ENGINE_COMPILE_ONLY is true, this can only be called when ENGINE_COMPILE_ONLY is false');
+      return null;
+    }
     const staging = this.bufferManager.acquireBuffer(
         size, GPUBufferUsage.COPY_DST | GPUBufferUsage.MAP_READ);
     this.ensureCommandEncoderReady();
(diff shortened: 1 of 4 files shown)


axinging:

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -827,49 +847,72 @@ export class WebGPUBackend extends KernelBackend {
...
+      this.asyncProgramInfos.push({program, bindings});
+    }
+    return output;
+  }
+
+  private recordAndSubmit(
+      program: webgpu_program.WebGPUProgram, bindings: GPUBindingResource[],
+      shouldTimeProgram: boolean) {
+    program.bindGroup = this.device.createBindGroup({
+      layout: (program.pipeline as GPUComputePipeline).getBindGroupLayout(0),
```
qjia7, line 890: Here we should have to check that the `program.pipeline` is an object of `GPUComputePipeline` not `Promise<GPUComputePipeline>`.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -355,8 +357,23 @@ export class WebGPUBackend extends KernelBackend {
...
+    const pipelines = await Promise.all(pipelinesPromise);
+    for (let i = 0; i < pipelines.length; i++) {
+      this.pipelineCache[asyncProgramInfos[i].shaderKey] = pipelines[i];
+    }
+  }
+
   public async getBufferData(buffer: GPUBuffer, size: number):
       Promise<ArrayBuffer> {
+    if (env().getBool('ENGINE_COMPILE_ONLY')) {
+      console.warn('This can only be called when ENGINE_COMPILE_ONLY is false');
```
qjia7, line 374: How about changing the message to `The data may be invalid since ENGINE_COMPILE_ONLY is set` since it's a warning? And `return null` can be removed?

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -827,49 +835,65 @@ export class WebGPUBackend extends KernelBackend {
       this.makeUniforms(programUniform)
     ];
```
qjia7, line 836: I seems that in `ENGINE_COMPILE_ONLY` mode, the `bindings`/ `programUniform` are not necessary. They all can be skipped. So it will be better to move them into `recordAndSubmit`.

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -129,7 +129,9 @@ export class WebGPUBackend extends KernelBackend {
   private dummyContext: GPUCanvasContext;
   private tensorDataPendingDisposal: DataId[] = [];
   private static nextDataId = 0;
-  private pipelineCache: {[key: string]: GPUComputePipeline};
+  private pipelineCache:
+      {[key: string]: GPUComputePipeline|Promise<GPUComputePipeline>};
+  private asyncProgramInfos: webgpu_program.WebGPUProgram[] = [];
```
qjia7, line 134: If you need a separate async pipeline cache, maybe just define it as `private asyncPipelineCache: {[key: string]: Promise<GPUComputePipeline>};` or reuse `pipelineCache`?

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -355,8 +357,23 @@ export class WebGPUBackend extends KernelBackend {
     return this.currentComputePass;
   }
 
+  // Check if parallel compilation is done.
+  async checkCompileCompletion() {
+    const asyncProgramInfos = this.asyncProgramInfos;
+    this.asyncProgramInfos = [];
```
qjia7, line 363: Will `asyncProgramInfos` also become `[]` since `this.asyncProgramInfos = []`?

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -355,8 +357,23 @@ export class WebGPUBackend extends KernelBackend {
     return this.currentComputePass;
   }
 
+  // Check if parallel compilation is done.
+  async checkCompileCompletion() {
```
qjia7, line 361: Should we rename it to `checkCompileCompletionAsync` to keep consistent with webgl https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/backend_webgl.ts#L1228?

tfjs-backend-webgpu/src/backend_webgpu.ts
axinging: We should tell user that data is invalid and how to avoid this. return null tells user that this data is invalid.

axinging (commented):
Done, PTAL

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -783,6 +803,52 @@ export class WebGPUBackend extends KernelBackend {
...
+    if (!parallelCompilation) {
+      this.recordAndSubmit(
+          program, output, inputs, tensorShapes, programDefinedUniform);
+    }
+    return output;
+  }
+
+  private recordAndSubmit(
+      program: webgpu_program.WebGPUProgram, output: TensorInfo,
+      inputs: TensorInfo[], tensorShapes: number[][],
```
qjia7, line 846: `tensorShapes` is useless?

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -783,6 +803,52 @@ export class WebGPUBackend extends KernelBackend {
...
+        // Returning dtype from tensorMap because it reflects dtype
+        // of underlying buffer, rather than abstract dtype.
+        dtype: this.tensorMap.get(input.dataId).dtype,
+        shape: input.shape,
+        name: program.variableNames[i]
+      };
+    });
+
+    program.shaderKey =
+        webgpu_program.makeShaderKey(program, tensorShapes, inputsData, output);
```
qjia7, line 827: Since the program.shaderKey has been assigned to the full shader key, do you still need to pass `program.shaderKey` to `compileProgram in L831?

qjia7: It seems that `tensorShapes` can also be removed since you have `inputsData` and `output`.

tfjs-backend-webgpu/src/backend_webgpu.ts
axinging: removed

tfjs-backend-webgpu/src/backend_webgpu.ts
axinging: Done

qjia7 (approved):
LGTM, thanks.
@Linchenn Please take a look. This is only functionally supported. We can use separate PRs to follow up if we work out a solution to better expose it in upper interface.

tfjs-backend-webgpu/src/flags_webgpu.ts
```diff
@@ -89,3 +89,6 @@ ENV.registerFlag('WEBGPU_CONV_SEPARATE_IM2COL_SHADER', () => false);
  * etc.). 'unary,conv2d' to print both unary and conv2d.
  */
 ENV.registerFlag('WEBGPU_PRINT_SHADER', () => '');
+
+/** Experimental flag, whether enter compile only phase. */
+ENV.registerFlag('ENGINE_COMPILE_ONLY', () => false);
```
Linchenn, line 94: In webgl backend, 'ENGINE_COMPILE_ONLY' has been used. We could either move 'ENGINE_COMPILE_ONLY' from webgl to core, or rename the flag here to 'WEBGPU_*'. 

We would unify the feature for webgl and webgpu eventually. Right now, since the feature usage are different between webgl and webgpu, (for example, webgpu does not need to call `getUniformLocations()`), we could have different names for two backends to avoid interfere.

What do you think? @qjia7 @axinging Feel free to say no, and that's just my thoughts.

qjia7: +1, @axinging Please change the flag name to `WEBGPU_ENGINE_COMPILE_ONLY`. Thanks.

Linchenn (approved):
LGTM, thanks!

axinging:
@Linchenn  @qjia7 Rename to WEBGPU_ENGINE_COMPILE_ONLY is done, PTAL

Linchenn:
Thanks, LGTM!

PR #7198, opened 2022-12-23
[webgpu] Let vectorized binary op return an overridable value on NaN

hujiajie:
This is like

  2c95a67 [webgpu] Further tweak vectorized NaN handling in binary ops

but for `valueForNaN`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7198)

### DIFF
--- a/tfjs-backend-webgpu/src/binary_op_webgpu.ts
+++ b/tfjs-backend-webgpu/src/binary_op_webgpu.ts
@@ -88,6 +88,7 @@ export class BinaryOpProgram implements WebGPUProgram {
     const opFnStr = `
     fn binaryOperation(a : ${dType}, b : ${dType}) -> ${dType} {
       let isNaN = false;
+      let valueForNaN = uniforms.NAN;
       {
         ${getBinaryOpString(this.op, this.isVec4)}
       }
--- a/tfjs-backend-webgpu/src/binary_op_util.ts
+++ b/tfjs-backend-webgpu/src/binary_op_util.ts
@@ -119,7 +119,6 @@ const MOD = `
 `;
 const MOD_VEC4 = `
   let isNaN = !vec4<bool>(b);
-  let valueForNaN = uniforms.NAN;
   var resultTemp = vec4<f32>(a % b);
   ${CHECK_NAN_SNIPPET_VEC4}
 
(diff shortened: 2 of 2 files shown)


tfjs-backend-webgpu/src/binary_op_webgpu.ts
```diff
@@ -88,6 +88,7 @@ export class BinaryOpProgram implements WebGPUProgram {
     const opFnStr = `
     fn binaryOperation(a : ${dType}, b : ${dType}) -> ${dType} {
       let isNaN = false;
+      let valueForNaN = uniforms.NAN;
```
gyagp, line 91: Currently for NOT_EQUAL_VEC4, valueForNaN is 1.0 instead of uniforms.NAN. Any specific reason?  @hujiajie @qjia7

hujiajie: The op returns booleans, even for NaN operands. I think that's by design.

(Well, the doc says nothing about this, it's just explicitly checked in the unit tests. Another mess is that currently this is defined in NOT_EQUAL, but not in EQUAL!)

gyagp: So do you think if uniforms.NAN is good enough for a default value? I'm not sure if we have more cases like (NOT_)EQUAL.

hujiajie: Yes, I think so.

gyagp (approved):

tfjs-backend-webgpu/src/binary_op_webgpu.ts
qjia7: I can't see any reason that we use `let valueForNaN = 1.0;` for `NOT_EQUAL_VEC4`. For WebGL, there is no NaN check for [`NotEqual`](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/kernels/NotEqual.ts#L23).
So my opinion is we directly use `uniforms.NAN` in `CHECK_NAN_SNIPPET_VEC4`. And don't need to define an additional variable `valueForNaN`.

hujiajie: 1. WGSL implementations may assume that NaNs and infinities are not present at runtime. In such an implementation, when an expression evaluation would produce an infinity or a NaN, an **indeterminate value** of the target type is produced instead. So, I feel explicit checks are necessary.
2. We want `i32(valueForNaN)` to evaluate to 1/true later, but `i32(uniforms.NAN)` is 0/false.

qjia7 (approved):

PR #7212, opened 2022-12-31
Adds function overloads to minimize.

MicahZoltu:
This will allow the caller to get correct type checking on the return values.

Note: This does not validate that the minimize function correctly returns the right type.  For example, if you changed `if (returnCost)` to `if (!returnCost)` you would not get a compiler error, so care must still be taken in ensuring this function does the right thing internally.

This change results in a number of tests failing to compile because they reuse variables (`let` instead of `const`).  I don't mind going through and fixing them all, but I wanted to get feedback on whether this is a desirable change first before I spend the time doing that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7212)

### DIFF
--- a/tfjs-core/src/optimizers/optimizer.ts
+++ b/tfjs-core/src/optimizers/optimizer.ts
@@ -56,6 +56,9 @@ export abstract class Optimizer extends Serializable {
    *
    * @doc {heading: 'Training', subheading: 'Optimizers'}
    */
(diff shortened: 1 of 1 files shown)


MicahZoltu:
Meh, nevermind.  Not worth doxxing myself just to provide a 3 line improvement.

PR #7257, opened 2023-01-09
[webgpu] Prepare for binary op restructuring

hujiajie:
The current code isn't great in that the vec4 shaders have diverged from the scalar ones more than necessary. The situation will be much improved by moving the NaN checking snippet in each op to a centralized place. Here is the common preparation work, so that following refactoring can be done on a per-op basis.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7257)

### DIFF
--- a/tfjs-backend-webgpu/src/binary_op_util.ts
+++ b/tfjs-backend-webgpu/src/binary_op_util.ts
@@ -42,8 +42,9 @@ export enum BinaryOpType {
 }
 
 const CHECK_NAN_SNIPPET = `
-  if (isnan(a)) { return a; }
-  if (isnan(b)) { return b; }
+  if isNaN { return valueForNaN; }
+  if isnan(a) { return valueForNaN; }
+  if isnan(b) { return valueForNaN; }
   `;
 
 const CHECK_NAN_SNIPPET_VEC4 = `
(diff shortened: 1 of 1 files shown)


tfjs-backend-webgpu/src/binary_op_util.ts
```diff
@@ -264,6 +318,10 @@ export function getBinaryOpString(
     case BinaryOpType.SUB:
       return SUB;
     default:
-      throw new Error(`BinaryType ${type} is not implemented!`);
+      // throw new Error(`BinaryType ${type} is not implemented!`);
```
gyagp, line 321: You may put NaN-insensitive ops in another do while, and keep the throw here.

tfjs-backend-webgpu/src/binary_op_util.ts
```diff
@@ -216,6 +217,59 @@ function getBinaryWithNanString(op: string, useVec4: boolean) {
...
+        doOpSnippet = POW_VEC4;
+        break;
+      default:
+        continue;
+    }
+    const checkNaNSnippet =
+        useVec4 ? CHECK_NAN_SNIPPET_VEC4 : CHECK_NAN_SNIPPET;
+    return `
+      ${doOpSnippet}
+      ${checkNaNSnippet}
```
gyagp, line 268: This introduces a duplicated check that is not necessary.

gyagp (approved):
Personally I don't think this PR is necessary, and we don't need to code with follow-up refactoring per op. 
Just approve here to unblock the overall progress.

tfjs-backend-webgpu/src/binary_op_util.ts
hujiajie: _"That's right. It's part of the plan." -- The Dark Forest_

tfjs-backend-webgpu/src/binary_op_util.ts
```diff
@@ -42,8 +42,9 @@ export enum BinaryOpType {
 }
 
 const CHECK_NAN_SNIPPET = `
-  if (isnan(a)) { return a; }
-  if (isnan(b)) { return b; }
+  if isNaN { return valueForNaN; }
+  if isnan(a) { return valueForNaN; }
+  if isnan(b) { return valueForNaN; }
```
qjia7, line 47: LGTM to only change above three lines to align with `CHECK_NAN_SNIPPET_VEC4`.
For other part, It seems that you are introducing more duplicated things. Like twice `checkNaNSnippet` check and some never running code. I understand you may clean the NaN ops to remove the extra `checkNaNSnippet` in following PRs. But I prefer you do it when everything is ready or do ops refactor first. I hope each PR is clean and don't bring some confusing things.

qjia7 (changes requested):

tfjs-backend-webgpu/src/binary_op_util.ts
hujiajie: Will do in the final cleanup, but not possible now.

```javascript
do {
  switch (type) {
    case BinaryOpType.FOOBAR:
      return FOOBAR;
    default:
      continue;
  }
  return `blah blah`;  // <-- still unreachable
} while (false);
```

tfjs-backend-webgpu/src/binary_op_util.ts
hujiajie: > LGTM to only change above three lines to align with CHECK_NAN_SNIPPET_VEC4.

Will do in the final cleanup, but not possible now. Current scalar implementations are lacking the definition of `resultTemp`.

PR #7277, opened 2023-01-17
[wasm] Add Diag kernel

chunnienc:
`tf.raw_ops.Diag` does not support boolean as the input, but tfjs-core has op tests for boolean tensors. So wasm kernel still supports boolean tensors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7277)

### DIFF
--- a/tfjs-backend-wasm/src/cc/kernels/Diag.cc
+++ b/tfjs-backend-wasm/src/cc/kernels/Diag.cc
@@ -1,0 +1,71 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


Linchenn (approved):
LGTM!

tfjs-backend-wasm/src/cc/kernels/Diag.cc
```diff
@@ -0,0 +1,71 @@
...
+
+namespace {
+
+template <typename T>
+inline void DiagImpl(const T* x_buf, int32_t x_size, T* out_buf) {
+  std::fill(out_buf, out_buf + x_size * x_size, 0);
+  for (int32_t i = 0; i < x_size; ++i) {
+    out_buf[x_size * i + i] = x_buf[i];
+  }
+}
```
mattsoulanille, line 39: This is a really nice implementation. For a while, I thought you'd have to do a lot of index calculations to account for diagonals of different dimensions, but that's not the case.

For some `k` dimensional tensor, 
```
output[i1,..., ik, i1,..., ik] = diagonal[i1, ..., ik]
```
But the index in the tensor's data array that `i1,..., ik` corresponds to is some unique number `d`. Then, we just have 
```
output[d, d] = diagonal[d]
```
for all `d` in the input, and this is just
```
out_buf[x_size * d + d] = x_buf[d]
```

mattsoulanille (approved):
LGTM

PR #7290, opened 2023-01-19
[wasm] Add BroadcastArgs kernel

chunnienc:
Considering the use case of this kernel and there is already an util for it, this kernel is executed completely in JS.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7290)


tfjs-backend-wasm/src/kernels/BroadcastArgs.ts
```diff
@@ -0,0 +1,43 @@
...
+  inputs: BroadcastArgsInputs,
+  backend: BackendWasm,
+}): TensorInfo {
+  const {inputs, backend} = args;
+  const {s0, s1} = inputs;
+
+  const s0Vals = backend.typedArrayFromHeap(s0);
+  const s1Vals = backend.typedArrayFromHeap(s1);
+
+  const broadcastShape = backend_util.assertAndGetBroadcastShape(
```
mattsoulanille, line 31: Maybe we should have a generic way to fall back to the CPU implementation for ops that aren't implemented in WASM. That would allow us to simply not implement this op, and it would use the fallback by default. IIRC we have something like this for the WebGL backend, but it's mostly used to avoid sending tensors to the GPU for really small ops.

tfjs-backend-wasm/src/kernels/BroadcastArgs.ts
```diff
@@ -0,0 +1,43 @@
...
+      Array.from(s0Vals), Array.from(s1Vals));
```
mattsoulanille, line 32: Nit: We can probably change the type of `assertAndGetBroadcastShape` to accept typed arrays.

mattsoulanille (approved):
LGTM with some minor nits, but see my comment about a more general solution.

tfjs-backend-wasm/src/kernels/BroadcastArgs.ts
chunnienc: It's not worthy - Only BroadcastArgs kernel (wasm, webgl, and webgpu) passes Int32Array to it, while most of the callers pass the shape array in number[]. Besides, not all the typed array can be the input, where the tensor type is guarded in core.

tfjs-backend-wasm/src/kernels/BroadcastArgs.ts
chunnienc: I'm thinking about the same thing. I may do it in a separate CL later.
This is not the first wasm kernel runs with cpu util/kernel function, and there was no general bridge for wasm-cpu kernel. I will revisit all the cpu function usages in wasm and find a general solution for them after I am familiar with more kernels.

Linchenn (approved):
LGTM

Issue #7289, opened 2023-01-19
Cases failed to run at https://js.tensorflow.org/api/latest/


axinging:
**Case 1: tf.MathBackendWebGL** 
Search "const customBackend = new tf.MathBackendWebGL(customCanvas);", scroll to the case, click run. It complains:
```
An error occured on line: 10
tf.MathBackendWebGL is not a constructor
```

**Case 2: createGPUBufferFromData** 
Search "createGPUBufferFromData", and run it under chrome canary(with  --enable-unsafe-webgpu), it complains:
```
An error occured on line: 17
Cannot read properties of undefined (reading 'createBuffer')
```

For case 2, it only runs in webgpu backend, possibly we need below two steps to fix it:
1, set webgpu backend:
```
await tf.setBackend('webgpu');
await tf.ready();
```
2, add tfjs webgpu backend:
Either support 'import '@tensorflow/tfjs-backend-webgpu';'
Or add: <script src="https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-webgpu"></script>

Linchenn:
@pyu10055 @mattsoulanille 

TLDR for the issue: A code snippet on our API website depends on tfjs-backend-webgpu, because the feature is for WebGPU backend. Since we haven't fully supported/announced WebGPU, we could either 1. hide the feature from the API website, or 2. Add `<script src="https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-webgpu"></script>`. What do you think?

mattsoulanille:
I think adding webgpu to the API site is a good idea, however, we should sign up for an [origin trial](https://developer.chrome.com/en/docs/web-platform/webgpu/#register-for-ot). Otherwise, it won't work, even if the browser supports it.

We may also want to note that these APIs require webgpu if we haven't already.

Linchenn:
@qjia7 @axinging 
FYI, will add the latest version of tfjs-backend-webgpu to our API website, https://github.com/tensorflow/tfjs-website/pull/488.

PR #7297, opened 2023-01-24
Export WebGL and CPU backends in union package

Linchenn:
In union package, we should expose all things exposed from tfjs-backend-webgl and tfjs-backend-cpu. Otherwise, users could not use `tf.MathBackendWebGL` to create WebGL backend instance from tfjs's union package, unless they import `tfjs-backend-webgl` directly.

In the next tfjs release, case1 of https://github.com/tensorflow/tfjs/issues/7289 would be fixed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7297)

### DIFF
--- a/tfjs/src/index_test.ts
+++ b/tfjs/src/index_test.ts
@@ -52,12 +52,10 @@ describe('Union package', () => {
   });
 
   it('has cpu backend', () => {
-    const backend = tf.findBackend('cpu');
-    expect(backend).not.toBeNull();
+    expect(tf.MathBackendCPU).not.toBeNull();
   });
 
   it('has webgl backend', () => {
-    const backend = tf.findBackend('webgl');
-    expect(backend).not.toBeNull();
+    expect(tf.MathBackendWebGL).not.toBeNull();
   });
 });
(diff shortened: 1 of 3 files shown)


tfjs/src/index_test.ts
```diff
@@ -52,12 +52,10 @@ describe('Union package', () => {
...
   it('has cpu backend', () => {
-    const backend = tf.findBackend('cpu');
-    expect(backend).not.toBeNull();
+    expect(tf.MathBackendCPU).not.toBeNull();
   });
 
   it('has webgl backend', () => {
-    const backend = tf.findBackend('webgl');
-    expect(backend).not.toBeNull();
+    expect(tf.MathBackendWebGL).not.toBeNull();
```
mattsoulanille, line 59: We should still run the original check, I think. It's checking if the cpu and webgl backends have been registered, whereas the new check is checking if their constructors are exported.

Linchenn: Make senses. Thank you!

mattsoulanille (approved):

pyu10055 (approved):

PR #7298, opened 2023-01-25
Fix comment in tile.ts

sonmh79:
Fixed comment to show different expression from examples

Hi, I visited [api docs](https://js.tensorflow.org/api/4.2.0/) to learn about tensorflow.js and found a bit ackward comment in tile part. I think it's not the way you intend so I tried to fix it in a better way !

Thanks :)

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7298)


mattsoulanille (approved):
Looks good to me. Thanks for the fix!

chunnienc (approved):
LGTM. Thanks for the contribution!

PR #7306, opened 2023-01-26
[wasm] Add ArgMin

chunnienc:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7306)


tfjs-backend-wasm/src/kernels/ArgMin.ts
```diff
@@ -0,0 +1,81 @@
...
+  const innerSize = input.shape[axes[0]];
+  wasmFunc(inputId, CppDType[input.dtype], outerSize, innerSize, outId);
+
+  if (inputWasTransposed) {
+    // dispose of the transposed tensor.
+    backend.disposeData(transposed.dataId);
+  }
+
+  return out;
+}
```
mattsoulanille, line 74: Nit: This looks similar to ArgMax's implementation. Can we factor it out? If other ops have patterns like this too, then maybe this is better to do in a different PR.

tfjs-backend-wasm/src/cc/kernels/ArgMin.cc
```diff
@@ -0,0 +1,76 @@
...
+
+namespace {
+
+template <typename T>
+void ArgMinImpl(const T* x, const int outer_size, const int inner_size,
+                int32_t* out_buf) {
+  for (int i = 0; i < outer_size; ++i) {
+    const int offset = i * inner_size;
+    T min_value = x[offset];
+    int min_index = 0;
```
mattsoulanille, line 34: Maybe we can factor out some of this array indexing code. It seems like several ops act on just the last dimension of the tensor. Could we have a function that applies another function to the last dimension (or maybe a more general one that works on the last n dimensions)?

Then, we could just pass it this function.

```c
void ArgMaxInner(const T* x, const int size, int32_t* out_buf ) {
  T min_value = x[0];
  int min_index = 0;
  for (int i = 1; i < size; i++) {
    t Value = x[i];
    if (value < min_value) {
      min_value = value;
      min_index = i;
  }
  out_buf[i] = min_index;
}
```

Probably better for a later PR, though.

mattsoulanille (approved):
LGTM. I have some ideas about how we might simplify some of these kernels that could be applied to future PRs, but this PR looks good to me.

pyu10055 (approved):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7306)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @chunnienc)

mattsoulanille:
@chunnienc Feel free to merge this whenever you like.

chunnienc:
> @chunnienc Feel free to merge this whenever you like.

I'm working on the common utils you mentioned. Will append to this PR and request your review again once ready.

tfjs-backend-wasm/src/cc/kernels/ArgMin.cc
chunnienc: Added

mattsoulanille (approved):
LGTM

PR #7322, opened 2023-01-28
Propagate errors from indexedDB database.put

mattsoulanille:
database.put can sometimes throw an error if the data being stored is too large. Propagate that error to the promise instead of letting it escape the callback.

### DIFF
--- a/tfjs-core/src/io/indexed_db.ts
+++ b/tfjs-core/src/io/indexed_db.ts
@@ -162,18 +162,31 @@ export class BrowserIndexedDB implements IOHandler {
           // First, put ModelArtifactsInfo into info store.
           const infoTx = db.transaction(INFO_STORE_NAME, 'readwrite');
           let infoStore = infoTx.objectStore(INFO_STORE_NAME);
-          const putInfoRequest =
+          let putInfoRequest: IDBRequest<IDBValidKey>;
+          try {
+            putInfoRequest =
               infoStore.put({modelPath: this.modelPath, modelArtifactsInfo});
+          } catch (error) {
+            reject(error);
+            return;
+          }
           let modelTx: IDBTransaction;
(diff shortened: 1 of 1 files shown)


mattsoulanille:
I tried testing this by mocking indexedDB, but it returns native object, which I can't spy on. The correct approach would be to write a fake, but that would be a lot of work. Let me know if you want me to do that.

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7322)* status: 0 of 1 approvals obtained (waiting on @Linchenn and @mattsoulanille)

___
*[`tfjs-core/src/io/indexed_db.ts` line 170 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7322#-NMtSCa03zs-8KKdyxfg:-NMtSCa03zs-8KKdyxfh:b-lk4els) ([raw file](https://github.com/tensorflow/tfjs/blob/f7495bd5c5c72b0545147a2bf9a81a756b84a8fc/tfjs-core/src/io/indexed_db.ts#L170)):*
> ```XML
>               infoStore.put({modelPath: this.modelPath, modelArtifactsInfo});
>           } catch (error) {
>             reject(error);
> ```

should it be, also for the other change
`return reject(error);`

Linchenn (approved):
LGTM, thanks!

mattsoulanille:
@pyu10055 Please take a look. Thanks!

pyu10055 (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7322)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @mattsoulanille)

PR #7326, opened 2023-01-31
[Draft] Add dataCached and readCached

mattsoulanille:
Add two new methods, `dataCached` (to backends) and `readCached` (to tensors), to enable synchronously reading a tensor's data if it's already cached on the CPU. If the data is not cached on the CPU, these methods return `undefined`. `dataCached` is necessary for ops like `reshape` that need a tensor's value immediately when they are called, and it avoids breaking existing models that pass constants to `reshape`.

For example, when `reshape` is executed in a graph model and it has a tensor value as the new shape, it does the following:
1. Calls `dataCached()` on the shape tensor and, if the value is available, returns the reshaped result immediately.
2.  If `dataCached()` returns `undefined`, then it calls `data()` on the shape tensor and returns the reshaped result in a promise.

Due to the synchronous check in (1), existing models that pass a constant tensor as the shape to `reshape` will still work synchronously.

### DIFF
--- a/tfjs-backend-cpu/package.json
+++ b/tfjs-backend-cpu/package.json
@@ -42,14 +42,6 @@
     "util": false,
     "crypto": false
   },
-  "sideEffects": [
-    "./dist/register_all_kernels.js",
-    "./dist/base.js",
-    "./dist/index.js",
-    "./src/register_all_kernels.mjs",
-    "./src/base.mjs",
-    "./src/index.mjs"
-  ],
   "resolutions": {
     "minimist": "1.2.6"
   }
--- a/tfjs-backend-cpu/src/backend_cpu.ts
+++ b/tfjs-backend-cpu/src/backend_cpu.ts
@@ -142,6 +142,7 @@ export class MathBackendCPU extends KernelBackend {
     return util.convertBackendValuesAndArrayBuffer(
         this.data.get(dataId).values, dtype);
   }
+  override readCached = this.readSync;
 
   bufferSync<R extends Rank, D extends DataType>(t: TensorInfo):
       TensorBuffer<R, D> {
--- a/tfjs-backend-wasm/src/backend_wasm.ts
+++ b/tfjs-backend-wasm/src/backend_wasm.ts
@@ -132,6 +132,8 @@ export class BackendWasm extends KernelBackend {
     return typedArrayFromBuffer(bytes.buffer, dtype);
   }
 
+  override readCached = this.readSync;
+
   /**
    * Dispose the memory if the dataId has 0 refCount. Return true if the memory
    * is released, false otherwise.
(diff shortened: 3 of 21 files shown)


mattsoulanille:
This also implements https://github.com/tensorflow/tfjs/pull/6309 . Otherwise, users will be unable to use models with async reshape since they are not detected as async by static analysis.

mattsoulanille:
@pyu10055 Do you know of any other ops that need immediate access to data stored in tensors (like reshape)  that I could apply this to? Thanks!

pyu10055 (changes requested):
> Reviewed 2 of 7 files at r1, 17 of 21 files at r2, 2 of 2 files at r3, 2 of 2 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7326)* status: 0 of 1 approvals obtained (waiting on @Linchenn and @mattsoulanille)

___
*[`tfjs-converter/src/operations/types.ts` line 89 at r6](https://reviewable.io/reviews/tensorflow/tfjs/7326#-NNIHdZG373f07h0-ujc:-NNIHdZG373f07h0-ujd:bnxcqmu) ([raw file](https://github.com/tensorflow/tfjs/blob/d15c670e814795ec6deecd626192458953e63ac3/tfjs-converter/src/operations/types.ts#L89)):*
> ```XML
> }
> 
> export interface InternalOpMaybeAsyncExecutor {
> ```

These three executor interfaces can be merge into the new one, please change this in another PR, since it will affect all executor code.
___
*[`tfjs-converter/src/operations/executors/transformation_executor.ts` line 59 at r6](https://reviewable.io/reviews/tensorflow/tfjs/7326#-NNIO6bq7e2eO6R-_Wpd:-NNIO6bq7e2eO6R-_Wpe:b5whxhv) ([raw file](https://github.com/tensorflow/tfjs/blob/d15c670e814795ec6deecd626192458953e63ac3/tfjs-converter/src/operations/executors/transformation_executor.ts#L59)):*
> ```XML
>         }
>         case 'Reshape': {
>           const shape = getParamValueOrPromise('shape', node, tensorMap,
> ```

There are other ops that convert the input tensor/tensors into a javascript array or primitives. You can look them up in the op_list json files and update them in a separate PR.
___
*[`tfjs-core/src/tensor.ts` line 417 at r6](https://reviewable.io/reviews/tensorflow/tfjs/7326#-NNIEBPDBNVcs9w-FVt-:-NNIEBPDBNVcs9w-FVt0:b-yv4hgt) ([raw file](https://github.com/tensorflow/tfjs/blob/d15c670e814795ec6deecd626192458953e63ac3/tfjs-core/src/tensor.ts#L417)):*
> ```XML
> 
>   /**
>    * Synchronously returns the values from the `tf.Tensor` if they cached on.
> ```

remove the `.` at the end

PR #7327, opened 2023-01-31
[wasm] Add Dilation2D

chunnienc:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7327)

### DIFF
--- a/tfjs-backend-wasm/src/kernels/Dilation2D.ts
+++ b/tfjs-backend-wasm/src/kernels/Dilation2D.ts
@@ -1,0 +1,94 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


mattsoulanille (approved):
LGTM!

tfjs-backend-wasm/src/kernels/Dilation2D.ts
```diff
@@ -0,0 +1,103 @@
...
+export function dilation2D(args: {
+  inputs: Dilation2DInputs,
+  attrs: Dilation2DAttrs,
+  backend: BackendWasm,
+}): TensorInfo {
+  const {inputs, backend, attrs} = args;
+  const {x, filter} = inputs;
+  const {strides, pad, dilations} = attrs;
+
+  if (x.dtype !== filter.dtype) {
```
Linchenn, line 62: Could we move this check to core? As I know, we typically do users' input check in core; otherwise the backends' behaviors could be different.

chunnienc: This is intentional: our cpu and webgl backend can take different types and do the type cast automatically, while wasm backend is very hard to do that. Therefore this check is just here to make the wasm implementation easier.

The TF Dilation2D op also has this constraint that x and filter must have the same type, so it should be fine in terms of correctness. However, it will be a breaking change to TFJS if we move this check to core level.

Linchenn: Thanks for explanation!

Linchenn (approved):
LGTM, thanks!

PR #7329, opened 2023-01-31
Parallelize e2e tests

mattsoulanille:
Break e2e tests up into multiple steps to run it more parallelly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7329)


e2e/scripts/test-ci.sh
```diff
@@ -42,27 +39,3 @@ cd ..
...
-yarn build
-cd ..
-
-node ../scripts/run_flaky.js "yarn run-browserstack --browsers=win_10_chrome --tags '$TAGS'"
-
-# Test script tag bundles
-node ../scripts/run_flaky.js "karma start ./script_tag_tests/tfjs/karma.conf.js --browserstack --browsers=bs_chrome_mac --testBundle tf.min.js"
-
-# Additional tests to run in nightly only.
-if [[ "$NIGHTLY" = true || "$RELEASE" = true ]]; then
```
Linchenn, line 58: Will this PR affect tests for `"$RELEASE" = true`? Because I saw you move these tests to yml with `nightlyOnly=true`, but I didn't see you add these tests for release.

mattsoulanille: Oh, good catch. I'll fix that.

mattsoulanille: Looks like this may be a deeper issue than I initially thought. `cloudbuild-release.yml` runs `e2e/scripts/release-e2e.sh` which runs `yarn test-ci`, expecting it to run the CI tests. However, with this PR, `yarn test-ci` no longer runs all the CI tests (some of them are written directly in the cloudbuild file). I'll need to think of a better way of doing this.

mattsoulanille:
Closed as this has been accomplished with gnu parallel instead.

Issue #7343, opened 2023-02-06
Can't run tflite backend in TensorFlow.js Model Benchmark


huningxin:
<em>Please make sure that this is a bug. As per our
[GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md),
we only address code/doc bugs, performance issues, feature requests and
build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): No
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Windows 10
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device: N/A
- TensorFlow.js installed from (npm or script link):  tf.js model benchmark
- TensorFlow.js version (use command below): 4.2.0
- Browser version: 109.0.5414.120 (Official Build) (64-bit) (cohort: Stable) 
- Tensorflow.js Converter Version: N/A

**Describe the current behavior**
1. Launch TF.js model benchmark: https://tfjs-benchmarks.web.app/local-benchmark/
2. Set backend to `tflite`
3. Click “Run benchmark"
4. The site reports error "Error: The texture is invalid. Also, please make sure the texture and the TFJS WebGL backend are using the same canvas. If you want to use your own custom canvas, you have to create and use the custom TFJS WebGL backend created from the canvas through 'new tf.MathBackendWebGL(customCanvas)'."

**Describe the expected behavior**

The benchmark should run workload by tflite backend and report performance numbers.

**Standalone code to reproduce the issue**
Provide a reproducible test case that is the bare minimum necessary to generate
the problem. If possible, please share a link to Colab/CodePen/any notebook.

**Other info / logs** Include any logs or source code that would be helpful to
diagnose the problem. If including tracebacks, please include the full
traceback. Large logs and files should be attached.

Console log:
tf-backend-webgl.js:7589 Uncaught (in promise) Error: The texture is invalid. Also, please make sure the texture and the TFJS WebGL backend are using the same canvas. If you want to use your own custom canvas, you have to create and use the custom TFJS WebGL backend created from the canvas through 'new tf.MathBackendWebGL(customCanvas)'.
    at MathBackendWebGL.createTensorFromGPUData (tf-backend-webgl.js:7589:23)
    at makeTensor (tf-core.js:5652:35)
    at Object.tensor (tf-core.js:5872:16)
[...]

PR #7344, opened 2023-02-06
SharedArrayBuffer is not GPU data

axinging:
Bug: https://github.com/tensorflow/tfjs/issues/7343

TODO: Add test case for ShardArrayBuffer.
```
describeWithFlags('tensor from ShardArrayBuffer', ALL_ENVS, () => {
  it('float32 tensor from ShardArrayBuffer', async () => {
    let length = 4;
    // ReferenceError: SharedArrayBuffer is not defined.
    let sharedFloats = new Float32Array(new SharedArrayBuffer(length * 4));
    let expected = [];
    for (let i = 0; i < length; i++) {
      sharedFloats[i] = i;
      expected[i] = i;
    }
    const a = tf.tensor(sharedFloats);
    tf.test_util.expectArraysEqual(await a.data(), expected);
  });
});

```

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7344)

### DIFF
--- a/tfjs-core/src/ops/tensor_ops_util.ts
+++ b/tfjs-core/src/ops/tensor_ops_util.ts
@@ -17,7 +17,7 @@
 
 import {ENGINE} from '../engine';
 import {Tensor} from '../tensor';
-import {TensorLike, TypedArray, WebGLData, WebGPUData} from '../types';
+import {getGPUDataType, GPUDataType, TensorLike, TypedArray, WebGLData, WebGPUData} from '../types';
 import {DataType} from '../types';
 import {assert, assertNonNegativeIntegerDimensions, flatten, inferDtype, isTypedArray, sizeFromShape, toTypedArray} from '../util';
 
@@ -33,9 +33,7 @@ export function makeTensor(
         `Please use tf.complex(real, imag).`);
   }
 
-  if (typeof values === 'object' &&
-      ('texture' in values ||
-       ('buffer' in values && !(values.buffer instanceof ArrayBuffer)))) {
+  if (getGPUDataType(values) !== GPUDataType.Unknown) {
     if (dtype !== 'float32' && dtype !== 'int32') {
       throw new Error(
           `Creating tensor from GPU data only supports ` +
--- a/tfjs-core/src/ops/tensor_ops_util.ts
+++ b/tfjs-core/src/ops/tensor_ops_util.ts
@@ -17,7 +17,7 @@
 
 import {ENGINE} from '../engine';
 import {Tensor} from '../tensor';
-import {getGPUDataType, GPUDataType, TensorLike, TypedArray, WebGLData, WebGPUData} from '../types';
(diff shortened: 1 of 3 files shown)


tfjs-core/src/types.ts
```diff
@@ -196,3 +196,26 @@ export interface WebGPUData {
...
+  if (typeof values !== 'object') {
+    // Unknown data type.
+    return GPUDataType.Unknown;
+  }
+  if ('texture' in values) {
+    return GPUDataType.WebGL;
+  } else if (
+      'buffer' in values && !(values.buffer instanceof ArrayBuffer) &&
+      !(values.buffer instanceof SharedArrayBuffer)) {
+    return GPUDataType.WebGPU;
```
Linchenn, line 217: Great catch, thanks you!

I am considering, rather than excluding cases, what about we do checks like:?
```js
if ('texture' in values && values.texture instanceof WebGLTexture) {
  return GPUDataType.WebGL;
} else if ( 'buffer' in values && values.buffer instanceof GPUBuffer) {
...
}
```

axinging: @Linchenn, done. PTAL.

BTW, I am trying to add a test case 'tensor from ShardArrayBuffer', however, chrome complains 'ReferenceError: SharedArrayBuffer is not defined.' It seems 'yarn karma start' doesn't support SharedArrayBuffer .  This can be workaround by https://github.com/Honry/webnn-samples/blob/tflite-deeplab/server.js. 

@qjia7 @xhcao @haoyunfeix @gyagp , PTAL

hujiajie: What about [this idiom](https://github.com/tensorflow/tfjs/blob/50f2a824dd901a4b8a4db883fb8f88d9525adde6/tfjs-core/src/ops/browser.ts#L86) (though I don't feel parenthesis is needed around the operand of `typeof`)?

axinging: Thanks, @hujiajie, it seems the bots turrn green.

@Linchenn , Could you please take another look?

Linchenn (approved):
Nice workaround. LGTM, thanks!

tfjs-core/src/ops/tensor_ops_util.ts
```diff
@@ -33,9 +33,7 @@ export function makeTensor(
         `Please use tf.complex(real, imag).`);
   }
 
-  if (typeof values === 'object' &&
-      ('texture' in values ||
-       ('buffer' in values && !(values.buffer instanceof ArrayBuffer)))) {
+  if (getGPUDataType(values) !== GPUDataType.Unknown) {
```
mattsoulanille, line 36:
```suggestion
  if (isWebGPUData(values) || isWebGLData(values)) {
```

mattsoulanille (changes requested):
Sorry for the late review!

mattsoulanille (approved):
LGTM

PR #7345, opened 2023-02-06
BaseRandomLayer Abstract Layer and RandomWidth Preprocessing Layer

RWallie:
BaseRandomLayer Abstract Layer and RandomWidth Preprocessing Layer 
Co-authored-by:
Natalie Umanzor (@numanzor) [umanzor2949@gmail.com](mailto:umanzor2949@gmail.com)
Silvia Kocsis (@Silvia42) [silvia.kocsis@gmail.com](mailto:silvia.kocsis@gmail.com)
Ryan Wallace (@RWallie) [ryanwallace1396@gmail.com](mailto:ryanwallace1396@gmail.com)

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7345)


tfjs-layers/src/layers/preprocessing/random_width.ts
```diff
@@ -0,0 +1,160 @@
...
+        }
+        this.adjustedWidth = this.widthFactor.dataSync()[0] * imgWidth;
+        this.adjustedWidth = Math.round(this.adjustedWidth);
+        const size: [number, number] = [this.imgHeight, this.adjustedWidth];
+        if (this.interpolation === 'bilinear') {
+          return image.resizeBilinear(inputs, size);
+        } else if (this.interpolation === 'nearest') {
+          const output = image.resizeNearestNeighbor(inputs, size);
+          return output;
+        } else {
```
mattsoulanille, line 152: Minor nit: This can be a switch statement.

Alternatively, `resizeBilinear` and `resizeNearestNeighbor` are called with the same signature here. You may be able to store the function itself in `this.interpolation` instead of just the string corresponding to it (although maybe `this.interpolation` is public and should be a string).

tfjs-layers/src/layers/preprocessing/random_width.ts
```diff
@@ -0,0 +1,160 @@
...
+ * tf methods unimplemented in tfjs: 'bicubic', 'area', 'lanczos3', 'lanczos5',
+ *                                   'gaussian', 'mitchellcubic'
+ *
+ */
+
+const INTERPOLATION_KEYS = ['bilinear', 'nearest'] as const;
+export const INTERPOLATION_METHODS = new Set(INTERPOLATION_KEYS);
+type InterpolationType = typeof INTERPOLATION_KEYS[number];
+
+export class RandomWidth extends BaseRandomLayer {
```
mattsoulanille, line 46: [It looks like the keras class extends `BaseImageAugmentationLayer`](https://github.com/keras-team/keras/blob/v2.11.0/keras/layers/preprocessing/image_preprocessing.py#L1898) that (I think) makes some of the indexing calculations generic among layers like this one. Should we do the same here, or is it not applicable in our implementation?

tfjs-layers/src/layers/preprocessing/random_width.ts
```diff
@@ -0,0 +1,160 @@
...
+            'float32', this.seed
+          );
+        } else {
+          this.widthFactor = randomUniform([1],
+            (1.0 + this.widthLower), (1.0 + this.widthUpper)
+          );
+        }
+        this.adjustedWidth = this.widthFactor.dataSync()[0] * imgWidth;
+        this.adjustedWidth = Math.round(this.adjustedWidth);
+        const size: [number, number] = [this.imgHeight, this.adjustedWidth];
```
mattsoulanille, line 146: Tiny nit: Using `as const` makes TypeScript correctly infer this as `const [number, number]` instead of `number[]`.
```suggestion
        const size = [this.imgHeight, this.adjustedWidth] as const;
```

tfjs-layers/src/layers/preprocessing/random_width_test.ts
```diff
@@ -0,0 +1,161 @@
...
+
+import { Tensor, reshape, range, image, Rank, zeros, randomUniform, tensor } from '@tensorflow/tfjs-core';
+import { describeMathCPUAndGPU, expectTensorsClose } from '../../utils/test_utils';
+
+import { RandomWidth, RandomWidthArgs } from './random_width';
+
+describeMathCPUAndGPU('RandomWidth Layer', () => {
+  it('Returns correct, randomly scaled width of Rank 3 Tensor', () => {
+    const rangeTensor = range(0, 16);
+    const inputTensor = reshape(rangeTensor, [4,4,1]);
```
mattsoulanille, line 23: Minor nit since `rangeTensor` only seems to be used once.
```suggestion
    const inputTensor = range(0, 16).reshape([4, 4, 1]);
```

tfjs-layers/src/layers/preprocessing/random_width.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  override call(inputs: Tensor<Rank.R3>|Tensor<Rank.R4>,
+    kwargs: Kwargs): Tensor[]|Tensor {
+
+    return tidy(() => {
+        const input = getExactlyOneTensor(inputs);
+        const inputShape = input.shape;
+        this.imgHeight = inputShape[inputShape.length - 3];
+        const imgWidth = inputShape[inputShape.length - 2];
```
mattsoulanille, line 130: Minor nit: You can use `Array.at()` to get python-style indexing from the end of the array.
```suggestion
        this.imgHeight = inputShape.at(-3);
        const imgWidth = inputShape.at(-2);
```

tfjs-layers/src/engine/base_random_layer.ts
```diff
@@ -0,0 +1,52 @@
...
+
+import { LayerArgs, Layer } from './topology';
+import { randomGamma, randomNormal} from '@tensorflow/tfjs-core';
+import { randomStandardNormal, randomUniform } from '@tensorflow/tfjs-core';
+import { ValueError } from '../errors';
+
+export declare interface BaseRandomLayerArgs extends LayerArgs {}
+
+export type RNGTypes = {
+  [key: string]: Function;
```
mattsoulanille, line 19: `Function` is a very wide type here. Can it be narrowed to be more specific?

tfjs-layers/src/engine/base_random_layer.ts
```diff
@@ -0,0 +1,52 @@
...
+  static className = 'RandomWidth';
+  randomGenerator: Function;
+  private rngType: string;
+
+  private readonly rngTypes: RNGTypes = {
+    gamma: randomGamma,
+    normal: randomNormal,
+    standardNormal: randomStandardNormal,
+    uniform: randomUniform
+  };
```
mattsoulanille, line 34: If this constant is not specific to this class, it can probably be moved outside of the class. If that's the case, you might also be able to base the `RNGTypes` type off of this constant.
```ts
const RNG_TYPES = {...};
type RNGTypes = typeof RNG_TYPES;
```

tfjs-layers/src/engine/base_random_layer.ts
```diff
@@ -0,0 +1,52 @@
...
+export type RNGTypes = {
+  [key: string]: Function;
+};
+
+export abstract class BaseRandomLayer extends Layer {
+  // A layer handle the random number creation and savemodel behavior.
+  /** @nocollapse */
+  static className = 'RandomWidth';
+  randomGenerator: Function;
+  private rngType: string;
```
mattsoulanille, line 27: This can reference RNGTypes to be more specific (I know they're both the `string` type right now, but that may change if the other change I mentioned can be applied).
```ts
private rngType: keyof RNGTypes
```

mattsoulanille (changes requested):
Thanks for the PR! I have a few questions and changes to suggest.

RWallie:
Thank you @mattsoulanille for reviewing our pull request. We are getting right on the changes!

tfjs-layers/src/layers/preprocessing/random_width.ts
RWallie: I believe the `BaseImageAugmentationLayer` was removed a week and a half ago. [The preprocessing layers extend `BaseRandomLayer`](https://github.com/keras-team/keras/blob/0046e0dcd5b7cac5bd7cc9cb5276f667b80229a8/keras/layers/preprocessing/image_preprocessing.py#L1637) and some of the methodology was moved directly in the preprocessing layers.

tfjs-layers/src/layers/preprocessing/random_width.ts
RWallie: resizeBilinear and resizeNearestNeighbor expect a mutable array of two numbers. using  `as const` assigns the readonly type

tfjs-layers/src/layers/preprocessing/random_width.ts
mattsoulanille: Thanks for the clarification. In that case, this looks good to me as-is.

RWallie:
We changed the abstract class layer `BaseRandomLayer` to map more directly to keras. We created a `RandomGenerator` class that contains the incrementing of seeds to handle pseudorandomness, as you mentioned. We also stored the random distribution functions as properties on `RandomGenerator` as they are [stored as methods in keras](https://github.com/keras-team/keras/blob/0046e0dcd5b7cac5bd7cc9cb5276f667b80229a8/keras/backend.py#L2073). An instance of `RandomGenerator` is stored as a property on BaseRandomLayer. We use this instance to directly access `randomUniform` in the `RandomWidth` preprocessing layer, [as does keras](https://github.com/keras-team/keras/blob/master/keras/layers/preprocessing/image_preprocessing.py#L1714). 

We previously had rng_types handling which random distribution would be selected in the preprocessing layers. We removed these - how we had them defined didn't appropriately relate to keras. As you said, in keras, the rng_types were there to determine whether the op was stateful or stateless - [we believe all the random distribution functions to be stateless](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/rand_util.ts#:~:text=if%20(seed,%7D), so we didn't incorporate rng_types when creating an instance of `RandomGenerator` in `BaseRandomLayer`. 

Some conditionals were removed / rewritten based on your suggestions.

Thanks again @mattsoulanille

mattsoulanille:
Sorry, I think I gave some bad suggestions in my review.

```
Safari 11.1.2 (Mac OS 10.13.6) RandomWidth Layer webgl1 {"WEBGL_VERSION":1,"WEBGL_CPU_FORWARD":false,"WEBGL_SIZE_UPLOAD_UNIFORM":0} Returns correct, randomly scaled width of Rank 3 Tensor FAILED
	TypeError: inputShape.at is not a function. (In 'inputShape.at(-3)', 'inputShape.at' is undefined) in http://bs-local.com:9876/context.html (line 39696)
	<Jasmine>
```

Looks like this Safari version does not support `Array.at` and we're not polyfilling it. You should revert back to what you had before (`someArray[someArray.length - someNumber]`).

You should be able to access the build logs if you join the [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list

Thanks for making the changes! I'll take a look and review them tomorrow.

RWallie:
Hey @mattsoulanille hope you are doing well. Just wanted to check in and see how things are looking with our latest changes to the PR. Thanks again!

mattsoulanille (changes requested):
Sorry I didn't get to this sooner.

RWallie:
No worries @mattsoulanille ! Thanks for the review!

chunnienc (changes requested):
I will take a closer look after @mattsoulanille 's comments and suggestions are resolved.

RWallie:
We cleaned up the `RandomGenerator` class and changed it to `RandomSeed` - It only handles the pseudorandomness now and the random sampler functions will be imported directly into the random preprocessing layers, as we did in `random_width.ts`.

We took the `seed` property off of `RandomWidth`, as you mentioned it's not necessary and is duplicating state. We made the `currentSeed` inside of `RandomSeed` public so it would be accessible when calling `randomUniform` inside the preprocessing layer.

We set up a `getConfig()` inside of `BaseRandomLayer` that saves the `seed`



Thank you @mattsoulanille @chunnienc !

RWallie:
Hey @mattsoulanille wanted to check in on the PR to see if you had any updates. Thanks!

tfjs-layers/src/layers/preprocessing/random_width.ts
```diff
@@ -123,32 +117,31 @@ export class RandomWidth extends BaseRandomLayer {
...
+      const input = getExactlyOneTensor(inputs);
+      this.imgHeight = input.shape[input.shape.length - 3];
+      const imgWidth = input.shape[input.shape.length - 2];
+
+      this.widthFactor = randomUniform([1],
+        (1.0 + this.widthLower), (1.0 + this.widthUpper),
+        'float32', this.randomGenerator.currentSeed
+      );
+
+      this.randomGenerator.next();
```
mattsoulanille, line 129: `this.randomGenerator.next()` increments the seed and then returns the new seed, so you can use its return value instead of incrementing it separately.

```suggestion
        'float32', this.randomGenerator.next());
```

If you want to use the current seed before incrementing it, change the implementation of `RandomGenerator.next` to return `seed++` (increments after getting the value) instead of `++seed` (increments before getting the value).

tfjs-layers/src/engine/base_random_layer.ts
```diff
@@ -19,10 +20,22 @@ export abstract class BaseRandomLayer extends Layer {
...
   static className = 'BaseRandomLayer';
-  protected randomGenerator: RandomGenerator;
+  protected randomGenerator: RandomSeed;
+  private seed?: number;
 
   constructor(args: BaseRandomLayerArgs) {
     super(args);
-    this.randomGenerator = new RandomGenerator(args.seed);
+    this.seed = args.seed;
+    this.randomGenerator = new RandomSeed(this.seed);
```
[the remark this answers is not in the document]
RWallie: The utility classes for the random sampler classes use `Math.random()` to assign the seed if none exists. [Here's the example from `randomUniform`'s `UniformRandom` utility class](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/rand_util.ts#L189)

RWallie:
Hey @mattsoulanille we removed the seed from `BaseRandomLayers` so we won't be duplicating state. We handled an undefined seed inside of `RandomSeed` . When calling `RandomSeed`'s next method, we return `undefined` if the `seed` is `undefined`. [Randomness gets handled inside of the utility classes for each random sampler function](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/rand_util.ts#L189) if the `seed` is `null` or `undefined` and generates a `seed` using `Math.random()`.

RWallie:
Hey @mattsoulanille - wanted to reach out to see how things are looking with our latest PR. Thank you!

tfjs-layers/src/backend/random_seed.ts
```diff
@@ -16,11 +16,14 @@
...
+  constructor(seed: number | undefined) { 
+    this.seed = seed; 
   }
-  next() { 
-    ++this.currentSeed; 
+  next(): number | undefined { 
+    if (this.seed === undefined) {
+      return undefined;
+    }
+    return this.seed++;
```
mattsoulanille, line 27: Note to other reviewer: When no seed is specified, the keras implementation does not choose a random seed. It just generates random values. Saving it yields something like this, with the seed set to `Null`:
```
{'name': 'random_width_3', 'trainable': True, 'dtype': 'float32', 'factor': (-0.2, 0.3), 'interpolation': 'bilinear', 'seed': None}
```
This is truly random, and there is no reproducibility here.

On the other hand, when the user sets the seed, it gets saved to the config.
```
{'name': 'random_width_2', 'trainable': True, 'dtype': 'float32', 'factor': (-0.2, 0.3), 'interpolation': 'bilinear', 'seed': 1}
```

mattsoulanille (approved):
LGTM. I'm sorry it took me so long to get to this.

AdamLang96:
@mattsoulanille Thanks for all the help on this!

mattsoulanille:
@chunnienc Please take a look when you get a chance. Thanks!

chunnienc (approved):

PR #7346, opened 2023-02-06
[wasm] Add ResizeBilinearGrad and ResizeNearestNeighborGrad kernels

chunnienc:
@mattsoulanille and reviewers: This compiler explorer example https://godbolt.org/z/zGr56TWeT shows that the shape (helper) class and lambda argument are zero-cost in our usage as we discussed. You can see that the generated assembly codes for `Operator1` (with these features) and `Operator2` (native implementation) are the same.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7346)


mattsoulanille (approved):
LGTM. The shape abstraction makes indexing much easier to read!

chunnienc:
> Thank you, nice Shape class.
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7346#-:-NNco3LE4SJxCd2ch-wp:ba2i43k)_ status: :shipit: complete! 2 of 1 approvals obtained (waiting on @chunnienc)
> 
> _[`tfjs-backend-wasm/src/cc/shape.h` line 57 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7346#-NNco-dC7eXMAYjy_uiJ:-NNco-dC7eXMAYjy_uiK:bu5d7uo) ([raw file](https://github.com/tensorflow/tfjs/blob/255539732117eca4d52b8b5290156e7312b1e8ff/tfjs-backend-wasm/src/cc/shape.h#L57)):_
> 
> > ```objc
> >   }
> > 
> >   inline T& operator[](size_t i) { return data_[i]; }
> > ```
> 
> boundary check on `i`?

The operator[] does not do boundary checking, while the `.at(i)` method checks the boundary and can throw an `std::out_of_range` error. This is a convention in C++ STL and almost all containers (std::vector, std::array, etc) implement this behavior. That's also why I expose two methods for different use cases.
Ref:
https://en.cppreference.com/w/cpp/container/array/operator_at
https://en.cppreference.com/w/cpp/container/array/at

PR #7371, opened 2023-02-13
webgpu: Support 2x4 output for Conv2DBackpropInput

qjia7:
Conv2DBackpropInput in ArPortraitDepth
26.32ms -> 14.17ms on CFL
6.91ms -> 4.70ms on TGL
To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7371)

PR #7386, opened 2023-02-16
Fine-tune packed BackpropInput

Linchenn:
Originally, we use `idyCVal` and `idyCVal2` (could only be 0.0 or 1.0) to indicate if `result.xy` and `result.zw` are valid and `result.xy * idyCVal` and `result.zw * idyCVal2` could return the results to avoid checking through if-branches.

If stride is  1, both `idyCVal` and `idyCVal2` are always `1.0` and the original solution performs well. However, if stride is 2, whenever, only one of `idyCVal` and `idyCVal2` is `1.0`, which means either computing `result.xy` or `result.zw` is wasting of time.

As a result, this PR adds the if-branches to check `idyCVal` and `idyCVal2` before computing, instead of always computing `result.xy * idyCVal` and `result.zw * idyCVal2`. This improves Conv2DBackpropInput ops in ArPortraitDepth ~40% and ArPortraitDepth model is improved from 23ms to 18.5ms (Benchmarked on LENOVO P620 2021).

Before the PR:
image

After the PR:
image

Reference https://github.com/tensorflow/tfjs/pull/7371.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7386)

### DIFF
--- a/tfjs-backend-webgl/src/conv_backprop_gpu.ts
+++ b/tfjs-backend-webgl/src/conv_backprop_gpu.ts
@@ -59,16 +59,13 @@ export class Conv2DDerFilterProgram implements GPGPUProgram {
                 continue;
               }
 
-              if (${isChannelsLast}) {
-                float dyValue = getDy(b, yR, yC, d2);
-                float xValue = getX(b, xR, xC, d1);
-                dotProd += (xValue * dyValue);
-              } else {
-                float dyValue = getDy(b, d2, yR, yC);
-                float xValue = getX(b, d1, xR, xC);
-                dotProd += (xValue * dyValue);
-              }
-
+              ${isChannelsLast?
+             `float dyValue = getDy(b, yR, yC, d2);
+              float xValue = getX(b, xR, xC, d1);
+              dotProd += (xValue * dyValue);` :
+             `float dyValue = getDy(b, d2, yR, yC);
+              float xValue = getX(b, d1, xR, xC);
+              dotProd += (xValue * dyValue);`}
             }
           }
         }
--- a/tfjs-backend-webgl/src/conv_backprop_packed_gpu.ts
+++ b/tfjs-backend-webgl/src/conv_backprop_packed_gpu.ts
@@ -64,36 +64,50 @@ export class Conv2DDerInputPackedProgram implements GPGPUProgram {
             int wCPerm = ${filterWidth} - 1 - wC;
 
             float dyC = float(dyCCorner + wC) / strides[1];
-            float idyCVal = dyC < 0.0 ? 0. :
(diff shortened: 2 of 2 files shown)


Linchenn (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7386)* status: 0 of 1 approvals obtained

___
*[`tfjs-backend-webgl/src/conv_backprop_packed_gpu.ts` line 78 at r1](https://reviewable.io/reviews/tensorflow/tfjs/7386#-NOQixbz1jed3yeC0yAS:-NOQixbz1jed3yeC0yAT:b6hfkvy) ([raw file](https://github.com/tensorflow/tfjs/blob/dcf35cdf2c30623e771b2670529c2230dd211abd/tfjs-backend-webgl/src/conv_backprop_packed_gpu.ts#L78)):*
> ```XML
>             if (idyCVal && idyCVal2) {
>               for (int d2 = 0; d2 < ${convInfo.outChannels}; d2 += 2) {
>                 vec4 wValue = getW(wRPerm, wCPerm, d1, d2);
> ```

Even though the three branches have the same codes, we still should not move the if-branches into the for loop. Otherwise the performance drops (for Conv2DBackpropInput op, drops ~10%), probably because the if-branches will be executed for each iteration (then threads probably would be stalled/synced in each iteration).



_Code quote:_
```XML
              for (int d2 = 0; d2 < ${convInfo.outChannels}; d2 += 2) {
                vec4 wValue = getW(wRPerm, wCPerm, d1, d2);
```

pyu10055 (approved):
great work, thanks!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7386#-:-NORV6dV9YOTnFSZvpHA:b6rtb8b)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @qjia7)

qjia7 (approved):
LGTM, thanks!

Linchenn:
Before this PR (with packed ConvBackpropInput):
image

After this PR:
ConvBackpropInput op's inference time drops (the top1 drops from 5.05 to 4.10)
image

qjia7:
I just saw a big regression on Intel devices (both CFL and ADL) with this PR.
Before (CFL)
conv2dBackpropInput_before
model_before
After (CFL)
conv2dBackpropInput_after
model_after

Before (ADL)
ADL_op_before
ADL_model_before
After (ADL)
ADL_op_after
ADL_model_after

qjia7:
It's surprising that there is such a big regressions for Intel devices. But the similar algorithm works well for webgpu. It may be a driver bug if it only happens on Intel devices. I can report an issue to our driver team. But it will be good if we can skip this optimization for Intel devices on TFJS level temporarily.

PR #7387, opened 2023-02-16
[webgl] Reduce JS map lookups in webgl backend

chunnienc:

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7387)

### DIFF
--- a/tfjs-backend-webgl/src/texture_manager.ts
+++ b/tfjs-backend-webgl/src/texture_manager.ts
@@ -27,11 +27,11 @@ export class TextureManager {
   private _numBytesAllocated = 0;
   private _numBytesFree = 0;  // How many bytes that have been allocated
                               // are available for reuse.
-  private freeTextures: {[shape: string]: Texture[]} = {};
+  private freeTextures: Record<string, Texture[]> = {};
+  private usedTextures: Record<string, Texture[]> = {};
   private logEnabled = false;
-  private usedTextures: {[shape: string]: Texture[]} = {};
 
-  constructor(private gpgpu: GPGPUContext) {}
+  constructor(private readonly gpgpu: GPGPUContext) {}
 
   acquireTexture(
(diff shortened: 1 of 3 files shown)


tfjs-backend-webgl/src/texture_manager.ts
```diff
@@ -25,13 +25,13 @@ export class TextureManager {
   private numUsedTextures = 0;
   private numFreeTextures = 0;
   private _numBytesAllocated = 0;
-  private _numBytesFree = 0;  // How many bytes that have been allocated
-                              // are available for reuse.
-  private freeTextures: {[shape: string]: Texture[]} = {};
+  // Number of bytes that have been allocated and available for reuse.
+  private _numBytesFree = 0;
+  private freeTextures?: Record<string, Texture[]> = {};
+  private usedTextures?: Record<string, Texture[]> = {};
```
mattsoulanille, line 31: Nit: Why is this optional (with `?`) but immediately assigned? Does it get deleted somewhere else?

mattsoulanille (approved):
LGTM with a nit.

tfjs-backend-webgl/src/texture_manager.ts
chunnienc: Yes. `TextureManager.dispose()` assigns nulls to them, and a lot of tests relies on this behavior (set to null after disposed) to be correct, at least they failed when I tried to make it always not null. There may be some additional guards/checks needed after we turn on the null check, but so far it's fine since it matches the old behavior.

mattsoulanille: Sounds good. Thanks!

chunnienc: Rolled back and added a todo comment

Linchenn (approved):

PR #7391, opened 2023-02-17
Add cache for parseNodeName

chunnienc:
This PR adds cache for parseNodeName - an expensive string manipulation function in graph executor. The time saved is around 2% to 3% for MobileNetV3 (tested on my laptop, may be imprecise).

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7391)

### DIFF
--- a/tfjs-converter/src/executor/execution_context.ts
+++ b/tfjs-converter/src/executor/execution_context.ts
@@ -39,6 +39,8 @@ export interface ExecutionContextInfo {
  * For model with branch logic, TensorFLow will generate Switch/Merge ops.
  */
 export class ExecutionContext {
+  public readonly parseNodeNameCache =
+      new Map<string, [string, number, string?]>();
   private rootContext = {id: 0, frameName: '', iterationId: 0};
   private contexts: ExecutionContextInfo[] = [this.rootContext];
   private lastId = 0;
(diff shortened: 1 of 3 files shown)


mattsoulanille (approved):
LGTM. I was going to recommend using a closure to memoize `parseNodeName` instead of adding it to `context` but it looks like you need to pass this context to other functions like `getParamValue`.

chunnienc:
Yeah. I think the context is suitable since it covers calls from parseNodeName itself and also getTensor and getParamValue, and it associate the lifecycle of the string with the graph, so that these string keys can be gc'ed with the graph.

Linchenn:
Thanks! I think this approach would benefit small models a lot, by introducing cache and cache lookup to avoid string operations.

Have you benchmarked some large models? For large models, the number of nodes may be significant, so I am not sure, when the cache is very large, whether the cache lookup is still faster than string operations?

chunnienc:
> Thanks! I think this approach would benefit small models a lot, by introducing cache and cache lookup to avoid string operations.
> 
> Have you benchmarked some large models? For large models, the number of nodes may be significant, so I am not sure, when the cache is very large, whether the cache lookup is still faster than string operations?

(Discussed offline) Yes it's always faster.
Benchmark: https://jsbench.me/8eleeumdpi/1

Linchenn (approved):
LGTM, thanks!

PR #7392, opened 2023-02-17
Add new tensor diposal algorithm for sync graph execution

chunnienc:
See code comments for more details.
It only works for sync execution since the new algorithm depends on a fixed node execution order.  The time complexity decreases from `O(NumEdges*NumTensors)` or more precise `O(NumTensorEdges)` to `O(NumTensors)`.
The time saved is around 5% for MobileNetV3 (again, tested on my laptop and may be imprecise).

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7392)


tfjs-converter/src/operations/executors/utils.ts
```diff
@@ -115,16 +114,33 @@ function getNodeNameWithContextId(name: string, contextId?: string): string {
...
+    const cachedResult = context.parseNodeNameCache.get(name);
+    if (cachedResult != null) {
+      return cachedResult;
+    }
+  }
   const parts = name.split(':');
+  let result: [string, number, string?];
   if (parts.length === 1) {
-    return [name, 0, undefined];
+    result = [name, 0, undefined];
```
mattsoulanille, line 133: This is from #7391, so that PR should be merged before this one.

tfjs-converter/src/executor/model_analysis.ts
```diff
@@ -146,26 +145,87 @@ export function getNodesInTopologicalOrder(
...
+  const nNodes = orderedNodes.length;
+  const nodeToOrder = new Map<Node, number>();
+  for (let i = 0; i < nNodes; ++i) {
+    nodeToOrder.set(orderedNodes[i], i);
+  }
+
+  // `liveUntil[i]` indicates that "all the intermediate tensors from
+  // `orderedNodes[i]` should be disposed after `orderedNodes[liveUntil[i]]`
+  // being executed."
+  const INF_LIFE = Math.floor(Number.MAX_SAFE_INTEGER / 2);
```
mattsoulanille, line 167: Nit: Why use MAX_SAFE_INTEGER / 2 instead of MAX_SAFE_INTEGER? I don't see any place where INF_LIFE would have something added to it, although I may have missed something.

tfjs-converter/src/executor/model_analysis.ts
```diff
@@ -146,26 +145,87 @@ export function getNodesInTopologicalOrder(
...
+ *     execution. That is, for a node `x`, `nodeLiveUntilMap[x]` indicates all
+ *     nodes which their intermediate tensors should be disposed after `x` being
+ *     executed.
+ */
+export function getNodeLiveUntilMap(orderedNodes: Node[]): Map<Node, Node[]> {
+  const nNodes = orderedNodes.length;
+  const nodeToOrder = new Map<Node, number>();
+  for (let i = 0; i < nNodes; ++i) {
+    nodeToOrder.set(orderedNodes[i], i);
+  }
```
mattsoulanille, line 162: Tiny nit, feel free to ignore, but this may be more readable.

```suggestion
  const nodeToOrder = new Map(orderedNodes.map((node, index) => [node, index]));
```

tfjs-converter/src/executor/model_analysis.ts
mattsoulanille, line 167: Nit: Pull this out of the function maybe? It probably doesn't matter.

tfjs-converter/src/executor/model_analysis.ts
```diff
@@ -146,26 +145,87 @@ export function getNodesInTopologicalOrder(
...
+
+  for (let nodeOrder = 0; nodeOrder < nNodes; ++nodeOrder) {
+    const node = orderedNodes[nodeOrder];
+    for (const child of node.children) {
+      const childOrder = nodeToOrder.get(child)!;
+      // Extend the node's life to at least its child's life.
+      liveUntil[nodeOrder] =
+          Math.max(liveUntil[nodeOrder], liveUntil[childOrder]);
+    }
+  }
```
mattsoulanille, line 187: Nit: This may be easier to read, but feel free to skip it.

```suggestion
  const liveUntil = orderedNodes.map((node, index) => {
    if (isControlFlow(node)) {
      return INF_LIFE;
    }
    return node.children
          .map(node => nodeToOrder.get(node)!)
          .reduce((a, b) => Math.max(a, b), index);
  });
```

tfjs-converter/src/executor/model_analysis.ts
```diff
@@ -50,12 +50,11 @@ export function getExecutionSubgraph(
...
   const seen = new Set<string>();
   const inputNodeNames =
-      Object.keys(inputs).map(name => parseNodeName(name)[0]);
+      new Set(Object.keys(inputs).map((name) => parseNodeName(name)[0]));
 
-  let initNodeNames: string[] = [];
-  if (initNodes != null) {
-    initNodeNames = initNodes.map(node => parseNodeName(node.name)[0]);
-  }
+  initNodes = initNodes || [];
```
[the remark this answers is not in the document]
chunnienc: Our compiler target in tsconfig is es2017, while the nullish operator is introduced in es2020 :(

tfjs-converter/src/executor/model_analysis.ts
```diff
@@ -146,26 +145,87 @@ export function getNodesInTopologicalOrder(
...
+export function getNodeLiveUntilMap(orderedNodes: Node[]): Map<Node, Node[]> {
+  const nNodes = orderedNodes.length;
+  const nodeToOrder = new Map<Node, number>();
+  for (let i = 0; i < nNodes; ++i) {
+    nodeToOrder.set(orderedNodes[i], i);
+  }
+
+  // `liveUntil[i]` indicates that "all the intermediate tensors from
+  // `orderedNodes[i]` should be disposed after `orderedNodes[liveUntil[i]]`
+  // being executed."
```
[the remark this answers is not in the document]
chunnienc: We use the node parent-child relationship to determine when is safe to dispose the tensor, but tensors from a node may not necessary to be "used" by its children. But overall I think this is a good rewrite. Will modify and update it.

tfjs-converter/src/executor/model_analysis.ts
```diff
@@ -146,26 +145,87 @@ export function getNodesInTopologicalOrder(
...
+  const INF_LIFE = Math.floor(Number.MAX_SAFE_INTEGER / 2);
+  const liveUntil = [...Array(nNodes).keys()];
+
+  for (let nodeOrder = 0; nodeOrder < nNodes; ++nodeOrder) {
```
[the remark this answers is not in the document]
chunnienc: I prefer to keep `nodeOrder` since the idea of "node order" is the key to this algorithm.

tfjs-converter/src/executor/model_analysis.ts
chunnienc: Nice catch. Yes it can just be MAX_SAFE_INTEGER here.

I was testing another algorithm before I sent out this PR and that algorithm can tweak the life by adding some offset to its value. That's why I made it a smaller value to be safe, but it turns out that the one you reviewed is good enough and the life here can just be the max value.

tfjs-converter/src/executor/model_analysis.ts
chunnienc: Just notice that this rewrite this incorrect. If a child node lives until INF_LIFE, the parent node should also live until INF_LIFE. That's for the control flow node so that it can always get tenors from its parents.

This is also the behavior of the old `checkNodesForDisposal` function, as far as I understand. Please double-check this if you can.

tfjs-converter/src/executor/model_analysis.ts
chunnienc: Is there any other functions use the live until info? I'm thinking to rewrite the old `checkNodesForDisposal` with this info, but I don't have any good ideas so far. Maybe later.

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -318,44 +330,74 @@ export class GraphExecutor implements FunctionExecutor {
...
+    }
+
+    for (const node of liveUntilNodes) {
+      for (const tensor of tensorMap[node.name]) {
+        if (!tensor || tensor.kept || tensorsToKeep.has(tensor.id)) {
+          continue;
+        }
+        tensor.dispose();
+      }
+    }
```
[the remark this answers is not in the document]
chunnienc: I intentionally make this function keep similar signature and name as the original `checkNodesForDisposal`, to maintain the parity of tensor disposing functions in sync and async graph executor. I want to make it easier for future maintainer to understand that they are the same, at least from the aspect of intentions and behaviors.

I think a better rewrite may be renaming both of them to something like "function disposeNodes...(...)", since the original function not only "checks" the tensors but also disposes tensors in it. What do you think?

For factoring out some stuff, I don't see huge gains on making a separate function for that single if check. I'm thinking to rewrite the async disposing algorithm to utilize the node dependency and reduce loops over tensor edges, but I haven't figured out a better algo (and I need to learn more about async control flow to make sure the new algo won't break it). Maybe we can have a better picture of which parts can be shared then.

tfjs-converter/src/executor/model_analysis.ts
```diff
@@ -178,18 +175,16 @@ export function getNodeLiveUntilMap(orderedNodes: Node[]): Map<Node, Node[]> {
...
-    for (const child of node.children) {
-      const childOrder = nodeToOrder.get(child)!;
-      // Extend the node's life to at least its child's life.
-      liveUntil[nodeOrder] =
-          Math.max(liveUntil[nodeOrder], liveUntil[childOrder]);
-    }
+    // Extend the node's life to at least its child's life.
+    liveUntil[nodeOrder] = Math.max(
+        liveUntil[nodeOrder],
+        ...node.children.map((child) => liveUntil[nodeToOrder.get(child)!]));
```
mattsoulanille, line 181: `Math.max` supports a limited number of arguments before running into `maximum call stack size exceeded`. In Node, I've experimentally found this limit to be around `123258`, which is rather small. I recommend using `reduce` here to avoid that limit.

```suggestion
    liveUntil[nodeOrder] = node.children.reduce((currentMax, child) => 
        Math.max(liveUntil[nodeToOrder.get(child)!], currentMax), liveUntil[nodeOrder]);
```

tfjs-converter/src/executor/model_analysis.ts
mattsoulanille: I was talking about the const `INF_LIFE`, but github highlighted a bunch of other lines too. It can probably stay in the function, though.

tfjs-converter/src/executor/model_analysis.ts
mattsoulanille: Nice catch. Like you said, I missed the `INF_LIFE` of children also affecting parents. I think this fixes it:
```ts
  // Make control flow nodes (and consequently their direct parents)
  // live forever since they're tricky to track correctly.
  const selfLifespans = orderedNodes.map((node, nodeOrder) => 
    isControlFlow(node) ? INF_LIFE : nodeOrder);

  // A node lives long enough to pass on its tensors to its children.
  // It lives until at least `max(node's position, childrens' positions)`.
  const liveUntil = orderedNodes.map((node, nodeOrder) =>
    node.children
      .map(node => selfLifespans[nodeToOrder.get(node)!])
      .reduce((a, b) => Math.max(a, b), selfLifespans[nodeOrder]));
```

tfjs-converter/src/executor/model_analysis.ts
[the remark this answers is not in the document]
mattsoulanille: We should fix this in the tsconfig. [I think we can set `lib` to `esnext` and `target` to `es2017` to have TypeScript automatically compile nullish coalescing to something compatible with `es2017`.](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMOcAUBDA-ALkWEAWwCMBTAJwBpFjd8izyBKRAbwChFFzSoRykxRFiyIMAbnYBfIA)

But for this PR, feel free to leave it as-is.

tfjs-converter/src/executor/model_analysis.ts
chunnienc: Then I'd just leave it here. This analysis does not expose the lifespan in numbers, so it's better to keep this intermediate value inside.

tfjs-converter/src/executor/model_analysis.ts
chunnienc: Nice. I didn't know Math.max uses recursion in it, thanks!

mattsoulanille (approved):

Linchenn (approved):
LGTM

PR #7413, opened 2023-02-24
Improve assertAndGetBroadcastShape

Linchenn:
According to https://jsbench.me/bmlehvit9v/1, push is far faster than unshift.

image
`assertAndGetBroadcastShape2` is after this PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7413)

### DIFF
--- a/tfjs-core/src/ops/broadcast_util.ts
+++ b/tfjs-core/src/ops/broadcast_util.ts
@@ -62,25 +62,25 @@ export function assertAndGetBroadcastShape(
   const result: number[] = [];
   const l = Math.max(shapeA.length, shapeB.length);
 
-  for (let i = 0; i < l; i++) {
-    let a = shapeA[shapeA.length - i - 1];
+  for (let i = l - 1; i >= 0; i--) {
+    let a = shapeA[i];
     if (a == null) {
       a = 1;
     }
-    let b = shapeB[shapeB.length - i - 1];
+    let b = shapeB[i];
     if (b == null) {
       b = 1;
     }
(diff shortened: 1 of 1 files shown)


tfjs-core/src/ops/broadcast_util.ts
```diff
@@ -62,25 +62,25 @@ export function assertAndGetBroadcastShape(
...
-      result.unshift(a);
+      result.push(a);
     } else if (a !== b) {
       const errMsg = `Operands could not be broadcast together with shapes ` +
           `${shapeA} and ${shapeB}.`;
       throw Error(errMsg);
     } else {
-      result.unshift(a);
+      result.push(a);
     }
```
chunnienc, line 84: After the for loop:
```
result.reverse();
return result;
```

chunnienc (changes requested):

tfjs-core/src/ops/broadcast_util.ts
chunnienc: or you can create a fixed length array beforehand:
```
const result = new Array(l);
```
and then assign the value with index `i`

Linchenn: Done. Array is faster than reverse.

chunnienc (approved):

mattsoulanille (approved):
LGTM. This does not appear to be using `push()` anymore, and instead is using array indexing. I assume this is faster than `push()` since it doesn't need to resize the array every so often.

PR #7428, opened 2023-02-28
[webgpu] Implement toPixels with graphics draw

axinging:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7428)

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -22,6 +22,7 @@ import {backend_util, BackendValues, buffer, DataStorage, DataType, engine, env,
 import {AdapterInfo} from './adapter_info';
 import {BufferManager} from './buffer_manager';
 import {TextureManager} from './texture_manager';
+import {compileTextureProgram} from './webgpu_graphics_program';
 import * as webgpu_program from './webgpu_program';
 import * as webgpu_util from './webgpu_util';
 
@@ -136,6 +137,8 @@ export class WebGPUBackend extends KernelBackend {
   private supportTimeQuery: boolean;
   private uniformPendingDisposal: BufferInfo[] = [];
   private uploadWaitMs = 0;
+  // This is only used by toPixels.
+  private drawTexturePipeline: GPURenderPipeline;
 
   private nextDataId(): number {
     return WebGPUBackend.nextDataId++;
@@ -751,12 +754,58 @@ export class WebGPUBackend extends KernelBackend {
     return {offset: 0, size: currentOffset, buffer: uniformBuffer};
   }
 
+  // Draw GPUTexture to GPUCanvasContext.
+  private drawTextureToCanvas(
+      gpuContext: GPUCanvasContext, texture: GPUTexture) {
+    if (!this.drawTexturePipeline) {
+      this.drawTexturePipeline = compileTextureProgram(this.device);
+    }
+
+    const bindGroup = this.device.createBindGroup({
+      layout: this.drawTexturePipeline.getBindGroupLayout(0),
+      entries: [
+        {
+          binding: 0,
+          resource: texture.createView(),
+        },
+      ],
+    });
+
+    const renderPassDescriptor: GPURenderPassDescriptor = {
+      colorAttachments: [
+        {
+          view: gpuContext.getCurrentTexture().createView(),
+          loadOp: 'load',
+          clearValue: {r: 0.0, g: 0.0, b: 0.0, a: 0.0},
+          storeOp: 'store',
+        },
+      ],
+    };
+
+    this.ensureCommandEncoderReady();
+    const pass =
+        this.currentCommandEncoder.beginRenderPass(renderPassDescriptor);
+    pass.setPipeline(this.drawTexturePipeline);
+    pass.setBindGroup(0, bindGroup);
+    pass.draw(6);
+    pass.end();
+  }
+
   public runWebGPUProgram(
       program: webgpu_program.WebGPUProgram, inputs: TensorInfo[],
       outputDtype: DataType, programDefinedUniform?: ProgramUniform,
-      output?: TensorInfo): TensorInfo {
+      output?: TensorInfo, outputTextureInfo?: TextureInfo,
+      gpuContext?: GPUCanvasContext): TensorInfo {
     if (!output) {
       output = this.makeTensorInfo(program.outputShape, outputDtype);
+    }
+    if (program.pixelsOpType === webgpu_program.PixelsOpType.TO_PIXELS) {
+      if (gpuContext == null || outputTextureInfo == null) {
+        throw new Error(
+            'GPUCanvasContext is null or output texture is not provided for ToPixels!');
+      }
+      const info = this.tensorMap.get(output.dataId);
+      info.resourceInfo = outputTextureInfo;
     }
     if (util.sizeFromShape(output.shape) === 0) {
       // Short-circuit the computation since the result is empty (has 0 in its
@@ -772,24 +821,26 @@ export class WebGPUBackend extends KernelBackend {
     // program size, program defined uniforms.
     let programUniform: ProgramUniform = [];
     let bufferShapes: number[][] = [];
-    if (!program.isFromPixels) {
+    const uniformsType = 'int32';
+    if (program.pixelsOpType == null) {
       programUniform.push(
           {type: 'float32', data: [NaN]}, {type: 'float32', data: [Infinity]});
       bufferShapes = inputs.concat(output).map(d => d.shape);
-      const uniformsType = 'int32';
       bufferShapes.map(d => {
         programUniform.push({type: uniformsType, data: d});
         const strides = util.computeStrides(d);
         programUniform.push({type: uniformsType, data: strides});
       });
-      if (program.size) {
-        const size = util.sizeFromShape(program.outputShape);
-        programUniform.push({
-          type: uniformsType,
-          data:
-              [program.outputComponent ? size / program.outputComponent : size]
-        });
-      }
+    } else {
+      const strides = util.computeStrides(output.shape);
+      programUniform.push({type: uniformsType, data: strides});
+    }
+    if (program.size) {
+      const size = util.sizeFromShape(program.outputShape);
+      programUniform.push({
+        type: uniformsType,
+        data: [program.outputComponent ? size / program.outputComponent : size]
+      });
     }
 
     const inputsData = inputs.map((input: TensorInfo, i: number) => {
@@ -860,8 +911,13 @@ export class WebGPUBackend extends KernelBackend {
       this.commandQueueOwnedIds.add(input.dataId);
     });
     this.commandQueueOwnedIds.add(output.dataId);
-
-    if (env().get('WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE') as
+    if (program.pixelsOpType === webgpu_program.PixelsOpType.TO_PIXELS) {
+      this.ensureComputePassEnded();
+      this.drawTextureToCanvas(
+          gpuContext, outputTextureInfo.texture as GPUTexture);
+      this.submitQueue();
+    } else if (
+        env().get('WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE') as
         number <= this.dispatchNumberInEncoder) {
       this.submitQueue();
     }
--- a/tfjs-backend-webgpu/src/kernels/ToPixels.ts
+++ b/tfjs-backend-webgpu/src/kernels/ToPixels.ts
@@ -1,0 +1,70 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use backend file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import {KernelConfig, KernelFunc, TensorInfo} from '@tensorflow/tfjs-core';
+import {ToPixels, ToPixelsAttrs, ToPixelsInputs} from '@tensorflow/tfjs-core';
+
+import {TextureInfo, WebGPUBackend} from '../backend_webgpu';
+import {ToPixelsProgram} from '../to_pixels_webgpu';
+
+export function toPixels(
+    args:
+        {inputs: ToPixelsInputs, backend: WebGPUBackend, attrs: ToPixelsAttrs}):
+    TensorInfo {
+  const {inputs, backend, attrs} = args;
+  const {$img} = inputs;
+  const {canvas, numChannels} = attrs;
+  const [height, width] = $img.shape.slice(0, 2);
+
+  const format = 'rgba8unorm';
+  const outShape = [height, width, 1];
+  const program = new ToPixelsProgram(outShape, $img.dtype, format);
+  canvas.width = width;
+  canvas.height = height;
+  const gpuContext = canvas.getContext('webgpu');
+  //  'rgba8unorm' is not supported yet as the context format
+  //  (https://bugs.chromium.org/p/chromium/issues/detail?id=1241369).
+  //  If supported, we can use single compute pass to transfer the input tensor
+  //  data to webgpu context canvas.
+  gpuContext.configure({
+    device: backend.device,
+    format: 'bgra8unorm',
+    compositingAlphaMode: 'opaque'
+  });
+
+  // ToPixelsProgram will frist write this texture, then this texture will be
+  // drawn into GPUCanvasContext.
+  const intermediateTexture = backend.device.createTexture({
+    format,
+    size: [width, height, 1],
+    usage: GPUTextureUsage.COPY_SRC | GPUTextureUsage.STORAGE_BINDING |
+        GPUTextureUsage.TEXTURE_BINDING
+  });
+
+  const textureInfo: TextureInfo =
+      {width, height, format: null, usage: null, texture: intermediateTexture};
+
+  const uniformData = [{type: 'uint32', data: [numChannels]}];
+  return backend.runWebGPUProgram(
+      program, [$img], 'int32', uniformData, null, textureInfo, gpuContext);
+}
+
+export const toPixelsConfig: KernelConfig = {
+  kernelName: ToPixels,
+  backendName: 'webgpu',
+  kernelFunc: toPixels as unknown as KernelFunc
+};
--- a/tfjs-backend-webgpu/src/kernels/ToPixels.ts
+++ b/tfjs-backend-webgpu/src/kernels/ToPixels.ts
@@ -36,6 +36,10 @@ export function toPixels(
   canvas.width = width;
   canvas.height = height;
   const gpuContext = canvas.getContext('webgpu');
+  if (!gpuContext) {
+    throw new Error(
+        `Please make sure this canvas has only been used for webgpu context!`);
+  }
   //  'rgba8unorm' is not supported yet as the context format
   //  (https://bugs.chromium.org/p/chromium/issues/detail?id=1241369).
   //  If supported, we can use single compute pass to transfer the input tensor
--- a/tfjs-backend-webgpu/src/to_pixels_webgpu.ts
+++ b/tfjs-backend-webgpu/src/to_pixels_webgpu.ts
@@ -1,0 +1,85 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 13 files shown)


axinging:
@qjia7 @xhcao @gyagp PTAL

tfjs-core/src/ops/to_pixels_test.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2020 Google Inc. All Rights Reserved.
+ * Copyright 2023 Google Inc.
```
xhcao, line 3: Do we need to modify it if we only modify the content, it is not a new file?

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -772,7 +823,7 @@ export class WebGPUBackend extends KernelBackend {
     // program size, program defined uniforms.
     let programUniform: ProgramUniform = [];
     let bufferShapes: number[][] = [];
-    if (!program.isFromPixels) {
+    if (program.pixelsOpType == null) {
```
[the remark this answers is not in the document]
axinging: program.pixelsOpType maybe null or undefined.  "==" will capture both

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -252,7 +256,9 @@ export class WebGPUBackend extends KernelBackend {
     }
     if ('texture' in tensorData.resourceInfo) {
       const textureInfo = tensorData.resourceInfo;
-      if (textureInfo.texture instanceof GPUTexture) {
+      // Texture from canvas should not be released.
+      if (textureInfo.texture instanceof GPUTexture &&
+          textureInfo.isCanvas !== true) {
```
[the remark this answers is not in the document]
axinging: This should be released here, updated

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -751,13 +757,58 @@ export class WebGPUBackend extends KernelBackend {
     return {offset: 0, size: currentOffset, buffer: uniformBuffer};
   }
 
+  // Draw GPUTexture to GPUCanvasContext.
+  private drawTexture(gpuContext: GPUCanvasContext, texture: GPUTexture) {
```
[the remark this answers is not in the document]
axinging: done

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -751,13 +757,58 @@ export class WebGPUBackend extends KernelBackend {
...
+          resource: texture.createView(),
+        },
+      ],
+    });
+
+    const renderPassDescriptor: GPURenderPassDescriptor = {
+      colorAttachments: [
+        {
+          view: gpuContext.getCurrentTexture().createView(),
+          loadOp: 'clear',
```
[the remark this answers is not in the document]
axinging: Good catch, thanks!

tfjs-backend-webgpu/src/kernels/ToPixels.ts
```diff
@@ -0,0 +1,91 @@
...
+ * =============================================================================
+ */
+
+import {KernelConfig, KernelFunc, TensorInfo, util} from '@tensorflow/tfjs-core';
+import {ToPixels, ToPixelsAttrs, ToPixelsInputs} from '@tensorflow/tfjs-core';
+
+import {TextureInfo, WebGPUBackend} from '../backend_webgpu';
+import {ToPixelsProgram} from '../to_pixels_webgpu';
+
+export const toPixelsConfig: KernelConfig = {
```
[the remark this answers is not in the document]
axinging: I am not sure if this is necessary, anyway, updated

tfjs-backend-webgpu/src/kernels/ToPixels.ts
```diff
@@ -0,0 +1,91 @@
...
+    args:
+        {inputs: ToPixelsInputs, backend: WebGPUBackend, attrs: ToPixelsAttrs}):
+    TensorInfo {
+  const {inputs, backend, attrs} = args;
+  const {$img} = inputs;
+  const {canvas, numChannels} = attrs;
+  const [height, width] = $img.shape.slice(0, 2);
+
+  const format = 'rgba8unorm';
+  const outShape = [height, width, getNumberChannel(format)];
```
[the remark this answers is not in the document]
axinging: We wil soon support bgra8unorm. I simply replace this with 4.

tfjs-backend-webgpu/src/kernels/ToPixels.ts
```diff
@@ -0,0 +1,91 @@
...
+    texture: intermediateTexture,
+    isCanvas: true
+  };
+
+  const size = util.sizeFromShape(program.outputShape);
+  const strides = util.computeStrides(program.outputShape);
+
+  const uniformData = [
+    {type: 'uint32', data: [size]}, {type: 'uint32', data: [numChannels]},
+    {type: 'uint32', data: [...strides]}
```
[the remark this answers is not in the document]
axinging: Yes, good catch! Updated.

tfjs-core/src/ops/to_pixels_test.ts
axinging: I  donot know. I revert this change.

axinging (commented):
Thanks @xhcao , Please take another look

tfjs-backend-webgpu/src/to_pixels_webgpu.ts
```diff
@@ -0,0 +1,88 @@
...
+  workgroupSize: [number, number, number] = [64, 1, 1];
+  workPerThread = 4;
+  type: DataType;
+  textureFormat: GPUTextureFormat;
+  pixelsOpType = PixelsOpType.TO_PIXELS;
+  size = true;
+
+  constructor(
+      outShape: number[], type: DataType, textureFormat: GPUTextureFormat) {
+    this.outputShape = outShape;
```
qjia7, line 38: We can pass an outShape without channel like `[height, width]`. Then we don't need to make `workPerThread` to `4` and calculate a `flatIndex `again in shader. And the shader will be like below:
```
${main('index')} {
  if (index < uniforms.size) {
    var rgba = vec4<f32>(0.0, 0.0, 0.0, 1.0);
    for (var d = 0; d < uniforms.numChannels; d = d + 1) {
      let value = f32(inBuf[index * uniforms.numChannels + d]);
      ${calculateResult}
    }
    let coords = getCoordsFromIndex(index);
    textureStore(outImage, vec2<i32>(coords.yx), rgba);
  }
}

File: tfjs-backend-webgpu/src/backend_webgpu.ts [hunk 10]
```diff
@@ -860,8 +908,13 @@ export class WebGPUBackend extends KernelBackend {
       this.commandQueueOwnedIds.add(input.dataId);
     });
     this.commandQueueOwnedIds.add(output.dataId);
-
-    if (env().get('WEBGPU_DEFERRED_SUBMIT_BATCH_SIZE') as
+    if (program.pixelsOpType === webgpu_program.PixelsOpType.TO_PIXELS) {
+      this.ensureComputePassEnded();
+      this.drawTextureToCanvas(
```
qjia7, line 913: How about moving `drawTextureToCanvas` to `ToPixels`?

tfjs-backend-webgpu/src/to_pixels_webgpu.ts
axinging: Done, thanks!

tfjs-backend-webgpu/src/backend_webgpu.ts
axinging: Currently all webgpu related code are kept inside backend_webgpu. So I prefer puting drawTextureToCanvas inside backend_webgpu. This also provide a opportunity to batch graphics and compute commands.

tfjs-backend-webgpu/src/kernels/ToPixels.ts
```diff
@@ -31,7 +31,7 @@ export function toPixels(
   const [height, width] = $img.shape.slice(0, 2);
 
   const format = 'rgba8unorm';
-  const outShape = [height, width, 4];
+  const outShape = [height, width, 1];
```
qjia7, line 34: nit: `[height, width, 1]` -> `[height, width]`?

tfjs-backend-webgpu/src/backend_webgpu.ts
qjia7: Got it, thanks.

qjia7 (approved):
LGTM, thanks!

qjia7:
@pyu10055 @mattsoulanille @Linchenn In today's meeting, we have a short discussion about `toPixels`. This PR provides an efficient way to use gpu to draw image to canvas behind flag `TO_PIXELS_RETURN_VOID`.

tfjs-core/src/ops/browser.ts
```diff
@@ -318,6 +317,24 @@ export async function toPixels(
...
+  // If the current backend has 'toPixels' registered, it has a more
+  // efficient way of drawing pixels, so we call that.
+  const kernel = getKernel(ToPixels, ENGINE.backendName);
+  if (canvas != null && kernel != null) {
+    const inputs: ToPixelsInputs = {$img};
+    const attrs: ToPixelsAttrs = {canvas, numChannels: depth};
+    const tensorPlaceholder = ENGINE.runKernel(
+        ToPixels, inputs as {} as NamedTensorMap, attrs as {} as NamedAttrMap);
+    (tensorPlaceholder as Tensor).dispose();
+    if (env().getBool('TO_PIXELS_RETURN_VOID')) {
```
mattsoulanille, line 329:
```suggestion
    if (!getData) {
```

mattsoulanille (commented):
I've only had a chance to look at the external facing API so far, but I want to share these comments now. I'll do a full review soon.

Thanks for working on this! It greatly simplifies the process of drawing model outputs to canvas.

axinging:
> I've only had a chance to look at the external facing API so far, but I want to share these comments now. I'll do a full review soon.
> 
> Thanks for working on this! It greatly simplifies the process of drawing model outputs to canvas.

Thanks @mattsoulanille ,Please take another look.

pyu10055 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7428)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @Linchenn, @mattsoulanille, @qjia7, and @xhcao)

___
*[`tfjs-core/src/ops/browser.ts` line 294 at r4](https://reviewable.io/reviews/tensorflow/tfjs/7428#-NPMNdBPolzwZyya6MjN-r4-294:-NQfNX2aCbbMXqC3n6av:b-y8xd5x) ([raw file](https://github.com/tensorflow/tfjs/blob/b0012ce7a8d9c8612839c2d6bf4cf1712646b0bb/tfjs-core/src/ops/browser.ts#L294)):*

Instead of using a flag, we can get better typescript types by adding another argument and using function signature overloading. This has the benefit that it doesn't break the existing API.

```suggestion
 * @param getData Whether to return the data of the tensor drawn to the canvas
 * or only draw to the canvas (and return null). For backends that run on the
 * GPU, set this to false to avoid copying the data back to the CPU. Defaults
 * to true.
 * @return Returns a promise that resolves when the canvas has been drawn to.
 * When canvas is non null and getData is false, returns Promise<void>.
 * Otherwise, returns canvas data via Promise<Uint8ClampedArray>.
 *
 * @doc {heading: 'Browser', namespace: 'browser'}
 */
export function toPixels(img: Tensor2D|Tensor3D|TensorLike,
                         canvas: HTMLCanvasElement,
                         getData: false): Promise<void>;                                                                                                                                          
export function toPixels(img: Tensor2D|Tensor3D|TensorLike,
                         canvas: HTMLCanvasElement,
                         getData: true): Promise<Uint8ClampedArray>;                                                                                                                              
export function toPixels(img: Tensor2D|Tensor3D|TensorLike,
                         canvas: HTMLCanvasElement): Promise<Uint8ClampedArray>;                                                                                                                  
export function toPixels(img: Tensor2D|Tensor3D|TensorLike
			): Promise<Uint8ClampedArray>;                                                                                                                                            
export async function toPixels(
    img: Tensor2D|Tensor3D|TensorLike,
    canvas?: HTMLCanvasElement,
    getData = true): Promise<Uint8ClampedArray | void> {

``` 

The return types are correctly inferred based on the arguments passed to it:

```typescript
toPixels(someTensor); // -> Promise<Uint8ClampedArray>
toPixels(someTensor, canvas); // -> Promise<Uint8ClampedArray>
toPixels(someTensor, canvas, true); // -> Promise<Uint8ClampedArray>
toPixels(someTensor, canvas, false); // -> Promise<void>
```

Then, in 5.0, we can flip the default so that if you pass a canvas, it does not return data (but if you don't pass a canvas, it still does).

What do you think of this approach?
</blockquote></details>

I like this idea better, nit on the naming of the last parameter, I will suggest `directDraw` instead of `getData`.

axinging:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7428)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @Linchenn, @mattsoulanille, @qjia7, and @xhcao)
> 
> _[`tfjs-core/src/ops/browser.ts` line 294 at r4](https://reviewable.io/reviews/tensorflow/tfjs/7428#-NPMNdBPolzwZyya6MjN-r4-294:-NQfNX2aCbbMXqC3n6av:b-y8xd5x) ([raw file](https://github.com/tensorflow/tfjs/blob/b0012ce7a8d9c8612839c2d6bf4cf1712646b0bb/tfjs-core/src/ops/browser.ts#L294)):_
> 
> Previously, mattsoulanille (Matthew Soulanille) wrote…
> I like this idea better, nit on the naming of the last parameter, I will suggest `directDraw` instead of `getData`.

Thanks @pyu10055  @mattsoulanille, API already changed as suggested by matt.
As to `directDraw` instead of `getData`: 1), directDraw doesn't imply return data or not. 2), toPixels itself implies draw. So we prefer getData.  WDYT?

mattsoulanille (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7428)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @axinging, @Linchenn, @qjia7, and @xhcao)

___
*[`tfjs-core/src/ops/browser.ts` line 294 at r4](https://reviewable.io/reviews/tensorflow/tfjs/7428#-NPMNdBPolzwZyya6MjN-r4-294:-NQmLnxZ5cXbwh42f20f:b-ychdst) ([raw file](https://github.com/tensorflow/tfjs/blob/b0012ce7a8d9c8612839c2d6bf4cf1712646b0bb/tfjs-core/src/ops/browser.ts#L294)):*

I like this idea better, nit on the naming of the last parameter, I will suggest `directDraw` instead of `getData`.  
</blockquote></details>

I agree with @axinging, and I think `getData` is better since it relates more to the fact that it's getting data from the tensor. Whether `getData`/ `directDraw` is true or false, it will be using the direct draw method if the backend supports it.

Additionally, if we call it `directDraw`, then when it's true, it will not return anything. This is opposite to `getData`, where if it's true, it will return the tensor data. There is a rare edge case where this difference matters, as it technically changes our API. If a user passes too many arguments, they are more likely to get different behavior with `directDraw`, since a truthy value will cause the function to return nothing. On the other hand, a truthy value to `getData` will preserve the original behavior.

```
const arguments = [someTensor, canvas, foo, bar, baz] // User has extra arguments for some reason. 
const result = toPixels(...arguments)
```
This is a minor issue, though.

tfjs-backend-webgpu/src/kernels/ToPixels.ts
```diff
@@ -0,0 +1,70 @@
...
+  const {$img} = inputs;
+  const {canvas, numChannels} = attrs;
+  const [height, width] = $img.shape.slice(0, 2);
+
+  const format = 'rgba8unorm';
+  const outShape = [height, width, 1];
+  const program = new ToPixelsProgram(outShape, $img.dtype, format);
+  canvas.width = width;
+  canvas.height = height;
+  const gpuContext = canvas.getContext('webgpu');
```
qjia7, line 38: What if  another context (`2d` or `webgl`) has been requested for `canvas` and `canvas.getContext('webgpu')` return null?

tfjs-core/src/ops/browser.ts
```diff
@@ -358,7 +394,7 @@ export async function toPixels(
     bytes[j + 3] = Math.round(rgba[3]);
   }
 
-  if (canvas != null) {
+  if (canvas != null && kernel == null) {
     canvas.width = width;
     canvas.height = height;
     const ctx = canvas.getContext('2d');
```
qjia7, line 400: @pyu10055 @mattsoulanille It's possible that `canvas.getContext('2d')` is null if a `webgl` or `webgpu`  context has been requested for the same canvas before the canvas passes into `toPixels`. Here we should check `ctx`'s value. And what's this API's expectation for `canvas`? A canvas which is never called `getContext` or a `2d` canvas?

qjia7 (changes requested):
I offline talked with team. One question is what's this API's purpose to be designed to? To process the input tensor to be a `[height, width, 4]` data so that it can be directly drawn into a canvas? If yes, this API mainly wants to pad the input data with shape `[height, width, 1/3/4]` into a data with shape `[height, width, 4]`. canvas is optional because we draw the result just by the way.
If above assumption is true, maybe another way to improve this API is to return a rank-3 tensor with shape `[height, width, 4]`:
```
export function toPixels(img: Tensor2D|Tensor3D|TensorLike): Tensor3D;
export function toPixels(img: Tensor2D|Tensor3D|TensorLike, canvas?: HTMLCanvasElement): Tensor3D;
```
Then, when user calls `toPixels`. He will get a tensor object, the underlying data can be hold on cpu or gpu. Then the user can simply draw the result to canvas based on his usage scenario.
```
const pixels = tf.toPixels(img); // This step helps user to process the data to make it ready  to draw to canvas.
const gpudata = pixels.dataToGPU();
// draw gpudata to canvas.
...
```
Or
```
const res = tf.toPixels(img, canvas);
```

mattsoulanille:
> I offline talked with team. One question is what's this API's purpose to be designed to? To process the input tensor to be a `[height, width, 4]` data so that it can be directly drawn into a canvas? If yes, this API mainly wants to pad the input data with shape `[height, width, 1/3/4]` into a data with shape `[height, width, 4]`. canvas is optional because we draw the result just by the way. If above assumption is true, maybe another way to improve this API is to return a rank-3 tensor with shape `[height, width, 4]`:
> 
> ```
> export function toPixels(img: Tensor2D|Tensor3D|TensorLike): Tensor3D;
> export function toPixels(img: Tensor2D|Tensor3D|TensorLike, canvas?: HTMLCanvasElement): Tensor3D;
> ```
> 
> Then, when user calls `toPixels`. He will get a tensor object, the underlying data can be hold on cpu or gpu. Then the user can simply draw the result to canvas based on his usage scenario.
> 
> ```
> const pixels = tf.toPixels(img); // This step helps user to process the data to make it ready  to draw to canvas.
> const gpudata = pixels.dataToGPU();
> // draw gpudata to canvas.
> ...
> ```
> 
> Or
> 
> ```
> const res = tf.toPixels(img, canvas);
> ```

I talked with Ping, and we think users will expect `toPixels` to the opposite of `fromPixels`. That is, it should take a tensor that represents an image and put it on the screen. Right now, this is very complicated to do using WebGL and WebGPU, and users are more likely to pass data through the CPU instead of writing their own shaders. If we can provide a function that does it for them, they are more likely to write efficient pipelines.

I agree that its return type is odd. If I were to make breaking changes, I would split this into two methods
* `toPixels` - Draw a `[height, width, 4]` tensor to the screen. Returns `Promise<void>`.
* `asImage` - Attempt to format a tensor as `[height, width, 4]` for drawing to the screen. Does not draw to the screen itself.

To avoid breaking the API, I think we should use the behavior I described in my comment above:
```typescript
toPixels(someTensor); // -> Promise<Uint8ClampedArray>, does not draw
toPixels(someTensor, canvas); // -> Promise<Uint8ClampedArray>, draws
toPixels(someTensor, canvas, true); // -> Promise<Uint8ClampedArray>, draws
toPixels(someTensor, canvas, false); // -> Promise<void>, draws
```

tfjs-core/src/ops/browser.ts
mattsoulanille: @Linchenn, I think you might have more context on this.

tfjs-core/src/ops/to_pixels_test.ts
```diff
@@ -15,154 +15,22 @@
...
-    // tslint:disable-next-line:no-any
-    expectPromiseToFail(() => tf.browser.toPixels({} as any), done);
-  });
-
-  it('accepts a tensor-like object', async () => {
-    const x = [[10], [20]];  // 2x1;
-    const data = await tf.browser.toPixels(x);
+describeWithFlags('toPixels no canvas, returns void', ALL_ENVS, () => {
+  toPixelsTestCase(toPixelsNoCanvas, false);
+});
```
mattsoulanille, line 24: `toPixelsNoCanvas` always returns data. I don't think we need this test case.

axinging: Done, thanks

tfjs-core/src/ops/browser.ts
axinging: I add a check as below:
+  if (!gpuContext) {
+    throw new Error(
+        `Please make sure this canvas has only been used for webgpu context!`);
+  }

Linchenn: We may not need to expect the canvas and context.

If WebGL backend could draw to the canvas (same canvas as the backend and uses the same context), then draw it directly; otherwise, use CPU way to draw to it, but we could send a warning for this.

PR #7441, opened 2023-03-02
[Prototype] Record WebGL commands to speed up subsequent model execution

Linchenn:
# Benchmark results
With this optimization, the models, which pass the correctness tests, are accelerated on average:

MacBook Pro 2019 | Linux Workstation | Samsung Galaxy S23 Ultra
-- | -- | --
55% | 45% | 34%

Specifically:

image

## Performance regression
However, almost all models' performance become worse on iPhone 14 pro and Google Pix 7 Pro. One possible reason is that all dispose methods are removed, causing high memory usage and thus speed regression.
image

## Unsupported models
The following models failed model correctness tests:
MoveNet-MultiPose
Coco-SSD
DeepLabV3
AutoML Object
USE - batchsize 30
USE - batchsize 1
TextToxicity
posenet
bodypix
speech-commands

# PR's implementation details
For the intermediate tensors to be kept, we are supposed to keep the tensor only and release internal textures, but, since this PR is an initial prototype, we just removed all dispose methods.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7441)

### DIFF
--- a/tfjs-converter/src/executor/graph_executor.ts
+++ b/tfjs-converter/src/executor/graph_executor.ts
@@ -357,13 +357,6 @@ export class GraphExecutor implements FunctionExecutor {
     return tidy(exe);
   }
 
-
-  public cleanOutputs() {
-    for (const output of this.placeHolderOutputs) {
-      engine().backend.cleanTensor(output.dataId);
-    }
-  }
-
   public cloneTensors(tensors: Tensor[]) {
     return tensors.map(tensor => {
       const newTensor = clone(tensor);
@@ -375,7 +368,6 @@ export class GraphExecutor implements FunctionExecutor {
   }
 
   public replay(inputs: NamedTensorMap) {
-    this.cleanOutputs();
     for (const prop in inputs) {
       if (this.placeHolderInputs.hasOwnProperty(prop)) {
         engine().backend.bindInputToPlaceHolder(


chunnienc:
CPU: 8-core AMD Ryzen 7 PRO 6850U with Radeon Graphics (-MT MCP-) Speed: 2695 MHz
|Model | 300 Runs Avg., Head (ms) | 300 Runs Avg., This PR (ms) |
|--|---| ---|
|MobileNetV3 (small_075) | 8.6 (+/- 0.2) | 6.7 (+/- 0.2) |
|FaceLandmarkDetection (attention_mesh) | 33.1 (+/- 0.8) |  25.2 (+/- 0.3) |

tfjs-backend-webgl/src/backend_webgl.ts
```diff
@@ -983,6 +1041,15 @@ export class MathBackendWebGL extends KernelBackend {
     }
 
     if (!env().get('ENGINE_COMPILE_ONLY')) {
```
chunnienc, line 1043: always false here

tfjs-backend-webgl/src/kernels/Identity.ts
```diff
@@ -18,15 +18,27 @@
...
 
-  return {dataId: x.dataId, shape: x.shape, dtype: x.dtype};
+
+  const texData = backend.texData.get(dataId);
+  const {dtype, shape, isPacked} = texData;
+  let program;
+  if (isPacked) {
+    program = new UnaryOpPackedProgram(shape, unary_op.CLONE);
+  } else {
+    program = new UnaryOpProgram(shape, unary_op.CLONE);
```
chunnienc, line 39: weak guarantee: identity op always exists on output tensor

chunnienc (commented):
This PR shows how pure webgl-backend compatible model can run with recording and save almost all JS framework time. It should work as expected with our assumptions (no tensor and texture disposal, no arbitrary outputs, no CPU model logics in model execution, etc) and benchmark cases. 

The benchmark results are reliable. Although there will be more checks, recording logics, and copies in the production-ready recording framework, it won't be too different from the benchmarks here since those recording framework time are considered to be significantly less than our existing framework time and model execution time.

While it should work as a proof-of-concept for the performance improvement, the internal API design is far away from what I (personally) expected and a robust, extendable, and production-ready state. The internal recording API may land differently from this after heterogeneous backends and kernels implementations are considered into the design.

PR #7505, opened 2023-03-21
Fix graph executor to not dispose output tensors

### LINKED ISSUE #7504: regression: tensor is disposed early
upgrade to `tfjs` **4.3.0** from `tfjs` **4.2.0** and immediately hit a regression issue.
tested in **browser** using `webgl` and in **nodejs** using `tensorflow` backends, both have same problem.

basically, some tensors are disposed immediately after model execution.

simple reproduction

```js
const tf = require('@tensorflow/tfjs-node');

async function main() {
  console.log(`tfjs: ${tf.version_core} backend: ${tf.getBackend()}`);
  const model = await tf.loadGraphModel('https://vladmandic.github.io/human/models/faceres.json');
  const inputShape = model.inputs[0].shape?.map((x) => Math.abs(x));
  const input = tf.randomUniform(inputShape, 0, 1, model.inputs[0].dtype);
  const res = model.execute(input);
  for (const t of res) console.log(t.shape, t.isDisposed);
}

main();
```

result:

    tfjs: 4.3.0 backend: tensorflow
    [ 1, 1 ] false
    [ 1, 1024 ] true
    [ 1, 100 ] false

likely introduced via #7392 and attempted to fix via #7445 (but issue still exists)
also related issue #7431

i've tried a workaround by explicitly listing output tensors, but it does not help:

```js
      const outputs = model?.outputs.map((o) => o.name);
      const res = model.execute(input, outputs);
```

runtime error is obvious, but just to make it easily searchable by other users here's a stack trace:

```log
node_modules/.pnpm/@tensorflow+tfjs-core@4.3.0/node_modules/@tensorflow/tfjs-core/dist/tf-core.node.js:3882
            throw new Error("Tensor is disposed.");
                  ^

Error: Tensor is disposed.
    at Tensor.throwIfDisposed (node_modules/.pnpm/@tensorflow+tfjs-core@4.3.0/node_modules/@tensorflow/tfjs-core/dist/tf-core.node.js:3882:19)
    at Tensor.<anonymous> (node_modules/.pnpm/@tensorflow+tfjs-core@4.3.0/node_modules/@tensorflow/tfjs-core/dist/tf-core.node.js:3761:30)
    at step (node_modules/.pnpm/@tensorflow+tfjs-core@4.3.0/node_modules/@tensorflow/tfjs-core/dist/tf-core.node.js:140:27)
    at Object.next (.pnpm/@tensorflow+tfjs-core@4.3.0/node_modules/@tensorflow/tfjs-core/dist/tf-core.node.js:89:53)
    at node_modules/.pnpm/@tensorflow+tfjs-core@4.3.0/node_modules/@tensorflow/tfjs-core/dist/tf-core.node.js:82:71
    at new Promise (<anonymous>)
    at __awaiter (node_modules/.pnpm/@tensorflow+tfjs-core@4.3.0/node_modules/@tensorflow/tfjs-core/dist/tf-core.node.js:68:12)
    at Tensor.data (node_modules/.pnpm/@tensorflow+tfjs-core@4.3.0/node_modules/@tensorflow/tfjs-core/dist/tf-core.node.js:3756:16)
```

chunnienc:
Fix #7504 
Added a new test case for this and verified that the old executor fails the test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7505)


tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -380,19 +382,23 @@ export class GraphExecutor implements FunctionExecutor {
...
 
-    for (const inputNodeName of liveUntilNodes) {
-      const tensors =
-          getTensorsForCurrentContext(inputNodeName, tensorMap, context);
+    for (const liveUntilThisNode of liveUntilNodes) {
+      if (isNonDisposableNode(liveUntilThisNode)) {
+        continue;
+      }
+      const tensors = getTensorsForCurrentContext(
+          liveUntilThisNode.name, tensorMap, context);
```
mattsoulanille, line 401: Nit on the naming: For a node `n_k` in the graph, `liveUntilNodes_k` is the set of nodes `{n_a, n_b, ...}` that will be disposed immediately after `n_k` is run. 

```suggestion
    for (const nodeToDispose of liveUntilNodes) {
      if (isNonDisposableNode(nodeToDispose)) {
        continue;
      }
      const tensors = getTensorsForCurrentContext(
          nodeToDispose.name, tensorMap, context);
```

tfjs-converter/src/executor/graph_executor.ts
```diff
@@ -380,19 +382,23 @@ export class GraphExecutor implements FunctionExecutor {
...
-      return;
+      liveUntilNodes?: Node[]) {
+    function isNonDisposableNode(node: Node) {
+      // Skip output nodes and any control flow nodes, since its dependency is
+      // tricky to track correctly.
+      return isControlFlow(node) || outputNodeNameSet.has(node.name);
     }
-    if (liveUntilNodes == null) {
+
+    if (isControlFlow(node) || liveUntilNodes == null) {
```
mattsoulanille, line 392: I see you moved the check for node disposability to check the liveUntilNodes (this makes sense since those are the ones we're disposing). Do we need this `isControlFlow(node)` check? That means we'll never dispose any nodes that live until a control flow node if I'm not mistaken.

Is this to make sure we always have a control flow node's inputs available (e.g. in the case of a while)?

chunnienc: > Is this to make sure we always have a control flow node's inputs available (e.g. in the case of a while)?
Yes. That is the behavior of the old algorithm. The branch here check whether the current node allow us to dispose the input of it, while the checks in the loop are for whether we should dispose the input itself. See https://github.com/tensorflow/tfjs/pull/7505/files#diff-44e0a825cd7c6f31c03d9333db5dc21a8937d66a5b28d719c699863eef96ad8dR329

But the behavior is changed. The old function does not allow disposing inputs of an output node no matter it's control flow or not, while this PR allows it. I don't see any current or future problems can have with it.


> Do we need this isControlFlow(node) check? 
No, bc a graph with control flow should not be sync. But I still keep it in case the upstream graph builder has some special cases.

mattsoulanille (approved):
LGTM

pyu10055 (approved):
thank you for fixing this bug, given this is not caught by our e2e model tests, can we add an e2e test for mobilenet graph model conversion to prevent future breakage.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7505#-:-NR4cx5c4r0m7faTFbZY:b7mp81m)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @chunnienc and @mattsoulanille)

vladmandic:
> This bug only occurs when the output nodes are not leaves or parents of leaves in the graph

Btw, isn't that somewhat typical for feature vectors?

chunnienc:
> Btw, isn't that somewhat typical for feature vectors?

Yes. But it turns out that all our (converted) models for testing and benchmarking have an additional "Identity" op node after each model output node, which makes them always leaves. It's something we need to fix on our side to make the tests more robust.

PR #7511, opened 2023-03-23
Fix latest npm package version retrieval in scripts

chunnienc:
This PR is for fixing the nightly tests.
Since our scripts use command `npm view <pkg> dist-tags.latest` to get the latest version in npm, they can only get the version with the "latest" tag instead of the one with the largest version number. It leads to some problems since our latest version (4.3.0) is deprecated and the "latest" tag is associated to an older version (4.2.0). This PR fix the latest version retrieval to get the global maximum version number instead of one with latest tag.

This PR is not verified with all release/ci scripts. I/we will send separate PRs if we have other problems with versioning when running the release/ci.

Nightly passed: https://pantheon.corp.google.com/cloud-build/builds;region=global/c27e24d2-ca5d-46d2-9f0c-875f66a2a5cb?project=learnjs-174218

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7511)

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -98,8 +98,10 @@ async function main() {
 
   if (phaseInt !== 0) {
     // Phase0 should be published and release branch should have been created.
-    const firstPackageLatestVersion =
-        $(`npm view @tensorflow/${phases[0].packages[0]} dist-tags.latest`);
+    const firstPackageVersions: string[] = JSON.parse(
+        $(`npm view @tensorflow/${phases[0].packages[0]} versions --json`));
+    const firstPackageLatestVersion = semver.rsort(firstPackageVersions)[0];
+
     releaseBranch = `${name}_${firstPackageLatestVersion}`;
 
     $(`git clone -b ${releaseBranch} ${urlBase}tensorflow/tfjs ${
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -28,6 +28,7 @@
 
 import * as argparse from 'argparse';
 import chalk from 'chalk';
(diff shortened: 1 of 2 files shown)


scripts/release.ts
```diff
@@ -120,8 +121,9 @@ async function main() {
     const packageJsonPath = `${dir}/${packageName}/package.json`;
     let pkg = `${fs.readFileSync(packageJsonPath)}`;
     const parsedPkg = JSON.parse(`${pkg}`);
-    const latestVersion =
-        $(`npm view @tensorflow/${packageName} dist-tags.latest`);
+    const packageVersions: string[] =
+        JSON.parse($(`npm view @tensorflow/${packageName} versions --json`));
+    const latestVersion = semver.rsort(packageVersions)[0];
```
mattsoulanille, line 127: Nit: Can we factor this and the other instance out into a `getVersions` or `getLatestVersion` function?

mattsoulanille (approved):
LGTM

Linchenn (approved):
LGTM, thanks!

scripts/release.ts
chunnienc: We already have these functions (`getNpmVersion(..., tag='latest')`, etc) in `release-util.ts`, which are tag based and for `publish-npm.ts`. I'm not sure whether we should change all the latest tag based function to use maximum version number or not, it's hard to trace them all. I think the best approach is to only fix them if we actually see an error when running the script since some of them may actually depend the latest tagged version.

PR #7527, opened 2023-03-28
[e2e] Fix golden file test indexing nonexistent tensor

mattsoulanille:
Use `.toEqual(jasmine.anything())` instead of `.not.toEqual(null)` to catch both `null` and `undefined`.

Use `model.outputs.map(output => output.name)` instead of `model.outputNodes` to get the output node name without the numerical index.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7527)


chunnienc (approved):
Did you verify that jasmine.anything() can fail the unexpected node name?

mattsoulanille:
> Did you verify that jasmine.anything() can fail the unexpected node name?

I verified that `expect(x).toEqual(jasmine.anything())` will fail if `x` is `null` or `undefined` and will pass when `x` is defined. I'm checking for missing node names when I map the `targetNodeNames` to the golden node details. Is that what you are asking, or did I misunderstand?

PR #7547, opened 2023-04-05
Update monorepo to 4.4.0.

mattsoulanille:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7547)


tfjs-backend-wasm/starter/parcel/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "@tensorflow/tfjs-backend-wasm": "3.14.0",
-    "@tensorflow/tfjs-core": "3.14.0"
+    "@tensorflow/tfjs-core": "4.4.0"
```
chunnienc, line 11: We shouldn't set core and wasm backend in different major versions.

tfjs-backend-wasm/starter/webpack/package.json
```diff
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "@tensorflow/tfjs-backend-wasm": "2.6.0",
```
chunnienc, line 11: same here

tfjs-backend-wasm/starter/parcel/package.json
mattsoulanille: Oh, that should not have happened. Nice catch! I'll take a look.

mattsoulanille:
Closing this PR to regenerate the 4.4.0 PR with a fix for the dependency version issue.

PR #7566, opened 2023-04-10
[prototype] Record-Replay API and implementation for production

chunnienc:
#### General
- [x] Make basic command types and tfjs-core level recording
- [x] Make `TensorPlaceholder`: a tensor/tensor info holder for passing tensors between commands
  - `TensorPlaceholder` naturally solves the issue of how to dispose/reuse/detach tensors. Those are no longer issues since tensors in TensorPlaceholder should be different (have different id) across multiple runs.
- [x] Make WebGLProgrammCommand to record most webgl
- [x] Make mechanism to let op select whether to use auto recorder or built-in recorder, depending on recording support of kernel 
- [ ] Modify sync graph executor to use `tf.<op>._record(...)` to record commands in graph execution

#### Benchmark
- [x] Benchmark non-recording model inference time (no performance loss in MobileNetV3)
- [ ] Benchmark MobileNetV3 with recording (use builtin recorder for all ops and assume correct)
- [ ] Benchmark MobileNetV3 with recording (use auto recorder for all ops, and make sure no performance loss with ClosureCommand) 

#### Compatibility
- [x] Make sure model inference without recording works with new tfjs-core and old backend (and vice versa)
- [ ] Make sure model inference without recording works with new tfjs-converter and old tfjs-core
- [ ] Make sure all combinations work

#### JIT prototype and exploration
- [ ] (JIT) Cache and reuse packed/unpacked versions of the same tensor in TensorPlaceholder to avoid redundant pack/unpack computations.
- [ ] (JIT) Options to record input/output pack/unpack state to make recorded command faster
- [ ] (JIT) Build simple model inference memory plan with static analysis (based on command in/out TensorPlaceholders)
- [ ] (JIT) Simple `WebGLProgramCommand` dynamic fuser

### Usage:
Remember to use local core and backend-webgl builds (of course you can use only one to test compatibility).
You may also set `tf.env().set('WEBGL_CPU_FORWARD', false)`. In that case `tf.add._record(x, y)` would produce different command, since the cpu forwarded one produces `ClosureCommand` for CPU computations while the non-forwarded one produces `WebGLProgramCommand`.
Currently only `tf.add` has builtin recording mode (conditionally used based on your choice of backend). All the other ops use auto recorder.
```js
x = tf.tensor1d([1, 1]);
y = tf.tensor1d([2, 2]);
let {tape, outputs} = tf.add._record(x, y);
outputs.print(); // Tensor[3, 3]. `outputs` is the output of `tf.add(x, y)`, it can be used as a normal tensor.

cmd = rec.tape.commands[0];
console.log('*** Recorded command: ***', cmd);

cmd.inputs[0].set(x);
cmd.inputs[1].set(tf.tensor1d([10, 20]));
cmd.execute();
cmd.outputs[0].toTensor().print(); // Tensor[11, 21]
cmd.outputs[0].reset() // Dispose tensor(s) held in the placeholder

cmd.inputs[0].set(tf.tensor1d([5, 5]))
cmd.execute();
cmd.outputs[0].toTensor().print(); // Tensor[15, 25]
x = cmd.outputs[0].releaseTensor(); 

// Now the output placeholder no longer holds the tensor.
// x can be used as a regular, independent tensor
```

### DIFF
--- a/tfjs-core/src/engine.ts
+++ b/tfjs-core/src/engine.ts
@@ -120,13 +120,21 @@ interface GradientTapeState {
   kernelDepth: number;
 }
 
+interface TensorPlaceholderConfig {
+  readonly allowOverwriteTensor: boolean;
+}
+
 class TensorPlaceholder {
+  static readonly DEFAULT_CONFIG: TensorPlaceholderConfig = {
+    allowOverwriteTensor: true,
+  };
   // A map from templateDataId to built TensorPlaceholder. When the refCount of
   // a TensorPlaceholder decreases to zero, its corresponding entry in the pool
   // will be deleted.
   private static pool =
       new WeakMap</*templateDataId=*/DataId, TensorPlaceholder>();
 
+  readonly config: TensorPlaceholderConfig;
   readonly template: TensorInfo;
   /** Unique id of this placeholder. */
   readonly pid: DataId;
@@ -136,7 +144,10 @@ class TensorPlaceholder {
   private refCount = 1;
   private disposed = false;
 
-  private constructor(template: TensorInfo|Tensor) {
+  private constructor(
+      template: TensorInfo|Tensor, config?: Partial<TensorPlaceholderConfig>) {
+    this.config = {...TensorPlaceholder.DEFAULT_CONFIG, ...(config ?? {})};
+
     this.template = {
       dataId: template.dataId,
       shape: [...template.shape],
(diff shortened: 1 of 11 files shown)


chunnienc:
OS: Win11 CPU: 8-core AMD Ryzen 7 PRO 6850U with Radeon Graphics (-MT MCP-) Speed: 2695 MHz
|Model | 200 Runs Avg., 4.4.0 (ms) | 200 Runs Avg., This PR (ms) |
|--|---| ---|
|MobileNetV3 (small_075) | 6.9 (+/- 0.2) | 5.8 (+/- 0.2) |
| FaceDetection | 5.5 (+/- 0.2) |  3.5 (+/- 0.3) |

tfjs-core/src/engine.ts
```diff
@@ -109,7 +110,410 @@ function isRegisteredKernelInvocation<T extends Tensor|Tensor[],
...
+  // Number of nested kernel calls. When kernel depth is greater than 1, we turn
+  // off the tape.
+  kernelDepth: number;
+}
+
+export interface TensorPlaceholderConfig {
+  readonly allowOverwriteTensor: boolean;
+}
+
+export class TensorPlaceholder {
```
chunnienc, line 128: @mattsoulanille

PR #7582, opened 2023-04-13
[e2e] Support set env flag from url

axinging:
For most tfjs app, URL parameter like tfjsflags=WEBGPU_PRINT_SHADER:binary will be recognized as setting tfjs environment flag value (If it has been registered). For e2e, this logic has been broken due to setEnvFlags will override values for all existing flags.

This change supports setting environment flag values from URL parameters in e2e, example: tfjsflags=WEBGPU_PRINT_SHADER:binary. When a flag is set in both GUI(tunable flag) and URL parameter, URL parameter will override flag from GUI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7582)

### DIFF
--- a/e2e/benchmarks/benchmark_util.js
+++ b/e2e/benchmarks/benchmark_util.js
@@ -109,8 +109,8 @@ function generateInputFromDef(inputDefs, isForGraphModel = false) {
         generatedRaw.dispose();
       } else {
         throw new Error(
-          `The ${inputDef.dtype} dtype of '${inputDef.name}' input ` +
-          `at model.inputs[${inputDefIndex}] is not supported.`);
+            `The ${inputDef.dtype} dtype of '${inputDef.name}' input ` +
+            `at model.inputs[${inputDefIndex}] is not supported.`);
       }
       tensorArray.push(inputTensor);
     });
@@ -162,8 +162,8 @@ function getPredictFnForModel(model, input) {
     predict = () => model.predict(input);
   } else {
     throw new Error(
-      'Predict function was not found. Please provide a tf.GraphModel or ' +
-      'tf.LayersModel');
+        'Predict function was not found. Please provide a tf.GraphModel or ' +
+        'tf.LayersModel');
   }
   return predict;
 }
@@ -241,8 +241,8 @@ async function timeModelInference(model, input, numRuns = 1) {
 async function timeInference(predict, numRuns = 1) {
   if (typeof predict !== 'function') {
     throw new Error(
-      'The first parameter should be a function, while ' +
-      `a(n) ${typeof predict} is found.`);
+        'The first parameter should be a function, while ' +
+        `a(n) ${typeof predict} is found.`);
   }
 
   const times = [];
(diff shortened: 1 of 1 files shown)


axinging:
@qjia7 @xhcao @gyagp PTAL

qjia7 (commented):
Hi Xing, can you remove the refactor related codes and make this PR only do the url flag? It's will be easier for review. Thanks.

qjia7 (commented):
This is useful to test some flags not in the benchmarks tunable list. 
For example, WebGL's `WEBGL_EXP_CONV`.
@mattsoulanille @Linchenn Please take a look, thanks.

e2e/benchmarks/benchmark_util.js
```diff
@@ -553,6 +557,40 @@ async function setEnvFlags(flagConfig) {
   }
 }
 
+/**
+ * Set flags from URL. URL should be in the format:
+ * ?tfjsflags=FLAG1:1,FLAG2:true.
+ */
+function setEnvFlagsFromUrlParameters() {
+  const TENSORFLOWJS_FLAGS_PREFIX = 'tfjsflags';
```
gyagp, line 565: Maybe it's better to call it envflags?

e2e/benchmarks/benchmark_util.js
```diff
@@ -553,6 +557,40 @@ async function setEnvFlags(flagConfig) {
...
+        console.error(err);
+      }
+    });
+  }
+}
+
+/**
+ * Converted a URL parameter to a typed value, such a boolean, number, string.
+ */
+function parseValue(value) {
```
gyagp, line 583: Can we export this function from environment.ts?

e2e/benchmarks/benchmark_util.js
```diff
@@ -502,8 +503,9 @@ const TUNABLE_FLAG_VALUE_RANGE_MAP = {
 /**
  * Set environment flags for testing.
  *
- * This is a wrapper function of `tf.env().setFlags()` to constrain users to
- * only set tunable flags (the keys of `TUNABLE_FLAG_TYPE_MAP`).
+ * This will first set tunable flags (the keys of `TUNABLE_FLAG_TYPE_MAP`). Then
+ * set URL parameter flags. If there are overlap, URL parameter flags will
+ * override tunable flags.
```
gyagp, line 508: We also need to change the description here accordingly.

mattsoulanille (approved):
LGTM

e2e/benchmarks/benchmark_util.js
axinging: This feature has been supportted by tfjs-core. We add this because e2e is an exception(e2e supports both url flags and tunable flags). For most app, this logic already works. I didn't see any other tfjs app requires this API.

e2e/benchmarks/benchmark_util.js
axinging: In tfjs-core, this is also called tfjsflags.

e2e/benchmarks/benchmark_util.js
axinging: Do you have any detailed suggestion here?

gyagp: LGTM

e2e/benchmarks/benchmark_util.js
gyagp: Then I'm fine.

e2e/benchmarks/benchmark_util.js
gyagp: I just hope to reuse that part of code. But I'm fine as it is now.

gyagp (approved):

Linchenn (approved):

Issue #7577, opened 2023-04-12
Parallel shader compilation is broken


shanumantesc:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): No
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): M1 Mac 13.3.1 (22E261)
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device: Reproducible on any device
- TensorFlow.js installed from (npm or script link):
- TensorFlow.js version (use command below): 4.2.0
- Browser version: Chrome Version 112.0.5615.49
- Tensorflow.js Converter Version: N/A

**Describe the current behavior**
- We use `ENGINE_COMPILE_ONLY` to speed up initial model loading time by not blocking on shader compilation status checks.
- After https://github.com/tensorflow/tfjs/pull/6913, even if we set `ENGINE_COMPILE_ONLY` env variable, we end up calling `getAttributeLocation` when we call `bindVertexProgramAttributeStreams`.
- This introduces a synchronization point for the graphics pipeline and we block on shader compilation. A side issue is that TFJS calls `getAttributeLocation` without first checking for `gl.LINK_STATUS`
- See screenshot below which contains Chrome trace of a model which is being run with `ENGINE_COMPILE_ONLY` (running on M1 mac Chrome)
<img width="399" alt="51667c9b-5733-488c-afac-126e5cd357e6" src="https://user-images.githubusercontent.com/71482681/231530895-d57bc361-b529-4edb-9fb6-97f84af951cf.png">

**Describe the expected behavior**
- TFJS should check `ENGINE_COMPILE_ONLY` before calling `bindVertexProgramAttributeStreams` so that we do not stall GPU pipeline. We can set up the VAO in `checkCompileCompletion`.

**Standalone code to reproduce the issue**
- Run any network with `ENGINE_COMPILE_ONLY` set to true, capture chrome perf trace, and see that `getAttributeLocation` is called for each call to `createProgram`

**Other info / logs** Include any logs or source code that would be helpful to
diagnose the problem. If including tracebacks, please include the full
traceback. Large logs and files should be attached.

Linchenn:
@shanumantesc Thanks for reporting this! I just verified the [fix](https://github.com/tensorflow/tfjs/pull/7587) and it should work now.

gaikwadrahul8:
Hi, @shanumantesc

[...]

PR #7587, opened 2023-04-14
[WebGL] Fix Vao usage for parallel compile feature

Linchenn:
For https://github.com/tensorflow/tfjs/issues/7577

### Problem
If we create VAO for each webgl program, then we have to set VAO for each program and bind it before execution. Right now, setting VAO happens in `createProgram` and `createProgram` is involved in compilation stage. However, setting VAO has blocking-call `getAttribLocation`, so setting VAO should be called after compilation stage.  

This should work similar as what we do for `getUniformLocations`. We moved `getUniformLocations` call out of `createProgram`/compilation-stage, because it has blocking-call.

#### before https://github.com/tensorflow/tfjs/pull/6913 is merged (tfjs 4.0.0)
image
`checkCompileCompletion` takes the majority time for warm up.

#### after https://github.com/tensorflow/tfjs/pull/6913 is merged (tfjs 4.1.0)
image

`checkCompileCompletion` takes the trivial time for warm up.
image
`getAttribLocation` is blocking in the compilation stage.

#### after this PR
image

`checkCompileCompletion` takes the majority time for warm up.

#### codes
I used the following codes to reproduce the problem
```html
<!DOCTYPE html>
<html>
  <body>

    // Local build
    <script src="https://unpkg.com/@tensorflow/tfjs-core@latest/dist/tf-core.js"></script>
    <script src="https://unpkg.com/@tensorflow/tfjs-converter@latest/dist/tf-converter.js"></script>
    <script src="./dist/bin/tfjs-backend-webgl/dist/tf-backend-webgl.js"></script>

    // After https://github.com/tensorflow/tfjs/pull/6913/files is merged
    <!-- <script src="https://unpkg.com/@tensorflow/tfjs-core@4.1.0/dist/tf-core.js"></script>
    <script src="https://unpkg.com/@tensorflow/tfjs-converter@4.1.0/dist/tf-converter.js"></script>
    <script src="https://unpkg.com/@tensorflow/tfjs-backend-webgl@4.1.0/dist/tf-backend-webgl.js"></script> -->

    // Before https://github.com/tensorflow/tfjs/pull/6913/files is merged
    <!-- <script src="https://unpkg.com/@tensorflow/tfjs-core@4.0.0/dist/tf-core.js"></script>
    <script src="https://unpkg.com/@tensorflow/tfjs-converter@4.0.0/dist/tf-converter.js"></script>
    <script src="https://unpkg.com/@tensorflow/tfjs-backend-webgl@4.0.0/dist/tf-backend-webgl.js"></script> -->


    <script>
      const url = `https://tfhub.dev/google/tfjs-model/imagenet/mobilenet_v3_small_075_224/classification/5/default/1`;
      model = tf.loadGraphModel(url, {fromTFHub: true});

      function normalWarmUp() {
        return model.predict(input);
      }

      function parallelWarmUp() {
        tf.env().set('ENGINE_COMPILE_ONLY', true);
        model.predict(input);
        tf.env().set('ENGINE_COMPILE_ONLY', false);
        tf.backend().checkCompileCompletion();
        tf.backend().getUniformLocations();
        return model.predict(input);
      }

      function cleanUp() {
        tf.backend().binaryCache = {};
      }

      async function benchmark(fn) {
        model = await model;
        const start = performance.now();
        fn().dataSync();
        return performance.now() - start;
      }

    </script>
  </body>
</html>

```

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7587)


Linchenn:
The problem is caused by https://github.com/tensorflow/tfjs/pull/6913, which is trying to bind VAO before each webgl program execution. I think the idea makes sense, even though TFJS's vertex shaders are same, because users may use the same canvas and write their own webgl program with different vertex shaders.

tfjs-backend-webgl/src/backend_webgl.ts
```diff
@@ -1277,6 +1277,7 @@ export class MathBackendWebGL extends KernelBackend {
 
   getUniformLocations() {
     for (const binary of Object.values(this.binaryCache)) {
+      this.gpgpu.buildVao(binary.webGLProgram);
```
Linchenn, line 1280: It does not make sense to add this code within `getUniformLocations` and theoratically we should move it to a new function, like 'setVao'. However, since we have told some users to use this to utilize parallel compilation feature, making `setVao` and requiring users to call it are breaking change for them.

tfjs-backend-webgl/src/gpgpu_math.ts
```diff
@@ -253,6 +254,7 @@ export function runProgram<T extends Tensor, K extends Tensor>(
         outTex.texture, outTexShape[0], outTexShape[1]);
   }
   gpgpu.setProgram(binary.webGLProgram);
```
pyu10055, line 256: should this two line be the same as just call gpgpu.buildVao(binary.webGLProgram)?

pyu10055 (commented):
thanks for fixing the issue!

tfjs-backend-webgl/src/gpgpu_math.ts
Linchenn: IICU, `buildVao` is to initialize VAO, while here it is using initialized VAO. Specifically, VAO initialization has to bind all attributes, while using initialized VAO does not need to.

pyu10055 (approved):

tfjs-backend-webgl/src/backend_webgl.ts
mattsoulanille: Nit: Can you add a TODO here explaining this? Thanks!

mattsoulanille (approved):
LGTM with a nit.

tfjs-backend-webgl/src/backend_webgl.ts
Linchenn: Done. Thanks!

mattsoulanille (approved):
LGTM. Thanks for adding the TODO.

PR #7598, opened 2023-04-18
Avoid allocating a large arraybuffer when loading weights

mattsoulanille:
The [loadWeights](https://github.com/mattsoulanille/tfjs/blob/avoid_large_arraybuffer/tfjs-core/src/io/weights_loader.ts#L82) function loads weights in 4MB chunks and then concatenates them into a single large ArrayBuffer. That ArrayBuffer is used for splitting the weights data back up into tensors. Allocating large ArrayBuffers (3.5GB) can be unstable on Chrome, so this PR avoids this allocation, instead slicing the weights out of the chunks manually.

The implementation wraps the array of weights (stored as `ArrayBuffer[]`) in a new `CompositeArrayBuffer` class. This class implements `slice` by copying the desired range out of the buffer(s) that it overlaps with.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7598)

### DIFF
--- a/tfjs-core/src/io/weights_loader.ts
+++ b/tfjs-core/src/io/weights_loader.ts
@@ -18,6 +18,7 @@
 import {env} from '../environment';
 
 import {NamedTensorMap} from '../tensor_types';
+import {TypedArray} from '../types';
 import * as util from '../util';
 import {decodeWeights} from './io_utils';
 import {monitorPromisesProgress} from './progress';
@@ -247,10 +248,19 @@ export class CompositeArrayBuffer {
   private bufferUniformSize?: number;
   public readonly byteLength: number;
 
-  constructor(buffers: ArrayBuffer | ArrayBuffer[]) {
-    if (buffers instanceof ArrayBuffer) {
+  constructor(buffers: ArrayBuffer | ArrayBuffer[] | TypedArray | TypedArray[]) {
+    // Normalize the `buffers` input to be `ArrayBuffer[]`.
+    if (!(buffers instanceof Array)) {
       buffers = [buffers];
     }
+    buffers = buffers.map((bufferOrTypedArray) => {
+      if (util.isTypedArray(bufferOrTypedArray)) {
+        return bufferOrTypedArray.buffer;
+      }
+      return bufferOrTypedArray;
+    });
+
+    // Skip setting up ranges if there are no buffers.
     if (buffers.length === 0) {
       return;
     }
--- a/tfjs-core/src/io/weights_loader.ts
+++ b/tfjs-core/src/io/weights_loader.ts
@@ -264,7 +264,7 @@ class CompositeArrayBuffer {
         this.bufferUniformSize = undefined;
       }
 
-      // Create the ranges, including their start and end poionts.
+      // Create the ranges, including their start and end points.
       const end = start + buffer.byteLength;
       this.ranges.push({buffer, start, end,});
       start = end;
@@ -349,6 +349,8 @@ class CompositeArrayBuffer {
     }
 
     // Otherwise, use a generic search function.
+    // This should almost never end up being used in practice since the weight
+    // entries should always be in order.
     const index = search(this.ranges, check);
     if (index === -1) {
       return -1;
@@ -369,6 +371,7 @@ class CompositeArrayBuffer {
  *     greater than the value passed to the function.
  */
 function search<T>(sortedArray: T[], compare: (t: T) => number): number {
+  // Binary search
   let min = 0;
   let max = sortedArray.length;
 
--- a/tfjs-core/src/io/weights_loader.ts
+++ b/tfjs-core/src/io/weights_loader.ts
@@ -212,24 +212,80 @@ export function weightsLoaderFactory(
     groupIndicesToFetch.forEach(i => {
       const numBuffers = manifest[i].paths.length;
 
-      let groupBytes = 0;
-      for (let i = 0; i < numBuffers; i++) {
-        groupBytes += buffers[bufferIndexOffset + i].byteLength;
-      }
+      // let groupBytes = 0;
+      // for (let i = 0; i < numBuffers; i++) {
+      //   groupBytes += buffers[bufferIndexOffset + i].byteLength;
+      // }
 
       // Create a buffer for the whole group.
-      const groupBuffer = new ArrayBuffer(groupBytes);
-      const groupByteBuffer = new Uint8Array(groupBuffer);
-      let groupBufferOffset = 0;
-      for (let i = 0; i < numBuffers; i++) {
-        const buffer = new Uint8Array(buffers[bufferIndexOffset + i]);
-        groupByteBuffer.set(buffer, groupBufferOffset);
-        groupBufferOffset += buffer.byteLength;
+      // const groupBuffer = new ArrayBuffer(groupBytes);
+      // const groupByteBuffer = new Uint8Array(groupBuffer);
+      // let groupBufferOffset = 0;
+      // for (let i = 0; i < numBuffers; i++) {
+      //   const buffer = new Uint8Array(buffers[bufferIndexOffset + i]);
+      //   groupByteBuffer.set(buffer, groupBufferOffset);
+      //   groupBufferOffset += buffer.byteLength;
+      // }
+
+
+
+      const weightsEntries = [...groupWeightsToFetch[i]];
+      weightsEntries.sort((a, b) => a.groupOffset - b.groupOffset);
+
+      let bufferIndex = 0;
+      let precedingBytes = 0;
+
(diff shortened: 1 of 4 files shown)


mattsoulanille:
@pyu10055 This isn't a full review request yet, but I'm interested in knowing which approach you think is better. Thanks!

chunnienc:
If the size of each chunk is fixed (4MB) except for the last one, you can get start and end offsets in O(1) by division and module.

mattsoulanille:
> If the size of each chunk is fixed (4MB) except for the last one, you can get start and end offsets in O(1) by division and module. No need to do binary search or two-pointers approach on sorted data.
> 
> In terms of API, I prefer option 2, which is a better implementation for separation of concerns. In the weight loader I just need to specify where to slice and don't need to worry about how to slice. You cam implement lazy/offline slicer If performance is a concern.

Unfortunately, there's no guarantee on the size of the chunks. We let people configure it when converting the model. They should all be the same size, but I'd even hesitate to assume that, since it seems a bit flaky.

I agree with you and also prefer option 2. I'll implement it as a binsearch for now, and if we need better perf, I can try to automatically detect the chunk size or make it check chunks near the last one read before doing a full binsearch.

mattsoulanille:
Looking at this again, it doesn't actually prevent us from storing the weights in a single ArrayBuffer. `ModelArtifacts` contains the `weightData` key, [which stores the model weights as a single ArrayBuffer](https://github.com/mattsoulanille/tfjs/blob/tfjs-ulm/tfjs-core/src/io/types.ts#L254). This is [constructed by IOHandlers like http](https://github.com/mattsoulanille/tfjs/blob/tfjs-ulm/tfjs-core/src/io/http.ts#L213), which load the weights and then concatenate them into a single ArrayBuffer.

tfjs-core/src/io/weights_loader.ts
```diff
@@ -245,3 +235,179 @@ export function weightsLoaderFactory(
...
   };
 }
+
+type BufferRange = {
+  start: number,
+  end: number,
+  buffer: ArrayBuffer,
+};
+
+export class CompositeArrayBuffer {
```
mattsoulanille, line 245: This will be used in another PR that enables large model weights to be stored in a list of ArrayBuffers. That's why it's exported here.

mattsoulanille:
> Looking at this again, it doesn't actually prevent us from storing the weights in a single ArrayBuffer. `ModelArtifacts` contains the `weightData` key, [which stores the model weights as a single ArrayBuffer](https://github.com/mattsoulanille/tfjs/blob/tfjs-ulm/tfjs-core/src/io/types.ts#L254). This is [constructed by IOHandlers like http](https://github.com/mattsoulanille/tfjs/blob/tfjs-ulm/tfjs-core/src/io/http.ts#L213), which load the weights and then concatenate them into a single ArrayBuffer.
> 
> I'll leave this PR as-is and create a new one to fix this issue.

I'm sending this out for review since it's easier to review separately from the other part of the large weights fix. I'll submit the other part, which integrates this code with the rest of the codebase, in a separate PR.

tfjs-core/src/io/weights_loader.ts
```diff
@@ -245,3 +235,180 @@ export function weightsLoaderFactory(
...
+      if (i !== buffers.length - 1 &&
+        buffer.byteLength !== this.bufferUniformSize) {
+        // Unset the buffer uniform size, since the buffer sizes are not
+        // uniform.
+        this.bufferUniformSize = undefined;
+      }
+
+      // Create the ranges, including their start and end points.
+      const end = start + buffer.byteLength;
+      this.ranges.push({buffer, start, end,});
```
chunnienc, line 284: nit: remove ',' or format to multiple lines

tfjs-core/src/io/weights_loader.ts
```diff
@@ -245,3 +235,180 @@ export function weightsLoaderFactory(
...
+
+    // Set the byteLenghth
+    if (this.ranges.length === 0) {
+      this.byteLength = 0;
+    }
+    this.byteLength = this.ranges[this.ranges.length - 1].end;
+  }
+
+  slice(start = 0, end = this.byteLength): ArrayBuffer {
+    // NaN is treated as zero for slicing. This matches ArrayBuffer's behavior.
```
chunnienc, line 296: convert start and end to Number with `Number(...)` before checking NaN?

tfjs-core/src/io/weights_loader.ts
```diff
@@ -245,3 +235,180 @@ export function weightsLoaderFactory(
     return weightsTensorMap;
   };
 }
+
+type BufferRange = {
```
chunnienc, line 239: Naming: range -> chunk/shard/partition
And all related variable and function names

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7598)* status: 0 of 1 approvals obtained (waiting on @chunnienc and @mattsoulanille)

___
*[`tfjs-core/src/io/weights_loader.ts` line 290 at r2](https://reviewable.io/reviews/tensorflow/tfjs/7598#-NTKHwuV5I4hk3pxoc30:-NTKHwuV5I4hk3pxoc31:b2kvmkm) ([raw file](https://github.com/tensorflow/tfjs/blob/076c8a2b8b2f9f6ee2db8b2b80827e0480e9b7fc/tfjs-core/src/io/weights_loader.ts#L290)):*
> ```XML
>       }
>     }
>     return outputBuffer
> ```

missing `;`
___
*[`tfjs-core/src/io/weights_loader.ts` line 292 at r2](https://reviewable.io/reviews/tensorflow/tfjs/7598#-NTKI8XiDXgr4-BmgPA7:-NTKI8XiDXgr4-BmgPA8:b-148luq) ([raw file](https://github.com/tensorflow/tfjs/blob/076c8a2b8b2f9f6ee2db8b2b80827e0480e9b7fc/tfjs-core/src/io/weights_loader.ts#L292)):*
> ```XML
>     return outputBuffer
>   }
>   private search(byteIndex: number) {
> ```

this could be improved if the searching is in order, I believe that is how our weights are setup.
___
*[`tfjs-core/src/io/weights_loader.ts` line 245 at r6](https://reviewable.io/reviews/tensorflow/tfjs/7598#gh-1171960442:-NTUi-6u8g06213ElrCP:bb0dhy) ([raw file](https://github.com/tensorflow/tfjs/blob/6a0754736b33075ffd0b47ea2c4fb8efc22f8465/tfjs-core/src/io/weights_loader.ts#L245)):*

This will be used in another PR that enables large model weights to be stored in a list of ArrayBuffers. That's why it's exported here.
</blockquote></details>

It might be good to be in a separate file
___
*[`tfjs-core/src/io/weights_loader.ts` line 272 at r9](https://reviewable.io/reviews/tensorflow/tfjs/7598#-NTUgq5O6_F7L1D1ndeA:-NTUgq5O6_F7L1D1ndeB:bx3s9tt) ([raw file](https://github.com/tensorflow/tfjs/blob/8287312a2446569908657ee86dba5bf0a4b1dfed/tfjs-core/src/io/weights_loader.ts#L272)):*
> ```XML
>     let start = 0;
> 
>     for (let i = 0; i < buffers.length; i++) {
> ```

start from 1?

chunnienc (approved):

tfjs-core/src/io/weights_loader.ts
mattsoulanille: Good point. That's a much better name. Fixed.

tfjs-core/src/io/weights_loader.ts
chunnienc: Update:
I assume you add these nan checks because you think there may be calls from JS now or future which ignores the typescript type check. In these way I'd suggest to do `start = Number(start)` since `isNaN('123')` returns `true` and `ArrayBuffer.prototype.slice` accepts the numbers in strings.

pyu10055 (approved):
> Reviewed 3 of 4 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7598)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @chunnienc and @mattsoulanille)

tfjs-core/src/io/weights_loader.ts
mattsoulanille: I added these NaN checks because some of the tests were failing (they intentionally gave it no datatype, so I think the tests themselves are correct). I'd like this to match `ArrayBuffer.slice` as closely as possible, so I [implemented your comment.](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/io/composite_array_buffer.ts#L89-L90)

Issue #7273, opened 2023-01-14
TypeError: env().platform.isTypedArray is not a function. (In 'env().platform.isTypedArray(a)', 'env().platform.isTypedArray' is undefined)


anglo2333:
<em>Please make sure that this is a bug. As per our
[GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md),
we only address code/doc bugs, performance issues, feature requests and
build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js):
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04):
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device:
- TensorFlow.js installed from (npm or script link):
- TensorFlow.js version (use command below):
- Browser version:
- Tensorflow.js Converter Version:

**Describe the current behavior**

**Describe the expected behavior**

**Standalone code to reproduce the issue**
Provide a reproducible test case that is the bare minimum necessary to generate
the problem. If possible, please share a link to Colab/CodePen/any notebook.

**Other info / logs** Include any logs or source code that would be helpful to
diagnose the problem. If including tracebacks, please include the full
traceback. Large logs and files should be attached.

etosolo322:
use the previous version, for example @4.1.0

asallem94:
I am running into this issue with a react-native project.

Rhysjc:
Experiencing the same issue with a new react-native project.

Rhysjc:
Having done some digging, I think #7181 had introduced this. I assume `isTypedArray` needs to be added to the [React Native platform](https://github.com/tensorflow/tfjs/blob/master/tfjs-react-native/src/platform_react_native.ts)

Rhysjc:
`tfjs-react-native` has `tfjs-core` as a peer dependency on version 3.7.0. I'm not too sure how the versioning of the packages here work, but assuming `tfjs` follows the same versions, downgrading `@tensorflow/tfjs` to `3.7.0` in your React Native project may resolve this. I don't have time to try it right now but I will do tomorrow. If this resolves it for anyone please comment.

huhm:
> use the previous version, for example @4.1.0

it works for me.   I used the tsjs-node

Rhysjc:
[...]

PR #7609, opened 2023-04-20
Support loading models with weights above 2GB on Chrome

mattsoulanille:
Chrome ArrayBuffers throw allocation errors above 2GB in size. This makes it impossible to load TFJS models above this size in Chrome (even with weight sharding) because model loading involves concatenating all the weights into a single ArrayBuffer.

This PR avoids this concatenation. Instead of slicing the weight tensors out of a single concatenated ArrayBuffer, it keeps the weight buffers in their original shards and slices them using the CompositeArrayBuffer class created in #7598.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7609)


tfjs-core/src/io/io_utils.ts
```diff
@@ -330,10 +336,15 @@ export function base64StringToArrayBuffer(str: string): ArrayBuffer {
 /**
  * Concatenate a number of ArrayBuffers into one.
  *
- * @param buffers A number of array buffers to concatenate.
+ * @param buffers An array of ArrayBuffers to concatenate, or a single
+ *     ArrayBuffer.
  * @returns Result of concatenating `buffers` in order.
  */
-export function concatenateArrayBuffers(buffers: ArrayBuffer[]): ArrayBuffer {
+export function concatenateArrayBuffers(buffers: ArrayBuffer[]
```
chunnienc, line 343: Can we move this to be part of `CompositeArrayBuffer`? like static method `CompositeArrayBuffer.join(buffers: ArrayBuffer[])` or through public method `new CompositeArrayBuffer(buffers).toArrayBuffer()`, which makes it easier to bridge CompositeArrayBuffer with native ArrayBuffer and pass CompositeArrayBuffer around in the future if needed.

tfjs-node/src/io/io_utils.ts
```diff
@@ -51,30 +51,3 @@ export function toArrayBuffer(buf: Buffer|Buffer[]): ArrayBuffer {
...
-        0 :
-        Buffer.byteLength(JSON.stringify(modelArtifacts.modelTopology), 'utf8'),
-    weightSpecsBytes: modelArtifacts.weightSpecs == null ?
-        0 :
-        Buffer.byteLength(JSON.stringify(modelArtifacts.weightSpecs), 'utf8'),
-    weightDataBytes: modelArtifacts.weightData == null ?
-        0 :
-        modelArtifacts.weightData.byteLength,
-  };
-}
```
mattsoulanille, line 80: This is duplicated in `tfjs-core/src/io/io_utils.ts`

tfjs-core/src/io/io_utils.ts
```diff
@@ -285,7 +291,7 @@ const useNodeBuffer = typeof Buffer !== 'undefined' &&
  */
 export function stringByteLength(str: string): number {
   if (useNodeBuffer) {
-    return Buffer.byteLength(str);
+    return Buffer.byteLength(str, 'utf8');
```
mattsoulanille, line 294: tfjs-node used `utf8` in its implementation, so I think it should also be here.

chunnienc (changes requested):

tfjs-core/src/io/io_utils.ts
mattsoulanille: I was considering that, and my original implementation actually used `new CompositeArrayBuffer(buffers).slice()`, but I removed it in favor of `concatenateArrayBuffers` because of an issue with the types in tfjs-converter tests ([here was my fix for it in the `spy_ops.ts` file](https://github.com/tensorflow/tfjs/pull/7609/commits/cf2493e67a24fd1f0e5496f409e84a50d3c453a7#diff-02ae9d1c277302d454a7f39b9dd0793fbaedcc497ce7262c0c3b6fe2c0a07b77), but it's a bit hacky).

I'm fine with using the converter spy_ops.ts fix if it'll make the core implementation cleaner. What do you think?

chunnienc: I took a look at the usage of `spyOnAllFunctions` in tests, and I think the test is something we should fix. I think a hacky way like what you did is probably fine.

In general, instead of automatically replace everything with spy using `spyOnAllFunctions`, I think we should explicitly create an ioSpy object which only contains the function we want to spy, so that we can make the test more controllable and reliable. There are some stuffs exported in io apparently should not be spied, like `getWeightSpecs`, which is a io helper function instead of a function to do io.

mattsoulanille: @chunnienc I've replaced `concatenateArrayBuffers` with `CompositeArrayBuffer.join` in tfjs-core and deprecated `concatenateArrayBuffers`. We can't replace it in other packages yet because that would introduce a breaking change. Downstream packages could not be used with an earlier version of tfjs-core that does not implement `CompositeArrayBuffer` (see #7273 for an example of why this is important). We can apply this change to all the packages in the next major release.

mattsoulanille: Actually, it's fine to use it in tests, since users will never run those. I'll swap `concatenateArrayBuffers` for `CompositeArrayBuffer.join` in the test files.

chunnienc (approved):

pyu10055 (approved):
> Reviewed 4 of 18 files at r1, 1 of 1 files at r2, 13 of 13 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7609)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @chunnienc)

PR #7625, opened 2023-04-25
webgpu: Change the Chrome macOS flags in BrowserStack

gyagp:
This is to ensure all the flags are necessary for WebGPU tests in BrowserStack on Chrome macOS based on SwiftShader.

### DIFF
--- a/tools/karma_template.conf.js
+++ b/tools/karma_template.conf.js
@@ -83,11 +83,17 @@ const CUSTOM_LAUNCHERS = {
     os_version: 'High Sierra',
     flags: [
       '--enable-unsafe-webgpu',
+      '--disable-vulkan-surface',
+
+      '--disable-vulkan-fallback-to-gl-for-testing',
+      '--disable-features=VaapiVideoDecoder',
+      '--ignore-gpu-blocklist',
+
       '--enable-features=vulkan',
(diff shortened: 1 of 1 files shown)


mattsoulanille (approved):
LGTM. These flags make more sense. Thanks!

gyagp:
> LGTM. These flags make more sense. Thanks!

I'm still doing tests to see what are the most meaningful and necessary flags. I will change the status to "Ready for review" once I figure it out.

gyagp:
@mattsoulanille Actually we only need "--use-webgpu-adapter=swiftshader" to enable swiftshader based WebGPU tests. However, in BrowserStack, some tests related to fromPixels would fail if we don't add flag "--disable-vulkan-fallback-to-gl-for-testing". I'm not sure about the reason, as I couldn't reproduce it with macOS at hand. I just filed https://github.com/tensorflow/tfjs/issues/7631 to track the issue. 
PTAL!

qjia7 (approved):
LGTM, thanks!

gyagp:
@mattsoulanille Do you need another look? I simplified the solution after your LGTM.

gyagp:
@mattsoulanille Let me merge this to unblock other efforts at hand.

PR #7626, opened 2023-04-25
Avoid changing 'action_env' between bazel run and test

mattsoulanille:
When action_env changes, Bazel needs to recompile everything, since the compilation output may have been changed by the shell variables specified in action_env. This PR avoids changing action_env between build and test commands, which prevents Bazel from re-building everything when switching between `bazel test` and `bazel run` commands.

Tested by running `yarn build` in tfjs-core followed by `yarn test`. It should not have to re-build the targets built in `yarn build`.


Linchenn (approved):
Very helpful, thanks! LGTM

fengwuyao (approved):
Thanks Mat!

PR #7628, opened 2023-04-25
Add Draw API

Linchenn:
This PR add Draw op to Core and implemented it in CPU backend.

We could get the following effects with a couple lines of codes:
```js
// Setup
tf.setBackend('cpu');
const imgElem = document.getElementById('salad');
const canvasElem = document.getElementById("display");
const image = tf.browser.fromPixels(imgElem);

// TFJS processing
const filter = tf.tensor4d(fval, [3,3,3,3], 'float32');
const normalizedImage = tf.div(image, 255);
const convRes = tf.conv2d(normalizedImage, filter, 1, 'same');
const res =  tf.clipByValue(convRes, 0, 1);

// Draw
const canvasOptions = {};
const drawOptions = {alpha: 0.5}
tf.browser.draw(res, canvasElem, canvasOptions, drawOptions);
```
image


Linchenn:
The tests for this PR is under CPU-backend, instead of Core, because different backend would use different context and has different implementation to validate the image data.

Linchenn:
@axinging Could you help review this PR? Hope the API design make sense for WebGPU backend.

tfjs-backend-cpu/src/kernels/Draw.ts
```diff
@@ -0,0 +1,92 @@
...
+  const alpha = drawOptions ?.alpha || 1;
+
+  const contextType = canvasOptions ?.contextType || '2d';
+  if (contextType !== '2d') {
+    throw new Error(`Context type ${
+        canvasOptions.contextType} is not supported by the CPU backend.`);
+  }
+  const ctx = canvas.getContext(contextType,
+    canvasOptions?.contextAttributes || {}) as CanvasRenderingContext2D ;
+
```
qjia7, line 38: Check whether `ctx` is null in case `getContext` is called before with a different `contextType`?

tfjs-backend-cpu/src/kernels/Draw.ts
```diff
@@ -0,0 +1,92 @@
...
+  const bytes = new Uint8ClampedArray(width * height * 4);
+
+  for (let i = 0; i < height * width; ++i) {
+    const rgba = [0, 0, 0, 255 * alpha];
+
+    for (let d = 0; d < depth; d++) {
+      const value = data[i * depth + d];
+
+      if (image.dtype === 'float32') {
+        if (value < 0 || value > 1) {
```
qjia7, line 52: Should we throw an error or just clamp it to the supported range? For webgl/webgpu backend, it's not possible to do this check and throw errors.

tfjs-core/src/ops/browser.ts
```diff
@@ -299,25 +325,10 @@ export async function toPixels(
     $img = cast(originalImgTensor, 'int32');
     originalImgTensor.dispose();
   }
-  if ($img.rank !== 2 && $img.rank !== 3) {
-    throw new Error(
-        `toPixels only supports rank 2 or 3 tensors, got rank ${$img.rank}.`);
-  }
+  validateImgTensor($img);
```
qjia7, line 328: Maybe add a warning in this API to tell users that this API will be deprecated and suggest them switch to the more powerful `draw` API.

tfjs-core/src/ops/browser.ts
```diff
@@ -371,4 +382,46 @@ export async function toPixels(
...
+ * @param contextOptions A object with options to get the context from the
+ *     canvas. If the canvas has created context, contextOptions would not make
+ * effects.
+ * @param drawOptions A object of options to customize drawing.
+ *
+ * @doc {heading: 'Browser', namespace: 'browser'}
+ */
+export function draw(
+    image: Tensor2D|Tensor3D|TensorLike, canvas: HTMLCanvasElement,
+    contextOptions?: ContextOptions, drawOptions?: DrawOptions): void {
```
qjia7, line 409: Is it simpler to merge `contextOptions` and `drawOptions` into `CanvasOptions`/`CanvasConfiguration`? For example, in webgpu, we have [GPUCanvasConfiguration](https://www.w3.org/TR/webgpu/#dictdef-gpucanvasconfiguration)
```
export interface CanvasOptions {
  /**
   * Optional. If it is not set, it would be variable based on the current
   * backend.
   */
  contextType?: string;
  /**
   * Optional.
   */
  contextAttributes?: WebGLContextAttributes;

  /**
   * Optional. A number in range [0-1]. If the image is a 2D tensor or a 3D
   * tensor with 1 or 3 channels, the alpha channels would set as its value;
   * otherwise, it would not make effects.
   */
  alpha?: number;
}

**Comment by qjia7** [comment]
> The tests for this PR is under CPU-backend, instead of Core, because different backend would use different context and has different implementation to validate the image data.

Maybe we can easily let each backend support both backend preferred canvas context + 2d canvas. Then we can always use 2d canvas to verify the correctness in core.

File: tfjs-backend-cpu/src/kernels/Draw.ts (hunk 1 above)
Comment (Linchenn) [reply tfjs-backend-cpu/src/kernels/Draw.ts:38]: Done.

File: tfjs-core/src/ops/browser.ts (hunk 4 above)
Comment (Linchenn) [reply tfjs-core/src/ops/browser.ts:409]: [GPUCanvasConfiguration](https://www.w3.org/TR/webgpu/#dictdef-gpucanvasconfiguration) has alphaMode, but it is different from the alpha here.

I mean either contextOptions or [GPUCanvasConfiguration](https://www.w3.org/TR/webgpu/#dictdef-gpucanvasconfiguration) are describing the setting of canvas or context. However, drawOptions is to customize the kernel computations. Separating them might be more clearer for users?

Also, do you think the name and the definition of `drawOptions` makes sense for WebGPU?

File: tfjs-core/src/ops/browser.ts (hunk 3 above)
Comment (Linchenn) [reply tfjs-core/src/ops/browser.ts:328]: Done. Thanks!

File: tfjs-backend-cpu/src/kernels/Draw.ts (hunk 2 above)
Comment (Linchenn) [reply tfjs-backend-cpu/src/kernels/Draw.ts:52]: For WebGL and WebGPU, I don't think we need to check values and we could just clamp them. For CPU, it's very convenient to make such checks and throwing an error would be helpful for users, because it's common that users pass values out of 255 and get a pure white canvas with no hit message. So, I think this throw-error feature is good to have.

**Comment by Linchenn** [comment]
> > The tests for this PR is under CPU-backend, instead of Core, because different backend would use different context and has different implementation to validate the image data.
> 
> Maybe we can easily let each backend support both backend preferred canvas context + 2d canvas. Then we can always use 2d canvas to verify the correctness in core.

Nice catch, done. Thanks!

File: tfjs-core/src/ops/browser.ts (hunk 4 above)
Comment (qjia7) [reply tfjs-core/src/ops/browser.ts:413]: I think it's ok to custom `alpha`. I don't have good candidate name for `drawOptions`. Leave it open if other reviewers have better suggestion or just keep the current name.

**Review by qjia7 — APPROVED** [review]
LGTM, thanks!

File: tfjs-core/src/ops/browser.ts (hunk 4 above)
Comment (Linchenn) [reply tfjs-core/src/ops/browser.ts:413]: Sorry, I mean `contextOptions` (updated my comment), does it makes senses for WebGPU? Should I add more optional fields, like [GPUCanvasConfiguration](https://www.w3.org/TR/webgpu/#dictdef-gpucanvasconfiguration)? I intend to wrap all canvas and context related settings into it.

File: tfjs-core/src/ops/browser.ts [hunk 5]
```diff
@@ -371,4 +386,46 @@ export async function toPixels(
...
 
+/**
+ * Draws a `tf.Tensor` to a canvas.
+ *
+ * When the dtype of the input is 'float32', we assume values in the range
+ * [0-1]. Otherwise, when input is 'int32', we assume values in the range
+ * [0-255].
+ *
+ * @param image A rank-2 tensor with shape `[height, width]`, or a rank-3 tensor
+ * of shape `[height, width, numChannels]`. If rank-2, draws grayscale. If
```
chunnienc, line 397: "draws grayscale" -> "draws in grayscale"
I feel like the comment can be written to be more clear. @mattsoulanille please help us with grammar.

tfjs-core/src/ops/draw_test.ts
```diff
@@ -0,0 +1,99 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC. All Rights Reserved.
```
chunnienc, line 3: no all rights reserved.
you can use https://github.com/chunnienc/tfjs-license-fix to add license headers.

tfjs-core/src/ops/draw_test.ts
```diff
@@ -0,0 +1,99 @@
...
+    const ctx = canvas.getContext('2d');
+
+    // tslint:disable-next-line:no-any
+    tf.browser.draw(img, canvas as any, {contextType: '2d'});
+    const actualData = ctx.getImageData().data;
+    const expectedData = data.map(e => Math.floor(e * 255));
+    expectArraysEqual(actualData, expectedData);
+  });
+
+  it('draw 2D image as gray scale', async () => {
```
chunnienc, line 73: "as gray scale" -> "in grayscale"

chunnienc (changes requested):

tfjs-core/src/ops/draw_test.ts
Linchenn: Done.

tfjs-core/src/ops/draw_test.ts
Linchenn: Done.

tfjs-core/src/ops/browser.ts
Linchenn: Done.

tfjs-core/src/ops/browser.ts
chunnienc: I agree that merging two options into one can cause some confusion in some overlapping but different purpose params like `alpha`. But I feel a little bit concerned when I see your example usage code:
```js
tf.browser.draw(res, canvasElem, undefined, {alpha: 0.5});
```
From the user's perspective it is hard to tell whether the first one or second one is drawOptions. We may hurt the readability and usability of this API. Just my personal concerns, throw it out for more discussion.

Also, if we choose to release it as two separate options params, consider to make it:
```js
function draw(
    image: Tensor2D|Tensor3D|TensorLike, canvas: HTMLCanvasElement, drawOptions?: DrawOptions, contextOptions?: ContextOptions)
```
if you think customizing `drawOptions` is a normal user feature but specifying  `contextOptions` is more like an feature for advanced users, swapping the order may be more easy to use since most of the users can just do `tf.browser.draw(res, canvas, { alpha: 1, ...});`

chunnienc (approved):
LGTM except for:
1. The doc comment can be improved. @mattsoulanille for native speaker review.
2. The API with two options looks okay to me, but we can have more discussion on that if you want online and/or offline.

tfjs-core/src/ops/browser.ts
Linchenn: 1. Since we would add `GPUCanvasConfiguration ` to `CanvasOptions`, I renamed `ContextOptions` to `CanvasOptions`.
2. For usage of the API with two objects, if users follow the coding style, it would be:
```js
const canvasOptions = {};
const drawOptions = {alpha: 0.5}
tf.browser.draw(res, canvasElem, canvasOptions, drawOptions);
```
3. I am open to suggestions for whether we want to merge the two options as `Options: {CanvasOptions, DrawOptions}`.

5.

tfjs-core/src/ops/browser.ts
```diff
@@ -371,4 +386,44 @@ export async function toPixels(
...
+ * [0-1]. Otherwise, when input is 'int32', we assume values in the range
+ * [0-255].
+ *
+ * @param image A rank-2 tensor with shape `[height, width]`, or a rank-3 tensor
+ * with shape `[height, width, depth]`. Rank-2 tensor would be drawn in
+ * grayscale. For rank-3 tensor, the depth must be 1, 3 or 4: if depth is 1, it
+ * would be drawn in grayscale; if the depth is 3, the three values of a pixel
+ * would be corresponding to R, G and B channels and alpha would be 1 (opaque)
+ * or customized in drawOptions; if the depth is 4, the four values of a pixel
+ * would be corresponding to R, G, B and A (alpha) channels.
```
mattsoulanille, line 402:
```suggestion
 * @param image The tensor to draw on the canvas. Must match one of
 * these shapes:
 *   - Rank-2 with shape `[height, width`]: Drawn as grayscale.
 *   - Rank-3 with shape `[height, width, 3]`: Drawn as RGB with alpha set in
 *     `drawOptions` (defaults to 1, which is opaque).
 *   - Rank-3 with shape `[height, width, 4]`: Drawn as RGBA.
```

tfjs-core/src/ops/browser.ts
```diff
@@ -371,4 +386,44 @@ export async function toPixels(
...
+ * @param canvas The canvas to draw to.
+ * @param canvasOptions A object to configure the canvas to draw to.
+ * @param drawOptions A object of options to customize drawing.
```
mattsoulanille, line 405: Should we describe the settings that people can change on these config and draw options here, or is this documented elsewhere?

Linchenn: This is at the definition of `drawOptions` and `canvasOptions` and tfjs-website would add them to the API. Similar as `Options` of [this API](https://github.com/tensorflow/tfjs/blob/tfjs-v4.4.0/tfjs-converter/src/executor/graph_model.ts#L600-L618).

tfjs-core/src/ops/draw_test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  getContext(type: '2d'): MockContext {
+    if (this.context == null) {
+      this.context = new MockContext();
+    }
+    return this.context;
+  }
+}
+
+describeWithFlags('Draw on 2d context', BROWSER_ENVS, () => {
```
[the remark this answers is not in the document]
Linchenn: As Jiajia suggested, we would implement and test 2d context for all backends soon.

tfjs-core/src/ops/browser.ts
```diff
@@ -359,6 +370,10 @@ export async function toPixels(
   }
 
   if (canvas != null) {
+    console.warn(
+        'tf.browser.toPixels is not efficient to draw tensor on canvas. ' +
+        'Please try tf.browser.draw instead.');
```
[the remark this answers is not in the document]
Linchenn: Nice catch. Added a variable in the browser.ts file for the boolean, since the file already has a variable `hasToPixelsWarned`, which share the coding style.

mattsoulanille (approved):

tfjs-core/src/ops/browser.ts
Linchenn: As collected feedbacks from you and WebML channel, get most votes for unified option object for this. Updated.

Jonathhhan:
Thanks for the nice lib. I also get the warning: `tf.browser.toPixels is not efficient to draw tensor on canvas. Please try tf.browser.draw instead.` It only works with `tf.setBackend('cpu');`, but then it is much slower than with `tf.setBackend('webgl')`. Are there any plans to implement `tf.browser.draw` for the webgl backend?

PR #7633, opened 2023-04-27
Add browser tests for CPU backend

Linchenn:
This PR adds browser test target for CPU backend. The test target does not include tests in Core, but the test target is added to nightly and presubmit tests.

Only added "bs_chrome_mac" for nightly and presubmit tests, because the CPU tests do not heavily relay on the platform.

Currently, CPU backend only has Node.js tests. We could provide browser tests for CPU backend, even though it does not need to be integrated into CI and nightly tests. eg. this would be helpful for Draw API tests for CPU backend.

### DIFF
--- a/tfjs-backend-cpu/BUILD.bazel
+++ b/tfjs-backend-cpu/BUILD.bazel
@@ -42,7 +42,7 @@ tfjs_web_test(
     headless = False,
     presubmit_browsers = [
         "bs_safari_mac",
-    ]
+    ],
 )
 
 tfjs_bundle(
--- a/tfjs-backend-cpu/BUILD.bazel
+++ b/tfjs-backend-cpu/BUILD.bazel
@@ -36,9 +36,16 @@ tfjs_web_test(
         "//tfjs-backend-cpu/src:tfjs-backend-cpu_test_bundle",
     ],
     args = [],
-    browsers = [],
+    browsers = [
+        "bs_chrome_mac",
+        "bs_firefox_mac",
+        "bs_android_10",
+        "win_10_chrome",
+    ],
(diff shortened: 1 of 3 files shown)


Linchenn:
With these changes, when I ran `bazel test :tfjs-backend-cpu_browser_test`, I got an error 'Could not resolve "./tests"'. Do you have ideas on it? @mattsoulanille

tfjs-backend-cpu/BUILD.bazel
```diff
@@ -29,6 +30,17 @@ nodejs_test(
...
 
+tfjs_web_test(
+    name = "tfjs-backend-cpu_browser_test",
+    srcs = [
+        "//tfjs-backend-cpu/src:tfjs-backend-cpu_test_bundle",
+    ],
+    args = [],
+    browsers = [],
+    headless = False,
+    presubmit_browsers = []
```
mattsoulanille, line 41: Nit: If this is empty, these tests won't run in presubmits. Is that intentional? Do the tests only work with `headless = False`?

Linchenn:
> With these changes, when I ran `bazel test :tfjs-backend-cpu_browser_test`, I got an error 'Could not resolve "./tests"'. Do you have ideas on it? @mattsoulanille

This has been fixed. Thanks Matt!

mattsoulanille (approved):

tfjs-backend-cpu/BUILD.bazel
Linchenn: Done. Thanks! Also removes core tests from this test target.

pyu10055 (approved):

PR #7634, opened 2023-04-27
webgpu: Fix potential tensor leak issue.

qjia7:


tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -201,9 +201,7 @@ export class WebGPUBackend extends KernelBackend {
    * @oaram force Optional, remove the data regardless of refCount
    */
   override disposeData(dataId: DataId, force = false): boolean {
-    if (this.tensorDataPendingDisposal.indexOf(dataId) >= 0) {
```
xhcao, line 204: I prefer that the code is not removed here. If removing this code, the array of `tensorDataPendingDisposal` could store multiple same copies `dataId`. I know the below code could handle the correctness that user calls dispose api sometimes. But why not directly return here.

xhcao (approved):
I have a nit comment. But if Yang prefer current code, please ignore my comment. Thanks.
LGTM

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -321,17 +323,20 @@ export class WebGPUBackend extends KernelBackend {
 
     this.commandQueueOwnedIds = new WeakSet<DataId>();
 
-    this.tensorDataPendingDisposal.forEach(d => {
-      this.releaseResource(d);
-      this.tensorMap.delete(d);
-    });
+    const disposeLength = this.tensorDataPendingDisposal.length;
+    for (var i = 0; i < disposeLength; i++) {
```
Linchenn, line 327: let

Linchenn (approved):

tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -322,17 +324,20 @@ export class WebGPUBackend extends KernelBackend {
...
 
-    this.tensorDataPendingDisposal.forEach(d => {
-      this.releaseResource(d);
-      this.tensorMap.delete(d);
-    });
+    const disposeLength = this.tensorDataPendingDisposal.length;
+    for (let i = 0; i < disposeLength; i++) {
+      const item = this.tensorDataPendingDisposal.pop();
+      if (this.disposeData(item)) {
+        engine().removeDataId(item, this);
```
gyagp, line 331: We shouldn't call this for internal tensors.

Linchenn: Ideally we have to distinguish internal or external tensor for webgpu backend. 

From my personal perspective, this might work because the assumption here 'removeDataId for internal tensor would not make effects' is always valid, because internal tensors are invisible from Core and its dataId should be unknown for Core.

qjia7: I am thinking another solution: provide a new `disposeDataInternal(dataId: DataId, force = false, internal = true): boolean` function, which will be almost same with original `disposeData`:
```
disposeDataInternal(dataId: DataId, force = false, internal = true): boolean {
    ... ...
    if (this.commandQueueOwnedIds.has(dataId)) {
      this.tensorDataPendingDisposal.push({dataId, internal});
      return false;
    }
    ...
}

override disposeData(dataId: DataId, force = false): boolean {
  return disposeDataInternal(dataId, force, false);
}
```
Then, all backend intermediate tensors call `disposeDataInternal` to dispose. And we can decide whether to call `engine().removeDataId ` based on the `item.internal`.
```
const item = this.tensorDataPendingDisposal.pop();
if(this.disposeData(item) && !item.internal)
{
  engine().removeDataId(item, this);
}
```
@gyagp @Linchenn How do you think? Or we just keep it unchanged like this PR for simplification.

Linchenn: Since `disposeDataInternal` is processing backend tensors, I think this does not necessarily need to be at Core. We could keep it just in WebGPU backend.

Also, if we go for this, I think we need to manually update all `disposeDataInternal` for all internal tensors and we have to make sure we didn't call `disposeDataInternal` for external tensors, right? This may involve some of your efforts. I am open to both ways.

qjia7: @gyagp @Linchenn Please review my latest changes again. I updated the PR description to help your understanding. Thanks.

Linchenn (approved):
LGTM, thanks!

gyagp (approved):

PR #7649, opened 2023-05-04
Remove duplicate Prod and SparseToDense ops from converter

### LINKED ISSUE #7648: Op Collision in OperationMapper
[OperationMapper](https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/src/operations/operation_mapper.ts#L45) stores a map of all the TFJS ops in the `opMappers` parameter. It generates this map by iterating all the ops in [op_list/](https://github.com/tensorflow/tfjs/tree/master/tfjs-converter/python/tensorflowjs/op_list) and uses `tfOpName` as the key. This key is not unique between ops in different categories.

There are two collisions in the current set of supported ops:
* [basic_math/Prod](https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/python/tensorflowjs/op_list/basic_math.json#L800) and [reduction/Prod](https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/python/tensorflowjs/op_list/reduction.json#L222)
  * The signature for reduction/Prod appears to be a subset of the signature for basic_math/Prod.
* [slice_join/SparseToDense](https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/python/tensorflowjs/op_list/slice_join.json#L364) and [normalization/SparseToDense](https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/python/tensorflowjs/op_list/normalization.json#L22)
  * The signatures between these two appear identical.

mattsoulanille:
Fixes #7648 


tfjs-converter/python/tensorflowjs/op_list/reduction.json
```diff
@@ -238,6 +238,12 @@
         "tfName": "keep_dims",
         "name": "keepDims",
         "type": "bool"
+      },
+      {
+        "tfName": "T",
+        "name": "dtype",
+        "type": "dtype",
+        "notSupported": true
```
mattsoulanille, line 246: @pyu10055 do you know what this 'T' attribute is for?

pyu10055 (approved):

tfjs-converter/src/operations/executors/basic_math_executor.ts
```diff
@@ -159,10 +159,6 @@ export const executeOp: InternalOpExecutor =
               getParamValue('x', node, tensorMap, context) as Tensor)];
         case 'Rsqrt':
           return [ops.rsqrt(getTensor(node.inputNames[0], tensorMap, context))];
-        case 'Prod':
```
chunnienc, line 162: For removing existing executor: are you sure that there are no old models converted with category alias?

chunnienc (approved):

tfjs-converter/src/operations/executors/basic_math_executor.ts
mattsoulanille: I checked an example model file, and we don't actually save the category of an op in the model.json. I don't think this should affect our model support, but good idea to check.

chunnienc: yeah `model.json` does not store the category name, my bad. Agreed that it shouldn't affected converted models.

PR #7550, opened 2023-04-05
Fix tfjs-release not updating all tfjs versions of subpackages

mattsoulanille:
Some TFJS packages, like wasm, have examples or demos in them. These usually depend on the parent package, but the parent package is not marked as to be updated when updating the subpackage dependency versions. For an example of this, see https://github.com/tensorflow/tfjs/pull/7547. This PR fixes that.

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs/7550)

PR #7656, opened 2023-05-04
Update monorepo to 4.5.0.

mattsoulanille:


tfjs-backend-wasm/starter/parcel/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "@tensorflow/tfjs-backend-wasm": "3.14.0",
-    "@tensorflow/tfjs-core": "3.14.0"
+    "@tensorflow/tfjs-core": "4.5.0"
```
chunnienc, line 11: why do we still update this version? I thought it's fixed in the previous version updater PR.

chunnienc (changes requested):

tfjs-backend-wasm/starter/parcel/package.json
mattsoulanille: Oh, #7550 hasn't been merged yet. Let me try again on that branch.

PR #7657, opened 2023-05-04
Update monorepo to 4.5.0.

mattsoulanille:


e2e/benchmarks/browserstack-benchmark/package.json
```diff
@@ -8,8 +8,8 @@
     "url": "https://github.com/tensorflow/tfjs"
   },
   "devDependencies": {
-    "@tensorflow/tfjs": "link:../../../tfjs",
-    "@tensorflow/tfjs-backend-wasm": "link:../../../link-package/node_modules/@tensorflow/tfjs-backend-wasm",
+    "@tensorflow/tfjs": "4.5.0",
+    "@tensorflow/tfjs-backend-wasm": "4.5.0",
     "@tensorflow/tfjs-vis": "link:../../../tfjs-vis",
```
mattsoulanille, line 13: I need to fix this in the release script. I think it's fine to do that for the next release, since I don't think anyone will use the browserstack benchmark tool. The dependency rewriter can't change tfjs-vis because it's not a monorepo package, i.e. it's not published in this PR.

pyu10055 (approved):

chunnienc (approved):
TFDF is still going to be released as alpha?

mattsoulanille:
> Is TFDF still going to be released as alpha?

I think that's what we're intending. @pyu10055 is this right?

mattsoulanille:
> 

Confirmed on chat.

PR #7660, opened 2023-05-05
webgpu: Optimize depthwise conv2d

qjia7:
This PR uses linear workgroup size to optimize depthwise conv2d vec4.

See 150% improvement for the time of FusedDepthwiseConv2D on
BlazePoseDetector.

### DIFF
--- a/tfjs-backend-webgpu/src/depthwise_conv2d_vec4_webgpu.ts
+++ b/tfjs-backend-webgpu/src/depthwise_conv2d_vec4_webgpu.ts
@@ -26,7 +26,7 @@ export class DepthwiseConv2DVec4Program implements WebGPUProgram {
   dispatchLayout: {x: number[]};
   dispatch: [number, number, number];
   variableNames = ['x', 'W'];
-  uniforms = 'pads : vec2<i32>, inDims : vec2<i32>,';
+  uniforms = 'pads : vec2<i32>, inDims : vec2<i32>, virtualWidth : i32,';
   workgroupSize: [number, number, number] = [64, 1, 1];
   workPerThread = 4;
   convInfo: backend_util.Conv2DInfo;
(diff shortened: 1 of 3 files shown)


qjia7:
See big improvement on Intel Gen9 GPU, for example DepthwiseConv2dNative in DeepLabV3 becomes 8.28ms from 21.06ms. No obvious changes for other GPUs.

tfjs-backend-webgpu/src/depthwise_conv2d_vec4_webgpu.ts
```diff
@@ -82,11 +90,16 @@ export class DepthwiseConv2DVec4Program implements WebGPUProgram {
...
-        let batch = i32(globalId.z) / uniforms.outShape[1];
-        let r = i32(globalId.z) % uniforms.outShape[1];
-        let c = i32(globalId.y) * ${this.workPerThread};
-        let d1 = i32(globalId.x) * 4;
+      ${main('index')} {
+        let width0 = uniforms.outShape[3] / 4;
+        let d1 = (index % width0) * 4;
+        var index1 = index / width0;
+        let width1 = uniforms.virtualWidth / ${this.workPerThread};
+        let c = (index1 %  width1) * ${this.workPerThread};
```
gyagp, line 98:
```suggestion
        let c = (index1 % width1) * ${this.workPerThread};
```

gyagp (approved):
LGTM % nit

qjia7:
depthwise_opt
I draw a picture to show the differences. In above picture, the shadow places are the data that threads in the same workgroup will process. We can see that linear workgroup size makes the processed data more contiguous in memory for threads in one workgroup. I guess it may be the reason that we see big improvement on Gen9 GPU. There are no obvious changes for other GPUs. Maybe other drivers can well handle this situation.

xhcao (approved):
LGTM. But I feel it is weird to introduce the new `virtualWidth` concept and the complicated computation of coordinate here. I guess the optimization is valid when `channels` is large or low-end devices.

qjia7:
> LGTM. But I feel it is weird to introduce the new `virtualWidth` concept and the complicated computation of coordinate here. I guess the optimization is valid when `channels` is large or low-end devices.

Thank you Xinghua. The `virtualWidth` is to make sure the correctness of the algorithm when switching to linear workgroup size. In current algorithm, each thread writes 4x4 data which requires the 4 data in the same col, the 4x4 data in the same row. It means that the width of the output shape needs to be divisible by 4. That's why I introduced the `virtualWidth = Math.ceil(this.outputShape[2] / this.workPerThread) * this.workPerThread; ` where `this.workPerThread = 4`.

Issue #7498, opened 2023-03-20
Bug in the layers.getLayer()


mahnehsilla:
(permalink at post' bottom)

This function has not been fully adapted to javascript, currently the TS definition is:

```
getLayer(name?:string, index?:string)
```

So the only way to get by index is 
```
getLayer(null, index)
```

## Suggestions

Very humble ideas, not a good coder.

Probably should be rewritten to
```
 getLayer(target:string|number)
```
Where target is name or index. 

Or an object of this kind:
```
 getLayer({name?:string, index?:number})
```
That errors out if the object keys length is 0.

-----------------------

## [Permalink to Code Bug](https://github.com/tensorflow/tfjs/blob/cce7f017759423f98f7add7a01c6f3d4b47b1633/tfjs-layers/src/engine/container.ts#L975-L996)

- TensorFlow.js version (use command below): 4.2.0

mahnehsilla:
Yes, I normally switch bw python keras and tfjs for the documentation as well (the python one is far more complete imho.) For example, the `tf.loadWeights` function isnt even in the documention of tfjs. I reported this to the forums I think.

I thought about a few contributions but I thought adding the tests could be annoying (I dont have much time.) 

I will try to add it though, this software is fantastic, and the only numpy equivalent for JS apart from DL @pyu10055

fengwuyao:
The issue has been fixed by https://github.com/tensorflow/tfjs/pull/7665
Thanks for pointing out! @mahnehsilla

PR #7665, opened 2023-05-05
Fix getLayer() API

fengwuyao:
Fix the bug:(https://github.com/tensorflow/tfjs/issues/7498) for getLayer() API.

Support input as string / number
- if type is `number`, we get layer from index.
- if type is `string`, we get layer from layer name.


tfjs-layers/src/engine/container.ts
```diff
@@ -972,27 +972,27 @@ export abstract class Container extends Layer {
    *    subclasses: ['LayersModel']
    * }
    */
-  getLayer(name?: string, index?: number): Layer {
-    if (index != null) {
-      if (this.layers.length <= index) {
+  getLayer(input?: string|number): Layer {
```
Linchenn, line 975: This might cause breaking change for `getLayer(null, 1)` usage.

We could make it like `getLayer(nameOrIndex: string|number, index?: number)`, similar as https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/engine.ts#L441

fengwuyao: Done! Thanks!

tfjs-layers/src/engine/container.ts
```diff
@@ -982,17 +982,17 @@ export abstract class Container extends Layer {
         return this.layers[index];
       }
     } else {
-      if (name == null) {
+      if (nameOrIndex == null) {
         throw new ValueError('Provide either a layer name or layer index');
       }
     }
 
     for (const layer of this.layers) {
```
Linchenn, line 990: If nameOrIndex is a number, we could add a branch for it.

Linchenn (changes requested):

Linchenn (approved):
LGTM, thanks!

tfjs-layers/src/engine/container_test.ts
```diff
@@ -385,9 +385,9 @@ describeMathCPUAndGPU('Container', () => {
 
   it('getLayer works by index', () => {
     const [container, layers] = createSimpleTwoLayerContainer();
-    expect(container.getLayer(null, 0)).toEqual(layers[0]);
-    expect(container.getLayer(null, 1)).toEqual(layers[1]);
-    expect(container.getLayer(null, 2)).toEqual(layers[2]);
+    expect(container.getLayer(0)).toEqual(layers[0]);
+    expect(container.getLayer(1)).toEqual(layers[1]);
+    expect(container.getLayer(2)).toEqual(layers[2]);
```
mattsoulanille, line 390: In addition to this new API, we should still test the `(null, index)` case since it's part of our public api.

mattsoulanille (changes requested):
Sorry I didn't review this earlier. I have just a few minor changes.

tfjs-layers/src/engine/container_test.ts
fengwuyao: Thanks! Done.

tfjs-layers/src/engine/container_test.ts
```diff
@@ -385,9 +385,9 @@ describeMathCPUAndGPU('Container', () => {
 
   it('getLayer works by index', () => {
     const [container, layers] = createSimpleTwoLayerContainer();
-    expect(container.getLayer(0)).toEqual(layers[0]);
-    expect(container.getLayer(1)).toEqual(layers[1]);
-    expect(container.getLayer(2)).toEqual(layers[2]);
+    expect(container.getLayer(null, 0)).toEqual(layers[0]);
+    expect(container.getLayer(null, 1)).toEqual(layers[1]);
+    expect(container.getLayer(null, 2)).toEqual(layers[2]);
   });
```
mattsoulanille: Nit: I didn't really make it clear in my last review. I think we should test both cases. I've added the other case here.
```suggestion
  });

  it('getLayer works by index when index is the first argument', () => {
    expect(container.getLayer(0)).toEqual(layers[0]);
    expect(container.getLayer(1)).toEqual(layers[1]);
    expect(container.getLayer(2)).toEqual(layers[2]);
  });
```

mattsoulanille (approved):
LGTM with a nit

PR #7666, opened 2023-05-06
[webgpu] More preparation for element-wise binary op restructuring

hujiajie:
The current code isn't great in that the vec4 shaders have diverged from the scalar ones more than necessary. Here is the common preparation work, so that following refactoring can be done on a per-op basis.


tfjs-backend-webgpu/src/binary_op_util.ts
```diff
@@ -208,13 +200,34 @@ export function getBinaryOpString(
       default:
         continue;
     }
+
+    let isNaN: string;
+    let dTypeN: string;
```
gyagp, line 205: What does suffix "N" mean?

tfjs-backend-webgpu/src/binary_op_util.ts
```diff
@@ -208,13 +200,34 @@ export function getBinaryOpString(
...
+      boolN = 'vec4<bool>';
+    } else {
+      isNaN = 'isnan';
+      dTypeN = 'f32';
+      boolN = 'bool';
+    }
+
     return `
+      let aIsNaN = ${isNaN}(a);
+      let aPostLegalization = select(a, ${dTypeN}(42), aIsNaN);
```
gyagp, line 219: It seems we don't need aPostLegalization at all.

hujiajie: My later patches will depend on `dTypeN(0)`, but `dTypeN` is not available in the global scope. One workaround is to use `a - a` instead, so it's critical to erase NaNs in `a` beforehand.

qjia7: I prefer you remove `aPostLegalization` and `bPostLegalization` in this PR. And use `a - a` in your needed PRs.

qjia7 (commented):
LGTM to unblock Jiajie's work. Personally , I prefer you can submit all binary NaN related changes in one PR so that reviewers can see your whole picture.

tfjs-backend-webgpu/src/binary_op_util.ts
hujiajie: Sorry I can't fully grab your idea. Can you show me an example of the unified `NOT_EQUAL`?

gyagp (approved):

gyagp:
Unblock this and will revisit once all the related PRs are merged.

PR #7673, opened 2023-05-09
[wasm] Add MaxPoolWithArgmax kernel

chunnienc:
Quick implementation based on existing pool3d library. Will look into utilizing xnnpack in the future.

### DIFF
--- a/tfjs-backend-wasm/src/cc/kernels/MaxPoolWithArgmax.cc
+++ b/tfjs-backend-wasm/src/cc/kernels/MaxPoolWithArgmax.cc
@@ -1,0 +1,127 @@
+/* Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


tfjs-backend-wasm/src/cc/kernels/MaxPoolWithArgmax.cc
```diff
@@ -0,0 +1,127 @@
...
+      break;
+    case DType::boolean:
+      MaxPoolWithArgmaxImpl(x_info.b(), pooled_info.b_write(),
+                            indexes_info.i32_write(), include_batch_index,
+                            pool_info);
+      break;
+    default:
+      util::warn("MaxPoolWithArgmax for tensor id failed. Unknown dtype %d",
+                 x_id, dtype);
+  }
```
mattsoulanille, line 123: I sort of wonder if this dtype switch could be a macro, but that would probably be unreadable and prone to bugs.

mattsoulanille (approved):
LGTM!

PR #7676, opened 2023-05-10
Fix golden integration tests not awaiting results

mattsoulanille:
Golden e2e model tests were not waiting for the asynchronous `tensor.data()` function to finish when checking results, making it hard to tell which test is failing. This PR fixes that.


pyu10055 (approved):

chunnienc (approved):

mattsoulanille:
@pyu10055 @chunnienc I needed to completely disable BrowserStack Chrome Mac golden model tests since they were consistently timing out with the async test fix in this PR. They work locally on my M1 machine. I'll re-enable them once they work on BrowserStack too.

I think we still have reasonable coverage from the other OSes we're testing.

qjia7:
Is it possible that the timing out is because it's running on virtual machine instead of physical machine? In virtual machine, webgl/webgpu are indeed running on CPU instead of GPU, which results they are really slow.

PR #7695, opened 2023-05-17
Do not throw an error when killing the verdaccio process

mattsoulanille:
Killing the verdaccio process throws an error because the disconnect event emits when the process is killed. We throw an error on a disconnect to catch any unexpected verdaccio disconnections.

Fix this by deregistering the disconnect handler before killing the verdaccio process.


mattsoulanille:
This was not caught by our verdaccio tests because those tests end the main process right after all the packages are published to Verdaccio, which is before the error appears. During the release, we publish to npm after publishing to verdaccio, so the process doesn't end and the error appears.

scripts/release-util.ts
```diff
@@ -625,7 +625,8 @@ export async function runVerdaccio(): Promise<ChildProcess> {
   // messaging works and verdaccio can tell node that it has started.
   // https://verdaccio.org/docs/verdaccio-programmatically/#using-fork-from-child_process-module
   const verdaccioBin = require.resolve('verdaccio/bin/verdaccio');
-  const serverProcess = fork(verdaccioBin, ['--config=e2e/scripts/verdaccio.yaml']);
+  const config = path.join(__dirname, '../e2e/scripts/verdaccio.yaml');
+  const serverProcess = fork(verdaccioBin, [`--config=${config}`]);
```
mattsoulanille, line 629: More reliable config directory resolution.

scripts/release-util.ts
```diff
@@ -647,15 +648,22 @@ export async function runVerdaccio(): Promise<ChildProcess> {
...
+
+  const onUnexpectedDisconnect = (err: unknown) => {
+    throw new Error(`Verdaccio process unexpectedly disconnected: ${err}`);
+  };
+  serverProcess.on('disconnect', onUnexpectedDisconnect);
+
+  const killVerdaccio = () => {
+    serverProcess.off('disconnect', onUnexpectedDisconnect);
+    serverProcess.kill();
+  }
```
chunnienc, line 660: nit: `};`

chunnienc (approved):

PR #7709, opened 2023-05-22
Recover the default backend

qjia7:
Creating tensor from WebGLData|WebGPUData is specific for webgl/webgpu backend. We should save the default backend and recover it at the end of the case. Otherwise, all other cases will switch to, for example `custom-webgl` backend.


Linchenn (approved):

mattsoulanille (approved):
LGTM. In the future, we might want to automatically reset the state after each example so we don't have to do this.

chunnienc (approved):
Thanks!

qjia7:
> LGTM. In the future, we might want to automatically reset the state after each example so we don't have to do this.

Glad to know. And other thing is do we have `tidy` to expose those tensors? I only see the webgpu case explicitly dispose tensors, but not others.

PR #7717, opened 2023-05-24
Add register name when register the class object

fengwuyao:
When register the class object:
- Generate the register name for the object.
- By default, the register name is "Custom>className"
- Construct a mapping between class object and register name.
- Add a getRegisterName() function to get registered name in mapping. If not registered, then return the name of input class.


tfjs-core/src/serialization.ts
```diff
@@ -151,15 +162,63 @@ export class SerializationMap {
...
+ * console.log(tf.serialization.GLOBALCUSTOMNAMES.get(MyCustomLayer));
  * ```
  *
  * @param cls The class to be registered. It must have a public static member
  *   called `className` defined and the value must be a non-empty string.
- *
+ * @param pkg The pakcage name that this class belongs to. This used to define
+ *     the key in GlobalCustomObject. If not defined, it defaults to `Custom`.
+ * @param name The name that user specified. It defaults to the actual name of
+ *     the class.
```
mattsoulanille, line 217:
```suggestion
 *     the class as specified by its static `className` property.
```

tfjs-core/src/serialization.ts
```diff
@@ -54,6 +54,17 @@ export declare type SerializableConstructor<T extends Serializable> = {
...
     (cls: SerializableConstructor<T>, config: ConfigDict) => T;
 
+// export const GLOBALCUSTOMOBJECT:
+//     {[key: string]: SerializableConstructor<Serializable>} = {};
+export const GLOBALCUSTOMOBJECT =
+    new Map<string, SerializableConstructor<Serializable>>();
+
+// export const GLOBALCUSTOMNAMES: {[key:
+// SerializableConstructor<Serializable>]: string} = {};
+export const GLOBALCUSTOMNAMES =
```
mattsoulanille, line 64: Do we need to export this?

tfjs-core/src/serialization.ts
```diff
@@ -54,6 +54,17 @@ export declare type SerializableConstructor<T extends Serializable> = {
 export declare type FromConfigMethod<T extends Serializable> =
     (cls: SerializableConstructor<T>, config: ConfigDict) => T;
 
+// export const GLOBALCUSTOMOBJECT:
+//     {[key: string]: SerializableConstructor<Serializable>} = {};
+export const GLOBALCUSTOMOBJECT =
```
mattsoulanille, line 59: Same here, do we need to export this?

fengwuyao: Right, we do not need to export this.

tfjs-core/src/serialization.ts
```diff
@@ -54,6 +54,13 @@ export declare type SerializableConstructor<T extends Serializable> = {
 export declare type FromConfigMethod<T extends Serializable> =
     (cls: SerializableConstructor<T>, config: ConfigDict) => T;
 
+const GLOBAL_CUSTOM_OBJECT =
```
pyu10055, line 57: add comments for explain the usages of these two global map

tfjs-core/src/serialization.ts
```diff
@@ -173,5 +228,47 @@ export function registerClass<T extends Serializable>(
...
+
+  const className = name;
+  const registerName = pkg + '>' + className;
+
+  if (GLOBAL_CUSTOM_OBJECT.has(registerName)) {
+    throw new Error(
+        `The name "${registerName}" has already been registered with class: ${
+            GLOBAL_CUSTOM_OBJECT.get(registerName)}`);
+  }
+  if (GLOBAL_CUSTOM_NAMES.has(cls)) {
```
pyu10055, line 247: this class registration check can be do at the beginning of the method.

pyu10055 (commented):
LGTM with nit

tfjs-core/src/serialization.ts
fengwuyao: SG, added the explanation and example usage

Linchenn (approved):
LGTM, thanks!

PR #7720, opened 2023-05-26
Update model saving for new date saved key

fengwuyao:
Add new key for TFJS model corresponding to Keras V3 format.
- Update the saving model function when save the model


tfjs-layers/src/model_save_test.ts
```diff
@@ -53,6 +53,14 @@ describeMathCPUAndGPU('LayersModel.save', () => {
     expect(handler.savedArtifacts.convertedBy).toEqual(null);
   });
 
+  it('Save model artifact with dateSaved key', async () => {
+    const model = tfl.sequential();
+    const handler = new IOHandlerForTest();
+    model.add(tfl.layers.dense({units: 1, inputShape: [3]}));
+    await model.save(handler);
+    expect(handler.savedArtifacts.dateSaved).not.toBeNull();
```
mattsoulanille, line 61: Nit: Can we test for the value being date-like? This regex isn't perfectly specific, but it approximates what we expect.
```suggestion
    expect(handler.savedArtifacts.dateSaved).toMatch(/\d+-\d+\d+@\d+:\d+:\d+/);```

mattsoulanille (approved):
LGTM with a nit

Linchenn:
Is this PR still valid? Otherwise, we could close it.

fengwuyao:
Thanks for pointing out. Removed since not needed.

PR #7724, opened 2023-05-30
webgpu: Move the readSync warning position

qjia7:
The warning should only happen when there is really data reading back from gpu to cpu.


Linchenn (approved):
LGTM, thanks!

gyagp (approved):
LGTM

Linchenn:
Thanks Jiajia for fixing it!

For future reference: the small synchronous models, like MobileNet, would no longer issue this warnings, even though they use dataSync at https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/src/operations/executors/utils.ts#L55.

PR #7678, opened 2023-05-11
[webgpu] Implement draw with graphics draw

axinging:
To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

PR #7744, opened 2023-06-07
Avoid draw API warning if the current backend does not have implementation

Linchenn:
Since Draw API on WebGL and [WebGPU](https://github.com/tensorflow/tfjs/pull/7678) backends are in progress, we could avoid the warning until they are launched.


tfjs-core/src/ops/browser.ts
```diff
@@ -372,10 +373,13 @@ export async function toPixels(
 
   if (canvas != null) {
     if (!hasToPixelsWarned) {
-      console.warn(
-          'tf.browser.toPixels is not efficient to draw tensor on canvas. ' +
-          'Please try tf.browser.draw instead.');
-      hasToPixelsWarned = true;
+      const backendName = getBackend();
+      if (backendName !== 'webgl' && backendName !== 'webgpu') {
```
qjia7, line 377: How about checking like below? Then we don't need to revisit it anymore.
``` 
 const kernel = getKernel(Draw, ENGINE.backendName);
  if (kernel != null) {
        console.warn(
            'tf.browser.toPixels is not efficient to draw tensor on canvas. ' +
            'Please try tf.browser.draw instead.');
  }

Linchenn: Nice catch, thanks!

mattsoulanille (approved):

qjia7 (approved):
LGTM, thanks!

PR #7746, opened 2023-06-08
[WebGPU] support SparseSegmentSum and SparseSegmentMean

xhcao:

### DIFF
--- a/tfjs-backend-webgpu/src/kernel_utils/sparse_segment_reduce.ts
+++ b/tfjs-backend-webgpu/src/kernel_utils/sparse_segment_reduce.ts
@@ -1,0 +1,77 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import {TensorInfo, TypedArray, util} from '@tensorflow/tfjs-core';
+
+import {WebGPUBackend} from '../backend_webgpu';
+import {fill} from '../kernels/Fill';
+import {SparseSegmentIdCountProgram, SparseSegmentMeanProgram, SparseSegmentSumProgram} from '../sparse_segment_reduce_webgpu';
+import {WebGPUProgram} from '../webgpu_program';
+
+export function sparseSegmentReduce(
+    input: TensorInfo, indices: TensorInfo, segmentIds: TensorInfo,
+    isSum = false, backend: WebGPUBackend): TensorInfo {
+  const inputSize = util.sizeFromShape(input.shape);
+  const segmentSize = inputSize / input.shape[0];
+  const dtype = input.dtype;
+
+  // Note that the current implementation assumes that segmentIds values are
+  // sorted.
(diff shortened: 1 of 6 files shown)


tfjs-backend-webgpu/src/kernel_utils/sparse_segment_reduce.ts
```diff
@@ -0,0 +1,77 @@
...
+  const segmentSize = inputSize / input.shape[0];
+  const dtype = input.dtype;
+
+  // Note that the current implementation assumes that segmentIds values are
+  // sorted.
+  const numIndices = util.sizeFromShape(indices.shape);
+  const $segmentIds = backend.readSync(segmentIds.dataId) as TypedArray;
+  const lastSegmentIdPlusOne =
+      numIndices > 0 ? $segmentIds[numIndices - 1] + 1 : 0;
+  const outputRows = lastSegmentIdPlusOne;
```
qjia7, line 38: `outputRows` -> `numSegments`?

tfjs-backend-webgpu/src/sparse_segment_reduce_webgpu.ts
```diff
@@ -0,0 +1,136 @@
...
+    this.dispatch = computeDispatch(
+        this.dispatchLayout, segmentIdsShape, this.workgroupSize);
+
+    this.shaderKey = 'sparseSegmentIdCountProgram';
+  }
+
+  getUserCode(): string {
+    const userCode = `
+    ${main('index')} {
+      if (index < uniforms.segmentIdsSize) {
```
qjia7, line 88: Maybe directly use `uniforms.segmentIdsShape` which is the input's shape  to avoid the extra uniform `segmentIdsSize`?

qjia7 (approved):
LGTM with two nits.

gyagp (approved):
LGTM

tfjs-backend-webgpu/src/sparse_segment_reduce_webgpu.ts
xhcao: Done

tfjs-backend-webgpu/src/kernel_utils/sparse_segment_reduce.ts
xhcao: `outputRows` is not equal with `numSegments` here. For example, numSegments = [1, 2, 10], the output shape is [11, x, x, x], not [3, x, x, x] here.

qjia7: See https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/sparse-segment-mean,
```
Output: Has same shape as data, except for dimension 0 which has size k, the number of segments.
```
Here `numSegments`, I mean `the number of segments`, which I referenced the name `numSegments` in `tf.unsortedSegmentSum(x, segmentIds, numSegments)`. It's a number not an array. What do you mean `numSegments = [1, 2, 10]`?

xhcao: const c = tf.tensor2d([[1,2,3,4], [-1,-2,-3,-4], [6,7,8,9]]);
const result = tf.sparse.sparseSegmentMean(c, tf.tensor1d([0, 1, 2], 'int32'), tf.tensor1d([0, 2, 10], 'int32'));
The result's shape is [11, 4], and `outputRows = `11`, if defined `numSegments = $segmentIds[numIndices - 1] + 1 = 11` here, I think it is confused and mostly we will think `numSegments  = 3` here, only three rows are filled values from `data`, other rows are initialized into zeros.

qjia7: I think 3 stands for `numSegmentIds` not `numSegments`, which are different. I just think `outputRows` is not a good name. But if you persist, I am ok you proceed without change.

xhcao: The `outputRows` comes from cpu backend, let us keep them the same name. Thanks.

PR #7749, opened 2023-06-09
[webgpu] Implement draw API

axinging:

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/Draw.ts
+++ b/tfjs-backend-webgpu/src/kernels/Draw.ts
@@ -31,7 +31,12 @@ export function draw(
   const {imageOptions} = options || {};
   const alpha = imageOptions ?.alpha || 1;
 
-  const format = 'rgba8unorm';
+  //  'rgba8unorm' should work on macOS according to
+  //  https://bugs.chromium.org/p/chromium/issues/detail?id=1298618. But
+  //  failed on macOS/M2. So use 'bgra8unorm' first when available.
+  const format = backend.device.features.has('bgra8unorm-storage') ?
+      'bgra8unorm' :
+      'rgba8unorm';
   const outShape = [height, width];
   const program = new DrawProgram(outShape, image.dtype, format);
   canvas.width = width;
--- a/tfjs-backend-webgpu/src/kernels/Draw.ts
+++ b/tfjs-backend-webgpu/src/kernels/Draw.ts
@@ -1,0 +1,79 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use backend file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import {KernelConfig, KernelFunc, TensorInfo} from '@tensorflow/tfjs-core';
+import {Draw, DrawAttrs, DrawInputs,} from '@tensorflow/tfjs-core';
+
+import {WebGPUBackend} from '../backend_webgpu';
+import {DrawProgram} from '../draw_webgpu';
+
+export function draw(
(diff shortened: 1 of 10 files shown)


axinging:
@qjia7 @xhcao @gyagp  @hujiajie PTAL

tfjs-backend-webgpu/src/kernels/Draw.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  if (canvasWebGPU) {
+    const canvas2dContext = canvas.getContext('2d');
+    if (!canvas2dContext) {
+      throw new Error(
+          `Please make sure this canvas has only been used for 2d or webgpu context!`);
+    }
+    canvas2dContext.drawImage(canvasWebGPU, 0, 0);
+  }
+  backend.disposeData(output.dataId);
```
qjia7, line 76: I am wondering: is it possible that we meet this [error](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/src/texture_manager.ts#L88) since this texture is generated by gpuContext.getCurrentTexture() not `textureManager`?
```
Cannot release a texture that was never provided by this texture manager

File: tfjs-backend-webgpu/src/webgpu_program.ts [hunk 2]
```diff
@@ -19,6 +19,11 @@ import {backend_util, DataType, DataTypeMap, env, Rank, TensorInfo, util} from '
 
 import {symbolicallyComputeStrides} from './shader_util';
 
+export enum PixelsOpType {
+  FROM_PIXELS,
+  TO_PIXELS
```
qjia7, line 24: To keep consistent, should we rename `TO_PIXELS` to `DRAW`?

tfjs-backend-webgpu/src/kernels/Draw.ts
axinging: You are right, this tensor should be marked as external.

tfjs-backend-webgpu/src/webgpu_program.ts
axinging: Done, PTAL

qjia7 (approved):

tfjs-backend-webgpu/src/kernels/Draw.ts
```diff
@@ -0,0 +1,85 @@
...
+      'rgba8unorm';
+  const outShape = [height, width];
+  const program = new DrawProgram(outShape, image.dtype, format);
+  canvas.width = width;
+  canvas.height = height;
+  const backendName = 'webgpu';
+  let gpuContext = canvas.getContext(backendName);
+  let canvasWebGPU;
+  if (!gpuContext) {
+    canvasWebGPU = new OffscreenCanvas(width, height);
```
qjia7, line 48: Sorry, forget one. Maybe we should cache the `canvasWebGPU` if needed. Otherwise, it will be created every time when we call the `Draw` API. And the `configure` only needs to be called once if the canvas doesn't change.

axinging: canvasWebGPU is a fallback when a 2d canvas is used in webgpu context. To support the cache, we should use width, height as key, and I don't know when to free this cache, may be need a new API to free. 
So I suggest donot support cache in this PR, maybe in next PR.

tfjs-core/src/ops/draw_test.ts
```diff
@@ -16,89 +16,201 @@
...
+    actualData = new Uint8ClampedArray(
+        ctx.getImageData(0, 0, ctx.canvas.width, ctx.canvas.height).data);
   }
+  return actualData;
 }
 
-class MockCanvas {
-  context: MockContext;
+const BROWSER_CPU_WEBGPU_ENVS: Constraints = {
+  predicate: (env) =>
```
Linchenn, line 40: IIUC, the Core is not supposed to know the backend, so we could filter out tests in the setup_tests of backends. In this case, we just need to filter out WebGL (done) and WASM backends. You could add an 'exclude' in tfjs-backend-wasm/src/setup_test.ts.

Linchenn (approved):
Thanks for the amazing implementation, LGTM except a nit.

tfjs-core/src/ops/draw_test.ts
axinging: It seems wasm uses a whitelist: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/setup_test.ts#L42, so we do not need to exclude it at all?

axinging:
@Linchenn, your comment is resolved, PTAL.

Linchenn:
> @Linchenn, your comment is resolved, PTAL.

LGTM, approved, thanks!

PR #7755, opened 2023-06-13
[Local benchmark tool] Add parallel compile option

Linchenn:
This PR replaces the name `Warm up time` with `1st inference time`, because the `Warm up time` name is confusing: the number of warm up runs could be many and `Warm up time` actually presents the `1st inference time`.

The parallel compilation feature has effects only for WebGL backend and WebGPU backend.

UI Preview:
image


e2e/benchmarks/local-benchmark/index.html
```diff
@@ -406,7 +412,27 @@ <h2>TensorFlow.js Model Benchmark</h2>
       tbody.appendChild(tr);
     }
 
-    async function warmUpAndRecordTime() {
+    async function measureFistPredictTime() {
```
chunnienc, line 415: nit: `measureFistPredictTime` -> `measureFirstPredictTime`

chunnienc (approved):

e2e/benchmarks/local-benchmark/index.html
Linchenn: Thanks!

e2e/benchmarks/benchmark_util.js
```diff
@@ -280,6 +280,63 @@ async function timeInference(predict, numRuns = 1) {
...
+    throw new Error('Parallel Compilation for TFlite is not supported.');
+  }
+
+  // First inference
+  let res = predict();
+  if (parallelCompile && res instanceof Promise) {
+    throw new Error(
+        'Parallel Compilation for async function is not supported.');
+  }
+  res = await res;
```
qjia7, line 332: Why we need an extra `res = await res`? And is it possible that res is a tensor array so that we can't directly call `tf.dispose(res)`?

qjia7: Got it, thanks.

qjia7 (approved):
LGTM!

PR #7748, opened 2023-06-08
webgpu: Refactor some helper functions of matmul/conv2d

qjia7:
To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

PR #7756, opened 2023-06-14
webgpu: work around the incorrect result on Intel device

qjia7:
The correctness of some models on benchmarks on Intel device become false by https://github.com/tensorflow/tfjs/pull/7748. It seems like a driver issue. And use `row * uniforms.wShape[3] / 4 + col / 4` instead of `(row * uniforms.wShape[3] + col) / 4` can fix it.


xhcao (approved):

gyagp:
Do you know the root cause? If they can be fixed easily, I'd rather have another PR to fix it, instead of revert.

qjia7:
Close it by #7783.

PR #7757, opened 2023-06-14
Add Tokenizer base class.

pforderique:
Add the Tokenizer base class from which BytePairEncoding and other future Tokenizers will inherit from.

### DIFF
--- a/tfjs-layers/src/layers/nlp/tokenizers.ts
+++ b/tfjs-layers/src/layers/nlp/tokenizers.ts
@@ -1,0 +1,127 @@
+/**
+ * @license
+ * Copyright 2018 Google LLC
+ *
+ * Use of this source code is governed by an MIT-style
+ * license that can be found in the LICENSE file or at
+ * https://opensource.org/licenses/MIT.
+ * =============================================================================
+ */
+
+/**
+ *  Tokenizer layers.
+ */
+
+/* Original source: keras-nlp/tokenizer.py */
+import { Tensor1D, serialization, tensor1d } from '@tensorflow/tfjs-core';
+
+import { Layer } from '../../engine/topology';
+import { NotImplementedError, ValueError } from '../../errors';
+
+export declare interface TokenizerOptions {
+  mode?: 'tokenize' | 'detokenize'
+}
+
+/**
+ * Base class for Tokenizers.
+ *
+ * Subclassers should always implement the `tokenize()` method, which will also
+ * be the default when calling the layer directly on inputs.
+ */
+export abstract class Tokenizer extends Layer {
+  /**
+   * Transform input tensors of strings into output tokens.
+   *
+   * @param inputs Input tensor.
+   * @param kwargs Additional keyword arguments.
+   */
+  abstract tokenize(inputs: Tensor1D): Tensor1D[];
+
+  /**
+   * Transform tokens back into strings.
+   *
+   * @param inputs Input tensor.
+   * @param kwargs Additional keyword arguments.
+   */
+  detokenize(inputs: Tensor1D[]): Tensor1D {
+    throw new NotImplementedError(
+      `No implementation of 'detokenize()' was found for
+      ${this.constructor.name}.`
+    );
+  }
+
+  /**
+   * Get the tokenizer vocabulary as a list of strings terms.
+   */
+  get vocabulary(): string[] {
+    throw new NotImplementedError(
+      `No implementation of 'vocabulary()' was found for
+      ${this.constructor.name}.`
+    );
+  }
+
+  /**
+   * Returns the total size of the token id space.
+   */
+  get vocabularySize(): number {
+    throw new NotImplementedError(
+      `No implementation of 'vocabularySize()' was found for
+      ${this.constructor.name}.`
+    );
+  }
+
+  /**
+   * Convert an integer id to a string token.
+   */
+  idToToken(id: number): string {
+    throw new NotImplementedError(
+      `No implementation of 'idToToken()' was found for
+      ${this.constructor.name}.`
+    );
+  }
+
+  /**
+   * Convert an integer id to a string token.
+   */
+  tokenToId(token: string): number {
+    throw new NotImplementedError(
+      `No implementation of 'tokenToId()' was found for
+      ${this.constructor.name}.`
+    );
+  }
+
+  override call(inputs: Tensor1D|Tensor1D[], kwargs: TokenizerOptions={mode: 'tokenize'}): Tensor1D|Tensor1D[] {
+    if (kwargs.mode === 'tokenize') {
+      if (inputs instanceof Array) {
+        throw new ValueError(`tokenize expects Tensor1D, not Tensor1D[].`);
+      }
+      return this.tokenize(inputs);
+    }
+
+    if (kwargs.mode === 'detokenize') {
+      if (!(inputs instanceof Array)) {
+        throw new ValueError(`detokenize expects Tensor1D[], not Tensor1D.`);
+      }
+      return this.detokenize(inputs);
+    }
+
+    throw new ValueError(`Input mode=${kwargs.mode} is not supported.`)
+  }
+}
+
+export class WhiteSpaceTokenizer extends Tokenizer {
+  /** @nocollapse */
+  static readonly className = 'WhiteSpaceTokenizer';
+
+  tokenize(inputs: Tensor1D): Tensor1D[] {
+    const stringInputs = inputs.dataSync() as unknown as string[];
+    return stringInputs.map(input => tensor1d(input.split(' ')));
+  }
+
+  override detokenize(inputs: Tensor1D[]): Tensor1D {
+    const stringInputs = inputs.map(input => input.dataSync() as unknown as string[]);
+    return tensor1d(stringInputs.map(str => str.join(' ')));
+  }
+}
+
+serialization.registerClass(WhiteSpaceTokenizer);
--- a/tfjs-layers/src/layers/nlp/tokenizers_test.ts
+++ b/tfjs-layers/src/layers/nlp/tokenizers_test.ts
@@ -1,0 +1,54 @@
+/**
+ * @license
+ * Copyright 2018 Google LLC
+ *
+ * Use of this source code is governed by an MIT-style
+ * license that can be found in the LICENSE file or at
+ * https://opensource.org/licenses/MIT.
+ * =============================================================================
+ */
+
+/**
+ * Unit Tests for Tokenizer Layers.
+ */
+
+import { Tensor1D, tensor1d } from '@tensorflow/tfjs-core';
+
+import { WhiteSpaceTokenizer } from './tokenizers';
+import { expectTensorsClose } from '../../../src/utils/test_utils';
+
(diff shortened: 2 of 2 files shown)


tfjs-layers/src/layers/nlp/tokenizers.ts
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * Use of this source code is governed by an MIT-style
+ * license that can be found in the LICENSE file or at
+ * https://opensource.org/licenses/MIT.
+ * =============================================================================
+ */
+
+/**
+ *  Tokenizer layers.
+ */
```
Linchenn, line 13: You could use the license as https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/kernels/MaxPoolWithArgmax.ts#L1-L16.

pforderique: done

tfjs-layers/src/layers/nlp/tokenizers_test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  it('tokenize', () => {
+    const inputData = tensor1d(['the quick brown fox']);
+    const expectedOutput = [tensor1d(['the', 'quick', 'brown', 'fox'])];
+
+    const tokenizeOutput = tokenizer.tokenize(inputData);
+    const callOutput = tokenizer.call(inputData) as Tensor1D[];
+
+    expect(tokenizeOutput.length).toBe(1);
+    expectTensorsClose(tokenizeOutput[0], expectedOutput[0]);
```
Linchenn, line 38: `expectTensorsClose` Nice catch!

Linchenn (approved):
LGTM!

tfjs-layers/src/layers/nlp/tokenizers.ts
```diff
@@ -0,0 +1,134 @@
...
+  mode?: 'tokenize' | 'detokenize'
+}
+
+/**
+ * Base class for Tokenizers.
+ *
+ * Subclassers should always implement the `tokenize()` method, which will also
+ * be the default when calling the layer directly on inputs.
+ */
+export abstract class Tokenizer extends Layer {
```
mattsoulanille, line 38: Let's add the rest of the Keras documentation here (as long as it's applicable). Thanks!

Usually we put the docs in exports_layers.ts, but that's not possible in this PR since `Tokenizer` is an abstract class (so we can't create a function that returns an instance of it). For future layers, we should add the corresponding function to `exports_layers.ts`:

[Example](https://github.com/tensorflow/tfjs/blob/tfjs-v4.7.0/tfjs-layers/src/exports_layers.ts#L370-L403), and on the [API site](https://js.tensorflow.org/api/latest/#layers.cropping2D).

We sometimes also include runnable examples for layers, which appear in the code boxes on the API website. That would have been a good place for the WhitespaceTokenizer example that Keras has in their docs, but since this is an abstract class, I think we should put that example here instead, just so people have a reference of how to implement their own tokenizer if they want to.

tfjs-layers/src/layers/nlp/tokenizers.ts
```diff
@@ -0,0 +1,134 @@
...
+  tokenToId(token: string): number {
+    throw new NotImplementedError(
+      `No implementation of 'tokenToId()' was found for
+      ${this.constructor.name}.`
+    );
+  }
+
+  override call(inputs: Tensor1D|Tensor1D[], kwargs: TokenizerOptions={mode: 'tokenize'}): Tensor1D|Tensor1D[] {
+    if (kwargs.mode === 'tokenize') {
+      if (inputs instanceof Array) {
```
mattsoulanille, line 102: To compare against Keras, I've set up a simple BytePairTokenizer in Python following [these docs](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/tokenizers/byte_pair_tokenizer.py#L245-L247).
```python
import keras_nlp
vocab = {"butter": 1, "fly": 2}
merge = ["b u", "t t", "e r", "bu tt", "butt er", "f l", "fl y"]
tokenizer = keras_nlp.tokenizers.BytePairTokenizer(vocab, merge)
```

It lets me pass a Scalar or a 1D tensor when tokenizing:
```python
>>> tokenizer.tokenize(tf.constant('butterfly'))
<tf.Tensor: shape=(2,), dtype=int32, numpy=array([1, 2], dtype=int32)>
>>> tokenizer.tokenize(tf.constant(['butterfly']))
<tf.RaggedTensor [[1, 2]]>
```
Note the difference in shapes:
```python
>>> tf.constant('butterfly').shape
TensorShape([])
>>> tf.constant(['butterfly']).shape
TensorShape([1])
```
I think we might want to also allow `Scalar` (from tfjs-core) as an input type to tokenize.

tfjs-layers/src/layers/nlp/tokenizers.ts
```diff
@@ -0,0 +1,134 @@
...
+   * Convert an integer id to a string token.
+   */
+  tokenToId(token: string): number {
+    throw new NotImplementedError(
+      `No implementation of 'tokenToId()' was found for
+      ${this.constructor.name}.`
+    );
+  }
+
+  override call(inputs: Tensor1D|Tensor1D[], kwargs: TokenizerOptions={mode: 'tokenize'}): Tensor1D|Tensor1D[] {
```
mattsoulanille, line 100: Python lets me pass just an array of strings or a string all on its own without converting them to tensors. Ping told me we have some utilities for dealing with these kinds of inputs, but they're not exported from TFJS yet. We could export them and use them in layers to make consuming inputs like these easier (but that can be a separate PR).
* [convertToTensor](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/tensor_util_env.ts#L98) is similar to [tf.convert_to_tensor](https://www.tensorflow.org/api_docs/python/tf/convert_to_tensor), which [BytePairTokenizer uses](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/tokenizers/byte_pair_tokenizer.py#L500)
* [TensorLike (and similar types)](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/types.ts#L141-L164) are types that aren't tensors, but which can be converted to tensors with convertToTensor.

I've also found some interesting behavior in the Keras version with tensors that have more than one dimension. If I use a 2d tensor, it flattens the inner dimensions by concatenating the strings, it seems. This is likely BPE implementation details.
```python
>>> tokenizer.tokenize(tf.constant([['butterfly', 'flybutterfly'], ['flyfly', 'butterbutter']]))
<tf.RaggedTensor [[1, 2, 2, 1, 2], [2, 2, 1, 1]]>
>>> tokenizer.tokenize(tf.constant(['butterflyflybutterfly', 'flyflybutterbutter']))
<tf.RaggedTensor [[1, 2, 2, 1, 2], [2, 2, 1, 1]]>
```

mattsoulanille, line 100: Setting the default value for kwargs is tricky since it's an object. If we set the default value to a full object and the user provides an object that's missing the `map` key but has some other (future) keys that we add, we no longer have `map` set to its intended default of `'tokenize'` (since the user-provided object replaced the default object).

Instead, we can destructure the object and set defaults for the destructured values

```suggestion
  override call(inputs: Tensor1D|Tensor1D[], {mode = 'tokenize'}: TokenizerOptions): Tensor1D|Tensor1D[] {
```

You'll have to update the body of the function too. If you need to call `super` with the same kwargs variable, you can instead do the destructuring in the body of the function, like this.
```typescript
  const {mode = 'tokenize'} = kwargs;
```

tfjs-layers/src/layers/nlp/tokenizers_test.ts
```diff
@@ -0,0 +1,61 @@
...
+ * Unit Tests for Tokenizer Layers.
+ */
+
+import { Tensor1D, tensor1d } from '@tensorflow/tfjs-core';
+
+import { WhiteSpaceTokenizer } from './tokenizers';
+import { expectTensorsClose } from '../../../src/utils/test_utils';
+
+describe('White Space Tokenizer', () => {
+  const tokenizer = new WhiteSpaceTokenizer();
```
mattsoulanille, line 28: Your class is stateless, so this isn't strictly necessary. However, it's usually good to put this in a `beforeEach()` block to make sure you get a new instance of `WhiteSpaceTokenizer` for each `it` spec. Otherwise, it will be reused across all of them, which may have unexpected results (e.g. maybe one of the tests accidentally changes the `tokenize` method and breaks all the other tests).


```suggestion
  let tokenizer: WhiteSpaceTokenizer;
  beforeEach(() => {
    tokenizer = new WhiteSpaceTokenizer();
  });
```

mattsoulanille (changes requested):
This looks good! I have a few minor changes, and some longer comments on some things I've discovered about BytePairTokenizer.

tfjs-layers/src/layers/nlp/tokenizers.ts
```diff
@@ -0,0 +1,134 @@
...
+    return stringInputs.map(input => tensor1d(input.split(' ')));
+  }
+
+  override detokenize(inputs: Tensor1D[]): Tensor1D {
+    const stringInputs = inputs.map(input => input.dataSync() as unknown as string[]);
+    return tensor1d(stringInputs.map(str => str.join(' ')));
+  }
+}
+
+serialization.registerClass(WhiteSpaceTokenizer);
```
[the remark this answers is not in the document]
pforderique: done

tfjs-layers/src/layers/nlp/tokenizers.ts
pforderique: done

tfjs-layers/src/layers/nlp/tokenizers_test.ts
pforderique: done

tfjs-layers/src/layers/nlp/tokenizers_test.ts
```diff
@@ -19,17 +19,38 @@
...
+    return stringInputs.map(input => tensor1d(input.split(' ')));
+  }
+
+  override detokenize(inputs: Tensor1D[]): Tensor1D {
+    const stringInputs = inputs.map(
+      input => input.dataSync() as unknown as string[]);
+    return tensor1d(stringInputs.map(str => str.join(' ')));
+  }
+}
+serialization.registerClass(SimpleTokenizer);
```
mattsoulanille, line 42: We're not saving or loading models that use this, so we don't need to register it (since it's just a test class).

```suggestion
```

mattsoulanille (approved):
LGTM with a couple of nits.

tfjs-layers/src/layers/nlp/tokenizers_test.ts
pforderique: done

tfjs-layers/src/layers/nlp/tokenizers.ts
pforderique: This will also be potentially resolved in a future PR as needed

PR #7600, opened 2023-04-19
[webgpu] Add getShader in backend

axinging:
Currently this only supports webgpu backend.

Example code:
```
  await tf.setBackend('webgpu');
  await tf.ready();

  const shader = tf.backend().getShader(() => {
    {
      const a = tf.tensor1d([1, 2, 3, 4, 5, 6], 'int32');
      const b = tf.tensor1d([1, 2, 3, 4, 5, 6], 'int32');
      const c = tf.add(a, b);
    };
    {
      const a = tf.tensor1d([1, 2, 3, 4, 5, 6], 'int32');
      const b = tf.pad1d(a, [2, 3]);
    };
  });

  for (const key in shader) {
    console.group(key);
    console.debug(shader[key]);
    console.groupEnd();
  }
```

Bug: https://github.com/tensorflow/tfjs/issues/7516

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

PR #7759, opened 2023-06-16
[Experiment] Improve downloadable cache for WebGPU backend 

Linchenn:
### Use Case
We want to improve a use case: users want to run one large model and they want to download a cache after the first model inference. Then they want to use the cache to speed up the first model inference in another environment.

## Approach
From my perspective, one downloadable cache is the shader strings, so we could save the map of shaderKey->fullShader and add 'exportCache' and 'importCache' API to export/import the map. This cache could reduce the time for `makeShader` (the string interpolation).

## Experiment
I took a relative large model, DeepLabV3, to prove the improvement. (A large model spends relatively more time in `makeShader`) 

### MakeShader time
I found `makeShader` takes 5.9ms of 572ms (first inference). (ran a couple times of this, `makeShader` takes 3~11ms).
image

### ImportCache time
I am using the `export.html` in this PR:
* Open the file and call `exportCache()`;
* Refresh the file and call `importCache()`;

Found the time to import shaders (copy the strings to the pipelineCache) is 4.4ms.

Even `importCache` is free, I feel this improvement may be trivial. What do you think? @qjia7 

### DIFF
--- a/export.html
+++ b/export.html
@@ -1,0 +1,40 @@
+<html>
+  <script src="https://unpkg.com/@tensorflow/tfjs-core@4.6.0"></script>
+  <script src="https://unpkg.com/@tensorflow/tfjs-converter@4.6.0"></script>
+  <script src="./dist/bin/tfjs-backend-webgpu/dist/tf-backend-webgpu.js"></script>
+
+  <script>
+    async function exportCache() {
+      window.localStorage.clear()
+      await tf.setBackend('webgpu');
+      await modelInference();
+      const cache = tf.backend().exportCache();
+      window.localStorage.setItem('webgpuCache', JSON.stringify(cache));
+    }
+
+    function importCache() {
+      const cache = JSON.parse(window.localStorage.getItem('webgpuCache'));
+      tf.backend().importCache(cache);
+    }
+
+    async function modelInference() {
+      const modelUrl = 'https://storage.googleapis.com/tfhub-tfjs-modules/tensorflow/tfjs-model/deeplab/pascal/1/quantized/2/1/model.json';
+      const input = tf.ones([1, 227, 500, 3], 'int32');
+      const model = await tf.loadGraphModel(modelUrl);
+      await model.predict(input).data();
+    }
+
+    async function benchmark(useCache = false) {
+      await tf.setBackend('webgpu');
+
+      const time1 = performance.now();
+      if (useCache) {
+        importCache();
+      }
+      const time2 = performance.now();
+      await modelInference();
+      const time3 = performance.now();
+      console.log(`Cache time: ${time2 - time1}, inference time: ${time3 - time2}`);
+    }
+  </script>
+</html>
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -60,6 +60,15 @@ export interface WebGPUTimingInfo extends TimingInfo {
   downloadWaitMs: number;
 }
 
+export type WebGPUProgramCache = {
+  shader: string,
+  pipeline: GPUComputePipeline|Promise<GPUComputePipeline>
+}
+
+export type ShaderCache = {
+  [key: string]: string
+}
+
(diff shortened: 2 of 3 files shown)


qjia7:
Personally, I think it's more like a new feature to easily dump all shaders source code by the interface `exportCache()` instead of performance improvement. From this point, I think it's an improvement for the current shader dump way which directly print shader source code in console. @axinging is working on another `getShader` way by #7600. I think we can work together to get a good solution.

BTW, for the use case you provided, in another environment, the user needs to download the shader cache from internet, right? If we  allow users to import external shaders to webgpu, will it bring security issue? And we also can't make sure the correctness of them.  Not sure whether it is acceptable for tfjs.

Linchenn:
> Personally, I think it's more like a new feature to easily dump all shaders source code by the interface `exportCache()` instead of performance improvement. From this point, I think it's an improvement for the current shader dump way which directly print shader source code in console. @axinging is working on another `getShader` way by #7600. I think we can work together to get a good solution.

Sure, my initiative is just the performance improvement, so I will take a look at Xing's PR. 

> BTW, for the use case you provided, in another environment, the user needs to download the shader cache from internet, right? If we allow users to import external shaders to webgpu, will it bring security issue? And we also can't make sure the correctness of them. Not sure whether it is acceptable for tfjs.

For our current use case, the users just save the cache and leave the page. Then they will open the page again and load the cache to speed up the model inference. The cache could be saved in the browser by the users.

pyu10055:
> > Personally, I think it's more like a new feature to easily dump all shaders source code by the interface `exportCache()` instead of performance improvement. From this point, I think it's an improvement for the current shader dump way which directly print shader source code in console. @axinging is working on another `getShader` way by #7600. I think we can work together to get a good solution.
> 
> Sure, my initiative is just the performance improvement, so I will take a look at Xing's PR.
> 
> > BTW, for the use case you provided, in another environment, the user needs to download the shader cache from internet, right? If we allow users to import external shaders to webgpu, will it bring security issue? And we also can't make sure the correctness of them. Not sure whether it is acceptable for tfjs.
> 
> For our current use case, the users just save the cache and leave the page. Then they will open the page again and load the cache to speed up the model inference. The cache could be saved in the browser by the users.

To clarify, the user would not leave the page, it is for webworkers, one webworker will execute the model, which include the shader creation and compilation, once the inference is done the worker will be killed. On the next inference, a new worker will be created, the goal is to reduce the shader warmup time for the new worker.

qjia7:
> > > Personally, I think it's more like a new feature to easily dump all shaders source code by the interface `exportCache()` instead of performance improvement. From this point, I think it's an improvement for the current shader dump way which directly print shader source code in console. @axinging is working on another `getShader` way by #7600. I think we can work together to get a good solution.
> > 
> > 
> > Sure, my initiative is just the performance improvement, so I will take a look at Xing's PR.
> > > BTW, for the use case you provided, in another environment, the user needs to download the shader cache from internet, right? If we allow users to import external shaders to webgpu, will it bring security issue? And we also can't make sure the correctness of them. Not sure whether it is acceptable for tfjs.
> > 
> > 
> > For our current use case, the users just save the cache and leave the page. Then they will open the page again and load the cache to speed up the model inference. The cache could be saved in the browser by the users.
> 
> To clarify, the user would not leave the page, it is for webworkers, one webworker will execute the model, which include the shader creation and compilation, once the inference is done the worker will be killed. On the next inference, a new worker will be created, the goal is to reduce the shader warmup time for the new worker.

Usually, the browser has the cache strategy to make sure that if the shader has been compiled once, the next compilation (refresh the page/relaunch the browser) will be speedup by the binary for the same shader source.
 To reduce the first inference time, the most efficiently way so far is to enable the parallel compilation. So you are considering how to further reduce the first inference time even with parallel compilation, right? Personally, I think using two workers may not help much. You need to add the extra time to communicate between workers (for example, collect all shader sources in first work and pass them? ). And since you already go through the whole model to get all shader sources, why not directly execute them in the first worker, but close it  and create a new worker for inference? It seems that the overall time is not reduced.

Maybe I don't understand the use case very well. If you find this way does help users speedup, I am glad to have it.

PR #7760, opened 2023-06-16
[BPE PR 2.1] BPE Utils

pforderique:
Included is bytesToUnicode, createStaticHashTable, and BytePairTokenizerCache with respective tests that will be needed for BPE.

### DIFF
--- a/tfjs-layers/src/layers/nlp/tokenizers_utils.ts
+++ b/tfjs-layers/src/layers/nlp/tokenizers_utils.ts
@@ -51,9 +51,11 @@ export function bytesToUnicode(): [Uint8Array, string[]] {
 }
 
 /**
- * HashTable extends Map and includes a `lookup` function that
+ * StaticHashTable extends Map and includes a `lookup` function that
  */
-class HashTable<K, V extends number|string> extends Map<K, V> {
+class StaticHashTable<K, V extends number|string> {
+  private _map: Map<K, V>;
+
   constructor(keys: K[], values: V[], private readonly defaultValue: V) {
     if (keys.length !== values.length) {
       throw new ValueError(`keys and values arrays must be same length.
@@ -67,14 +69,11 @@ class HashTable<K, V extends number|string> extends Map<K, V> {
       keyValPairs.push([key, val]);
     }
 
-    super(keyValPairs);
+    this._map = new Map(keyValPairs);
   }
 
-  override get(key: K) {
-    if (!this.has(key)) {
-      this.set(key, this.defaultValue);
-    }
-    return super.get(key);
+  get(key: K): V {
+    return this._map.get(key) || this.defaultValue;
   }
 
   lookup(keys: Tensor[]): Tensor[] {
@@ -91,8 +90,8 @@ class HashTable<K, V extends number|string> extends Map<K, V> {
   }
 }
 
-export function createHashtable<T1, T2 extends number|string>(
-  keys: T1[], values: T2[], defaultVal: T2): HashTable<T1, T2> {
+export function createStaticHashtable<T1, T2 extends number|string>(
+  keys: T1[], values: T2[], defaultVal: T2): StaticHashTable<T1, T2> {
 
-  return new HashTable(keys, values, defaultVal);
+  return new StaticHashTable(keys, values, defaultVal);
 }
--- a/tfjs-layers/src/layers/nlp/tokenizers_utils.ts
+++ b/tfjs-layers/src/layers/nlp/tokenizers_utils.ts
@@ -86,10 +86,10 @@ class StaticHashTable<K, V extends number|string> {
         innerValues.push(this.get(key));
       }
 
-      return tensor(innerValues as V[], t.shape);
+      return tensor(innerValues, t.shape);
     });
 
-    return await Promise.all(values)
+    return Promise.all(values);
   }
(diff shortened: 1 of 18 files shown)


tfjs-layers/src/layers/nlp/tokenizers_utils.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  get(key: K): V {
+    return this._map.get(key) || this.defaultValue;
+  }
+
+  lookup(keys: Tensor[]): Tensor[] {
+    const values = keys.map(tensor => {
+      const innerValues: V[] = [];
+      for (const key of tensor.dataSync() as unknown as K[]) {
```
mattsoulanille, line 82: We should avoid dataSync since it's not efficient for WebGPU. You can use `await tensor.data()` instead. That'll make the whole function async, but that's okay I think. I expect the model pipeline will be async as well.

tfjs-layers/src/layers/nlp/tokenizers_utils.ts
```diff
@@ -0,0 +1,139 @@
...
+        innerValues.push(this.get(key));
+      }
+
+      return tensor1d(innerValues as string[]|number[]);
```
mattsoulanille, line 86: For more general support (although we don't need it right now), we can preserve the shape of each tensor in the list (although I'm pretty sure they'll all be 1d). 
```suggestion
      return tensor(innerValues as V[], tensor.shape);
```

tfjs-layers/src/layers/nlp/tokenizers_utils.ts
```diff
@@ -0,0 +1,139 @@
...
+    });
+
+    return values;
+  }
+}
+
+export function createStaticHashtable<T1, T2 extends number|string>(
```
mattsoulanille, line 93: Tiny nit: Maybe use `K` and `V` instead of `T1` and `T2`.

tfjs-layers/src/layers/nlp/tokenizers_utils.ts
```diff
@@ -0,0 +1,139 @@
...
+  const chars = cs.map(n => String.fromCharCode(n));
+
+  // TODO(orderique): Verify same functionality.
+  const bytes = Uint8Array.from(bs);
+
+  return [bytes, chars];
+}
+
+/**
+ * StaticHashTable extends Map and includes a `lookup` function that
```
mattsoulanille, line 54: I don't think this extends map anymore.

tfjs-layers/src/layers/nlp/tokenizers_utils.ts
```diff
@@ -0,0 +1,139 @@
...
+  constructor() {
+    this._cache = new Map();
+  }
+
+  /**
+   * Insert token <=> encoded outputs pairs.
+   */
+  insert(keys: Tensor|string[], values: string[]): BytePairTokenizerCache {
+    const arrKeys = keys instanceof Tensor ?
+      keys.dataSync() as unknown as string[] : keys;
```
mattsoulanille, line 125: If this is not called often, then dataSync is fine.

tfjs-layers/src/layers/nlp/tokenizers_utils.ts
```diff
@@ -0,0 +1,139 @@
...
+    arrKeys.forEach((key, idx) => this._cache.set(key, values[idx]));
+    return this;
+  }
+
+  /**
+   * Look up the encoded outputs of given tokens.
+   */
+  lookup(keys: Tensor|string[]): string[] {
+    const arrKeys = keys instanceof Tensor ?
+      keys.dataSync() as unknown as string[] : keys;
```
mattsoulanille, line 136: If lookup is used often, you may want to make this async and use `await data` instead of dataSync.

mattsoulanille (changes requested):
The tests look good! I have a few minor comments about the implementation.

tfjs-layers/src/layers/nlp/tokenizers_utils.ts
```diff
@@ -0,0 +1,139 @@
...
+      const key = keys[idx];
+      const val = values[idx];
+      keyValPairs.push([key, val]);
+    }
+
+    this._map = new Map(keyValPairs);
+  }
+
+  get(key: K): V {
+    return this._map.get(key) || this.defaultValue;
```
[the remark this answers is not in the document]
mattsoulanille: You could try adding a test where the map contains a zero and the default value is non-zero to see this error surface and make sure my idea to fix it works.

tfjs-layers/src/layers/nlp/tokenizers_utils.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Cache that stores the encoded result of seen tokens.
+ *
+ * The cache key is string tensor or python strings, and the value is split
+ * tokens joined by whitespace. For example, "dragonfly" => "dragon fly"
+ *
+ * Examples:
+ *
+ * ```
```
mattsoulanille, line 107:
```suggestion
 * ```js
```

Linchenn (approved):
LGTM!

mattsoulanille (approved):
LGTM!

PR #7774, opened 2023-06-22
[BPE PR 2.4] BPE Class + Merge utilities 

pforderique:
This PR includes the base BPE class implementation including private merge functions that will be needed for the tokenize() function. 

This PR only includes one test for the BPE setup configuration and no test suite for the merge functions since they are private.

PR #7770, opened 2023-06-21
[BPE PR 2.2] All BPE Utils

pforderique:
Add all of the outside utilities needed by the BPE class including all utils from PR 2.1 + removeStringsFromInputs + createAltsForUnsplittableTokens + splitStringsForBpe, and a custom regexSplit function for working with Tensor[] (in place of Ragged Tensors). 

NOTE:
The keras [implementation](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/tokenizers/byte_pair_tokenizer.py#L52) uses regex that includes \p{L} and \p{N}, used to represent all letters and numbers in any script, respectively. I was unable to find a way to support this in TS in a multi-line regex pattern, so I had to manually include some additional letters in this [implementation](https://github.com/pforderique/tfjs/blob/all-bpe-utils/tfjs-layers/src/layers/nlp/tokenizers_utils.ts#L183). 

This should remain a bug as to support all the necessary characters - hopefully in a cleanup PR (BPE PR 3).

PR #7780, opened 2023-06-23
[BPE PR 2.5] BytePairTokenizer tokenize function.

pforderique:
This PR implements the tokenize functionality of BytePairTokenizer and associated test cases.

Depends on BPE PR #7770 and #7774.


Linchenn:
Overall LGTM, great work! Could you help add some comments for BPE, like https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/tokenizers/byte_pair_tokenizer.py#L202?

tfjs-layers/src/layers/nlp/tokenizers.ts
```diff
@@ -436,37 +424,108 @@ export class BytePairTokenizer extends Tokenizer {
...
+        for (let i = 0; i < pad.length; i++) {
+          if (pad[i] === undefined) {
+            break;
+          }
+          newRow.push(tokenArr[i]);
+        }
+        return newRow;
+      });
+    }
+
```
Linchenn, line 527: I might miss something, do we need to consider scalar input? https://github.com/keras-team/keras-nlp/blob/740ecababa77622359025ec82d21b978d8d95831/keras_nlp/tokenizers/byte_pair_tokenizer.py#L551C6-L551C6

pforderique: I believe the reason we chose not to for now is because this `tokenize` method is only being used by the `call` method, which doesn't accept scalar inputs. I can make this change just for `tokenize` in the polish PR later on however, so we can have that extra functionality.

Linchenn: Got it. We could add it when `call` supports it. Thanks!

mattsoulanille (changes requested):

tfjs-layers/src/layers/nlp/tokenizers_utils.ts
```diff
@@ -242,18 +249,16 @@ export function regexSplit(
   });
 }
 
-export async function tensorToStringArr(input: Tensor): Promise<string[]> {
-  return await input.data() as unknown as string[];
+export function tensorToArr(input: Tensor): unknown[] {
+  return input.dataSync() as unknown as unknown[];
 }
```
mattsoulanille, line 254: You might be able to use [`tensor.arraySync()` instead](https://js.tensorflow.org/api/latest/#tf.Tensor.arraySync), although it returns a `number[]`, so maybe not.

Linchenn (approved):

tfjs-layers/src/layers/nlp/tokenizers.ts
```diff
@@ -436,37 +458,108 @@ export class BytePairTokenizer extends Tokenizer {
...
+    const flatTokens = rawTokens.reduce((acc, e) => acc.concat(e), []);
+
+    // Check cache.
+    const cacheLookup = this.cache.lookup(flatTokens);
+    const cacheMask = cacheLookup.map(e => e === '');
+
+    const emptyFlatTokensMask = flatTokens.map(e => e !== '');
+    const hasUnseenWords = cacheMask.map(
+      (bool, idx) => bool && emptyFlatTokensMask[idx]).some(e => e);
```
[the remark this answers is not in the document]
pforderique: Good catch! I've realized that `some()` can break early, so I've rewritten this to skip some of the middle `map` calls.

tfjs-layers/src/layers/nlp/tokenizers.ts
```diff
@@ -436,37 +458,108 @@ export class BytePairTokenizer extends Tokenizer {
...
+
+      token2D.map(tokenArr => {
+        const pad: number[] = Array(maxLength - tokenArr.length).fill(0);
+
+        const newRow = [];
+        for (let i = 0; i < pad.length; i++) {
+          if (pad[i] === undefined) {
+            break;
+          }
+          newRow.push(tokenArr[i]);
```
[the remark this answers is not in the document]
pforderique: Ah yes, I was trying to both truncate and pad each array element as needed, but I agree - I don't think this was doing what I thought it was. I've learned how to use `tf.slice` and `tf.pad` since writing this so I've changed this to work directly on tensors now and added coverage tests for this branch.

tfjs-layers/src/layers/nlp/tokenizers_utils.ts
pforderique: Sounds good! I've added this to my notes so I can come back to this once the return type of `arraySync` has been updated.

pforderique:
@mattsoulanille comments have been resolved. Thanks!

tfjs-layers/src/layers/nlp/tokenizers_test.ts
```diff
@@ -198,4 +198,33 @@ describe('BytePairTokenizer', () => {
...
+
+    const tokenizeOutput = tokenizer.tokenize(inputData);
+
+    expect(tokenizeOutput.length).toBe(3);
+    expectTensorsClose(tokenizeOutput[0], expectedOutput[0]);
+    expectTensorsClose(tokenizeOutput[1], expectedOutput[1]);
+    expectTensorsClose(tokenizeOutput[2], expectedOutput[2]);
+  });
+
+  it('tokenize with does not leak memory', () => {
```
mattsoulanille, line 218: Nit
```suggestion
  it('tokenize does not leak memory', () => {
```

mattsoulanille (approved):
LGTM

PR #7788, opened 2023-06-26
BPE Detokenize

pforderique:
Implements the detokenize method for BytePairEncoding.

Note:
Kera’s detokenize() currently uses Python’s byte strings (b’my text here’) so their implementation is a bit different than what I have here. I’ve added a test case for now that works on both implementations, highlighting that `detokenize(tokenize(input))` should equal the original `input`, but there might be something I’m missing due to the omitted part of the implementation. This can be pushed off to the clean up PR in the future, but should be okay for now.


mattsoulanille:
LGTM once #7780 is merged.

mattsoulanille (approved):

Linchenn (approved):
LGTM!

PR #7790, opened 2023-06-28
[gpt2pre 1.1] Start End Packer Base class

pforderique:
Implements the base for the StartEndPacker class including it's `getConfig()` method.
This is PR 1/2 of the full StartEndPacker implementation. 

`call()`  and its respective documentation will be added in [gpt2pre 1.2] (PR 2/2).

### DIFF
--- a/tfjs-layers/src/layers/nlp/models/start_end_packer.ts
+++ b/tfjs-layers/src/layers/nlp/models/start_end_packer.ts
@@ -1,0 +1,111 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+/**
+ *  Start End Packer implementation based on `tf.layers.Layer`.
+ */
+
+/* Original source: keras-nlp/start_end_packer.py */
+import { Tensor, serialization } from '@tensorflow/tfjs-core';
+
+import { Layer, LayerArgs } from '../../../engine/topology';
+import { NotImplementedError } from '../../../errors';
+
+export declare interface StartEndPackerArgs extends LayerArgs {
+  /**
+   * Integer. The desired output length.
+   */
+  sequenceLength: number;
+
+  /**
+   * Integer or string. The ID or token that is to be placed at the start of
+   * each sequence. The dtype must match the dtype of the input tensors to the
+   * layer. If undefined, no start value will be added.
+   */
+  startValue?: number|string;
+
+  /**
+   * Integer or string. The ID or token that is to be placed at the end of each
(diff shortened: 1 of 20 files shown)


tfjs-layers/src/layers/nlp/models/start_end_packer.ts
```diff
@@ -0,0 +1,111 @@
+/**
```
mattsoulanille, line 1: This is in the `models/` directory, but on keras_nlp, it's in `preprocessing/`. Was this on purpose?

tfjs-layers/src/layers/nlp/models/start_end_packer.ts
```diff
@@ -0,0 +1,111 @@
...
+  private padValue: number|string;
+  private returnPaddingMask: boolean;
+
+  constructor(args: StartEndPackerArgs) {
+    super(args);
+
+    this.sequenceLength = args.sequenceLength;
+    this.startValue = args.startValue;
+    this.endValue = args.endValue;
+    this.padValue = args.padValue;
```
mattsoulanille, line 90: Do we need defaults for these in case they are null in `args`? Since we don't use strictNullChecks, TS will not complain about it, but it may fail in practice. We also might want tests for this as well.

tfjs-layers/src/layers/nlp/models/start_end_packer_test.ts
```diff
@@ -0,0 +1,39 @@
...
+      startValue: 10,
+      endValue: 20,
+      padValue: 100,
+    });
+    const config = startEndPacker.getConfig();
+
+    expect(config.sequenceLength).toEqual(512);
+    expect(config.startValue).toEqual(10);
+    expect(config.endValue).toEqual(20);
+    expect(config.padValue).toEqual(100);
```
mattsoulanille, line 37: Really small nit: You can save the input to StartEndPacker and make sure that the output of getConfig has the same values for the properties that the input has. Feel free to leave it as-is if you prefer.

```ts
const config = {
  sequenceLength: 512,
  ...,
};

...
const outputConfig = startEndPacker.getConfig();
expect(outputConfig).toEqual(jasmine.objectContaining(config));

**Review by mattsoulanille — CHANGES_REQUESTED** [review]

File: tfjs-layers/src/layers/nlp/models/start_end_packer.ts (hunk 2 above)
Comment (pforderique) [reply tfjs-layers/src/layers/nlp/models/start_end_packer.ts:90]: These variables can be left as `None` in Keras, so I've changed the declaration to explicitly state that these can be `undefined`.

File: tfjs-layers/src/layers/nlp/models/start_end_packer.ts (hunk 1 above)
Comment (pforderique) [reply tfjs-layers/src/layers/nlp/models/start_end_packer.ts:1]: Not done on purpose - thanks for catching it!

**Review by Linchenn — APPROVED** [review]

**Comment by pforderique** [comment]
@mattsoulanille comments addressed!

File: tfjs-layers/src/layers/nlp/preprocessing/start_end_packer.ts [hunk 4]
```diff
@@ -76,9 +76,9 @@ export class StartEndPacker extends Layer {
   static readonly className = 'StartEndPacker';
 
   private sequenceLength: number;
-  private startValue: number|string;
-  private endValue: number|string;
-  private padValue: number|string;
+  private startValue: number|string|undefined;
+  private endValue: number|string|undefined;
+  private padValue: number|string|undefined;
```
mattsoulanille, line 81: Small style nit:
```suggestion
  private startValue?: number|string;
  private endValue?: number|string;
  private padValue?: number|string;
```

mattsoulanille (approved):
LGTM!

PR #7792, opened 2023-06-30
[webgpu] Update INT_DIV

hujiajie:

### DIFF
--- a/tfjs-backend-webgpu/src/binary_op_util.ts
+++ b/tfjs-backend-webgpu/src/binary_op_util.ts
@@ -23,9 +23,9 @@ export enum BinaryOpType {
   DIV,
   ELU_DER,
   EQUAL,
+  FLOOR_DIV,
   GREATER,
   GREATER_EQUAL,
-  INT_DIV,
   LESS,
   LESS_EQUAL,
   LOGICAL_AND,
@@ -56,6 +56,13 @@ const EQUAL = `
   let one = sign(b) * 0 + 1;
   let resultTemp = select(zero, one, a == b);
 `;
+const FLOOR_DIV = `
+  let remainder =
+      select(a % b, round(a % b), (round(a) == a) & (round(b) == b));
+  let quotient = (a - remainder) / b;
+  let resultTemp =
+      round(select(quotient, quotient - 1, sign(remainder) == -sign(b)));
+`;
 const GREATER = `
   let zero = sign(a) * 0 + 0;
   let one = sign(b) * 0 + 1;
@@ -66,36 +73,6 @@ const GREATER_EQUAL = `
   let one = sign(b) * 0 + 1;
   let resultTemp = select(zero, one, a >= b);
 `;
-
-const INT_DIV = `
-  let s = sign(a) * sign(b);
-  let ia = i32(round(a));
-  let ib = i32(round(b));
-  return f32(idiv(ia, ib, s));
-`;
-const INT_DIV_VEC4 = `
-  let ia = vec4<i32>(round(a));
-  let ib = vec4<i32>(round(b));
-  let cond = ib != vec4<i32>(0);
-  var resultTemp = vec4<i32>(0);
-  let s = sign(a) * sign(b);
-
-  // Windows (D3D) wants guaranteed non-zero int division at compile-time.
-  if (cond[0]) {
-    resultTemp[0] = idiv(ia[0], ib[0], s[0]);
-  }
-  if (cond[1]) {
-    resultTemp[1] = idiv(ia[1], ib[1], s[1]);
-  }
-  if (cond[2]) {
-    resultTemp[2] = idiv(ia[2], ib[2], s[2]);
-  }
-  if (cond[3]) {
-    resultTemp[3] = idiv(ia[3], ib[3], s[3]);
-  }
-  return vec4<f32>(resultTemp);
-`;
-
 const LESS = `
   let zero = sign(a) * 0 + 0;
   let one = sign(b) * 0 + 1;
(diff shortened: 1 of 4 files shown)


qjia7 (approved):
LGTM, thanks!

tfjs-backend-webgpu/src/binary_op_util.ts
```diff
@@ -66,36 +66,14 @@ const GREATER_EQUAL = `
...
-  let s = sign(a) * sign(b);
-  let ia = i32(round(a));
-  let ib = i32(round(b));
-  return f32(idiv(ia, ib, s));
+  let numerator = round(a);
+  let denominator = round(b);
+  let remainder = round(numerator % denominator);
+  let quotient = round((numerator - remainder) / denominator);
+  let resultTemp =
+      select(quotient, quotient - 1, sign(remainder) == -sign(denominator));
```
gyagp, line 75: Several questions and comments:
1) round() in WGSL has a special definition: "Result is the integer k nearest to e, as a floating point value. When e lies halfway between integers k and k + 1, the result is k when k is even, and k + 1 when k is odd." Will this introduce a diff between CPU and WebGPU solution?
2) For "let remainder = round(numerator % denominator);" Is round() necessary?
3) What's the result of "9 / (-2)" in this code?
4) numerator -> dividend, denominator -> divisor?
5) Please ensure we have enough unit test cases for them, especially when dividend and/or divisor are halfway between integers, the results from "(numerator - remainder) / denominator" are halfway between integers, etc.

hujiajie: > Several questions and comments:
> 
> 1. round() in WGSL has a special definition: "Result is the integer k nearest to e, as a floating point value. When e lies halfway between integers k and k + 1, the result is k when k is even, and k + 1 when k is odd." Will this introduce a diff between CPU and WebGPU solution?

This is not clearly mentioned in the TensorFlow.js API doc. Even if the CPU rounding mode differs from the GPU backend, I cannot tell which one is more correct, or if either is fine.

> 2. For "let remainder = round(numerator % denominator);" Is round() necessary?

No, but I feel that's more aligned with neighboring lines.

> 3. What's the result of "9 / (-2)" in this code?

-5

> 4. numerator -> dividend, denominator -> divisor?

The API doc prefers numerator/denominator.

> 5. Please ensure we have enough unit test cases for them, especially when dividend and/or divisor are halfway between integers, the results from "(numerator - remainder) / denominator" are halfway between integers, etc.

gyagp: Different languages may have different results.
[python]
(-9) % 2 -> 1
9 % (-2)  -> -1

[JavaScript]
(-9) % 2 -> -1
9 % (-2)  -> 1

So your result will be at least different from the JS's result. 
With round(), we introduce another level of difference between CPU and GPU, at least for the quotient calculation.

hujiajie: Please note `INT_DIV` is exclusively used by `floorDiv()`, and the doc states the result should be floored. This matches with the current CPU behavior (`Math.floor(a / b)`).

qjia7: JiaJie, if `INT_DIV` is only used for `floorDiv`, will it be better to rename `INT_DIV` to `FLOOR_DIV`?

jzm-intel:
I have concern that the whole implementation of `INT_DIV` and `floorDiv` may conflict their expected semantic. The implementation, before and after this PR, is not coherent with `floor(x/y)`, which is the definition of `floorDiv(x, y)`. It only works correctly for integer input, and actually only tested on `int32`. 
For example, input `(5.9, 2)` will result in `6.0 / 2.0 = 3.0` instead of `floor(5.9 / 2.0) = floor(2.95) = 2.0`, and input `(3.7, 1.9)` will result in `4.0 / 2.0 = 2.0` instead of `floor(3.7 / 1.9) = floor(1.947368...) = 1.0`.
I think the `round` of two inputs in original implementation are intend for handling integer inputs passing to WebGL as float, and we should revisit them.

qjia7 (commented):
Good work, Jiajie. Please also add some test cases for `float32` type in tfjs-core.

gyagp (approved):
I defer to Zhaoming's review on this.

qjia7:
@jzm-intel Do you have any concern with the latest change?

tfjs-backend-webgpu/src/kernels/FloorDiv.ts
```diff
@@ -21,9 +21,11 @@ import {BinaryOpType} from '../binary_op_util';
 import {binaryKernelFunc} from '../kernel_utils/kernel_funcs_utils';
 import {floorDivImplCPU} from '../kernel_utils/shared';
 
-export const floorDiv =
-    binaryKernelFunc({opType: BinaryOpType.INT_DIV,
-                      cpuKernelImpl: floorDivImplCPU, dtype: 'int32'});
+export const floorDiv = binaryKernelFunc({
+  opType: BinaryOpType.FLOOR_DIV,
+  cpuKernelImpl: floorDivImplCPU,
+  dtype: 'int32'
```
jzm-intel, line 27: Why does it state as 'int32'? This is the last thing that confuse me....

jzm-intel (approved):
LGTM, thanks

tfjs-backend-webgpu/src/kernels/FloorDiv.ts
hujiajie: This is the result dtype.

jzm-intel: Thanks, still LGTM.

PR #7794, opened 2023-06-30
[gpt2pre 3] Preprocessor Layer

pforderique:
Implements the base layer for processor layers, such as the GPT2 Preprocessor layer.

Update:
~~Some methods were left unimplemented and are pending discussion~~
Based on discussion with @mattsoulanille, `presets` was removed, and `fromConfig` and `getConfig` was implemented instead.

### DIFF
--- a/tfjs-layers/src/layers/nlp/models/preprocessor.ts
+++ b/tfjs-layers/src/layers/nlp/models/preprocessor.ts
@@ -1,0 +1,130 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+/* Original source: keras-nlp/models/preprocessor.py */
+import { serialization } from '@tensorflow/tfjs-core';
+
+import { Layer, LayerArgs } from '../../../engine/topology';
+import { Kwargs } from '../../../types';
+import { Tokenizer } from '../tokenizers';
+import { NotImplementedError } from 'tfjs-layers/src/errors';
+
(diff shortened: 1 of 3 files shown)


tfjs-layers/src/layers/nlp/models/preprocessor.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { Layer, LayerArgs } from '../../../engine/topology';
+import { Tokenizer } from '../tokenizers';
+import { Kwargs } from '../../../types';
+import { deserializeKerasObject, serializeKerasObject } from '../../../utils/generic_utils';
+
+/**
+ * Base class for model Preprocessors.
+ */
+export class Preprocessor extends Layer {
```
mattsoulanille, line 29: You likely need to register this layer with the global serialization map (e.g. https://github.com/tensorflow/tfjs/blob/master/tfjs-layers/src/layers/padding.ts#L237)

mattsoulanille (commented):
This looks good, but can we have a simple test for this where it loads either a fake tokenizer (which you define and register in the test code) or the BytePairTokenizer? The reason for this is because it has a non-trivial `fromConfig` function that initializes an additional layer.

tfjs-layers/src/layers/nlp/models/preprocessor.ts
pforderique: I added a test case for deserializing the set `tokenizer` in a Preprocessor Layer, but I am running into: 
```Error: Unknown object: BytePairTokenizer. This may be due to one of the following reasons:
        1. The object is defined in Python, in which case it needs to be ported to TensorFlow.js or your JavaScript code.
        2. The custom object is defined in JavaScript, but is not registered properly with tf.serialization.registerClass().
            at deserializeKerasObject```
Even though BytePairTokenizer is registered [here](https://github.com/pforderique/tfjs/blob/main/tfjs-layers/src/layers/nlp/tokenizers.ts#L572). Not sure what I am missing here.

pforderique: I've gone ahead and removed the test case for round trip serialization with a set tokenizer because the `Preprocessor` class does not take a tokenizer as a parameter, and therefore can't reconstruct a tokenizer from its config. I will be adding this test to the `GPT2Preprocessor` class, which DOES take in a tokenizer.

I also went ahead and made a small change to how BytePairTokenizer is serialized, since a JS `Map` needs more information than just keys to be initialized.

mattsoulanille (approved):
LGTM

Linchenn (approved):
LGTM!

PR #7806, opened 2023-07-06
[gpt2pre 2] GPT2 Tokenizer Class

pforderique:
Implements the `GPT2Tokenizer` that extends `BytePairEncoding`.

This is a standalone PR with no dependencies or implementation followups.

 `GPT2Tokenizer` will be used by ` GPT2Preprocessor` in a future PR.

NOTE:
This class does not implement `presets()` as discussed earlier this week.

PR #7814, opened 2023-07-10
[gpt2pre 4] GPT2Preprocessor Layer

pforderique:
This PR implements the `GPT2Preprocessor` class with its `call` method.
Dependencies: #7791 #7806 

**[UPDATE 7-11-23]**
After experimenting with adding generic input/output types to  `Layer`, I noticed the refactoring might be a bit more involved than expected due to a lot of util methods expecting `Tensor|Tensor[]` or `SymbolicTensor|SymbolicTensor[]` as their inputs. I have documented this at [go/TFJS-layer-generics](http://goto.google.com/TFJS-layer-generics) and will be using the original design outlines below. 

**[UPDATE 7-10-23]**
After discussion with @mattsoulanille today (7-10-23), I will be exploring changes to the base `Layer` class to allow different input and output types to the `call` method. This will allow us to support input strings for `Tokenizers` and any other special i/o needed by the other NLP layers.

**[ORIGINAL]**
Here is the original plan I had, as implemented in this PR:
I had to deviate from the reference implementation a bit more here because the types can get tricky. Since `GPT2Preprocessor` extends the `Layer` class (it really extends `Preprocessor`, which extends `Layer`), its overridden `call()` method must return `Tensor|Tensor[]`. 

However, the Keras implementation allows a bit more flexibility by returning the result of `pack_x_y_sample_weight()`, which returns a packed tuple of x, y, and sample_weight.

Designs Considered:
* I thought about making the TS version `packXYSampleWeight()` simply return a `Tensor` of x, y, and sample_weight so that it can be called and directly returned from `call()` like Keras does, but since x is often used as an object of type `{tokenIds: Tensor|Tensor[], paddingMask: Tensor[]}`, this cannot be wrapped in a tensor. 
* Flattening the results into one large `Tensor[]`  is also an option, but since tokenIds will most like be a `Tensor[]` itself (since it's the result of calling` tokenize()` on a tensor of strings), that would mean `call` would have to support a return type of `Tensor|Tensor[]|(Tensor[]|Tensor)[]` which starts to get confusing. Not to mention that distinguishing which output is at which position of the tensor would not be super clear.
* Create a new method `callAndReturnPaddingMask` that just returns a tuple of the `tokenIds` and `paddingMask` and have `call` return the first element (tokenIds) and ignore trying to return y and sample_weight since they aren't mutated by the class. This option is more feasible, however looking ahead, it seems that the `GPT2CasualLMPreprocessor` class does change y and sample_weight eventually, so it's probably a good idea to figure out how to support this now.
* **[DESIGN CHOSEN]** I've decided to opt for making a new method `callAndPackArgs` that returns an array of the arguments packed similarly to the Keras version and `call` will simply return a subset of these outputs to keep the class layer compatible.

Since GPT2 layers always pass in x as a type of `{tokenIds: Tensor|Tensor[], paddingMask: Tensor[]}`, I've declared this as a `PreprocessorOutput` interface and removed Array checking in `packXYSampleWeight`.

### DIFF
--- a/tfjs-layers/src/layers/nlp/models/gpt2/gpt2_preprocessor.ts
+++ b/tfjs-layers/src/layers/nlp/models/gpt2/gpt2_preprocessor.ts
@@ -1,0 +1,218 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+/**
+ * GPT-2 preprocessor layer.
+ */
+
+/* Original source: keras-nlp/models/gpt2/gpt2_preprocessor.py */
+import { Tensor, Tensor2D, serialization } from '@tensorflow/tfjs-core';
+
+import { LayerArgs } from '../../../../engine/topology';
+import { Preprocessor } from '../preprocessor';
+import { GPT2Tokenizer } from './gpt2_tokenizer';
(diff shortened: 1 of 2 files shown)


mattsoulanille (changes requested):
Looks good. I just have some more tests to add.

mattsoulanille (approved):
LGTM

mattsoulanille:
@Linchenn Please take a look when you get a chance. Thanks!

Linchenn (approved):
LGTM!

PR #7815, opened 2023-07-11
Add the e2e tests from keras to tfjs

fengwuyao:
Add the E2E test flow from Keras to TFJS.
- Build the Keras model, get the shape and value of the model and convert the model to TFJS layers model.
- Load the TFJS layers model and extract shape and values.
- Compare the shape and values.

### DIFF
--- a/e2e/custom_module/blazeface/yarn.lock
+++ b/e2e/custom_module/blazeface/yarn.lock
@@ -117,46 +117,50 @@
   resolved "https://registry.yarnpkg.com/@tensorflow-models/blazeface/-/blazeface-0.0.5.tgz#6261fcc5de61a283ea409b69f6a851f44a809826"
   integrity "sha1-YmH8xd5hooPqQJtp9qhR9EqAmCY= sha512-bIMJDV2CD8rr7v9OrIDdDZVh9sbk0EhlZdHaDFpgWUw3Lp21wUN7GSifOJVE5hV2ngmpzM+J12OH8bG1GZoDag=="
 
+"@tensorflow/tfjs-backend-cpu@link:../../../link-package/node_modules/@tensorflow/link-package/node_modules/@tensorflow/tfjs-backend-cpu":
+  version "0.0.0"
+
 "@tensorflow/tfjs-backend-cpu@link:../../../link-package/node_modules/@tensorflow/tfjs-backend-cpu":
   version "0.0.0"
+  uid ""
 
 "@tensorflow/tfjs-backend-wasm@link:../../../link-package/node_modules/@tensorflow/tfjs-backend-wasm":
   version "0.0.0"
+  uid ""
 
 "@tensorflow/tfjs-backend-webgl@link:../../../link-package/node_modules/@tensorflow/tfjs-backend-webgl":
   version "0.0.0"
+  uid ""
 
 "@tensorflow/tfjs-converter@link:../../../link-package/node_modules/@tensorflow/tfjs-converter":
   version "0.0.0"
+  uid ""
 
 "@tensorflow/tfjs-core@link:../../../link-package/node_modules/@tensorflow/tfjs-core":
   version "0.0.0"
+  uid ""
 
 "@tensorflow/tfjs-data@link:../../../link-package/node_modules/@tensorflow/tfjs-data":
   version "0.0.0"
+  uid ""
 
 "@tensorflow/tfjs-layers@link:../../../link-package/node_modules/@tensorflow/tfjs-layers":
   version "0.0.0"
+  uid ""
 
 "@tensorflow/tfjs@link:../../../tfjs":
   version "0.0.0"
-  dependencies:
-    "@tensorflow/tfjs-backend-cpu" "link:../../../link-package/node_modules/@tensorflow/tfjs-backend-cpu"
-    "@tensorflow/tfjs-backend-webgl" "link:../../../link-package/node_modules/@tensorflow/tfjs-backend-webgl"
-    "@tensorflow/tfjs-converter" "link:../../../link-package/node_modules/@tensorflow/tfjs-converter"
-    "@tensorflow/tfjs-core" "link:../../../link-package/node_modules/@tensorflow/tfjs-core"
-    "@tensorflow/tfjs-data" "link:../../../link-package/node_modules/@tensorflow/tfjs-data"
-    "@tensorflow/tfjs-layers" "link:../../../link-package/node_modules/@tensorflow/tfjs-layers"
-    argparse "^1.0.10"
-    chalk "^4.1.0"
-    core-js "3"
-    regenerator-runtime "^0.13.5"
-    yargs "^16.0.3"
+  uid ""
 
 "@types/color-name@^1.1.1":
   version "1.1.1"
   resolved "https://registry.yarnpkg.com/@types/color-name/-/color-name-1.1.1.tgz#1c1261bbeaa10a8055bbc5d8ab84b7b2afc846a0"
   integrity "sha1-HBJhu+qhCoBVu8XYq4S3sq/IRqA= sha512-rr+OQyAjxze7GgWrSaJwydHStIhHq2lvY3BOC2Mj7KnzI7XK0Uw1TOOdI9lDoajEbSWLiYgoo4f1R51erQfhPQ=="
+
+"@types/emscripten@~0.0.34":
+  version "0.0.34"
+  resolved "https://registry.yarnpkg.com/@types/emscripten/-/emscripten-0.0.34.tgz#12b4a344274fb102ff2f6c877b37587bc3e46008"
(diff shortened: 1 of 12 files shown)


e2e/custom_module/blazeface/yarn.lock
```diff
@@ -117,47 +117,51 @@
   resolved "https://registry.yarnpkg.com/@tensorflow-models/blazeface/-/blazeface-0.0.5.tgz#6261fcc5de61a283ea409b69f6a851f44a809826"
   integrity "sha1-YmH8xd5hooPqQJtp9qhR9EqAmCY= sha512-bIMJDV2CD8rr7v9OrIDdDZVh9sbk0EhlZdHaDFpgWUw3Lp21wUN7GSifOJVE5hV2ngmpzM+J12OH8bG1GZoDag=="
 
+"@tensorflow/tfjs-backend-cpu@link:../../../link-package/node_modules/@tensorflow/link-package/node_modules/@tensorflow/tfjs-backend-cpu":
```
Linchenn, line 120: Is it intended to update yarn.lock files?

Linchenn (changes requested):
Also, as we discussed, we need to compare inference results of Keras and TFJS.

e2e/custom_module/blazeface/yarn.lock
fengwuyao: yarn.lock files are automatically updated after running `yarn test-ci`

e2e/integration_tests/keras_to_tfjs_create_save_predict.py
```diff
@@ -0,0 +1,114 @@
...
+  result_data_path = os.path.join(
+      _tmp_dir, model_name + '.result-from-keras-data.json')
+  result_shape_path = os.path.join(
+      _tmp_dir, model_name + '.result-from-keras-shape.json')
+  with open(result_data_path, 'w') as f:
+    f.write(json.dumps(keras_predict_data))
+  with open(result_shape_path, 'w') as f:
+    f.write(json.dumps(keras_predict_shape))
+
+
```
Linchenn, line 101: Too many break lines.

e2e/integration_tests/keras_to_tfjs_create_save_predict.py
```diff
@@ -0,0 +1,114 @@
...
+  model = keras.Sequential()
+  d = keras.layers.Dense(100, activation='relu')
+  model.add(d)
+  model.add(keras.layers.Dense(50, activation='elu'))
+  model.add(keras.layers.Dense(24))
+  model.add(keras.layers.Activation(activation='elu'))
+  model.add(keras.layers.Dense(8, activation='softmax'))
+  model.build(input_shape=[1, 200])
+
+
```
Linchenn, line 39: many break lines.

Linchenn (approved):
LGTM excepts small lints.

e2e/integration_tests/keras_to_tfjs_create_save_predict.js
```diff
@@ -0,0 +1,114 @@
...
+const tf = require('@tensorflow/tfjs');
+const tfc = require('@tensorflow/tfjs-core');
+const tfconverter = require('@tensorflow/tfjs-converter');
+const tfl = require('@tensorflow/tfjs-layers');
+const tfjsNode = require('@tensorflow/tfjs-node');
+const ndarray = require('ndarray');
+const fs = require('fs');
+const join = require('path').join;
+const TEST_DATA_DIR = "keras_to_tfjs_create_save_predict_data";
+const _ = require("zlib")
```
mattsoulanille, line 36: This is unused, however, `_` is the name of an [existing javascript library](https://underscorejs.org/). This may be slightly confusing. Can it be removed or replaced with just `require("zlib");`? Alternatively, `const _zlib = require("zlib");`?

e2e/integration_tests/keras_to_tfjs_create_save_predict.js
```diff
@@ -0,0 +1,114 @@
...
+
+  const p = tfjsNode.io.fileSystem(file_path)
+
+  const model = await tf.loadLayersModel("file://" + tfjsNode.io.fileSystem(file_path).path + "/model.json");
+
+  const xsDataPath = join(TEST_DATA_DIR, `${model_name}.xs-data.json`);
+  const xsShapesPath = join(TEST_DATA_DIR, `${model_name}.xs-shapes.json`);
+  const dataArray = fs.readFileSync(xsDataPath, "utf-8", (_, data) => {
+    const d = JSON.parse(data);
+    return d
```
mattsoulanille, line 48: Nit: Missing semicolon and this can probably just be as follows:
```suggestion
    return JSON.parse(data);
```

e2e/integration_tests/keras_to_tfjs_create_save_predict.js
```diff
@@ -0,0 +1,114 @@
...
+  });
+  const shapeArray = fs.readFileSync(xsShapesPath, "utf-8", (_, data) => {
+    const shape = JSON.parse(data);
+    return shape
```
mattsoulanille, line 52: nit
```suggestion
    return JSON.parse(data);
```

e2e/integration_tests/keras_to_tfjs_create_save_predict.js
```diff
@@ -0,0 +1,114 @@
...
+  });
+
+  const parsedShape = JSON.parse(shapeArray);
+  const parsedData = JSON.parse(dataArray);
```
mattsoulanille, line 56: It looks like these were already `JSON.parse`d above. Why do we need to parse them again?

e2e/integration_tests/keras_to_tfjs_create_save_predict.js
```diff
@@ -0,0 +1,114 @@
...
+  const A = zip(parsedData, parsedShape);
```
mattsoulanille, line 57: Can we use a different variable name than `A`? It's not very descriptive.

e2e/integration_tests/keras_to_tfjs_create_save_predict.js
```diff
@@ -0,0 +1,114 @@
+/**
```
mattsoulanille, line 1: Should we write this file in TypeScript instead of JavaScript?

e2e/integration_tests/keras_to_tfjs_create_save_predict.js
```diff
@@ -0,0 +1,114 @@
...
+  const parsedShape = JSON.parse(shapeArray);
+  const parsedData = JSON.parse(dataArray);
+  const A = zip(parsedData, parsedShape);
+  const res = [];
+  for (let i = 0; i < A.length; i++) {
+    const value = A[i][0];
+    const shape = A[i][1];
+    res.push(ndarray(value, shape));
+  }
+  let z;
```
mattsoulanille, line 64: `z` is undefined for the next line if res.length is not 1. Should we throw an error in this case?

e2e/integration_tests/keras_to_tfjs_create_save_predict.js
```diff
@@ -0,0 +1,114 @@
...
+
+  for (let i = 0; i < shape.length; i++) {
+    const size = shape[i];
+    const chunk = array.slice(currentIndex, currentIndex + size);
+    reshapeArray.push(chunk);
+    currentIndex += size;
+  }
+
+  return reshapeArray;
+}
```
mattsoulanille, line 97: This function does not seem to be used anywhere.

e2e/integration_tests/keras_to_tfjs_create_save_predict.ts
```diff
@@ -0,0 +1,88 @@
+/**
```
mattsoulanille, line 1: It would be better to name this file `keras_to_tfjs_create_save_predict_test.ts` since it contains tests (we should also rename the existing `create_save_predict.ts` file to end with `tests.ts`).

e2e/package.json
```diff
@@ -64,5 +64,8 @@
   "license": "Apache-2.0",
   "engines": {
     "yarn": ">= 1.0.0"
+  },
+  "dependencies": {
+    "ndarray": "^1.0.19"
```
mattsoulanille, line 69: We already have `tf.tensor` for high-dimensional arrays. Can we use that instead?

e2e/integration_tests/keras_to_tfjs_create_save_predict.js
```diff
@@ -0,0 +1,114 @@
...
+  });
+
+  const parsedShape = JSON.parse(shapeArray);
+  const parsedData = JSON.parse(dataArray);
+  const A = zip(parsedData, parsedShape);
+  const res = [];
+  for (let i = 0; i < A.length; i++) {
+    const value = A[i][0];
+    const shape = A[i][1];
+    res.push(ndarray(value, shape));
```
mattsoulanille, line 62: Can we create a tfjsNode.tensor here instead of using ndarray?

mattsoulanille (commented):
Looks good. I just have some minor comments and changes.

PR #7838, opened 2023-07-12
[gpt2bb impl] Backbone class (1/1)

pforderique:
Implements the [Backbone](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/backbone.py#L25) class. 

I am not sure how this class is supposed to be deserialized from just the info that `getConfig` returns since `ContainerArgs` (needed by the super class `LayersModel`) requires `inputs` and `outputs` fields which are SymbolicTensors, but it seems that [GPT2Backbone](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/gpt2/gpt2_backbone.py#L159) initializes the parent class with different types.


Linchenn (approved):

mattsoulanille (approved):
Approved. We can figure out how getConfig will work when we implement it.

PR #7839, opened 2023-07-12
[gpt2bb spec] PositionEmedding layer

pforderique:
Add PositionEmbedding specs.

### DIFF
--- a/tfjs-layers/src/layers/nlp/modeling/position_embedding.ts
+++ b/tfjs-layers/src/layers/nlp/modeling/position_embedding.ts
@@ -25,6 +25,7 @@ import { Tensor, Tensor1D, Tensor2D, serialization } from '@tensorflow/tfjs-core
 import { Shape } from '../../../keras_format/common';
 import { Layer, LayerArgs } from '../../../engine/topology';
 import { NotImplementedError } from '../../../errors';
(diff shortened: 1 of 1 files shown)


tfjs-layers/src/layers/nlp/modeling/position_embedding.ts
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Combine with a token embedding.
+ * ```js
+ * const seqLength = 50;
+ * const vocabSize = 5000;
+ * const embedDim = 128;
+ * const inputs = tf.input({shape: [seqLength]});
+ * const tokenEmbeddings = tf.layers.embedding({
+ *    inputDim=vocabSize, outputDim=embedDim
+ * }).apply(inputs)
```
Linchenn, line 74: missed ';'

Linchenn (approved):
LGTM except a nit.

tfjs-layers/src/layers/nlp/modeling/position_embedding.ts
pforderique: Good catch! Thanks for the review

mattsoulanille (approved):

PR #7840, opened 2023-07-13
RGB to grayscale for tf.image

yiyihg:


jinjingforever (approved):
This is great! Thanks!

tfjs-core/src/ops/image/rgb_to_grayscale.ts
```diff
@@ -0,0 +1,84 @@
+/**
```
Linchenn, line 1: Let's update the license to our latest one: https://github.com/tensorflow/tfjs/blob/master/e2e/integration_tests/types.ts#L1-L16

PR #7841, opened 2023-07-13
Increase epsilon values for draw tests

Linchenn:
In the browserstack tests for `Firefox 90.0 (Mac OS 10.13)`, the difference between source values and 'draw+read back' could be 6.0, while the codes are correct. This PR increases the epsilon values from 3.0 to 6.0.


fengwuyao (approved):

qjia7 (approved):
LGTM, thanks!

PR #7848, opened 2023-07-17
Support custom model conversion

fengwuyao:
Update the way we store the weight names into bin files. Using folder structure in `weights.h5` to name.
e.g. `_layer_checkpoint_dependencies/dense/vars/0`

Having a followup [PR](https://github.com/tensorflow/tfjs/pull/7872) to update the loading weights.


tfjs-converter/python/tensorflowjs/converters/keras_h5_conversion.py
```diff
@@ -288,18 +288,34 @@ def h5_v3_merged_saved_model_to_tfjs_format(h5file, meta_file, config_file,split
...
+  return model_json, groups
 
-  for index, layer_name in enumerate(layer_names):
-    group_of_weights = model_weights[layer_name]
-    group = _convert_v3_group(group_of_weights, layer_name)
-    if group:
+def _convert_v3_group_structure_to_weights(groups, group, split_by_layer, indent=""):
+  for key in group.keys():
+    if isinstance(group[key], h5py.Group):
+      _convert_v3_group_structure_to_weights(groups, group[key], split_by_layer, indent + key + "/")
```
mattsoulanille, line 297: I was worried this was discarding the directory structure of the h5 file, but I see now that it's encoded in the indent by /.

mattsoulanille: On second thought, json supports nested structures. Is there a reason we're flattening the weight locations and using their paths to denote hierarchy instead of maintaining the hierarchy in the json object?

fengwuyao: Thanks for review. I think we are storing the weights into bin files which will have `name` and `data`. Keras v3 format is using folder structure to represent which layer the weights belongs to. Therefore, I think we can use the same way to represent the `name` in bin files. And assign weights back to layers based on specific name when we load the model. And I think in v2 we are flatten the weights as well.

mattsoulanille: That makes sense. Thanks for the explanation!

mattsoulanille (approved):

Linchenn (approved):
LGTM

Issue #7865, opened 2023-07-20
A model with empty weights doesn't load


thekevinscott:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js): Yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Mac
- TensorFlow.js installed from (npm or script link): `npm`
- TensorFlow.js version (use command below): `4.8.0`

**Describe the current behavior**
A `model.json` with an empty array of weights fails to be loaded:

```
$ tf.loadLayersModel('/path/to/my/model');
        var key = Object.keys(weights)[0].split('/');
                                         ^
TypeError: Cannot read properties of undefined (reading 'split')
```

**Why would I want a model with empty weights?**

I maintain a tool, UpscalerJS, for doing image upscaling in Javascript. It has a number of integration tests that require a functioning Tensorflow.js model, and it's helpful to keep that model as simple possible (a single upsampling layer) to assert behavior without significant test latency.

At some point tests were working, but a version upgrade somewhere along the line broke things. I haven't yet been able to isolate the version things changed.

**Describe the expected behavior**

I would expect a model with an empty weights array to load.

**Standalone code to reproduce the issue**

https://codesandbox.io/p/sandbox/jolly-firefly-2g63jk

Or, run the following code in Node:

```
const tf = require("@tensorflow/tfjs-node");
const fs = require("fs");

(async () => {
  const modelPath = `/${__dirname}/model`;
  const modelPathWithFile = `file://${modelPath}`;
  if (!fs.existsSync(modelPath)) {
    const model = tf.sequential();
    const scale = 2;
    model.add(
      tf.layers.upSampling2d({
        size: [scale, scale],
        dataFormat: "channelsLast",
        inputShape: [null, null, 3],
      })
    );

    model.compile({ loss: "meanSquaredError", optimizer: "sgd" });

    await model.save(modelPathWithFile);
  }

  await tf.loadLayersModel(`${modelPathWithFile}/model.json`);
})();
```

gaikwadrahul8:
Hi, @thekevinscott 

Thank you for bringing this issue to our attention and I tried to replicate the same issue from my end and I'm also getting the same error message which you've mentioned above so we'll have to dig more into this issue and will update you soon

Here is output from above code snippet for reference :

```
gaikwadrahul-macbookpro:test-7865 gaikwadrahul$ node index.js
[...]

PR #7868, opened 2023-07-22
Fix bug where a model with empty weights fails to load

thekevinscott:
First draft at addressing #7865.

This slightly refactors the `Container.loadWeights` function by pulling the "is this model a keras model" into a standalone function. That function will first assert that the model has non-empty weights before checking the first weight's name (which is subsequently used to determine keras-ness or not).

This also adds a unit test in `model_save_test.ts` enforcing that a model with empty weights can still be loaded.


fengwuyao (approved):
Looks good to me. Thanks!

pyu10055 (approved):

thekevinscott:
Just merged in latest master. Is there something I need to do prior to merge, or is this in y'all's hands now?

Neilblaze (approved):

PR #7860, opened 2023-07-19
Implement EinsumDense layer

pforderique:
Implements the `EinsumDense` layer from [Keras](https://github.com/keras-team/keras/blob/v2.13.1/keras/layers/core/einsum_dense.py#L32-L212) core. 

Depends on #7842 .

PR #7875, opened 2023-07-26
Implement MultiHeadAttention Layer

pforderique:
Implements the MultiHeadAttention layer from [Keras](https://github.com/keras-team/keras/blob/v2.13.1/keras/layers/attention/multi_head_attention.py) attention layers.

NOTE:
* This implementation does not support RaggedTensors yet.
* The `Softmax` layer was changed to support masking. Let me know if having a separate class for this is better. Regardless, the current change should be non-breaking.

Depends on #7860.


mattsoulanille (approved):
LGTM

mattsoulanille:
@Linchenn Please take a look when you get a chance. Thanks!

Linchenn (approved):
LGTM!

PR #7883, opened 2023-07-31
Fix SliceUpdate tests

pforderique:
After discussion with @Linchenn, I found out that using a zero-value epsilon in `expectTensorsClose` causes some tests for devices with low precision to fail. This PR changes the tests to use the default epsilon value.


Linchenn (approved):
Thanks, LGTM!

mattsoulanille (approved):

mattsoulanille:
Force-merging this since its tests passed and the failure is due to an unrelated tfjs-node issue.

PR #7884, opened 2023-07-31
Build node addon before running tests

mattsoulanille:
Fix a race condition in CI where the tests may be run before the node addon is built.


Linchenn (approved):
Thanks for fixing it!

fengwuyao (approved):
Thanks Matt!

PR #7882, opened 2023-07-31
Implement CachedMultiHeadAttention layer

pforderique:
Implements the CachedMultiHeadAttention layer from [Keras-NLP](https://github.com/keras-team/keras-nlp/blob/v0.6.0/keras_nlp/layers/modeling/cached_multi_head_attention.py#L21) attention layers.

Depends on #7875 (MultiHeadAttention implementation)

PR #7890, opened 2023-08-04
Implement TransformerDecoder layer

pforderique:
Implements the `TransformerDecoder` layer from [KerasNLP](https://github.com/keras-team/keras-nlp/blob/v0.6.1/keras_nlp/layers/modeling/transformer_decoder.py).

Dependencies: #7875 #7882 (MHA and CMHA layers)

Note: This implementation does not support cross-attention yet, but can be added in a future "polish" PR.

### DIFF
--- a/tfjs-layers/src/layers/nlp/modeling/transformer_decoder.ts
+++ b/tfjs-layers/src/layers/nlp/modeling/transformer_decoder.ts
@@ -20,13 +20,19 @@
  */
 
 /* Original source: keras_nlp/layers/modeling/transformer_decoder.py */
-import { Tensor, Tensor1D, Tensor2D, serialization } from '@tensorflow/tfjs-core';
+import { Tensor, add, serialization, tidy } from '@tensorflow/tfjs-core';
 
-import { Layer, LayerArgs, } from '../../../engine/topology';
-import { NotImplementedError } from '../../../errors';
-import { InitializerIdentifier } from '../../../initializers';
+import { Activation, getActivation, serializeActivation } from '../../../activations';
+import { Layer, LayerArgs, SymbolicTensor, } from '../../../engine/topology';
+import { ValueError } from '../../../errors';
+import { Initializer, InitializerIdentifier, getInitializer, serializeInitializer } from '../../../initializers';
 import { ActivationIdentifier } from '../../../keras_format/activation_config';
 import { Shape } from '../../../keras_format/common';
+import { Dense, Dropout } from '../../core';
+import { LayerNormalization } from '../../normalization';
+
+import { CachedMultiHeadAttention } from './cached_multihead_attention';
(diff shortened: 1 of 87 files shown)


tfjs-layers/src/layers/nlp/modeling/transformer_decoder.ts
```diff
@@ -212,22 +250,84 @@ export class TransformerDecoder extends Layer {
...
+      this.decoderSequenceShape, this.decoderSequenceShape);
+
+    this.selfAttentionLayernorm =
+      new LayerNormalization({epsilon: this.layerNormEpsilon});
+
+    this.selfAttentionLayernorm.build(this.decoderSequenceShape);
+    this.selfAttentionDropout = new Dropout({rate: this.dropout});
+
+    // Cross attention layers are optional.
+    // TODO(pforderique): Add cross attention layers.
```
mattsoulanille, line 285: As you mentioned here, we can skip cross attention since gpt2 doesn't require it.

tfjs-layers/src/layers/nlp/modeling/transformer_decoder_test.ts
```diff
@@ -0,0 +1,177 @@
...
+    const tModel = model({inputs: decoderInput, outputs});
+
+    const decoderSequence = randomUniform([2, 4, 6]);
+    const label = randomUniformInt([2, 4, 1], 0, 10).asType('float32');
+
+    tModel.compile({loss: 'sparseCategoricalCrossentropy', optimizer: 'adam'});
+    const loss = tModel.trainOnBatch(decoderSequence, label);
+
+    expect(await loss).toBeGreaterThan(0);
+  });
```
mattsoulanille, line 107: It's great that training works too (at least for this layer if not the whole model yet)!

tfjs-layers/src/layers/nlp/modeling/transformer_decoder_test.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    const params: Array<[string, boolean]> = [
+      ['without_norm_first', false],
+      ['with_norm_first', true],
+    ];
+
+    for (const [testcaseName, normalizeFirst] of params) {
+      testValidCallWithoutCrossAttention(testcaseName, normalizeFirst);
+    }
+  });
```
mattsoulanille, line 67: Nice use of `describe` within  another `describe`. I might break this for loop out into two discrete test cases just because it's such a small number of tests, but feel free to leave it as is.

mattsoulanille (approved):
LGTM!

Linchenn (approved):

tfjs-layers/src/layers/nlp/modeling/transformer_decoder_test.ts
pforderique: Thanks for the review! comments have been addressed

PR #7894, opened 2023-08-07
Implement GPT2 Backbone

pforderique:
Implements the `GPT2Backbone` layer based off [KerasNLP](https://github.com/keras-team/keras-nlp/blob/v0.6.1/keras_nlp/models/gpt2/gpt2_backbone.py). 

Dependencies: #7890, #7887

Note: `presets` is unfinished and is pending conversation about how to properly load in weights, vocab, merges, architecture info, etc.


Linchenn (approved):

tfjs-layers/src/layers/nlp/models/gpt2/gpt2_backbone.ts
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * Example usage:
+ * ```js
+ * const tokenIds = tf.ones([1, 12]), dtype="int32");
+ * const paddingMask = tf.tensor(
+ *  [[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0]], 'int32');
+ *
+ * # Pretrained GPT-2 decoder.
+ * model = GPT2Backbone.fromPreset("gpt2_base_en");
+ * model.apply(inputData, {paddingMask});
```
mattsoulanille, line 111: We might write this part of the API that supports automatic preset loading in tfjs-models. Let's see what makes the most sense as we work on model loading, though.

tfjs-layers/src/layers/nlp/models/gpt2/gpt2_backbone.ts
```diff
@@ -0,0 +1,252 @@
...
+          vocabularySize: 50257,
+          numLayers: 12,
+          numHeads: 12,
+          hiddenDim: 768,
+          intermediateDim: 3072,
+          dropout: 0.1,
+          maxSequenceLength: 1024,
+        },
+        preprocessorConfig: {},
+        weightsUrl: 'https://storage.googleapis.com/keras-nlp/models/gpt2_base_en/v1/model.h5',
```
mattsoulanille, line 240: We'll likely need to change all the weights urls to point to our converted model weights, but we can do that later.

mattsoulanille (approved):
LGTM with some longer term questions for us to think about in model loading.

PR #7899, opened 2023-08-10
[webgpu] Fix cast error

qjia7:
When cast a bool tensor to a float32 tensor, it's just an identity operation, see [here](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/src/kernels/Cast.ts#L67-L68), which adds the underlying tensor data refCount and reuse the same gpu buffer. However, if we use the casted float32 tensor as the conv2d's input, error happens due to that we expect a f32 data but the original buffer data is i32. This PR fixes it by converting the input data to f32 data since our calculation is based on f32.

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/conv2d_webgpu_test.ts
+++ b/tfjs-backend-webgpu/src/kernels/conv2d_webgpu_test.ts
@@ -286,6 +286,19 @@ describeWebGPU('conv2d vec4', () => {
         ]));
   });
 
+  it('bool cast float32 + conv2d', async () => {
+    const a1 = tf.tensor4d([3, 2, 3, 4, 3, 2], [1, 2, 3, 1]);
+    const b1 = tf.tensor4d([2, 2, 2, 1, 4, 1], [1, 2, 3, 1]);
+    // Generate a bool tensor and have data in GPU.
+    const a = tf.greater(a1, b1);
+    // Cast the bool tensor to a float32 tensor.
+    const x = tf.cast(a, 'float32') as tf.Tensor4D;
+    const w = tf.tensor4d([1, 2, 3, 4, 5, 6, 7, 8], [2, 2, 1, 2]);
+    const res = tf.conv2d(x, w, 1, 0);
+    const resData = await res.data();
+    test_util.expectArraysClose(resData, new Float32Array([6, 8, 10, 12]));
+  });
+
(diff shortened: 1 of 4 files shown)


tfjs-backend-webgpu/src/kernels/conv2d_webgpu_test.ts
```diff
@@ -286,6 +286,19 @@ describeWebGPU('conv2d vec4', () => {
         ]));
   });
 
+  it('bool cast float32 + conv2d', async () => {
```
qjia7, line 289: FYI. Due to the problem shader is `conv2d_mm_webgpu.ts`, and to use this test to reproduce the error, we still need to comment out the [im2col path](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/src/kernels/Conv2D_impl.ts#L313-L332). Otherwise, it can't go to `Conv2DMMProgram` since the test size is too small.

gyagp (approved):

qjia7:
@Linchenn @mattsoulanille, Kai reported a bug to us and this is the fixing. But it seems that bots are failing due to some reason and not related with my PR. Can you help resolve the bots issue to unblock me? Thanks.

tfjs-backend-webgpu/src/kernels/conv2d_webgpu_test.ts
pyu10055: you can disable im2col with the flag WEBGPU_CONV_SEPARATE_IM2COL_SHADER, right?

pyu10055 (approved):

tfjs-backend-webgpu/src/kernels/conv2d_webgpu_test.ts
qjia7: No, Currently if `workgroupsBy32x32 <= thresholdToIncreaseWorkgroups` is satisfied, it will also go to the im2col path. So that flag can't let us skip the im2col path.

qjia7: im2col path currently is controlled by `WEBGPU_CONV_SEPARATE_IM2COL_SHADER` and `WEBGPU_THRESHOLD_TO_INCREASE_WORKGROUPS_FOR_MATMUL`. So disable both of them. Thanks for your suggestion.

PR #7901, opened 2023-08-11
fix converter tests

fengwuyao:
Update the dependencies and fix the broken converter tests.


mattsoulanille (approved):
Lgtm. Thanks!

tfjs-converter/python/tensorflowjs/converters/keras_h5_conversion_test.py
```diff
@@ -135,8 +135,6 @@ def testConvertModelWithNestedLayerNames(self):
 
     # Check meta-data in the artifact JSON.
     self.assertEqual(model_json['format'], 'layers-model')
-    self.assertEqual(model_json['generatedBy'],
-                     'keras v%s' % tf.keras.__version__)
```
pyu10055, line 139: can you double check what is the version number we are saving into the `generatedBy` field? You should still match up those in the test. thanks

pyu10055 (approved):

PR #7926, opened 2023-08-21
Optimize Einsum to BatchMatMul implementation

Linchenn:

### DIFF
--- a/tfjs-core/src/backends/einsum_util.ts
+++ b/tfjs-core/src/backends/einsum_util.ts
@@ -20,8 +20,9 @@
  * based on Einstein summation.)
  */
 
+import {dispose, matMul, Rank, reshape, transpose} from '../base';
 import {Tensor} from '../tensor';
-import {assert} from '../util_base';
+import {arraysEqual, assert} from '../util_base';
 
 const ARROW = '->';
 const ARROW_REGEX = /->/g;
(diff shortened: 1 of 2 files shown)


tfjs-core/src/ops/einsum_test.ts
```diff
@@ -262,4 +262,26 @@ describeWithFlags('einsum', ALL_ENVS, () => {
     const y = tensor2d([[0, 1], [2, 3], [4, 5]]);
     expect(() => tf.einsum('ij,jk->ik->i', x, y)).toThrowError(/exactly one/);
   });
+
+  it('reduce einsum to batch matmul', async () => {
```
pyu10055, line 266: how does this test ensure it is reduce to matmul op? it only check the results are the same, right?

Linchenn: Decoupled `checkReducibleToMatMul` out of einsum implementation to ensure this.

An alternative approach is to make `spyOn` to tf.matMul  to check if matMul is called. However, the methods/properties of module `tf` is immutable, so we could not use `spyOn(tf, 'matMul')`. To make this work, we need to have some mechanism similar as tfjs-converter's op-executor, which allows to insert [fake dependency](https://github.com/tensorflow/tfjs/blob/master/tfjs-converter/src/operations/executors/arithmetic_executor_test.ts#L74). However, the efforts are non-trivial.

mattsoulanille (changes requested):

PR #7932, opened 2023-08-23
[WebGPU] Fix shader key for ScatterProgram

Linchenn:
## Issue
This PR fixes the following error because a TensorScatterUpdate op is trying to reuse a cached pipeline but it should not, because the size of their uniforms (the [`strides`](https://github.com/tensorflow/tfjs/blob/20ceb6f1f710271f428078b56ebddfcdfc1e1428/tfjs-backend-webgpu/src/kernels/TensorScatterUpdate.ts#L66), the length of which is dynamic) are different.
image

## Solution
We could also add such uniform information into [TensorScatterUpdate kernel's shader key](https://github.com/tensorflow/tfjs/blob/20ceb6f1f710271f428078b56ebddfcdfc1e1428/tfjs-backend-webgpu/src/scatter_webgpu.ts#L51) instead of all kernel's shader key.

## Alternatives considered
We could also add uniform informations into shader key to fix it, but it's unnecessary to apply it to all shader keys since this uniform problem only happens, from my perspective, for scatter program.


qjia7:
>In theory, we could also add such uniform information into [TensorScatterUpdate kernel's shader key](https://github.com/tensorflow/tfjs/blob/20ceb6f1f710271f428078b56ebddfcdfc1e1428/tfjs-backend-webgpu/src/scatter_webgpu.ts#L51) instead of all kernel's shader key. However, the uniformData is passed directly into [runWebGPUProgram](https://github.com/tensorflow/tfjs/blob/20ceb6f1f710271f428078b56ebddfcdfc1e1428/tfjs-backend-webgpu/src/kernels/TensorScatterUpdate.ts#L69), and is not passed to ScatterProgram constructor, so [TensorScatterUpdate kernel's shader key](https://github.com/tensorflow/tfjs/blob/20ceb6f1f710271f428078b56ebddfcdfc1e1428/tfjs-backend-webgpu/src/scatter_webgpu.ts#L51) could not know the size of uniforms at runtime and thus could not add such uniform information.

Does it work to change it as below?    
```
this.shaderKey = `scatter_${indicesRank}_${updatesRank}_${
        this.sliceDimGreaterThanOne}_${outputDtype}_${sumDupeIndices}`;
```
to
```
this.shaderKey = `scatter_${indicesRank}_${updatesRank}_${
        this.sliceDimGreaterThanOne}_${outputDtype}_${sumDupeIndices}_${strides.length}`;
```

Linchenn:
> Does it work to change [it](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/src/scatter_webgpu.ts#L51) as below?

Nice catch! It works and just updated the PR.

qjia7 (approved):
LGTM, thanks!

pyu10055 (approved):

PR #7933, opened 2023-08-24
Implement Sampler and TopKSampler

pforderique:
Implements the [TopKSampler](https://github.com/keras-team/keras-nlp/blob/v0.6.1/keras_nlp/samplers/top_k_sampler.py) from Keras NLP.

### DIFF
--- a/tfjs-layers/src/layers/nlp/samplers.ts
+++ b/tfjs-layers/src/layers/nlp/samplers.ts
@@ -1,0 +1,226 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+/**
+ * Sampler classes.
+ */
+
+/* Original source: keras_nlp/samplers/sampler.py */
+import { Tensor, serialization, softmax, tensor, topk, where, zerosLike } from '@tensorflow/tfjs-core';
+import { sliceUpdate } from './utils';
+
+export type NextFn =
+  (prompt: Tensor, cache: Tensor, index: number) => [Tensor, Tensor, Tensor];
+
+export interface SamplerArgs {
+  /**
+   * Float.
+   * Used to control the randomness of the sampling.
+   * The higher the temperature, the more diverse the samples.
+   * Defaults to `1.0`.
+   */
+  temperature?: number;
+}
+
+/**
+ * Base sampler class.
+ *
+ * This base class can be extended to implement different auto-regressive
+ * sampling methods. Subclasses can either:
+ *
+ * - Override the `getNextToken()` method, which computes the next token
+ * based on a probability distribution over all possible vocab entries.
+ *
+ * - Override `apply()`, if the sampling method needs additional information
+ * beyond the next tokens probability distribution to sample a sequence.
+ * Please check available subclass samplers for examples.
+ *
+ * Examples:
+ * // TODO(pforderique): Add examples.
+ */
+export abstract class Sampler {
+  temperature: number;
+
+  constructor(args: SamplerArgs) {
+    this.temperature = args.temperature ?? 1.0;
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


tfjs-layers/src/layers/nlp/samplers.ts
```diff
@@ -0,0 +1,226 @@
...
+    cache?: Tensor,
+    index?: number,
+    mask?: Tensor,
+    endTokenId?: number,
+    hiddenStates?: Tensor
+  ): Tensor {
+    const maxLength = prompt.shape[prompt.shape.length - 1];
+    index = index ?? 0;
+    mask = mask == null ? zerosLike(prompt).cast('bool') : mask.cast('bool');
+    cache = cache ?? tensor([]);
```
mattsoulanille, line 97: Tiny style nit

```suggestion
    cache: Tensor = tensor([]),
    index: number = 0,
    mask?: Tensor,
    endTokenId?: number,
    hiddenStates?: Tensor
  ): Tensor {
    const maxLength = prompt.shape[prompt.shape.length - 1];
    mask = mask == null ? zerosLike(prompt).cast('bool') : mask.cast('bool');
```

tfjs-layers/src/layers/nlp/samplers.ts
```diff
@@ -0,0 +1,226 @@
...
+        nextToken,
+      );
+      // Update the prompt with the next token.
+      nextToken = nextToken.expandDims(-1);
+      prompt = sliceUpdate(prompt, [0, index], nextToken);
+
+      index += 1;
+      iter += 1;
+    }
+    return prompt;
```
mattsoulanille, line 135: This needs to use tf.tidy inside the while loop. Otherwise, it keeps every intermediate tensor, which in gpt2's case is 1024 loops of tensors.

tfjs-layers/src/layers/nlp/samplers.ts
```diff
@@ -0,0 +1,226 @@
...
+ * Examples:
+ * // TODO(pforderique): Add examples.
+ */
+export class TopKSampler extends Sampler {
+  k: number;
+  seed: number;
+
+  constructor(args: TopKSamplerArgs) {
+    super(args);
+    this.k = args.k;
```
mattsoulanille, line 189:
```suggestion
    this.k = args.k ?? 5;
```

tfjs-layers/src/layers/nlp/samplers.ts
```diff
@@ -0,0 +1,226 @@
...
+    this.seed = args.seed;
+  }
+
+  override getNextToken(probabilities: Tensor): Tensor {
+    // Filter out top-k tokens.
+    const k = 5;
+    const {values, indices} = topk(probabilities, k, false);
```
mattsoulanille, line 196:
```suggestion
    const {values, indices} = topk(probabilities, this.k, false);
```

tfjs-layers/src/layers/nlp/samplers.ts
```diff
@@ -0,0 +1,226 @@
...
+      const samplesArr = [];
+
+      for (const probDistribution of probabilitiesArr) {
+        let sample =
+          Math.random() * probDistribution.reduce((a, b) => a + b, 0);
+        const sampleIdx = probDistribution.findIndex(val => {
+          return (sample -= val) <= 0;
+        });
+        samplesArr.push(sampleIdx);
+      }
```
mattsoulanille, line 209: I feel like this can be done in `O(log(n))`, but that likely doesn't matter since the rest of the model takes much longer.
```suggestion
      // TODO(mattSoulanille): Investigate whether this should use binsearch.
      const probabilitiesArr = probabilities.arraySync() as number[][];
      const samplesArr = [];

      for (const probDistribution of probabilitiesArr) {
        let sample =
          Math.random() * probDistribution.reduce((a, b) => a + b, 0);
        const sampleIdx = probDistribution.findIndex(val => {
          return (sample -= val) <= 0;
        });
        samplesArr.push(sampleIdx);
      }
```

tfjs-layers/src/layers/nlp/samplers.ts
mattsoulanille: Here's a somewhat hacky version I used. It's probably better to use local variables in the tidy and return them from the tidy function. Then, assign the outside variables to the returned ones.

https://github.com/mattsoulanille/tfjs/blob/gpt2_test2/tfjs-layers/src/layers/nlp/samplers.ts#L115-L142

pforderique: got it! For learning purposes, why do we have to manually dispose the oldCache/oldPrompt? I was under the impression the tidy would take care of it. Thanks!

tfjs-layers/src/layers/nlp/samplers.ts
pforderique: sounds good! I'll leave the TODO - only thing is I'm not sure we can expect the distribution to be sorted in any way

Issue #7929, opened 2023-08-23
It seems the OP Conv2DBackpropInput will use more time in the latest versoin of tfjs in the Intel devices 


zhiliang029:
When we use tfjs to do VB, we find the version of tfjs 4.10.0 will use more time to get the result than the version of 3.11.0 on Intel i7-7700  HD630 graphics. The following is the result of the benchmark.

the version of tfjs 3.11.0
![image](https://github.com/tensorflow/tfjs/assets/122433353/b7f7212c-9500-4c51-a5be-dc26536f78af)

the version of tfjs 4.10.0
![image](https://github.com/tensorflow/tfjs/assets/122433353/db8adc1e-eebe-4953-9137-d31eacedd927)

the OP Conv2DBackpropInput will use more time on the version of tfjs 4.10.0

qjia7:
@Linchenn  Some additional information:
1. backend: webgl
2. For tfjs 3.11.0, it sill uses `Conv2DDerInputProgram` instead of `Conv2DDerInputPackedProgram`.

So it seems `Conv2DBackpropInput`'s shape in user's model runs faster on the non-packed program than the packed one.

@zhiliang029 Can you share the `Conv2DBackpropInput`'s details information in `individual` mode? Just choose the `kernelTiming` in control panel to `individual` instead of `aggregate`. Then only copy `Conv2DBackpropInput`'s data here. It will be helpful to know the shape you used for `Conv2DBackpropInput`. Then we can adjust the condition to switch to `Conv2DDerInputProgram` or `Conv2DDerInputPackedProgram`.

zhiliang029:
![image](https://github.com/tensorflow/tfjs/assets/122433353/e1775c58-a37b-4e0d-977b-f1c977dde08c)

Linchenn:
It's weird that packed version is slower than unpacked one. I would guess it's from the changes of this file, https://github.com/tensorflow/tfjs/pull/7386/files#diff-985bc9b238aea25dd6e5f275d50d9bee56b105a584a005f14f8716288954f01a. 

Could you help me test whether it's improved on intel devices, if you revert the changes in this file?

Linchenn:
Benchmark results:
After pack optimization - 2.3ms
Revert the [changes](https://github.com/tensorflow/tfjs/pull/7386/files#diff-985bc9b238aea25dd6e5f275d50d9bee56b105a584a005f14f8716288954f01a) - 1.6ms
Before pack optimization - 0.7ms

[...]

PR #7934, opened 2023-08-24
[WebGL] Flag to disable pack optimization for Conv2dTranspose

Linchenn:
Fix https://github.com/tensorflow/tfjs/issues/7929.


qjia7 (approved):
Let's have fine tuning when we have enough theoretical proof that in some conditions, the packed version is not good for some cases. Now it's good for me just expose this flag to users so that they can decide whether to use the packed one based on their situations.

pyu10055 (approved):

PR #7942, opened 2023-09-05
[tfjs-core] fix gather gradient when batchDims is 1

### LINKED ISSUE #7494: tf.gather causes reshape error during gradient computation
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow.js):
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): MacOS
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device: N.A.
- TensorFlow.js installed from (npm or script link): npm
- TensorFlow.js version (use command below): 4.2.0
- Browser version: N.A.
- Tensorflow.js Converter Version: N.A.

When using `tf.gather`, the gradient computation throws an error like this:
```
Error: Invalid TF_Status: 3
Message: Input to reshape is a tensor with 64 values, but the requested shape has 4096
```

64 here is the batch size and 46*46=4096.

The reshape error occurred in [gatherGradConfig](https://github.com/tensorflow/tfjs/blob/087a02a2848db8ba1950c6894fc76d8be3936f11/tfjs-core/src/gradients/GatherV2_grad.ts#L27).

Replacing `tf.gather` usage with my own custom implementation worked and the gradient computes successfully using the tensor.

**Code where the error occurs:**

> input is tensor of shape: [ 64, 2 ], indices is tensor of shape [ 64, 1 ]

```ts
// input: Tensor {
//   shape: [ 64, 2 ],
//   dtype: 'float32',
//   size: 128,
//   strides: [ 2 ],
//   rankType: '2',
// }
// indices: Tensor {
//   shape: [ 64, 1 ],
//   dtype: 'int32',
//   size: 64,
//   strides: [ 1 ],
//   rankType: '2',
// }

const result = tf.gather(input, indices, 1, 1);
// I also tried the following, same error
// const result = tf.gather(input, indices.arraySync(), 1, 1);

// result shape: [ 64, 1 ]

const grads = optimizer.computeGradients(() => {
  // computation using result
}

// Error: Invalid TF_Status: 3
// Message: Input to reshape is a tensor with 64 values, but the requested shape has 4096
```

**Replacing with custom tf.gather implementation does not result in an error, and the gradient computation is behaving as expected:**

```ts
function gatherOwn(input: tf.Tensor, indices: tf.Tensor): tf.Tensor {
  const batchSize = input.shape[0];
  const tensors = input.split(batchSize, 0);
  const output = tf.stack(tensors.map((tensor, i) => tensor.squeeze().gather(indices.arraySync()[i])));
  return output;
}

const result = gatherOwn(input, indices);

const grads = optimizer.computeGradients(() => {
  // computation using result
}

// No error, gradient computes as expected
```

Using tf.gather should not cause the gradient computation to throw an error.

I can provide a standalone code, but it will take some time on my side.

N.A.

paradite:
This PR fixes #7494, where error is encountered when computing gradient of `tf.gather` with `batchDims=1`.

I am opening a PR with a naive and unoptimized fix first to see if the CI passes, and gather feedback on the direction of the fix.

If the CI passes, I will work on optimizing the logic if needed.

I am not really familiar with the maths behind the original implementation in `derX` of `gatherGradConfig`, and I couldn't get it to work within the `derX` function, I am open to suggestions on better and more optimized way to fix the issue.

### DIFF
--- a/tfjs-core/src/gradients/GatherV2_grad.ts
+++ b/tfjs-core/src/gradients/GatherV2_grad.ts
@@ -36,18 +36,7 @@ export const gatherGradConfig: GradConfig = {
 
     const derX = () => {
       const paramsShape = x.shape;
-      console.log("TCL ~ x:", x.shape, x.arraySync())
-      // console.log("TCL ~ dy:", dy.shape, dy.arraySync())
-      // console.log("TCL ~ batchDims, parsedAxis:", batchDims, parsedAxis)
       const indicesSize = indices.size;
-      // console.log("TCL ~ indicesSize:", indicesSize)
-
-      let $batchDims = batchDims;
-
-      if (batchDims == null) {
-        $batchDims = 0;
-      }
-      console.log("TCL ~ $batchDims:", $batchDims)
 
       const outerShape = paramsShape.slice(0, parsedAxis);
       const outerDims = outerShape.length;
@@ -68,15 +57,12 @@ export const gatherGradConfig: GradConfig = {
       const valuesTranspose = transpose(values, transposeDims);
       let paramsGrad = unsortedSegmentSum(
           valuesTranspose, reshapedIndices as Tensor1D, x.shape[parsedAxis]);
-      console.log("TCL ~ paramsGrad:", paramsGrad.shape, paramsGrad.dataSync())
       const invertTransposeDims = getUndoAxesPermutation(transposeDims);
       paramsGrad = transpose(paramsGrad, invertTransposeDims);
-      console.log("TCL ~ paramsGrad:", paramsGrad.shape, paramsGrad.dataSync())
       return paramsGrad;
     };
 
     const derXIndividual = (x: Tensor, indices: Tensor, dy: Tensor) => {
-      console.log("TCL ~ dy:", dy, dy.shape, dy.dataSync())
       return function(): Tensor {
         const paramsShape = x.shape;
         const indicesSize = indices.size;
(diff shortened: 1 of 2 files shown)


mattsoulanille:
Thanks for the contribution! It looks like you're splitting the batched tensor, applying the gradient function to each batch, and then stacking them together again. This seems reasonable to me.

The original math for `derX` is also not really clear to me. For `y = tf.gather(x, indices, axis, batchDim)`, I'd expect `dy/dx` to just copy the selected values specified by the `indices` variable from the `dy` variable and be zero everywhere else. However, it seems to be using `unsortedSegmentSum`. I think this is to account for repeated indices. There's also some complicated transpose logic to make the input work for `unsortedSegmentSum`. It might be possible to get the indices of these two ops to work correctly for arbitrary batch dimensions, but I'm not sure. I'm fine with your approach, and we can revisit it if we need better performance in the future.

There's probably a way to apply your approach for generic batchDims, but I'm not going to block this PR because of that.

Also, it looks like CI is failing due to code linting. I'll push a patch to fix that.

mattsoulanille (approved):
LGTM

paradite:
Thank you @mattsoulanille for the help and review. I will take note of the lint issue for future PRs.

Hi @pyu10055, can I get your blessing and help to merge this?

pyu10055 (approved):

PR #7951, opened 2023-09-12
Make MultiHeadAttention use masks from query and value tensors

mattsoulanille:
Add optional [Keras masks](https://www.tensorflow.org/guide/keras/understanding_masking_and_padding) to tfjs tensors. Enable them for tfjs-layers on layers that emit them. Use the masks of query and value input tensors in MultiHeadAttention to compute the correct mask automatically.

### DIFF
--- a/tfjs-layers/src/engine/topology.ts
+++ b/tfjs-layers/src/engine/topology.ts
@@ -751,19 +751,19 @@ export abstract class Layer extends serialization.Serializable {
    */
   protected assertInputCompatibility(inputs: Tensor|Tensor[]|SymbolicTensor|
                                      SymbolicTensor[]): void {
-    inputs = generic_utils.toList(inputs);
+    const inputsList = generic_utils.toList(inputs);
     if (this.inputSpec == null || this.inputSpec.length === 0) {
       return;
     }
     const inputSpec = generic_utils.toList(this.inputSpec);
-    if (inputs.length !== inputSpec.length) {
+    if (inputsList.length !== inputSpec.length) {
       throw new ValueError(
           `Layer ${this.name} expects ${inputSpec.length} inputs, ` +
-          `but it received ${inputs.length} input tensors. ` +
+          `but it received ${inputsList.length} input tensors. ` +
           `Input received: ${inputs}`);
     }
-    for (let inputIndex = 0; inputIndex < inputs.length; inputIndex++) {
-      const x = inputs[inputIndex];
+    for (let inputIndex = 0; inputIndex < inputsList.length; inputIndex++) {
+      const x = inputsList[inputIndex];
       const spec: InputSpec = inputSpec[inputIndex];
       if (spec == null) {
         continue;
@@ -954,20 +954,8 @@ export abstract class Layer extends serialization.Serializable {
     // Ensure inputs are all the same type.
     const inputsList = generic_utils.toList(inputs);
 
-    let allAreSymbolic = true;
-    for (const input of inputsList) {
-      if (!(input instanceof SymbolicTensor)) {
-        allAreSymbolic = false;
-        break;
-      }
-    }
-    let noneAreSymbolic = true;
-    for (const input of inputsList) {
-      if (input instanceof SymbolicTensor) {
-        noneAreSymbolic = false;
-        break;
-      }
-    }
+    const allAreSymbolic = checkAllSymbolic(inputs);
+    const noneAreSymbolic = checkNoneSymbolic(inputs);
 
     if (allAreSymbolic === noneAreSymbolic) {
       throw new ValueError(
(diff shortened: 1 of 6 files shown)


tfjs-layers/src/engine/topology.ts
```diff
@@ -1018,7 +1018,31 @@ export abstract class Layer extends serialization.Serializable {
...
+          const outputMask = this.computeMask(inputs as Tensor | Tensor[]);
+          if (output instanceof Array && outputMask instanceof Array) {
+            if (output.length !== outputMask.length) {
+              throw new Error(`${this.name} output ${output.length} tensors `
+                + `but ${outputMask.length} masks for those tensors`);
+            }
+            for (let i = 0; i < output.length; i++) {
+              output[i].kerasMask = outputMask[i];
+            }
+          } else if (outputMask instanceof Array) {
```
pyu10055, line 1034: what if the array only contains one mask?

tfjs-layers/src/layers/nlp/multihead_attention.ts
```diff
@@ -813,12 +813,20 @@ export class MultiHeadAttention extends Layer {
...
+      }
+      if (valueMask != null) {
+        const mask = valueMask.expandDims(1); // Shape is [B, 1, S]
+        autoMask = autoMask ? logicalAnd(autoMask, mask) : mask;
+      }
       if (useCausalMask) {
         // the shape of the causal mask is [1, T, S]
         const mask = this.computeCausalMask(query, value);
-        autoMask = mask;
+        autoMask = autoMask ? logicalAnd(autoMask, mask) : mask;
```
pyu10055, line 828: how is this associated with the Topology computeMask logic?

tfjs-layers/src/engine/topology.ts
mattsoulanille: I think that should be an error. If there's only one mask for all the tensors, it should be returned as a `tensor` instead of a `[tensor]`.

mattsoulanille: Actually, keras does not seem to broadcast masks at all. Each output tensor needs its own mask:
https://github.com/keras-team/keras/blob/master/keras/engine/base_layer.py#L2893-L2898

tfjs-layers/src/layers/nlp/multihead_attention.ts
mattsoulanille: Earlier layers (Embedding) have computeMask called to compute the mask for their output tensors. This layer uses those masks.

tfjs-layers/src/engine/topology.ts
```diff
@@ -1642,3 +1660,29 @@ export function getSourceInputs(
...
+                          ): tensors is Tensor | Tensor[] {
+  let noneAreSymbolic = true;
+  for (const tensor of generic_utils.toList(tensors)) {
+    if (tensor instanceof SymbolicTensor) {
+      noneAreSymbolic = false;
+      break;
+    }
+  }
+  return noneAreSymbolic;
+}
```
mattsoulanille, line 1688: I moved this here so I could write them as type guards (`tensors is Tensor...`)

tfjs-layers/src/layers/nlp/multihead_attention_test.ts
```diff
@@ -188,101 +188,6 @@ describe('MultiHeadAttention', () => {
     expectTensorsNotClose(queryKernel, outputKernel, 1e-6);
   });
 
-  describeMathCPU('High Dimensional Attention', () => {
```
mattsoulanille, line 191: Moved

tfjs-layers/src/layers/nlp/multihead_attention_test.ts
```diff
@@ -298,52 +203,60 @@ describe('MultiHeadAttention', () => {
...
+        Array<boolean[]>(5).fill([true, false, false]),
+        [[true, true, true]].concat(
+          Array<boolean[]>(4).fill([false, false, false]))
+      ]);
+    });
+
     /**
      * Test that the value and causal masks are taken into account.
      */
-    function testValueMask(testcaseName: string, useCausalMask: boolean) {
```
mattsoulanille, line 305: Refactored to explicitly declare both tests instead of using a loop to declare them.

tfjs-layers/src/layers/nlp/multihead_attention_test.ts
```diff
@@ -482,6 +395,101 @@ describe('MultiHeadAttention', () => {
   // TODO(pforderique): Test serialization.
 });
 
+describeMathCPU('High Dimensional Attention', () => {
```
mattsoulanille, line 398: Moved some tests from above to here.

fengwuyao (approved):

pyu10055 (approved):

Issue #7408, opened 2023-02-22
wasm backend kernel `Fill` dtype not optional


sklum:
For the `Fill` kernel, the `webgl` backend [uses](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/kernels/Fill.ts#L29) `util.inferDtype` to determine data type which has the functionality described in the docs: `The type of an element in the resulting tensor. Defaults to 'float'.`

The `wasm` backend [does not use](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-wasm/src/kernels/Fill.ts#L24) `util.inferDtype`, and as written the `dtype` argument is not optional (you'll get an unrecognized dtype `undefined` error if unspecified).

Happy to open a PR that fixes this if that'd be helpful.

chunnienc:
Hi @sklum,

Thanks for your report! I can confirm it's a bug as you described. Feel free to open a PR if you could!

gaikwadrahul8:
Hi, @sklum 

Apologize for the delayed response and just to confirm, May I know do have workaround to take care of this issue ? If yes, would you like to submit PR from your end or do you want to take care of this issue from our end ? Thank you!

sklum:
Hi @gaikwadrahul8.  I haven't had time to get a PR up although I expect to within a week or two.  Feel free to implement if you can get it done before then.

The workaround is simply to specify the `dtype` (i.e. treat the parameter as non-optional).

gaikwadrahul8:
Hi, @sklum 

Please take your time and submit PR to take care of this issue from your end and our team will review your PR and if everything looks good team will merge your PR. Thank you!

sklum:
Hi @chunnienc, sorry for the delay in fixing this.  Turns out you fixed it in another commit.  Please see the referenced PR and let me know if you'd like me do anything else.

PR #7953, opened 2023-09-12
Infer dtype in wasm fill

sklum:
This makes the `dtype` handling in `wasm` follow the same codepath as the `webgl` backend.

Originally, I had [this issue](https://github.com/tensorflow/tfjs/issues/7408) but it was fixed by @chunnienc's commit:  https://github.com/tensorflow/tfjs/commit/ff6739dd2b0fd5af59d35a9dbdf698a525b1791f#diff-ec2bd7b410dc39732c6db76ce69e87aa309d69fb0d53d355f5e43aa055dd3b7bR44.  I discovered the fix while testing this PR, but thought I would open it anyway as it may be nice as a sanity check to have both `Fill`s have the same behavior.

Alternatively, I can open a PR that eliminates [the util.inferDtype call](https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgl/src/kernels/Fill.ts#L29) within the `webgl` backend as it's now a no-op.

### DIFF
--- a/tfjs-backend-wasm/src/kernels/Fill.ts
+++ b/tfjs-backend-wasm/src/kernels/Fill.ts
@@ -15,13 +15,17 @@
  * =============================================================================
  */
 
-import {KernelConfig, KernelFunc} from '@tensorflow/tfjs-core';
+import {KernelConfig, KernelFunc, util} from '@tensorflow/tfjs-core';
 import {Fill, FillAttrs} from '@tensorflow/tfjs-core';
 
 import {BackendWasm} from '../backend_wasm';
 
 export function fill(args: {attrs: FillAttrs, backend: BackendWasm}) {
-  const {attrs: {shape, value, dtype}, backend} = args;
+  const {attrs: {shape, value}, backend} = args;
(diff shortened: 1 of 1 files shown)


tfjs-backend-wasm/src/kernels/Fill.ts
```diff
@@ -15,13 +15,17 @@
...
 import {Fill, FillAttrs} from '@tensorflow/tfjs-core';
 
 import {BackendWasm} from '../backend_wasm';
 
 export function fill(args: {attrs: FillAttrs, backend: BackendWasm}) {
-  const {attrs: {shape, value, dtype}, backend} = args;
+  const {attrs: {shape, value}, backend} = args;
+  let {attrs: {dtype}} = args;
+
+  dtype = dtype || util.inferDtype(value)
```
chunnienc, line 27: nit: add semicolon

chunnienc (approved):
LGTM. Please make sure the comment is resolved before merging this PR.

sklum:
Should be good now!

pyu10055 (approved):

Issue #7972, opened 2023-09-23
Enable importExternalTexture for webgpu-fromPixels


Linchenn:
It seems we disabled `importExternalTexture` previously because of webgpu issue, https://github.com/tensorflow/tfjs/pull/6868, but I saw WebGPU is [advertising](https://pr-7324-static-dot-dcc-staging.uc.r.appspot.com/blog/from-webgl-to-webgpu/#video-frame-processing) this feature recently, so could we enable this feature too?

Linchenn:
@gyagp @shaoboyan

gyagp:
Yes, we can turn it on again. Just uploaded https://github.com/tensorflow/tfjs/pull/7976 to enable it and PTAL.

PR #7976, opened 2023-09-25
webgpu: Enable importExternalTexture

gyagp:
Issues related to importExternalTexture were fixed in Chrome, so that we may enable this feature by default again.
Note that its Linux implementation still has problem, and WebGPU support on Linux hasn't been officially released yet.
Bug #7972

### DIFF
--- a/tfjs-backend-webgpu/src/kernels/FromPixels.ts
+++ b/tfjs-backend-webgpu/src/kernels/FromPixels.ts
@@ -63,10 +63,8 @@ export function fromPixels(args: {
       [pixels.width, pixels.height];
   const outputShape = [height, width, numChannels];
 
-  // Disable importExternalTexture temporarily as it has problem in spec and
-  // browser impl
   const importVideo =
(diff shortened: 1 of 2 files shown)


tfjs-backend-webgpu/src/kernels/FromPixels.ts
```diff
@@ -63,10 +63,8 @@ export function fromPixels(args: {
       [pixels.width, pixels.height];
   const outputShape = [height, width, numChannels];
 
-  // Disable importExternalTexture temporarily as it has problem in spec and
-  // browser impl
   const importVideo =
-      false && env().getBool('WEBGPU_IMPORT_EXTERNAL_TEXTURE') && isVideo;
+      env().getBool('WEBGPU_IMPORT_EXTERNAL_TEXTURE') && isVideo;
```
qjia7, line 67: I remember we have a related test case. Do you need to change the corresponding test?

gyagp: Yes, I just added the fix to the test case. Thanks!

Linchenn (approved):
Thank you Yang!

PR #7994, opened 2023-10-09
Stream weights to the GPU when loading a model

mattsoulanille:
When downloading model weight data, slice it into weight tensors and push them to the GPU eagerly. This avoids storing an extra copy of the weights on CPU, allowing for larger models (1.3B to possibly ~6.7B or larger) to be loaded without causing a V8 OOM crash.

When streaming the weights, check `CPU_HANDOFF_SIZE_THRESHOLD` or `WEBGPU_CPU_HANDOFF_SIZE_THRESHOLD` to determine whether the weight should be sent to GPU or remain on CPU.

This feature is guarded by the `streamWeights` option in `LoadOptions`. Since most of TFJS's graph model saving relies on the CPU copy of the model, model saving is disabled when the model was streamed (i.e. it will throw an error since the weights ArrayBuffer is missing).

### DIFF
--- a/tfjs-core/src/io/types.ts
+++ b/tfjs-core/src/io/types.ts
@@ -544,7 +544,7 @@ export interface LoadOptions {
    * Whether to stream the model directly to the backend or cache all its
    * weights on CPU. Useful for large models.
    */
-  streamWeights?: boolean
+  streamWeights?: boolean;
 }
 
 /**
--- a/tfjs-core/src/io/types.ts
+++ b/tfjs-core/src/io/types.ts
@@ -259,10 +259,10 @@ export declare interface ModelArtifacts {
 
   /**
    * Returns a stream of the weights. Some models are too large to fit in
-   * V8's memory heap, and `streamWeights` loads their weights without storing
+   * V8's memory heap, and `getWeightStream` loads their weights without storing
    * them all in memory at the same time.
    */
-  streamWeights?: () => ReadableStream<ArrayBuffer>;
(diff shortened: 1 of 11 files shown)


tfjs-core/src/io/types.ts
```diff
@@ -532,6 +539,12 @@ export interface LoadOptions {
    * With this func you can convert the weight file name to any URL.
    */
   weightUrlConverter?: (weightFileName: string) => Promise<string>;
+
+  /**
+   * Whether to stream the model directly to the backend or cache all its
+   * weights on CPU. Useful for large models.
+   */
+  streamWeights?: boolean;
```
pyu10055, line 547: this flag is the same name as the function `streamWeights?: () => ReadableStream<ArrayBuffer>`?

pyu10055 (approved):

tfjs-core/src/io/types.ts
mattsoulanille: Yes, but they're part of different interfaces. One is on the `ModelArtifacts` and is the actual function that, when called, will start streaming the weights. It's internal, and users shouldn't really need to access it.

The other is on the `LoadOptions` interface, and it's used to configure whether the model loader should stream the weights or load them normally. It's exposed to the user.

I can rename the one on `ModelArtifacts` to something like `startWeightsStream` or `streamWeightsData`.

Linchenn (approved):
Thanks Matt!

PR #8001, opened 2023-10-12
webgpu: fix browser check

crowlKats:
This change is due to this not working in Deno's NPM compatibility.

### DIFF
--- a/tfjs-backend-webgpu/src/webgpu_util.ts
+++ b/tfjs-backend-webgpu/src/webgpu_util.ts
@@ -155,10 +155,7 @@ export function GPUBytesPerElement(dtype: DataType): number {
 }
 
 export function isWebGPUSupported(): boolean {
-  return ((typeof window !== 'undefined') ||
-          //@ts-ignore
-          (typeof WorkerGlobalScope !== 'undefined')) &&
-      !!navigator.gpu;
+  return !!globalThis.navigator.gpu;
 }
 
 export function assertNotComplex(


tfjs-backend-webgpu/src/webgpu_util.ts
```diff
@@ -155,10 +155,7 @@ export function GPUBytesPerElement(dtype: DataType): number {
 }
 
 export function isWebGPUSupported(): boolean {
-  return ((typeof window !== 'undefined') ||
-          //@ts-ignore
-          (typeof WorkerGlobalScope !== 'undefined')) &&
-      !!navigator.gpu;
+  return !!globalThis.navigator.gpu;
```
qjia7, line 158: It seems that the functionality is equal before and after the change. Can you explain more why it can fix the Deno's NPM compatibility issue?

qjia7: I am ok with this change since it makes the code simpler. Just curious how it fix the issue you mentioned.

qjia7 (approved):

tfjs-backend-webgpu/src/webgpu_util.ts
crowlKats: Node doesnt have `window`, as such in our NPM compat we remove the object. However, we still have the global scope, so navigator is present, but not the `window` object itself.

Linchenn (approved):

PR #8041, opened 2023-10-30
docs:fix to README where the provided example prompts type errors in …

chaxus:
When I first started using `@tensorflow/tfjs`, I first followed the code examples in the `README.md`. But it hints at a type error in `typescript`.

Then I found out that the return type of `model.predict` might be an `array`, but `array` don't have `print` methods. So I made a judgment to resolve the type error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,12 @@ Add the following code to an HTML file:
       model.fit(xs, ys).then(() => {
         // Use the model to do inference on a data point the model hasn't seen before:
         // Open the browser devtools to see the output
-        model.predict(tf.tensor2d([5], [1, 1])).print();
+        const result = model.predict(tf.tensor2d([5], [1, 1]))
+        if (Array.isArray(result)) {
+          result.forEach(item => item.print())
+        } else {
+          result.print()
+        }
       });
     </script>
   </head>
@@ -153,7 +158,12 @@ const ys = tf.tensor2d([1, 3, 5, 7], [4, 1]);
 // Train the model using the data.
 model.fit(xs, ys).then(() => {
   // Use the model to do inference on a data point the model hasn't seen before:
-  model.predict(tf.tensor2d([5], [1, 1])).print();
+  const result = model.predict(tf.tensor2d([5], [1, 1]))
+  if (Array.isArray(result)) {
+    result.forEach(item => item.print())
+  } else {
+    result.print()
+  }
 });
 ```
 


verden11:
I ran into a similar issue. Would it not be better to update the types of the function? i.e. adding
```ts
predict(x: Tensor, args: ModelPredictArgs = {}): Tensor;
predict(x: Tensor[], args: ModelPredictArgs = {}): Tensor[];
```
just above already existing
```ts
predict(x: Tensor|Tensor[], args: ModelPredictArgs = {}): Tensor|Tensor[] {
```
line in `tfjs-layers/src/engine/training.ts` should do the trick and no `README.md` changes would be needed

chaxus:
> I ran into a similar issue. Would it not be better to update the types of the function? i.e. adding
> 
> ```ts
> predict(x: Tensor, args: ModelPredictArgs = {}): Tensor;
> predict(x: Tensor[], args: ModelPredictArgs = {}): Tensor[];
> ```
> 
> just above already existing
> 
> ```ts
> predict(x: Tensor|Tensor[], args: ModelPredictArgs = {}): Tensor|Tensor[] {
> ```
> 
> line in `tfjs-layers/src/engine/training.ts` should do the trick and no `README.md` changes would be needed

I am not sure if I will return Tensor[] when the argument is Tensor. If the argument is Tensor, the result will only be Tensor. The argument is Tensor[] and only returns Tensor[]. Then I think it would be better if you modified it this way.

verden11:
> > I ran into a similar issue. Would it not be better to update the types of the function? i.e. adding
> > ```ts
> > predict(x: Tensor, args: ModelPredictArgs = {}): Tensor;
> > predict(x: Tensor[], args: ModelPredictArgs = {}): Tensor[];
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > just above already existing
> > ```ts
> > predict(x: Tensor|Tensor[], args: ModelPredictArgs = {}): Tensor|Tensor[] {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > line in `tfjs-layers/src/engine/training.ts` should do the trick and no `README.md` changes would be needed
> 
> I am not sure if I will return Tensor[] when the argument is Tensor. If the argument is Tensor, the result will only be Tensor. The argument is Tensor[] and only returns Tensor[]. Then I think it would be better if you modified it this way.

Can it ever be that the result is an array if you pass a single Tensor? (or the other way around, can a result be a single tensor if you as an array of tensors to predict?)

Issue #7454, opened 2023-03-07
No way to create new react-native app with tfjs-react-native because of dependency issues???


rbmdotdev:
<details><summary>Click to expand!</summary> 
 
 <details><summary>Click to expand!</summary> 
 
 ### Issue Type

Bug

### Have you reproduced the bug with TF nightly?

No

### Source

source

### Tensorflow Version

tfjs-react-native-0.8.0

### Custom Code

Yes

### OS Platform and Distribution

all

### Mobile device

all

### Current Behaviour?

```shell
I hope I have just missed something dumb but I don’t think so as I have been struggling to get tfjs-react-native working for the last two days.

version 0.8.0 of this library seems to be unacceptably out of date as far as its dependencies go resulting in it not being possible.....as far as I can tell.....to instantiate a new react-native / expo app and add it as a dependency.  You have to jump through some major hoops to find expo and react SDK versions that might possibly support the versions of react and react-native this library expects and even then it still won't work for me.

There is at least 1 other ticket #7323 mentioning the fact that the libraries dependencies are so out of date that it looks like the library is not maintained at all.

The response to that ticket seems to be 'no it is maintained we will update it sometime'.  Can this please get resolved?
```

### Standalone code to reproduce the issue

```shell
follow this tutorial to create a new expo app https://docs.expo.dev/get-started/create-a-new-app/

then do npm i @tensorflow/tfjs-react-native

It will refuse to install due to conflicting react and react-native dependencies
```

### Relevant log output

_No response_</details></details>

gaikwadrahul8:
Hi, @rbmdotdev 

Apologize for the delayed response and I see this PR [#7451 ](https://github.com/tensorflow/tfjs/pull/7451) got merged but that PR does not attempt to update the demo app or any expo dependencies. These will be updated in a later PR. you can check this [comment](https://github.com/tensorflow/tfjs/pull/7451#issue-1612353879). Thank you!

rbmdotdev:
Should the library work with just react-native and no expo right now?

Episodex:
I have the same issue. Can we use this with some older versions of expo and other packages? I'm trying to find the sweet spot of packages versions but still no luck. Can somebody provide the versions of expo, react, react-native etc that can run 0.8.0?

mattsoulanille:
[...]

Issue #7323, opened 2023-01-30
tfjs-react-native package not updated and contains obsolete dependencies


vladmandic:
The package `@tensorflow/tfjs-react-native` has the latest version **0.8.0** which was published in November 2021.  
It still contains dependencies on TFJS 3.11.0 as well as pretty much obsolete versions of React, Expo (-camera, -gl, etc.) and other key packages.

Last mentions are pointing towards demo project at <https://github.com/tensorflow/tfjs-examples/blob/master/react-native/image-classification/expo/package.json> to work around dependency issues.  

But a) using such old versions introduces many security vulnerabilities, b) its not compatible with many other newer components

Is the package still maintained?

Linchenn:
@vladmandic Yes, it is still maintained, but we may do version updates in a couple of months later, since updating React and Expo may cause breaking.

vladmandic:
i understand it may not be updated exactly at the same time as `tfjs-core`,  
but right now latest published version is 15 months old?

Linchenn:
Yes, but we definitely will keep them up to date later.

rbmdotdev:
^ I am currently blocked by this dependency issue, can we get this pr merged and pushed to npm?

paradite:
While we wait for the PR to be merged, I want to share a walkaround I have been using for a few months:

Just copy the entire `platform_react_native.ts` file ([latest 4.2.0 version here](https://github.com/tensorflow/tfjs/pull/7415/files#diff-c117d2eecd67428a4e7bc46157769a45a54da43c97250da546a9c93feea83834)) into your project and initialize the RN runtime on your own in `App.tsx` with `import './xxx/platform_react_native';`.

It is only a few lines of changes and everything will work as expected with latest dependencies.

ItamarGefen:
> While we wait for the PR to be merged, I want to share a walkaround I have been using for a few months:
> 
> Just copy the entire `platform_react_native.ts` file ([latest 4.2.0 version here](https://github.com/tensorflow/tfjs/pull/7415/files#diff-c117d2eecd67428a4e7bc46157769a45a54da43c97250da546a9c93feea83834)) into your project and initialize the RN runtime on your own in `App.tsx` with `import './xxx/platform_react_native';`.
> 
> It is only a few lines of changes and everything will work as expected with latest dependencies.

[...]

PR #7415, opened 2023-02-24
Update tfjs-react-native to use tfjs 4.2.0

mattsoulanille:
Add missing function for `isTypedArray` to `platform_react_native.ts`.

Update the integration tests app to use expo and the latest version of react native.

To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

PR #8062, opened 2023-11-10
Update tfjs-react-native dependencies

paradite:
Fix https://github.com/tensorflow/tfjs/issues/7454 and fix https://github.com/tensorflow/tfjs/issues/7323

This is a follow-up on discussion in https://github.com/tensorflow/tfjs/pull/7415 to fix and update the dependencies for tfjs-react-native library, and allow it to work out of the box without tinkering or patching.

The package with new dependencies has been tested (published using a different package name [tfjs-react-native-para-patch](https://www.npmjs.com/package/tfjs-react-native-para-patch)) using a sample app I created at https://github.com/paradite/tfjs-sample-expo-app

I also updated the docs on how to handle the potential peer conflicts in future when the other dependencies eventually get updated (using `yarn` would not result in error even if the peer dependency conflicts with the direct dependency).

I will be continuing to work on this app to reach feature parity of the old sample app in the tf.js repo and submit another PR to replace it. The new app will be based on Expo and eliminate the need to having iOS and Android native code.

---

Just a side note, `@tensorflow/tfjs-react-native` has not been updated in the past two years and obviously does not get published together with other packages, so someone might need to take a look at the process to publish a new version: https://www.npmjs.com/package/@tensorflow/tfjs-react-native?activeTab=versions. In my testing, running npm run build is sufficient to get it working.


mattsoulanille (approved):
Thank you for taking this on! These changes look good to me. As for publishing, we can add tfjs-react-native back to our publishing flow once you feel it's ready. Like you said, without native modules, `npm run build` should be sufficient.

paradite:
@pyu10055 can I get your blessing for approval and merge? Thanks!

mattsoulanille (approved):

pyu10055 (approved):

paradite:
Hi @mattsoulanille @pyu10055, thanks for the help. I think the package at this state can be published on npm as it should work out of the box, and help to address the installation errors with the old version. 

I didn't look closely at the publishing process but I believe it is a manual process. Please do help to publish a new package whenever it is convenient. Let me know if there is anything I can help with. 

After a new package is published, I could test it with my sample app and proceed with next step on updating the sample app in tfjs repo.

Issue #4704, opened 2021-02-17
Support use in Worklets, specifically AudioWorklets


space-pope:
<em>Please make sure that this is a feature request. As per our [GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md), we only address code/doc bugs, performance issues, feature requests and build/installation issues on GitHub. tag:feature_template</em>

**System information**
- TensorFlow.js version (you are using): 3.0.0
- Are you willing to contribute it (Yes/No): would if I could, but the scope's a bit beyond me right now

**Describe the feature and the current behavior/state.**

A version of tfjs that works within `WorkletGlobalScope` (for my use case, `AudioWorkletGlobalScope`) would be ideal for processing data off the main thread. Network requests [are disabled](https://github.com/WebAudio/web-audio-api/issues/1439) in this scope, and the global object (`self`) is undefined. The latter prevents a tfjs distribution from even being imported in this scope, and the former would prevent functions like `loadGraphModel` from working, as they rely on fetching data over the network.

In addition, TensorFlow and models can't be loaded on the main thread and transferred to an existing `WorkletNode` via `MessagePort` since the objects of course contain functions and hence can't be cloned.

**Will this change the current api? How?**

Yes, in that model loading functions will need to be altered/added so that they accept prefetched data. I'm unclear on how much altering the use of `self` would change the API, or even how feasible that is.

**Who will benefit with this feature?**

Anyone looking to process data off the main thread in a `Worklet`. In my case, I'd like to do realtime audio processing using TensorFlow models in an [`AudioWorklet`](https://developers.google.com/web/updates/2017/12/audio-worklet).

**Any Other info.**

rthadur:
cc @pyu10055 @lina128

gaikwadrahul8:
Hi, @space-pope 

Apologize for the delayed response and we are re-visiting our older feature requests and checking whether those feature requests implemented or not as of now and I seems like we haven't implemented this feature as of now if I'm not wrong so May I know are you still looking for this feature in TFJs ?

If someone wants to contribute for this feature then you're always welcome and please feel free to do and please refer these links [Ref-1](https://github.com/tensorflow/tfjs/blob/master/CONTRIBUTING.md), [Ref-2](https://github.com/tensorflow/tfjs/blob/master/CONTRIBUTING_MISSING_OP.md) . Thank you!

space-pope:
[...]

PR #8092, opened 2023-12-08
Fix: env() does not always return platform

jbransen:
In attempts at running TensorFlow in an `AudioWorklet` (see #4704 ), I ran into the issue where `env().platform` is `undefined`. This is the pragmatic fix, since the fallback is supposed to be removed in the future. (and to actually run TFJS in an AudioWorklet, I needed some more patches to the WASM backend loading, but that's unrelated to this PR).


pyu10055 (approved):

oneil512:
Any progress on this?

PR #8392, opened 2024-10-02
webgpu: fix: conditionally call deprecated GPUAdapter.requestAdapterInfo

mattvr:
## Context

`GPUAdapter.requestAdapterInfo` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/GPUAdapter/requestAdapterInfo) and non-standard.

It's in use by `tfjs-backend-webgpu`, but [the result is optional](https://github.com/tensorflow/tfjs/blob/3daf152cb794f4da58fce5e21e09e8a4f89c8f80/tfjs-backend-webgpu/src/backend_webgpu.ts#L136), and only used for a [single optimization](https://github.com/tensorflow/tfjs/blob/3daf152cb794f4da58fce5e21e09e8a4f89c8f80/tfjs-backend-webgpu/src/backend_webgpu.ts#L149). Yet if called in an environment which doesn't support it, it will throw.

By checking the feature's available before calling it, this unbreaks Deno runtime support for `tfjs-backend-webgpu`, which doesn't support this feature and throws.

## Testing/Verification

`yarn test` and `yarn lint` succeed. Verified basic example now works in Deno.

```ts
import * as tf from "@tensorflow/tfjs-core";
import "@tensorflow/tfjs-backend-webgpu";

await tf.ready();
tf.randomGamma([2, 2], 1).print();
```

## Related

https://github.com/denoland/deno/issues/22029

### DIFF
--- a/tfjs-backend-webgpu/src/base.ts
+++ b/tfjs-backend-webgpu/src/base.ts
@@ -57,7 +57,10 @@ if (isWebGPUSupported()) {
     };
 
     const device: GPUDevice = await adapter.requestDevice(deviceDescriptor);
-    const adapterInfo = await adapter.requestAdapterInfo();
+    const adapterInfo =
+      'requestAdapterInfo' in adapter
+        ? await adapter.requestAdapterInfo()
+        : undefined;
     return new WebGPUBackend(device, adapterInfo);
   }, 3 /*priority*/);
 }


tfjs-backend-webgpu/src/base.ts
```diff
@@ -57,7 +57,10 @@ if (isWebGPUSupported()) {
     };
 
     const device: GPUDevice = await adapter.requestDevice(deviceDescriptor);
-    const adapterInfo = await adapter.requestAdapterInfo();
+    const adapterInfo =
+      'requestAdapterInfo' in adapter
+        ? await adapter.requestAdapterInfo()
+        : undefined;
```
petamoriken, line 63:
```suggestion
      'info' in adapter
        ? adapter.info
        : await adapter.requestAdapterInfo();
```

mattsoulanille:
I'm seeing this error in the build:
```
tfjs-backend-webgpu/src/base.ts:64:38 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'GPUAdapterInfo'.
  Type '{}' is missing the following properties from type 'GPUAdapterInfo': __brand, vendor, architecture, device, description

64     return new WebGPUBackend(device, adapterInfo);
                                        ~~~~~~~~~~~
```

Maybe our WebGPU types from NPM are out of date? Does `yarn build` work in the `tfjs-backend-webgpu` directory for you?

mattvr:
@mattsoulanille Thank you for reviewing, yeah – I'm seeing that too with the previous commit. Oudated WebGPU types was it!

The latest commit fixes the error. It keeps the `adapter.requestAdapterInfo()` fallback for backwards compatibility, to prevent breaking older environments. This required an `any` typecast though, let me know if you'd like any changes.

mattsoulanille:
Thanks! The `any` is probably okay since it gets immediately passed to another function that has a defined type for that argument, so the `any` type shouldn't leak anywhere.

mattsoulanille (approved):

tfjs-backend-webgpu/src/base.ts
```diff
@@ -57,7 +57,12 @@ if (isWebGPUSupported()) {
     };
 
     const device: GPUDevice = await adapter.requestDevice(deviceDescriptor);
-    const adapterInfo = await adapter.requestAdapterInfo();
+    const adapterInfo =
+      'info' in adapter
+        ? adapter.info
+        : 'requestAdapterInfo' in adapter
+          ? await (adapter as any).requestAdapterInfo()
```
mattsoulanille, line 64: You'll need `/* tslint:disable-next-line:no-any */` right before this, or it may be better to cast `(adapter as unknown as GPUAdapterInfo)` instead.
```suggestion
          ? await (adapter as unknown as GPUAdapterInfo).requestAdapterInfo()
```

mattsoulanille (changes requested):
Looks like the linter isn't happy with `any`.

tfjs-backend-webgpu/src/base.ts
mattsoulanille: Actually, this is wrong, my bad. The promise result should be `GPUAdapterInfo`. Just adding the `/* tslint:disable-next-line:no-any */` should fix it.

mattvr:
@mattsoulanille Sorry for the back-and-forth, could've sworn `build` and `test` succeeded before I submitted the last two commits. Should be fixed now! 🤞

mattsoulanille:
No worries! Our build and test scripts are not particularly easy to use (sorry). There's a `yarn lint` command you can run at the root of the repo to check this.

mattsoulanille (approved):

PR #8425, opened 2024-10-25
[tfjs-node] replace deprecated utils

tharvik:
after been [marked as deprecated in node 22](https://github.com/nodejs/node/pull/50488), node 23 [removed nearly all its `util.is*` functions](https://github.com/nodejs/node/pull/52744). so using the node backend in 23 fails quite loudly with some "TypeError: util_1.isNullOrUndefined is not a function".

this PR replace the few places where theses functions were used with [node's recommended fixes](https://nodejs.org/docs/latest-v22.x/api/util.html#utilisnullorundefinedobject). or with nullish-coalescing where appropriate.

### DIFF
--- a/tfjs-node/src/kernels/TopK.ts
+++ b/tfjs-node/src/kernels/TopK.ts
@@ -16,7 +16,6 @@
  */
 
 import {KernelConfig, scalar, TopK, TopKAttrs, TopKInputs} from '@tensorflow/tfjs';
-import {isNullOrUndefined} from 'util';
 
 import {createTensorsTypeOpAttr, NodeJSKernelBackend} from '../nodejs_kernel_backend';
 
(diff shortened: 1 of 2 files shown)


andresribeiro:
+1

Ananthusubramanian:
+1

domdomegg:
@mattsoulanille can you review this? The fix looks correct and it blocks being able to use tfjs-node on Node >=23.

barroudjo:
Same thing, on node 23 (and above) this is absolutely needed. Why is it not merged after 5 months ?

bubbatls:
+1 
Please

mattsoulanille (approved):

mattsoulanille:
Thanks! We're currently changing our CI config, so I've run this locally to verify it passes.

chr4ss12:
why is this still not released? 


Added this to my project:




// nice hack to support tensorflow,
// in future tensorflow package upgrade should not need these functions anymore.
// https://github.com/tensorflow/tfjs/pull/8425
const util = require('util');
if (!util.isNullOrUndefined) {
  util.isNullOrUndefined = (val: any): val is null | undefined => {
    return val === null || val === undefined;
  };
}

if (!util.isArray) {
  util.isArray = (val: any): val is any[] => {
    return Array.isArray(val);
  };
}

PR #8483, opened 2024-12-19
[WebGPU] Make dataToGPU upload to GPU if data is on CPU

mattsoulanille:

### DIFF
--- a/tfjs-backend-webgpu/src/backend_webgpu.ts
+++ b/tfjs-backend-webgpu/src/backend_webgpu.ts
@@ -594,8 +594,9 @@ export class WebGPUBackend extends KernelBackend {
    * @param dataId The source tensor.
    */
   override readToGPU(dataId: DataId): GPUData {
-    const srcTensorData = this.tensorMap.get(dataId);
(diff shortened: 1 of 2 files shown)


tfjs-backend-webgpu/src/backend_webgpu.ts
```diff
@@ -594,16 +594,19 @@ export class WebGPUBackend extends KernelBackend {
...
 
     if (dtype === 'complex64') {
       throw new Error('Does not support reading buffer for complex64 dtype.');
     }
 
     if (resource == null) {
       if (values != null) {
-        throw new Error('Data is not on GPU but on CPU.');
+        this.uploadToGPU(dataId);
+        srcTensorData = this.tensorMap.get(dataId);
```
Linchenn, line 608: Seems we don't need this line, wdyt?

Linchenn (approved):
Thanks!

tfjs-backend-webgpu/src/backend_webgpu.ts
mattsoulanille: Nice catch

PR #8517, opened 2025-02-15
[tfjs-core] do not hang on invalid browser files

tharvik:
`tf.io.browserFiles` doesn't fail when loading invalid files. or rather, it fails, but never rejects the promise, making the `IOHandler.load` hang forever.

wrapping the `JSON.parse` call in a try/catch and rejecting accordingly did the trick. and a small test to try it out.

note: `FileReader.readAsText()` is a callback way to go around reading files. the promise-based `Blob.text()` would make `BrowserFile.load` simpler and safer (but felt ouf-of-scope here).


mattsoulanille (approved):
Thanks!

PR #8524, opened 2025-02-28
Add support Pluggable Devices

### LINKED ISSUE #8040: Support PluggableDevices
**System information**
- TensorFlow.js version (you are using): 4.6.0
- Are you willing to contribute it (Yes/No): Probably not, I definitely lack the expertise for that if it's not a really simple thing

Both tensorflow python and [libtensorflow](https://github.com/intel/intel-extension-for-tensorflow/blob/main/docs/install/install_for_cpp.md#load) provide the option to load a pluggable device driver, like Intel's extensions for tensorflow, where as TensorFlow.js on Node.js currently seems to be vendor locked on NVIDIA GPUs, making it kinda less of a `tfjs-node-gpu` than a `tfjs-node-nvidia`.

**Will this change the current api? How?**

There'd probably be an additional api to load the driver `.so` and there'd maybe be more kinds of devices available.

**Who will benefit with this feature?**

Basically anyone that uses tfjs node and has any hardware that is not NVIDIA, but support pluggable devices, notably Intel GPUs (also optimized code for their CPUs) and Apple devices, AMD still doesn't have a pluggable device driver I think, but uses a custom fork of tensorflow.

**Any Other info.**

As (although I can't find any confirmation, other than the log format indicating it) tfjs-node seems to be built on libtensorflow, it seems to you'd "just" need to call the C API (`TF_LoadPluggableDeviceLibrary(<lib_path>, status);`) from libtensorflow for loading the driver and there you go.

cromefire:
Fixes #8040


cromefire:
Seems good now. At least it loads the library. Can't test actually fully loading the a lib end to end right now as my target system has not enough memory to compile the tensorflow extension and I'll have to find a way to cross-compile it first (issues with missing instructions and libc versions and the like...). It does try to load the extension correctly though and as it's only the libtensorflow C path from there, it'd say it looks good.

PR #8714, opened 2026-06-01
Validate weight file paths stay within the model directory

adilburaksen:
In `tfjs-node`, `NodeFileSystem.loadWeights` (`tfjs-node/src/io/file_system.ts`) joins each path from the model's weights manifest onto the model directory and reads it, without verifying the result stays inside that directory. The manifest is untrusted input, so a path containing `..` or an absolute path can read a file outside the model directory.

This resolves each weight file path and requires it to remain within the model directory before reading. Relative paths under the model directory are unaffected.


adilburaksen:
@mattsoulanille mind giving this a review? It validates that weight file paths stay within the model directory when loading, so a manifest can't reference ../ out to arbitrary files. Small check in the load path — happy to adjust if there's a legitimate case for weights living outside the model dir.

PR #8722, opened 2026-06-13
fix: reduce batched matMul gradients to input shapes (#8721)

### LINKED ISSUE #8721: batched matMul gradient has wrong shape
`
tf.grads((a, b) => tf.matMul(a, b))([tf.tensor([[[1,2],[3,4]], [[5,6],[7,8]]]), tf.tensor([[9,10],[11,12]])])[1].print()
`
fails with error 
Uncaught Error: Error in gradient for op BatchMatMul. The gradient of input 'b' has shape '2,2,2', which does not match the shape of the input '2,2'

should be easy to fix:
https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/gradients/BatchMatMul_grad.ts#L33
the result should be reduces over the batch axes by using some helper function like:
```
            const reduceToShape = function (res, b) {
              const reduceAxes = getReductionAxes(b.shape, res.shape);
              if (reduceAxes.length > 0) {
                  res = sum$3(res, reduceAxes);
              }
              return res;
            };
```

PrateekKrishna:
## Description

Fixes #8721

### What is the problem this PR is trying to solve?
During backpropagation for `BatchMatMul` where inputs rely on broadcasting (e.g., tensor `a` has a batch dimension, tensor `b` does not), the gradients calculated previously retained the broadcasted shape. This caused a shape mismatch error in TensorFlow.js because the gradient shape no longer matched the original un-batched input shape.

### How does this PR solve the problem?
- Added a `reduceToShape` helper function in `tfjs-core/src/gradients/BatchMatMul_grad.ts`.
- Utilized `getReductionAxes` and `sum` to reduce the calculated gradients along the broadcasted axes, ensuring the returned gradients always match the original shapes of `a` and `b`.

### What testing has been done?
- Added a new unit test in `tfjs-core/src/ops/mat_mul_test.ts` (`'gradients: batched matMul with broadcasted shapes (Issue #8721)'`) that replicates the failing code from the issue.
- Verified that the new test passes and confirms the output gradient shapes equal the input tensor shapes.
- Ran the full `yarn test` suite in `tfjs-core` locally to ensure no existing operations or gradients were broken. All tests passed successfully.
image


PrateekKrishna:
I have signed the google CLA!

PR #8741, opened 2026-08-17
Pin GitHub Actions to full-length commit SHAs

herdiyana256:
Every third-party action in these workflows is referenced by a mutable tag (`@v2`, `@v4`, `@v0`, `@0.14.0`, `@v7`). GitHub resolves a tag to whatever commit it points at when the workflow runs, so if a tag is moved or the action's repository is compromised, the workflow executes that new code with the permissions and secrets the job holds.

The sharpest exposure is `deploy-benchmark-prod.yml`. It runs on every push to `master` and hands a production Firebase deploy credential to `FirebaseExtended/action-hosting-deploy@v0`:

```yaml
on:
  push:
    branches: [master]
steps:
  - uses: actions/checkout@v2
  - uses: FirebaseExtended/action-hosting-deploy@v0
    with:
      firebaseServiceAccount: "${{ secrets.FIREBASE_SERVICE_ACCOUNT_JSTENSORFLOW }}"
```

A `@v0` tag is a moving target by design, so a single upstream tag repoint, or a compromise of that action's repo, runs attacker-controlled code with that credential and no change to this repository is needed. Same mechanism for every other tag-pinned action here, just with lower-value tokens (the default `GITHUB_TOKEN` and build context). In practice that means arbitrary code execution in CI with the job's secrets the moment one of these tags moves.

This pins each external action to the commit SHA its current tag resolves to, with the version kept in a trailing comment for readability and future bumps:

```yaml
  - uses: actions/checkout@11d5960 # v4
  - uses: FirebaseExtended/action-hosting-deploy@500ac62 # v0
```

The SHAs are exactly what the tags resolve to today, so there is no behavior change, a job can just only ever run the exact commit that was reviewed. The two benchmark deploy workflows also drop the deprecated `actions/checkout@v2` (Node 12, unsupported on current runners) onto the same checkout v4 commit already used elsewhere in the repo.

Pinned references:

- `actions/checkout` -> `11d5960` (v4)
- `actions/setup-node` -> `49933ea` (v4)
- `actions/stale` -> `6f05e42` (v7)
- `bazel-contrib/setup-bazel` -> `e8776f5` (0.14.0)
- `FirebaseExtended/action-hosting-deploy` -> `500ac62` (v0)

Follows GitHub's Actions hardening guidance and OpenSSF Scorecard's Pinned-Dependencies check.


herdiyana256:
Gentle bump on this, open since mid-August with no review yet. It pins the third-party GitHub Actions across the workflows to full-length commit SHAs (they're tag-pinned today), closing the mutable-tag supply-chain risk, most notably deploy-benchmark-prod which runs with FIREBASE_SERVICE_ACCOUNT on every push to master. All checks were green. @mattsoulanille would you be able to take a look when you get a chance?

